PR #3, opened 2018-11-26
Android Authenticator PoC 

awesome-manuel:
The idea is to have a separate app that has an [Authenticator](https://developer.android.com/training/id-auth/custom_auth), [ContentProvider](https://developer.android.com/guide/topics/providers/content-providers) and optional [Firebase Messaging](https://firebase.google.com/docs/cloud-messaging/) for Matrix. See also https://developer.android.com/training/id-auth/custom_auth#DistributeService

The advantage is that all Android apps connecting to a Matrix HS can use these services instead of each including the whole matrix-android-sdk (in possibly outdated version). Also encryption can be handled in a central place instead of in each app. The apps don't even get access to the users' crypto keys.

The is also the only app that needs to have different flavor for stores without Firebase/GCM support.

Also multi-account support could be simplified by this approach.

This pull request is just a proof of concept and still has issues of course, but you can already create an account in Android system settings.


MTRNord:
It would be awesome to document this as SimpleMatrix uses this too. SimpleMatrix currently provides a org.matrix user account type

bmarty:
This is probable a good idea, but it should maybe be handled in a dedicated GitHub project.

PR #18, opened 2018-12-30
Hacks to get it running on API 19

z3ntu:
At least you can take inspiration from this :wink: 

I was able to log in and view chats and send a message

References:
https://developer.android.com/studio/build/multidex
https://stackoverflow.com/questions/27986204/cant-convert-to-color-type-0x2-error-when-inflating-layout-in-fragment-but-onl

### DIFF
--- a/app/src/main/res/drawable/bg_splash.xml
+++ b/app/src/main/res/drawable/bg_splash.xml
@@ -10,7 +10,8 @@
     <item android:gravity="top">
         <shape android:shape="rectangle">
             <size android:height="80dp" />
-            <solid android:color="?colorPrimary" />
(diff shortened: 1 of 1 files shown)


z3ntu:
https://github.com/vector-im/riotX-android/commit/946ae32cd803b96518392f1eda13df5ac5662e03 Did most the necessary work but it still crashes (see second "References" link) - I've updated the PR to show that change

bmarty:
Hi @z3ntu ,
Thanks for the PR. To merge it, I need a sign off as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off please.
Benoît

PR #23, opened 2019-01-14
homeserver input field

f0x52:
I guess this would work? I just edited it in vim idk android stuff

### DIFF
--- a/app/src/main/res/layout/activity_login.xml
+++ b/app/src/main/res/layout/activity_login.xml
@@ -43,6 +43,16 @@
                 android:inputType="textPassword"
                 android:maxLines="1" />
 
+            <EditText
+                android:id="@+id/hsField"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="10dp"
+                android:ems="10"
+                android:hint="Homeserver"
+                android:inputType="textUri"
+                android:maxLines="1" />
+
         </LinearLayout>
     </ScrollView>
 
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/activity_login.xml
```diff
@@ -43,6 +43,16 @@
                 android:inputType="textPassword"
                 android:maxLines="1" />
 
+            <EditText
+                android:id="@+id/hsField"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginTop="10dp"
+                android:ems="10"
+                android:hint="Homeserver"
```
krombel: Would it be possible to directly use strings here? (especially useful for translation)

f0x52: In which way? I just copied how the username and password fields work

krombel: Then a separate PR for i10n has to be created either way. Just saw the diff and not what was done for the other input fields

ganfra: @krombel:  Yes, at the moment, some strings are still hard-coded, and should be cleaned. 
I'll probably do that this week.

ganfra: @f0x52 : could you please reproduce the behavior actual Riot has on this screen with a checkbox to show/hide additional auth settings, including homeserver, with default to matrix server? 
Thanks for your work!

jcgruenhage: @ganfra I think this should also be merged as is, and the behaviour Riot has can be implemented in a later PR. This PR is already immensely useful for testers as it is.

jcgruenhage: (Under the assumption that it works, of course)

f0x52: @ganfra I quickly made this to speed up easier testing, I don't really know android development so I probably can't help with more complicated behaviour

ganfra: No problem, I just made it quickly on my side. Thanks for your work anyway :)

PR #25, opened 2019-02-02
Update Gradle wrapper to 5.1.1

14mRh4X0r:
This enables Gradle to work over IPv6-only deployments, like at FOSDEM.

Side note: The project does not seem to build this way (due to a linting failure), but I can't tell whether that's a regression due to obvious reasons.


bmarty:
Thanks for the PR, replaced by #89 
Lint analysis seems ok now

PR #87, opened 2019-03-13
get timestamp for versioncode from last git commmit

Bubu:
This makes the versioncode for an apk reproducible. Building the same
commit/code will produce an apk with the same VC.

This is a requirement for being able to build pre-release version by f-droid.


app/build.gradle
```diff
@@ -15,9 +15,18 @@ def versionMajor = 0
 def versionMinor = 1
 def versionPatch = 0
 
+def getGitTimetamp() {
```
bmarty: A `s`is missing :)
Also I prefer this format, more readable: https://github.com/vector-im/riot-android/blob/develop/vector/build.gradle#L193

app/build.gradle
```diff
@@ -15,9 +15,18 @@ def versionMajor = 0
...
+    def stdout = new ByteArrayOutputStream()
+    exec {
+        commandLine 'git', 'show', '-s', '--format=%ct'
+        standardOutput = stdout
+    }
+    return stdout.toString() as Integer
```
bmarty: Timestamp should be `Long`

app/build.gradle
```diff
@@ -15,9 +15,18 @@ def versionMajor = 0
...
+}
+
 def generateVersionCodeFromTimestamp() {
     // It's unix timestamp divided by 10: It's incremented by one every 10 seconds.
```
bmarty: This comment should be updated too

app/build.gradle
Bubu: > A sis missing :)

:see_no_evil: 

>  Also I prefer this format, more readable: https://github.com/vector-im/riot-android/blob/develop/vector/build.gradle#L193

What exactly do you mean here? The syntax for the gradle function? I can change that, sure.

app/build.gradle
Bubu: I deliberately left the comment in place, I think it's still valid. How would you change it?

app/build.gradle
bmarty: Yes, it was the format of the gradle function, thanks for the change

app/build.gradle
bmarty: The comment can stay like that. The point is that the value doesn't change anymore if there is no new commit.

bmarty:
Thanks for the update, all I need now to merge this is a Sign-Off please.

Bubu:
This should do it?

bmarty:
Yes, thanks, I will merge the PR soon

PR #138, opened 2019-05-16
Send reaction view quick react and picker

### LINKED ISSUE #114: Reactions: Send reaction event when user reacts
From riot-web created by [jryans](https://github.com/jryans): vector-im/riot-web#9572  

Reaction events are sent in response to 2 UI actions:

* Choosing a new emoji type from the picker
* Clicking on an existing type displayed below a message to increment an existing type

### LINKED ISSUE #116: Reactions: Increment existing reaction type on click
From riot-web created by [jryans](https://github.com/jryans): vector-im/riot-web#9486  

When you click one of the existing reaction types displayed below a message, increment that type and highlight it.

BillCarsonFr:
Fixes #114 
Fixes #116 

Adding Ability to send reaction via quick react and picker + increment existing

Known Limitations:
 - Can only add reaction (not possible to undo/remove existing)
 - Quick reaction is only adding (e.g. if you disagree a previously agreed event, app will not undo old and add new)
- Not using the new send_relation API as not available (code created but not tested), uses the regular send event
- No Optimistic update. Intermediate states e.g. currently sending reaction not handled perfectly (might experience delay) -> Optimistic send in phase 2

+ fix / Error when to many reactions in cells (more than placeholders -8-)
+ fix / DefaultTimeline quick map access was not shifted when items inserted at given index (reaction was added to wrong message)

quick_react

picker_react

increment

### DIFF
--- a/vector/src/main/java/im/vector/riotredesign/features/home/room/detail/timeline/item/AbsMessageItem.kt
+++ b/vector/src/main/java/im/vector/riotredesign/features/home/room/detail/timeline/item/AbsMessageItem.kt
@@ -30,6 +30,7 @@ import com.airbnb.epoxy.EpoxyAttribute
 import im.vector.riotredesign.R
 import im.vector.riotredesign.core.utils.DimensionUtils.dpToPx
 import im.vector.riotredesign.features.home.AvatarRenderer
+import im.vector.riotredesign.features.home.room.detail.timeline.TimelineEventController
 import im.vector.riotredesign.features.reactions.widget.ReactionButton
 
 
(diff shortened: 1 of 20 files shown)


vector/src/main/java/im/vector/riotredesign/features/home/room/detail/timeline/item/AbsMessageItem.kt
```diff
@@ -107,6 +123,10 @@ abstract class AbsMessageItem<H : AbsMessageItem.Holder> : BaseEventItem<H>() {
         }
     }
 
+    override fun unbind(holder: H) {
```
ganfra: A small miss?

BillCarsonFr: Yep, will fix in uno_reaction PR. thx

PR #140, opened 2019-05-20
Use CS API to send relation

BillCarsonFr:
Use CS API instead of send to react to an event

+ Fix scroll on login view when hs field is hidden


bmarty:
According to @BillCarsonFr , we will need to rewrite it if we want to merge it.

PR #175, opened 2019-06-10
Feature/crypto

ganfra:
This PR brings crypto to RiotX.
It allows to decrypt/encrypt messages, use SAS verification and keys backup (only from settings screen).
It has some limitations: 
- Failures are not yet well handled
- It does send encrypted messages even when unknown devices are in the room (without warning).
- There are probably some other issues
- There are still code to clean up


matrix-sdk-android/build.gradle
```diff
@@ -45,8 +45,8 @@ android {
 
         debug {
             // Set to true to log privacy or sensible data, such as token
+            // TODO Set to false
```
BillCarsonFr: done?

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/events/model/Event.kt
```diff
@@ -81,10 +84,147 @@ data class Event(
...
+     * @return true if this event is encrypted.
+     */
+    fun isEncrypted(): Boolean {
+        return TextUtils.equals(type, EventType.ENCRYPTED)
+    }
+
+    /**
+     * Update the clear data on this event.
+     * This is used after decrypting an event; it should not be used by applications.
+     * It fires kMXEventDidDecryptNotification.
```
BillCarsonFr: Does it?

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/actions/EnsureOlmSessionsForDevicesAction.kt
```diff
@@ -0,0 +1,158 @@
...
+
+
+    suspend fun handle(devicesByUser: Map<String, List<MXDeviceInfo>>): Try<MXUsersDevicesMap<MXOlmSessionResult>> {
+        val devicesWithoutSession = ArrayList<MXDeviceInfo>()
+
+        val results = MXUsersDevicesMap<MXOlmSessionResult>()
+
+        val userIds = devicesByUser.keys
+
+        for (userId in userIds) {
```
BillCarsonFr: I see a lot of potential NPE code. Maybe we could reduce usage of !!

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/events/model/Event.kt
ganfra: aha, iOS comment copy/pasted twice!

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/actions/EnsureOlmSessionsForDevicesAction.kt
ganfra: Yes, it's converting java to kotlin which leads to this, but it should be clean.

PR #179, opened 2019-06-13
Crypto: Delete device

bmarty:
Add delete device capability to RiotX

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/auth/data/LoginFlow.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/auth/data/LoginFlow.kt
@@ -16,8 +16,18 @@
 
 package im.vector.matrix.android.internal.auth.data
 
+import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
 
+/**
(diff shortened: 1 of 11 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/auth/data/LoginFlow.kt
```diff
@@ -16,8 +16,18 @@
...
-internal data class LoginFlow(val type: String,
-                     val stages: List<String>)
\ No newline at end of file
+internal data class LoginFlow(
+
+        @Json(name = "type")
+        val type: String? = null,
+
+        @Json(name = "stages")
+        val stages: List<String>
```
BillCarsonFr: to be  safe, might be better to set that as optional (?)
Could be renamed in something like InteractiveAuth? https://matrix.org/docs/spec/client_server/r0.5.0#user-interactive-authentication-api

bmarty: Added optional to stage and renamed to "InteractiveAuthenticationFlow"

PR #188, opened 2019-06-18
Disambiguation of display names

bmarty:
This PR implements adding userId after the display name when there are several members with the same display name in a room.

Prefer using TimelineEvent.getDisambiguatedDisplayName() when appropriate now.

Also "sender" has been renamed to "senderId" for code clarity


matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/timeline/TimelineEvent.kt
```diff
@@ -53,6 +54,18 @@ data class TimelineEvent(
         }
     }
 
+    fun getDisambiguatedDisplayName(): String {
+        return if (isUniqueDisplayName) {
+            senderName ?: root.senderId
+        } else {
+            senderName?.let {
+                it + " (" + root.senderId + ")"
```
ganfra: Use string interpolation?

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/timeline/TimelineEvent.kt
```diff
@@ -53,6 +54,18 @@ data class TimelineEvent(
...
+            }
+                    ?: root.senderId
```
ganfra: Let the root.senderId be outside of the if as you write it twice.

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/timeline/TimelineEvent.kt
bmarty: done

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/timeline/TimelineEvent.kt
bmarty: done

PR #227, opened 2019-06-26
Encrypt Realm databases

BillCarsonFr:
Add encryption for realm databases (auth / session / crypto).

On creation a random key is generated, this key is then encrypted using the system KeyStore.
The encrypted key is stored in shared preferences.
When the database is opened again, the encrypted key is taken from the shared pref, then the Keystore is used to decrypt the key. The decrypted key is passed to the RealConfiguration.

On android >=M, the KeyStore generates an AES key to encrypt/decrypt the database key, and the encrypted key is stored with the initialization vector in base64 in the shared pref.
On android <M, the KeyStore cannot create AES keys, so a public/private key pair is generated, then we generate a random secret key. The database key is encrypted with the secret key; The secret key is encrypted with the public RSA key and stored with the encrypted key in the shared pref


ganfra:
LGTM

PR #241, opened 2019-06-28
min API 19 - tested

### LINKED ISSUE #237: Test app on Android API 16

bmarty:
There is a lot to do to clean resource/color, but not P1

Fixes #237


bmarty:
Fix #237

PR #258, opened 2019-07-01
Fix / signout sign-in not working

BillCarsonFr:
FIx / logout then login not working


vector/src/main/java/im/vector/riotredesign/features/MainActivity.kt
```diff
@@ -66,6 +67,10 @@ class MainActivity : VectorBaseActivity() {
                 override fun onSuccess(data: Unit) {
                     Timber.w("SIGN_OUT: success, start app")
                     sessionHolder.clearActiveSession()
+                    WorkManager.getInstance(applicationContext).also {
```
bmarty: to do in SDK?

BillCarsonFr: done

PR #262, opened 2019-07-02
Update timeline on new session

BillCarsonFr:
Added mechanism to detect when a new key is received in order to refresh current timeline.

How to:
Enter encrypted room on phone, go to other account where a key share popup appears then verify device 
-> Before: you had to exit then enter the room to see devcrypted event
-> Now: Timeline will be refreshed with clear content


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/RoomDecryptorProvider.kt
```diff
@@ -34,6 +34,16 @@ internal class RoomDecryptorProvider @Inject constructor(
     // A map from algorithm to MXDecrypting instance, for each room
     private val roomDecryptors: MutableMap<String /* room id */, MutableMap<String /* algorithm */, IMXDecrypting>> = HashMap()
 
+    private val newSessionListeners = ArrayList<NewSessionListener>()
+
+    fun addNewSessionListener(listener: NewSessionListener) {
+        if (!newSessionListeners.contains(listener)) newSessionListeners.add(listener)
```
bmarty: Why not use a Set?

PR #299, opened 2019-07-04
Fix / Push worker could launch concurrent syncs

BillCarsonFr:
Fix / Concurrent syncs
Start initial sync then enters foreground, if a push is received the SyncWorker is started and will start another initial sync (token null)
-> Now worker use the same sync task as sync thread on the task executor queue


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/sync/job/SyncWorker.kt
```diff
@@ -47,50 +45,45 @@ internal class SyncWorker(context: Context,
...
-                {
-                    if (it is Failure.ServerError
-                            && it.error.code == MatrixError.UNKNOWN_TOKEN) {
-                        sessionParamsStore.delete(params.userId)
-                        Result.failure()
-                    } else {
-                        Timber.i("Sync work failed $it")
-                        Result.retry()
+
+        val latch = CountDownLatch(1)
```
ganfra: Maybe using runBlocking would be cleaner

bmarty: ok, P2 :)

PR #306, opened 2019-07-04
Convert some callback functions to suspend

Dominaezzz:
Since this project is in Kotlin, I thought it would be a good idea to use the awesome async features of the language.
I've converted some of the functions that take callbacks like `fun call(callback: MatrixCallback<String>)`  to `suspend call(): String`. This will eventually make it easier to write a nicer version of  https://github.com/vector-im/riotX-android/blob/498b1f2b06e9a62cb1a57c37d480e8ebaff4fe1a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/EncryptEventWorker.kt#L71 
For the function callers, I've mostly assumed code needs to finish on the main thread.

### Pull Request Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/RoomService.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/RoomService.kt
@@ -17,7 +17,6 @@
 package im.vector.matrix.android.api.session.room
 
 import androidx.lifecycle.LiveData
-import im.vector.matrix.android.api.MatrixCallback
 import im.vector.matrix.android.api.session.room.model.RoomSummary
 import im.vector.matrix.android.api.session.room.model.create.CreateRoomParams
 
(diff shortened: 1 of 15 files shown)


bmarty:
Hi @Dominaezzz,

After discussion with the team, we want to keep the SDK compatible with a Java application. So we cannot expose suspend functions.

We only use coroutines internally in the SDK.

Also the code seems less clear like that, but it's a personal point of view...

Benoit

Dominaezzz:
Ah that's a shame, I was really hoping to integrate suspension into the SDK and utilize Kotlin's powerful async features. Especially now that Android is Kotlin first and this is an Android-specific SDK. I reckon most people using this sdk would be using Kotlin as well, since it's the primary language and if they're not using Kotlin yet, it is very easy to mix Kotlin and Java in a project as you might have experienced in the original Riot android project. I think it could be wasted opportunity in this rewrite. But if full java compatibility is strictly desired, fair enough.

Instead of suspending functions, would returning `Deferred<T>` or `Job` be do-able? These types can be consumed from Java and is more ergonomic for Kotlin than callbacks.
So like `fun call(): Deferred<String>` instead of `fun call(callback: MatrixCallback<String>)`. It would aid in minimizing awkward interactions between async and sync code. Also prevent callback hell. (Not a great fan of callbacks, if you can't tell).

The code may seem less clear because I kept the changes minimal, so you may not immediately appreciate the benefits of using suspension. I've already spotted a few bugs around async handling that could have been prevent with suspension.

PR #309, opened 2019-07-05
Rework Crypto using Try

bmarty:
Use of Try to avoid throwing Exceptions


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
```diff
@@ -319,11 +323,20 @@ internal class MXMegolmDecryption(private val credentials: Credentials,
...
-                                        val inboundGroupSession = olmDevice.getInboundGroupSession(body?.sessionId, body?.senderKey, body?.roomId)
 
                                         val payloadJson = HashMap<String, Any>()
                                         payloadJson["type"] = EventType.FORWARDED_ROOM_KEY
-                                        payloadJson["content"] = inboundGroupSession!!.exportKeys()!!
+
+                                        olmDevice.getInboundGroupSession(body?.sessionId, body?.senderKey, body?.roomId)
+                                                .fold(
+                                                        {
+                                                            // TODO
```
BillCarsonFr: TODO ?

bmarty: it will send event without useful content...
There is lot of ? in this code, we should rework it...

PR #325, opened 2019-07-10
Accept non unicode reactions

### LINKED ISSUE #307: Reactions: Reinstate the ability to react with non-unicode keys
From riot-ios created by [manuroe](https://github.com/manuroe): vector-im/riot-ios#2553  

Discussed at https://github.com/vector-im/riot-web/issues/10256.

Web PR with an hardcoded 100px width limit: https://github.com/matrix-org/matrix-react-sdk/pull/3186

BillCarsonFr:
Fixes #307 

There is still an issue due to font & compatibility
As the reactions widget was supposed to display only reaction, they use the emoji font, but this font displays spaces in a strange way (like 3 times bigger in some devices, and 0 in others)

-> We probably need to move to EmojiCompat (will replace emojis with Spans inside the text), but it's a bigger task (side effects on emoji picker?)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,10 +5,10 @@ Features:
  - Display read receipts in timeline (#81)
 
 Improvements:
- -
+ - Reactions: Reinstate the ability to react with non-unicode keys (#307)
 
 Other changes:
- -
+ - 
 
 Bugfix:
  - Fix text diff linebreak display (#441)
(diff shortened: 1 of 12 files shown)


BillCarsonFr:
Modified to use EmojiCompat lib, this allows to use EmojiCompat to process strings and add EmojiSpans when emojis are detected

PR #334, opened 2019-07-11
Feature/general perf

ganfra:
This PR should brings better perf
- It removes the use of UUID as primaryKey for events 
- It uses Lazy<> services in session
- Use a single thread for Sync


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/helper/TimelineEventEntityHelper.kt
```diff
@@ -69,6 +71,15 @@ internal fun TimelineEventEntity.updateSenderData() {
     this.senderMembershipEvent = senderMembershipEvent
 }
 
+internal fun TimelineEventEntity.Companion.nextId(realm: Realm): Long{
```
BillCarsonFr: Does this work if we create instances in multiple process?

PR #337, opened 2019-07-11
Add ".debug" to the applicationId to be able to install the app along with the prod version

bmarty:


vector/src/gplay/debug/google-services.json
```diff
@@ -0,0 +1,40 @@
...
+    "firebase_url": "https://vector-alpha.firebaseio.com",
+    "project_id": "vector-alpha",
+    "storage_bucket": "vector-alpha.appspot.com"
+  },
+  "client": [
+    {
+      "client_info": {
+        "mobilesdk_app_id": "1:912726360885:android:4ef8f3a0021e774d",
+        "android_client_info": {
+          "package_name": "im.vector.riotx.debug"
```
bmarty: only diff with the other file is here

PR #338, opened 2019-07-12
Text in green when encrypting

bmarty:
Also do some cleanup


vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/item/AbsMessageItem.kt
```diff
@@ -153,13 +161,12 @@ abstract class AbsMessageItem<H : AbsMessageItem.Holder> : BaseEventItem<H>() {
         return true
     }
 
-    protected fun View.renderSendState() {
-        isClickable = informationData.sendState.isSent()
-        alpha = if (informationData.sendState.isSent()) 1f else 0.5f
+    protected fun renderSendState(root: View, textView: TextView?) {
+        root.isClickable = informationData.sendState.isSent()
+        textView?.setTextColor(colorProvider.getMessageTextColor(informationData.sendState))
```
BillCarsonFr: no more alpha for pending messages?

PR #380, opened 2019-07-16
Fix a crash in notificationwhen display name is empty

bmarty:
Also some code cleanup

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/DefaultSession.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/DefaultSession.kt
@@ -94,11 +94,11 @@ internal class DefaultSession @Inject constructor(override val sessionParams: Se
     }
 
     override fun requireBackgroundSync() {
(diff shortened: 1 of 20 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/DefaultSession.kt
```diff
@@ -94,11 +94,11 @@ internal class DefaultSession @Inject constructor(override val sessionParams: Se
     }
 
     override fun requireBackgroundSync() {
-        SyncWorker.requireBackgroundSync(context, sessionParams.credentials.userId)
+        SyncWorker.requireBackgroundSync(context, myUserId)
     }
 
     override fun startAutomaticBackgroundSync(repeatDelay: Long) {
-        SyncWorker.automaticallyBackgroundSync(context, sessionParams.credentials.userId, 0, repeatDelay)
+        SyncWorker.automaticallyBackgroundSync(context, myUserId, 0, repeatDelay)
```
BillCarsonFr: i think this is some dead code (is it used anymore?)

PR #383, opened 2019-07-17
Pass filter to room directory screen or create room screen

bmarty:
When filtering in room directory:
 -> pass the filter to the room creation form
When filtering the room list:
 -> pass the filter to the room creation form
 -> pass the filter to the room directory list

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/roomdirectory/createroom/CreateRoomViewModel.kt
+++ b/vector/src/main/java/im/vector/riotx/features/roomdirectory/createroom/CreateRoomViewModel.kt
@@ -16,6 +16,7 @@
 
 package im.vector.riotx.features.roomdirectory.createroom
 
+import androidx.fragment.app.FragmentActivity
 import com.airbnb.mvrx.*
 import com.squareup.inject.assisted.Assisted
 import com.squareup.inject.assisted.AssistedInject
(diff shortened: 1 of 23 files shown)


vector/src/main/java/im/vector/riotx/features/roomdirectory/createroom/CreateRoomViewModel.kt
```diff
@@ -39,8 +41,13 @@ class CreateRoomViewModel @AssistedInject constructor(@Assisted initialState: Cr
 
         @JvmStatic
         override fun create(viewModelContext: ViewModelContext, state: CreateRoomViewState): CreateRoomViewModel? {
-            val fragment: CreateRoomFragment = (viewModelContext as FragmentViewModelContext).fragment()
-            return fragment.createRoomViewModelFactory.create(state)
+            val activity: FragmentActivity = (viewModelContext as ActivityViewModelContext).activity()
+
+            return when (activity) {
+                is CreateRoomActivity    -> activity.createRoomViewModelFactory.create(state)
```
BillCarsonFr: Two first cases are doing the same thing?

bmarty: cast is not the same

PR #384, opened 2019-07-17
Feature/edit e2e

### LINKED ISSUE #143: Support edits in e2e rooms
From riot-ios created by [manuroe](https://github.com/manuroe): vector-im/riot-ios#2449

### LINKED ISSUE #152: Message Edition: Reply (m.in_reply_to) messages edition
Reply messages are a bit special due to the "fallback format"
https://matrix.org/docs/spec/client_server/r0.4.0.html#fallback-for-m-text-m-notice-and-unrecognised-message-types

Replies are prefixed with the following content in the body:
```
> <@alice:example.org> This is the original body

This is where the reply goes
```

And this one in the formatted body:
```
<mx-reply>
  <blockquote>
    <a href="https://matrix.to/#/!somewhere:domain.com/$event:domain.com">In reply to</a>
    <a href="https://matrix.to/#/@alice:example.org">@alice:example.org</a>
    <br />
    <!-- This is where the related event's HTML would be. -->
  </blockquote>
</mx-reply>
This is where the reply goes.
```

As these part are not explicitly entered by the user, they should not be presented for edition

### LINKED ISSUE #314: Edits history: Display original event
From riot-ios created by [manuroe](https://github.com/manuroe): vector-im/riot-ios#2559  

The API will be updated as follows https://github.com/matrix-org/synapse/issues/5595.

Web bug: https://github.com/vector-im/riot-web/issues/10200

BillCarsonFr:
Fixes #143
Fixes #314
Fixes #152

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/action/ViewEditHistoryViewModel.kt
+++ b/vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/action/ViewEditHistoryViewModel.kt
@@ -28,6 +28,7 @@ import im.vector.matrix.android.api.session.room.model.message.isReply
 import im.vector.matrix.android.internal.crypto.algorithms.olm.OlmDecryptionResult
 import im.vector.riotx.core.platform.VectorViewModel
 import im.vector.riotx.features.home.room.detail.timeline.helper.TimelineDateFormatter
+import timber.log.Timber
 import java.util.*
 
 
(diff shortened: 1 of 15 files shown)


vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/action/ViewEditHistoryViewModel.kt
```diff
@@ -79,16 +82,36 @@ class ViewEditHistoryViewModel @AssistedInject constructor(@Assisted
...
+                                        senderKey = result.senderCurve25519Key,
+                                        keysClaimed = result.claimedEd25519Key?.let { k -> mapOf("ed25519" to k) },
+                                        forwardingCurve25519KeyChain = result.forwardingCurve25519KeyChain
+                                )
+                            } catch (e: MXCryptoError) {
+
+                            }
+                        }
+
+                        if (it.eventId == eventId) {
```
bmarty: I'm not comfortable with the `apply` here, because, there is 3 different `eventId` in the class...

BillCarsonFr:
I need to create a new task to handle decryption error when doing incremental edit aggregation.
Currently if the keys are not known when doing edit aggregation, the replace will be ignore and lost (the message won't be marked as edited, and in some cases showing the orginal event).
The EventRelationsAggregationTask need somehow to keep track of failed to decrypt relation, and listen for new keys in order to retry

PR #421, opened 2019-07-21
Remove Try from Tasks

Dominaezzz:

### Pull Request Checklist

So I noticed `Try<...>`  is used as a return type from suspending functions. Unfortunately, when exceptions are handled this way, [coroutine cancellation](https://kotlinlang.org/docs/reference/coroutines/cancellation-and-timeouts.html#cancellation-is-cooperative) doesn't work. `CancellationException` needs to be thrown as normal and not swallowed by `Try`. As it is now, most of the `Cancellable`s don't actually cancel properly because of this.

So I've done a some of the work of removing them from suspend functions in particular.
So instead of having,
```
suspend fun execute(...): Try<...>

execute(...).fold(..., ...)
```

I've changed it to,
```
suspend fun execute(...): ...

Try { execute(...) }.fold(..., ...)
// or runCatching { execute(...) }.fold(..., ...)
```

This will also reduce the number of allocations. :smile: 
This still isn't perfect but it's better than before. I wanted to keep this PR small. I'll slowly move `Try` up to non-suspending code, where it works properly.


ganfra:
Thanks for the work, used it on https://github.com/vector-im/riotX-android/pull/460

PR #444, opened 2019-07-29
Basic Message Failure support + Resend (text only)

BillCarsonFr:
- Does not yet support resending media, but failed media won't break any more the send media queue (can upload new images)
- Updated menu ux for failed to send messages with option to resend / remove
+ clean worker inputs when starting new independent task in unique queue

image

image


matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/events/model/Event.kt
```diff
@@ -81,6 +84,7 @@ data class Event(
 
     var mxDecryptionResult: OlmDecryptionResult? = null
     var mCryptoError: MXCryptoError.ErrorType? = null
+    var sendState: SendState = SendState.UNKNOWN
```
ganfra: Why did you pass it in Event class? We should avoid adding more "var" not in spec...

PR #464, opened 2019-08-07
Split apk

bmarty:
Split APK per arch.
Also cleanup unwanted library

APK size is reduced from about 31Mo to about 18Mo each


bmarty:
Not tested on x86

PR #495, opened 2019-08-18
Slide-in reply icon is distorted

### LINKED ISSUE #423: Slide-in reply icon is distorted
The swipe to reply icon looks a little stretched vertically.

You can see the difference compared to the reply icon at the bottom left above the keyboard here:

This is on a Pixel XL, Android 9.

BillCarsonFr:
Fixes #423 

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Other changes:
  -
 
 Bugfix:
- -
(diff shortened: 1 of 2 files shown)


bmarty:
Please add an entry in CHANGES.md

PR #522, opened 2019-08-27
Fix / regression on e2e reply and edit of reply

BillCarsonFr:
Fixes a regression on e2e replies


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/relation/DefaultRelationService.kt
```diff
@@ -208,10 +208,10 @@ internal class DefaultRelationService @AssistedInject constructor(@Assisted priv
         return TimelineSendEventWorkCommon.createWork<EncryptEventWorker>(sendWorkData, true)
     }
 
-    private fun createSendEventWork(event: Event): OneTimeWorkRequest {
+    private fun createSendEventWork(event: Event, startChain: Boolean = true): OneTimeWorkRequest {
```
bmarty: As discussed IRL, maybe avoid a default value for param startChain to avoid mistake in the future :/

BillCarsonFr: Done

PR #524, opened 2019-08-28
Add unread indent on room list

### LINKED ISSUE #485: Rooms with unread message(s) are not always distinguishable
If there is no notification in the room for isntance.

On Riot-Android we use a vertical bar on the left of the room cell.

https://matrix.to/#/!tTIucwUZLRtKnXeurb:matrix.org/$15654541203011NAckT:cervoi.se?via=matrix.org&via=privacytools.io&via=feneas.org

BillCarsonFr:
Fixes #485

Display a unread indicator (green left indent) on rooms with unread messages.
Therefore rooms with unread messages that are not notifications are now showing the indicator, showing that there are some new messages here (even if not important enough to trigger a notification)

image

-> It's a bit strange on 'home' tab because all list items are showing it (maybe not showing it here?)

Implem is done by passing the readService to the summary mapper, but it might be better to compute & persist on update like for notification count? instead of live computing?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@ Features:
  -
 
 Improvements:
- -
+ - Add unread indent on room list (#485)
 
 Other changes:
  -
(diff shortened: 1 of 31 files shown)


bmarty:
LGTM, thanks, but yes, maybe wait for @ganfra's feedback about passing the readService to the summary mapper.

PR #538, opened 2019-09-02
Reduce release build log level

BillCarsonFr:
Enhance log management.
- Timber Debug tree only planted in Debug build
- Reduced release log level (Waring+), can be extended with lab settings

### DIFF
--- a/vector/src/main/res/values/strings_riotX.xml
+++ b/vector/src/main/res/values/strings_riotX.xml
@@ -2,5 +2,6 @@
 <resources>
 
     <!-- Strings not defined in Riot -->
+    <string name="labs_allow_extended_logging">Help developers by allowing extended logging.</string>
 
 </resources>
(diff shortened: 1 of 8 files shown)


vector/src/main/res/values/strings_riotX.xml
```diff
@@ -2,5 +2,6 @@
 <resources>
 
     <!-- Strings not defined in Riot -->
+    <string name="labs_allow_extended_logging">Help developers by allowing extended logging.</string>
```
bmarty: In the summary, it will be better to explicitly inform the user that message content or any other private data will not be logged, even if the is set to true. So I propose:
"Help developers by allowing extended logging, when you send a Rageshake."
"Even when set to true, the application does not log message contents or other private data."

And move this settings next to the rageshake setting?

PR #552, opened 2019-09-11
Save draft of a message when exiting a room with non empty composer (#329)

### LINKED ISSUE #329: Persist text in composer when leaving room screen
Currently, the draft is lost.

> We can also ask the user what he wants to do:
> - do not quit the screen
> - save the draft
> - discard the draft

After internal discussion and @violoncelloCH comment, we just save the draft

> - also display a toast to notify/warn that the message was not sent

this idea is also rejected

The related message in case of quoting/replying/editing should be saved to

This may be managed by the SDK, maybe using the local echo with a new sentState DRAFT, or on a completely separated db table

bmarty:
Fixes #329 

Feature: save the composer content as a draft, for each room.

The implementation is indeed a persistence of the SendMode in DB, with some specific rules. A new DraftService is added to the SDK.

Draft are saved each time the composer text is updating (to be robust to random crash), with a 100ms debouncing.

When editing an event, a new draft is created, which means we can start editing multiple event without loosing anything. When editing, close the edit composer cancel the draft and the previous draft is restored.

Reply/Quote/Regular move shares the same draft, which means user can start typing text, then choose to quote another message or reply to another message with keeping the text entered in the composer.

Also, the room list displays a badge for rooms containing some draft (CC @nadonomy):

image

Known limitation:
- pills are not restored. But they are not send too, so we will handle this later on.

Possible feature:
- [X] The catchup screen should display rooms with draft

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/send/DraftService.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/send/DraftService.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2019 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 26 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/send/DraftService.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun saveDraft(draft: UserDraft)
+
+    /**
+     * Delete the last draft, basically just after sending the message
+     */
+    fun deleteDraft()
+
+    /**
+     * Return the current drafts if any, as a live data
+     * The draft list can contain one draft for {regular, reply, quote} and an arbitrary number of {edit} drafts
```
bmarty: comment is not true anymore

PR #560, opened 2019-09-17
Display a "No network" banner when the device has no network

### LINKED ISSUE #559: Infinite sync progress indicator when there is no network
When the network is not available (airplane mode, or other coverage reason) , the progress should be replaced by a banner indicating that there is no network.

Maybe using the `SyncState` ?

bmarty:
In Room list and in room detail screen

Fixes #559 

image


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/sync/job/SyncThread.kt
```diff
@@ -97,6 +97,8 @@ internal class SyncThread @Inject constructor(private val syncTask: SyncTask,
 
             if (!networkConnectivityChecker.isConnected() || state == SyncState.PAUSED) {
```
ganfra: Be careful, you should only go for it when no connection, not when paused

Issue #238, opened 2019-06-27
m.notice messages trigger push notifications


turt2live:
Suddenly I'm getting notifications for #matrix-spam:matrix.org and #riot-spam:matrix.org only on RiotX. My push rules should be preventing this.

richvdh:
I'm suffering this too.

bmarty:
Seems that we finally find the root cause of this (https://github.com/vector-im/riotX-android/commit/a4af6c1c99c0b9182fc4d906941c8338d796b41a)

bmarty:
Fixed in #573

PR #573, opened 2019-09-20
Clean up push rules management and fixes several issues

bmarty:
This PR fixes several issues:
- Clear cache was losing the push rules. Push rules are now retrieved from the sync, so are always up to date
- (Should fix) m.notice trigger notification (#238)
- simplify db entity by removing useless userId field
- Fix TU compilation error

It also introduce @UserId, to inject the userId as a String

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/algorithms/megolm/MXMegolmDecryptionFactory.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/algorithms/megolm/MXMegolmDecryptionFactory.kt
@@ -16,7 +16,6 @@
 
 package im.vector.matrix.android.internal.crypto.algorithms.megolm
 
-import im.vector.matrix.android.api.auth.data.Credentials
 import im.vector.matrix.android.internal.crypto.DeviceListManager
 import im.vector.matrix.android.internal.crypto.MXOlmDevice
 import im.vector.matrix.android.internal.crypto.OutgoingRoomKeyRequestManager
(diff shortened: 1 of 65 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/sync/SyncResponseHandler.kt
```diff
@@ -33,7 +33,7 @@ internal class SyncResponseHandler @Inject constructor(private val roomSyncHandl
                                                        private val cryptoService: DefaultCryptoService,
                                                        private val initialSyncProgressService: DefaultInitialSyncProgressService) {
 
-    fun handleResponse(syncResponse: SyncResponse, fromToken: String?, isCatchingUp: Boolean): Try<SyncResponse> {
+    suspend fun handleResponse(syncResponse: SyncResponse, fromToken: String?, isCatchingUp: Boolean): Try<SyncResponse> {
```
ganfra: If it becomes suspendable, we should remove Try

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/algorithms/megolm/MXMegolmDecryptionFactory.kt
```diff
@@ -16,18 +16,19 @@
...
 import im.vector.matrix.android.internal.crypto.actions.EnsureOlmSessionsForDevicesAction
 import im.vector.matrix.android.internal.crypto.actions.MessageEncrypter
 import im.vector.matrix.android.internal.crypto.store.IMXCryptoStore
 import im.vector.matrix.android.internal.crypto.tasks.SendToDeviceTask
+import im.vector.matrix.android.internal.di.UserId
 import im.vector.matrix.android.internal.util.MatrixCoroutineDispatchers
 import javax.inject.Inject
 
-internal class MXMegolmDecryptionFactory @Inject constructor(private val credentials: Credentials,
+internal class MXMegolmDecryptionFactory @Inject constructor(@UserId
```
ganfra: I prefer having annotations on same line for constructor properties...

bmarty: done

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/sync/SyncResponseHandler.kt
bmarty: ok, done. Also removing the return value, which is useless

PR #591, opened 2019-09-26
Fix issue with image orientation

### LINKED ISSUE #517: Picture uploads are unreliable, pictures are shown in wrong aspect ratio on desktop client
Every time I send a picture on RiotX, I hold my breath. Often times they time out, and then messages and pictures are sent out of order. Yesterday, a picture upload was marked "canceled" (with no option for me to do anything about it) and then started flitting between "timed out" and continuing to upload. It finally succeeded, but is a stretched, blurry mess when I view it from the web client.

bmarty:
Attempt to fix #517

### DIFF
--- a/vector/src/main/java/im/vector/riotx/core/di/VectorComponent.kt
+++ b/vector/src/main/java/im/vector/riotx/core/di/VectorComponent.kt
@@ -26,7 +26,6 @@ import im.vector.matrix.android.api.session.Session
 import im.vector.riotx.EmojiCompatFontProvider
 import im.vector.riotx.EmojiCompatWrapper
 import im.vector.riotx.VectorApplication
-import im.vector.riotx.core.images.ImageTools
 import im.vector.riotx.core.pushers.PushersManager
 import im.vector.riotx.core.utils.DimensionConverter
 import im.vector.riotx.features.configuration.VectorConfiguration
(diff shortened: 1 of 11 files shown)


vector/src/main/java/im/vector/riotx/core/images/ImageTools.kt
```diff
@@ -0,0 +1,76 @@
...
+import android.database.Cursor
+import android.net.Uri
+import android.provider.MediaStore
+import androidx.exifinterface.media.ExifInterface
+import timber.log.Timber
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class ImageTools @Inject constructor(private val context: Context) {
```
ganfra: No need to have scope @Singleton: there is no state and no heavy init

vector/src/main/java/im/vector/riotx/core/di/VectorComponent.kt
```diff
@@ -67,6 +68,8 @@ interface VectorComponent {
 
     fun dimensionConverter(): DimensionConverter
 
+    fun imageTools(): ImageTools
```
ganfra: As you remove scope, you won't need to be exposed here neither

vector/src/main/java/im/vector/riotx/core/images/ImageTools.kt
bmarty: ok, done

PR #596, opened 2019-09-28
more a11y tweaks

pvagner:
I'm using RiotX with a screen reader for a few months and I'm tweaking it a bit to make it more accessible to screen reader users.
I have noticed you have added some accessibility related improvements recently too. In fact I did some things very similar to you.
By knowing this I have rebased my stuff over your recent accessibility related work and here is the result.
The big features for screen reader users are better accessibility of read receipts and access to reactions emoji chooser.

### DIFF
--- a/vector/src/main/res/layout/fragment_login.xml
+++ b/vector/src/main/res/layout/fragment_login.xml
@@ -87,6 +87,7 @@
                     android:layout_marginTop="8dp"
                     android:background="?attr/selectableItemBackground"
                     android:scaleType="center"
+                    android:contentDescription="@string/a11y_show_password"
                     android:src="@drawable/ic_eye_black"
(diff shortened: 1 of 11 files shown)


vector/src/main/res/layout/fragment_login.xml
```diff
@@ -87,6 +87,7 @@
                     android:layout_marginTop="8dp"
                     android:background="?attr/selectableItemBackground"
                     android:scaleType="center"
+                    android:contentDescription="@string/a11y_show_password"
```
BillCarsonFr: for attributes that are changed in code we usually put a tools: instead of android:

bmarty: I do it

PR #618, opened 2019-10-11
Mark all as read

### LINKED ISSUE #396: "Mark all as read" option
All rooms marked as read in one click.

On Riot Android, this is an option in the top right menu in the home

bmarty:
Mark all room of the current screen as read (Fixes #396)
Also fixes test compilation issue


vector/src/main/java/im/vector/riotx/features/home/room/list/RoomListFragment.kt
```diff
@@ -78,6 +79,19 @@ class RoomListFragment : VectorBaseFragment(), RoomSummaryController.Listener, O
         injector.inject(this)
     }
 
+    override fun getMenuRes() = R.menu.room_list
```
ganfra: I think it should be handled in the host activity, as we use RoomList in different contexts 
(for handling incoming share as exemple)

bmarty: I've changed the way it's done. Please review changes

PR #624, opened 2019-10-12
Make tasks responsible for dispatching.

Dominaezzz:

### Pull Request Checklist

In this PR I have made `Task`s responsible for which dispatcher they run in.
In terms of coroutines, it doesn't make sense for caller to dictate where a task should execute, the task should know it needs to place blocking IO in the IO dispatcher, crypto in the crypto dispatcher, etc.
In some cases, tasks might need to use mutliple dispatchers depending on what it is doing, in which case, `executionThread` stops making sense.

Somewhat related:
I also noticed that `Dispatchers.IO` (dynamic 64 threads) is used for everything instead of `Dispatchers.Default` (fixed n threads, n == number of cores). Doing this might be less efficient, since threads won't be recycled, app becomes more resource heavy.

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/OutgoingRoomKeyRequestManager.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/OutgoingRoomKeyRequestManager.kt
@@ -300,7 +300,6 @@ internal class OutgoingRoomKeyRequestManager @Inject constructor(
                 .configureWith(SendToDeviceTask.Params(EventType.ROOM_KEY_REQUEST, contentMap, transactionId)) {
                     this.callback = callback
                     this.callbackThread = TaskThread.CALLER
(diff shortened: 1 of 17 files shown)


ganfra:
Thanks for you work. You are talking about using Default but in your PR you set computation as the "default" one?

Dominaezzz:
Yes I use computation as the default but computation is set to `Dispatchers.IO`.
Computation should point to `Dispatchers.Default` but I haven't made any changes regarding that in this PR because I wanted to know why it was used before changing it.

ganfra:
Actually, most of the tasks are doing network calls and db insertions, so it was more convenient to use IO by default. But as you said, each task should know which dispatcher to use, so shouldn't we totally remove it from TaskExecutor?

Dominaezzz:
Good point. I'll just remove it.

Dominaezzz:
I've rearranged to make sure the callback dispatcher is not as "busy". Also removed dispatcher.

PR #629, opened 2019-10-22
Fix / event mapper persist the clear type in type

BillCarsonFr:
Fix bug when some messages in encrypted rooms appears as malformed events
(rageshake 6356)

The issue is that the event mapper persists the clear type as the type in DB, leading to an event with a clear type m.room.message but with an encrypted content.
Not sure what lead to this case as the sync handler does not decrypt before persisting, but it might be possible that the in memory event is modified by another 'task' before the persisting occurs?(Event is not immutable)


ganfra:
That's why event should be immutable...

PR #639, opened 2019-10-24
Build and run test on CI

bmarty:
WIP, first build should fail


bmarty:
If CI is happy, so do we!

PR #651, opened 2019-10-29
Markdown off

### LINKED ISSUE #412: markdown set off by default
Markdown is good. I like markdown. I use markdown. But many users do not understand his. They want send `1*2*3` but it broken. They try send samba url `\\host\share` - and it broken... Admins must describe to users how they can disable this feature.

I think by default we need switch off markdown for all new users. Users, which know what  is this "markdown" - will enable it in settings.

bmarty:
Small cleanup and disable markdown by default
Fixes #412

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ Improvements 🙌:
  -
 
 Other changes:
- -
+ - Markdown set to off by default (#412)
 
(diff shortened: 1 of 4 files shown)


bmarty:
Yes, why not, but power users will enable md I guess

PR #667, opened 2019-11-05
Import change form https://github.com/matrix-org/matrix-android-sdk/pull/505

bmarty:


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/store/db/Helper.kt
```diff
@@ -51,8 +50,7 @@ fun <T : RealmObject> doRealmQueryAndCopy(realmConfiguration: RealmConfiguration
  */
 fun <T : RealmObject> doRealmQueryAndCopyList(realmConfiguration: RealmConfiguration, action: (Realm) -> Iterable<T>): Iterable<T> {
     return Realm.getInstance(realmConfiguration).use { realm ->
-        val result = action.invoke(realm)
-        realm.copyFromRealm(result)
+        action.invoke(realm).let { realm.copyFromRealm(it) }
```
ganfra: I find it less elegant...

bmarty: Yes, I wanted to keep the reading from left to right, because I could have inline the parameter of copyFromRealm...

PR #689, opened 2019-11-15
Login and Registration

bmarty:
Remaining work:
- [x] Button style (wired)
- [x] Registration screens
- [ ] Test
- [x] Password reset request
- [ ] Remaining TODO
- [x] Doc for signin password
- [x] Doc for sso

Known issues so far:
- [x] Login SSO: homeserver provided in wellknown credential is not used (See TODO)
- [x] Mandatory dummy action is not handled (connect on synapse demo)
- [x] Do Not Keep Activity: crash when BG-FG the app. Developer error, persist the data in the viewModel
- [x] Create a User at initial sync, to avoid empty drawer top-left for accounts without any rooms (which is the case for new account (9d0188cbf1c2cc173306893a5ca18180cf64ce68)
- [x] Check the homeserver CS version when user enter it. RiotX support only this minimal conf:
r0.5.0, "m.require_identity_server": false
image

- [ ] Add animation on FragmentTransition
- [x] Crash: Signin/Other/Back


vector/src/main/java/im/vector/riotx/features/login/LoginActivity.kt
```diff
@@ -18,64 +18,202 @@ package im.vector.riotx.features.login
...
+    }
 
-        loginViewModel.selectSubscribe(this, LoginViewState::asyncLoginAction) {
-            if (it is Success) {
-                val intent = HomeActivity.newIntent(this)
-                startActivity(intent)
-                finish()
+    private fun handleRegistrationNavigation(flowResult: FlowResult) {
+        // Complete all mandatory stage first
+        val mandatoryStages = flowResult.missingStages.filter { it.mandatory }
```
Dominaezzz: Looks like you only need `firstOrNull` here.

bmarty: Yes, indeed :)

bmarty:
I think the PR can start to be reviewed and more important the code is ready to be tested

Issue #201, opened 2019-06-20
Handle navigation to room via room alias


bmarty:
Request: https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-directory-room-roomalias

bmarty:
To check: navigation on room alias does not work.
The app should be able to handle normal permalink, like 
`<a href=\"https://matrix.to/#/#riotx:matrix.org\">#riotx:matrix.org</a>`
or convert # text, like
`#riotx:matrix.org`

Also pill support would be the cherry on the cake

bmarty:
Done in #786

PR #706, opened 2019-11-25
Feature/handle matrix to

ganfra:
This PR adds support for opening matrix.to link from outside the app (#57) and map room alias to roomId for joined room (#201 )


matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/RoomService.kt
```diff
@@ -58,4 +59,9 @@ interface RoomService {
      * Mark all rooms as read
      */
     fun markAllAsRead(roomIds: List<String>, callback: MatrixCallback<Unit>): Cancelable
+
+    /**
+     * Resolve a room alias to a room ID.
+     */
+    fun getRoomIdByAlias(roomAlias: String, searchOnServer: Boolean, callback: MatrixCallback<Optional<String>>): Cancelable
```
bmarty: Why using a Optional here?

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/query/RoomSummaryEntityQueries.kt
```diff
@@ -31,9 +31,21 @@ internal fun RoomSummaryEntity.Companion.where(realm: Realm, roomId: String? = n
     return query
 }
 
+internal fun RoomSummaryEntity.Companion.findByAlias(realm: Realm, roomAlias: String): RoomSummaryEntity? {
+    val roomSummary = realm.where<RoomSummaryEntity>()
+            .equalTo(RoomSummaryEntityFields.CANONICAL_ALIAS, roomAlias)
+            .findFirst()
```
bmarty: I think this is not necessary because the canonical alias is also in the alias list, no?

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/alias/RoomAliasDescription.kt
```diff
@@ -14,25 +14,20 @@
...
 
-import im.vector.matrix.android.api.MatrixCallback
-import im.vector.matrix.android.api.util.Cancelable
-import io.reactivex.SingleEmitter
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
 
-internal class MatrixCallbackSingle<T>(private val singleEmitter: SingleEmitter<T>) : MatrixCallback<T> {
+@JsonClass(generateAdapter = true)
+data class RoomAliasDescription(
```
bmarty: internal?

vector/src/main/AndroidManifest.xml
```diff
@@ -63,7 +63,13 @@
         <activity android:name=".features.roomdirectory.RoomDirectoryActivity" />
         <activity android:name=".features.roomdirectory.roompreview.RoomPreviewActivity" />
         <activity android:name=".features.home.room.filtered.FilteredRoomsActivity" />
-        <activity android:name=".features.home.room.detail.RoomDetailActivity" />
+        <activity
+            android:name=".features.home.room.detail.RoomDetailActivity"
+            android:parentActivityName=".features.home.HomeActivity">
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value=".features.home.HomeActivity" />
```
bmarty: Is it ok If we open the RoomDetailActivity from a notification? Today we already go back to the Home. We should check that.

vector/src/main/res/values/strings.xml
```diff
@@ -1768,5 +1768,6 @@ Not all features in Riot are implemented in RiotX yet. Main missing (and coming
     <string name="permissions_rationale_msg_keys_backup_export">Riot needs permission to save your E2E keys on disk.\n\nPlease allow access on the next pop-up to be able to export your keys manually.</string>
 
     <string name="no_network_indicator">There is no network connection right now</string>
+    <string name="permalink_malformed">Your matrix.to link was malformed</string>
```
bmarty: -> strings_RiotX.xml

bmarty:
Also the case correct room but unknown event does not work, we get an infinite loading

Ex: https://matrix.to/#/!QtykxKocfZaZOUrTwp:matrix.org/$INVALIDID:matrix.org?via=matrix.org

Issue #428, opened 2019-07-24
Add to f-droid


DamienCassou:
Hi,

I would like to try riotX but don't have Play Store or even Google services. Would it be possible to have riotX on f-droid.org?

krombel:
While RiotX is not part of the repository on f-droid.org I can suggest my F-Droid repos which contain the current development builds.
There are two flavors: One which is based on using Google Services but one as well which does not rely on those and would therefore be compatible to f-droid.org.
You might checkout https://fdroid.krombel.de for the links to the flavor you want to try.

DamienCassou:
> While RiotX is not part of the repository on f-droid.org I can suggest my F-Droid repos which contain the current development builds.

Would it be possible to get your repository mentioned in RiotX README file? That way, people know they can trust it.

Bubu:
RfP is open here: https://gitlab.com/fdroid/rfp/issues/1035

Right now this is pulling libraries from two `unkown maven repos` [here](https://github.com/vector-im/riotX-android/blob/develop/build.gradle#L42) and [here](https://github.com/vector-im/riotX-android/blob/develop/build.gradle#L51)

For https://repo.adobe.com/nexus/content/repositories/public/ I have no idea where I'd find the source for that. For the piasy stuff that shoudl really just be published to jcenter instead it's own repo, there's a closed issue about this here: https://github.com/Piasy/BigImageViewer/issues/113

neVERberleRfellerER:
Link to source of diff_match_patch is in pom (https://repo.adobe.com/nexus/content/repositories/public/diff_match_patch/diff_match_patch/current/diff_match_patch-current.pom). But there is a catch: it links to Google Code (http://code.google.com/p/google-diff-match-patch) which redirects to github and import squashed everything into one commit. This is quite problematic, because jar was published in 2013 and it is not possible to get the original source from 2013. Waybackmachine works and shows commit history (https://web.archive.org/web/20160531121855/http://code.google.com:80/p/google-diff-match-patch/source/list) Using commits in my favourite web search engine, I found this:
https://github.com/GerHobbelt/google-diff-match-patch/blob/master/java/name/fraser/neil/plaintext/diff_match_patch.java
which ends with last commit relevant to Java from original Google Code repo. But I guess that this is too weak to be useful.

bmarty:
Waiting for an official F-Droid version of RiotX, feel free to use @krombel 's repositories: https://fdroid.krombel.de/, which provide a nightly build gracefully

[...]

PR #760, opened 2019-12-11
Use diff_match_patch sources as dependency

BillCarsonFr:
Include diff-match-patch sources as a dependency and build from source for fdroid 

Part of #428


bmarty:
@BillCarsonFr as @Bubu says, we have to handle https://github.com/vector-im/riotX-android/blob/develop/build.gradle#L42 to fix #428

PR #792, opened 2019-12-30
Feature/stabilization

ganfra:
Fix some issues and update libs


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/helper/ChunkEntityHelper.kt
```diff
@@ -129,12 +92,15 @@ internal fun ChunkEntity.add(roomId: String,
         }
     }
 
+    val isUnlinked = isUnlinked
```
BillCarsonFr: ?

matrix-sdk-android/build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
         jcenter()
     }
     dependencies {
```
BillCarsonFr: More lib can be updated here 
androidx.recyclerview:recyclerview -> 1.2.0-xx ? (may fix some reported RS?)

PR #806, opened 2020-01-05
call /join/{roomIdOrAlias} instead of /rooms/{roomId}/join

### LINKED ISSUE #697: No Known Servers error on trying to join upgraded room
cc @turt2live

During the mass matrix community room upgrade, I attempted to click on a couple tombstone links but hit No Known Servers on every one.

I've sent a rageshake for my attempt on Synapse Announcements.

Bubu:
The former endpoint doesn't work for joining over federation, the
server_name parameter is ignored.

Fixes #697

### Pull Request Checklist


anoadragon453:
The original endpoint ([`/rooms/{roomId}/join`](https://matrix.org/docs/spec/client_server/unstable#post-matrix-client-r0-rooms-roomid-join)) does indeed not take a `server_name` parameter, while the proposed replacement endpoint ([`/rooms/{roomId}/join`](https://matrix.org/docs/spec/client_server/unstable#post-matrix-client-r0-rooms-roomid-join)) does.

Assuming that the code correctly grabs the server_name of the sender of the tombstone when following the link in an upgraded room, then this code should fix the linked issue.

bmarty:
> The original endpoint ([`/rooms/{roomId}/join`](https://matrix.org/docs/spec/client_server/unstable#post-matrix-client-r0-rooms-roomid-join)) does indeed not take a `server_name` parameter, while the proposed replacement endpoint ([`/rooms/{roomId}/join`](https://matrix.org/docs/spec/client_server/unstable#post-matrix-client-r0-rooms-roomid-join)) does.
> 
> Assuming that the code correctly grabs the server_name of the sender of the tombstone when following the link in an upgraded room, then this code should fix the linked issue.

This is not the case.
So I think this PR is not enough to fix #697.
Also some names has to be updated because roomId can now be a room Alias (for instance JoinRoomTask.Params.roomId)

Bubu:
> So I think this PR is not enough to fix #697.

I think that part is working. it definitely does fix #697. (At least in all test cases I could come up with.) The server_name is populated at that point.

>  Also some names has to be updated because roomId can now be a room Alias (for instance JoinRoomTask.Params.roomId)

Agreed. I think we also get a problem here then: https://github.com/vector-im/riotX-android/blob/develop/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/membership/joining/JoinRoomTask.kt#L49

Because the following code probably assumes we do have a roomID and not an alias.

I don't think it's currently ever called with an alias, but we do need to call it with aliases to actually join rooms from elsewhere. See #808 

@bmarty Do you have advice on how to fix this properly that I could take a stab at? Or do you want to just take it yourself?

bmarty:
Thanks for the update.
Let's keep `roomId` for the moment, we will improve this later.

PR #818, opened 2020-01-08
Exclude play-services-oss-licenses library from F-Droid build (#814)

### LINKED ISSUE #814: play-services-oss-licenses is a proproetary library
This dependency is as far as I can determine non-free:
`implementation 'com.google.android.gms:play-services-oss-licenses:17.0.0'`

The import path is com.google.android.gms.* which is a pretty good sign, that this part of the proprietary play services,

So at least for the f-droid build this cannot be included.

Ref #428

bmarty:
Fixes #814 

@Bubu please let me know if it's ok for you.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,7 @@ Improvements 🙌:
 
 Other changes:
  - Change the way RiotX identifies a session to allow the SDK to support several sessions with the same user (#800)
+ - Exclude play-services-oss-licenses library from F-Droid build (#814)
 
 Bugfix 🐛:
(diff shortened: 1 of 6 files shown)


Bubu:
Looks good from in that no proprietary code is included in the f-droid build.

What I can't comment on is if that third party license notice is required to be present in the app and whether this will create a (legal) problem if that's missing from the f-droid build.

bmarty:
Actually there are 2 third party licences screens in the app. One we manage manually and this one which I remove from f-Droid build. So I think it's ok

PR #829, opened 2020-01-10
Feature/stab

### LINKED ISSUE #827: version code scheme broken
A non-universal build of an earlier version should never have a higher VC than a universal build for a newer version.

When trying to upgrade from riotX-0.12-dev armv8 to riotX-0.13-dev universal I was confronted with a downgrade from 177798783 to 157858035. This should really not be a downgrade.

bmarty:
Fixes #827 

This PR contains also some other little fixes


vector/build.gradle
```diff
@@ -24,12 +24,16 @@ static def getGitTimestamp() {
...
-    return (getGitTimestamp() / 10).toInteger()
+    // It's unix timestamp, minus timestamp of October 3rd 2018 (first commit date) divided by 100: It's incremented by one every 100 seconds.
+    // plus 20_000_000 for compatibility reason with the previous timestamp
+    // Note that the result will be multiplied by 10 when adding the digit for the arch
+    return ((getGitTimestamp() - 1_538_524_800 ) / 100).toInteger() + 20_000_000
 }
 
 def generateVersionCodeFromVersionName() {
-    return versionMajor * 1_00_00 + versionMinor * 1_00 + versionPatch
+    // plus 4_000_000 for compatibility reason with the previous timestamp
```
Bubu: There's no previous timestamp in this case.

bmarty: I'm thinking about @krombel repository

Bubu: The logic is correct, I think, just the comment should not reference a timestamp for the release build but just compatibility with the previous version code generation? Or am I missing something here?

bmarty: yes, you're right, I update the comment

PR #892, opened 2020-01-23
Import Indonesian resource check from TravisCI to Weblate pipeline

bmarty:
Note that for this time, Buildkite will complain, but the fix has already be done in Weblate, so the step will be ok after the next Weblate sync


bmarty:
It should be all green now

PR #916, opened 2020-01-29
Negotiate E2E by default for DMs (#907)

### LINKED ISSUE #907: Negotiate E2E by default for DMs
From riot-web created by [jryans](https://github.com/jryans): vector-im/riot-web#12005  

When the cross-signing labs flag is on, we want to create DMs as E2E by default. There are no UX changes or toggles in the create DM flow for this. Since not all users support E2E yet (bots, bridges, people with only non-E2E clients), we need to be a bit careful.

When creating the DM room, check if the other user has shared device keys. If yes, then create an encrypted room, otherwise fallback to unencrypted.

bmarty:
Closes #907: Tested OK with regular user (so e2e on by default) and @riot-bot:matrix.org (so e2e off by default)

Also:
- Convert VerificationTxState to a sealed class
- Make CreateRoomParams a regular data class
- Cleanup

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/create/CreateRoomTask.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/create/CreateRoomTask.kt
@@ -17,9 +17,12 @@
 package im.vector.matrix.android.internal.session.room.create
 
 import com.zhuinden.monarchy.Monarchy
+import im.vector.matrix.android.api.extensions.orTrue
+import im.vector.matrix.android.api.session.crypto.crosssigning.CrossSigningService
 import im.vector.matrix.android.api.session.room.failure.CreateRoomFailure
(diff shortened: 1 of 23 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/create/CreateRoomTask.kt
```diff
@@ -49,12 +52,41 @@ internal class DefaultCreateRoomTask @Inject constructor(
         private val readMarkersTask: SetReadMarkersTask,
         @SessionDatabase
         private val realmConfiguration: RealmConfiguration,
+        private val crossSigningService: CrossSigningService,
+        private val deviceListManager: DeviceListManager,
         private val eventBus: EventBus
 ) : CreateRoomTask {
 
     override suspend fun execute(params: CreateRoomParams): String {
+        val createRoomParams = params
```
BillCarsonFr: not sure to understand that

PR #948, opened 2020-02-03
Feature/artifacts downloading

bmarty:
This PR add a tool to the project to download buildkite artifacts, in order to speed up the release of RiotX. It does not include change on the RiotX code.


tools/release/download_buildkite_artifacts.py
```diff
@@ -0,0 +1,119 @@
+#
```
richvdh: it's generally best to start python scrpts with a `#!` line:

```suggestion
#!/usr/bin/env python
#
```

tools/release/download_buildkite_artifacts.py
```diff
@@ -0,0 +1,119 @@
...
+                    action="store_true")
+parser.add_argument('-s',
+                    '--simulate',
+                    help="simulate action, do not create folder or download any file.",
+                    action="store_true")
+
+args = parser.parse_args()
+
+# parser has checked that the build was an int, convert to String for the rest of the script
+buildStr = str(args.build)
```
richvdh: python convention is to use underscores: `build_str` etc

tools/release/download_buildkite_artifacts.py
```diff
@@ -0,0 +1,119 @@
...
+ORG_SLUG = "matrix-dot-org"
+PIPELINE_SLUG = "riotx-android"
+
+### Arguments
+
+parser = argparse.ArgumentParser(description='Download artifacts from Buildkite.')
+parser.add_argument('-t',
+                    '--token',
+                    help='The buildkite token.')
+parser.add_argument('-b',
```
richvdh: you might want to set `required=True` if this is required.

PR #955, opened 2020-02-04
Feature/room history in e2e

### LINKED ISSUE #803: Messages sent when user is invited are not decryptable when the invited user joins the room
Creating a room that is 
- Encrypted
- History visibility set to on invite   

Users can not decrypt messages sent before they joined the room

image

bmarty:
Fix #803


matrix-sdk-android/src/main/java/im/vector/matrix/android/api/Matrix.kt
```diff
@@ -62,6 +66,7 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
             WorkManager.initialize(context, Configuration.Builder().build())
         }
         ProcessLifecycleOwner.get().lifecycle.addObserver(backgroundDetectionObserver)
+        matrixConfigurationHolder.matrixConfiguration = matrixConfiguration.copy()
```
ganfra: No need for that. You should inject your configuration in the DI tree like we do with the Context. Look at the MatrixComponent

bmarty: Done

PR #976, opened 2020-02-07
Use View events

bmarty:
This is a sanity rework.
No effective change on the UI/UX

A few questions for @BillCarsonFr as TODOs in the code :)


vector/src/main/java/im/vector/riotx/features/crypto/verification/VerificationBottomSheetViewEvents.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package im.vector.riotx.features.crypto.verification
+
+import im.vector.riotx.core.platform.VectorViewEvents
+
+/**
+ * Transient events for the verification bottom sheet
+ */
+sealed class VerificationBottomSheetViewEvents : VectorViewEvents {
+    // data class Loading(val message: CharSequence? = null) : RoomDirectoryViewEvents()
```
ganfra: I don't think we should let these lines... at least let them inherits the right class

vector/src/main/java/im/vector/riotx/features/settings/devices/VectorSettingsDevicesFragment.kt
```diff
@@ -64,35 +62,19 @@ class VectorSettingsDevicesFragment @Inject constructor(
...
-                                        transactionId = action.transactionId!!
-                                ).show(childFragmentManager, "REQPOP")
-                            }
-                        }
-                    }
+                is DevicesViewEvents.Loading            -> showLoading(it.message)
+                is DevicesViewEvents.Failure            -> showFailure(it.throwable)
+                is DevicesViewEvents.RequestPassword    -> maybeShowDeleteDeviceWithPasswordDialog()
+                is DevicesViewEvents.PromptRenameDevice -> displayDeviceRenameDialog(it.deviceInfo)
+                is DevicesViewEvents.ShowVerifyDevice   -> {
```
BillCarsonFr: For the view event, the device id is not needed, the transactionId is enough

vector/src/main/java/im/vector/riotx/features/crypto/verification/VerificationBottomSheetViewEvents.kt
bmarty: ok, removed

vector/src/main/java/im/vector/riotx/features/settings/devices/VectorSettingsDevicesFragment.kt
bmarty: ok, thanks

PR #977, opened 2020-02-10
Show confirmation dialog before deleting a message

onurays:
Screen Shot 2020-02-10 at 22 35 24


vector/src/main/java/im/vector/riotx/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -1108,7 +1108,14 @@ class RoomDetailFragment @Inject constructor(
                 showSnackWithMessage(getString(R.string.copied_to_clipboard), Snackbar.LENGTH_SHORT)
             }
             is EventSharedAction.Delete                     -> {
-                roomDetailViewModel.handle(RoomDetailAction.RedactAction(action.eventId, context?.getString(R.string.event_redacted_by_user_reason)))
+                val layout = requireActivity().layoutInflater.inflate(R.layout.dialog_delete_event, null)
+                AlertDialog.Builder(requireActivity())
+                        .setView(layout)
```
bmarty: Please add a title "Delete message"

vector/src/main/java/im/vector/riotx/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -1108,7 +1108,14 @@ class RoomDetailFragment @Inject constructor(
...
+                        .setPositiveButton(R.string.delete) { _, _ ->
+                            roomDetailViewModel.handle(RoomDetailAction.RedactAction(action.eventId, context?.getString(R.string.event_redacted_by_user_reason)))
```
bmarty: User are complaining that the user language is leaked because of the reason we pass here.
It was not specified in the issue, but we can take the opportunity of this PR to add:
- a checkbox "include a reason", checked by default,
- an TextInputLayout with the reason in the dialog, hint "reason for redacting", pre-filled with R.string.event_redacted_by_user_reason
If the check box is unchecked or the reason is empty, we set null for the reason

See R.layout.dialog_report_content for a similar dialog

vector/src/main/java/im/vector/riotx/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -1108,7 +1108,14 @@ class RoomDetailFragment @Inject constructor(
                 showSnackWithMessage(getString(R.string.copied_to_clipboard), Snackbar.LENGTH_SHORT)
             }
             is EventSharedAction.Delete                     -> {
-                roomDetailViewModel.handle(RoomDetailAction.RedactAction(action.eventId, context?.getString(R.string.event_redacted_by_user_reason)))
+                val layout = requireActivity().layoutInflater.inflate(R.layout.dialog_delete_event, null)
```
bmarty: Maybe add a `private fun` to handle this action and avoid having too big methods

bmarty: Normally there is no need to create a layout for a simple dialog message. Just use setMessage of the AlertDialog.Builder. But for this case, we will improve the dialog, see below.

vector/src/main/java/im/vector/riotx/features/home/room/detail/RoomDetailFragment.kt
onurays: The title and content of the dialog are set as the Riot Web.

vector/src/main/java/im/vector/riotx/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -793,6 +794,24 @@ class RoomDetailFragment @Inject constructor(
...
+        val reasonCheckBox = layout.findViewById<MaterialCheckBox>(R.id.deleteEventReasonCheck)
+        val reasonInput = layout.findViewById<TextInputEditText>(R.id.deleteEventReasonInput)
+
+        reasonCheckBox.setOnCheckedChangeListener { _, isChecked -> reasonInput.isEnabled = isChecked }
+
+        AlertDialog.Builder(requireActivity())
+                .setView(layout)
+                .setTitle(R.string.delete_event_dialog_title)
+                .setPositiveButton(R.string.remove) { _, _ ->
+                    val reason = if (reasonCheckBox.isChecked) reasonInput.text.toString().takeIf { it.isNotBlank() } else null
```
bmarty: Would be nicer to use two `takeIf()` statements:
```
reasonInput.text.toString()
    .takeIf { reasonCheckBox.isChecked }
    .takeIf { it.isNotBlank() }
```

onurays: This is brilliant suggestion.

PR #1010, opened 2020-02-13
Attachment process

bmarty:
This PR improve the attachment process. Ganfra started it and I finished it.

This is now possible to:
- Send multiple attachment to several rooms. If both encrypted and clear rooms are selected, the media will be sent twice, once in clear and once encrypted, then events will be sent, also encrypted or in clear depending on the room.
- Compact image before sending them. A library has been used, the result are sometimes strange. Especially the dimensions are not respected. Let see if we can improve this later on.
- Preview image before sending them
- Crop/rotate image before sending them
- Delete image form the list before sending them

Note that other attachment types (not previewable attachments) are sent to the room(s) before the previewable types are edited. This is maybe a problem as the user may think that if he cancels the media edition nothing will be sent. The use case of selecting file and image are not common, so let's stay like that for the moment.

The PR review should include lots of tests, especially to ensure that we never send clear events or clear file in a e2e room


steef435:
Haven't really played with it yet, but first impressions: nice! Feels very fluid :).

Quickly ran into this: the screen whites out and aborts the sharing when I start sharing a picture from an image viewer app (so by intent) and then click on the edit image button. I've sent a rageshake. Will test more.

steef435:
Thoughts after playing it for a while (as an end user, I haven't seen the code):
1. Three hurrays for cropping functionality 🎉🎉🎉
2. When I initiate an image share from a picture viewer or a file manager (so by intent?), and then click on the edit image button, RiotX crashes. I've sent a rageshake yesterday.
3. All images and files I share are (still) stored in a RiotX data folder which is readily accessible by file managers and photo viewers. When sharing a document to an encrypted room, I see this as a bit of a security leak, since the files are now stored unencrypted in a folder I only know about because the files start showing up in a different app. Signal for instance only caches encrypted files IIUC.

Some UX things I thought about:
- When sending an image to a single room: first I get a confirmation dialog, and then I also get another fragment where I can choose to edit the picture or send. Maybe the confirmation dialog is redundant here?

- In general, I think it would  be nice if this "image overview" screen would show the recipients. Signal and WhatsApp both show it at the bottom of the similar screen.


 > Note that other attachment types (not previewable attachments) are sent to the room(s) before the previewable types are edited. This is maybe a problem as the user may think that if he cancels the media edition nothing will be sent. The use case of selecting file and image are not common, so let's stay like that for the moment.
- I couldn't trigger this using my file manager. I get a "couldn't handle share data" snack from the file manager. Not that I mind because I agree it's an uncommon use case.

- When selecting multiple rooms to send an attachment to, I think it would be nice if the list of selected rooms is shown on top. Just like the breadcrumbs when you are selecting participants for a new room. It gets hard to keep track of selected rooms otherwise.

- Really minor but might as well mention it: if you choose to share an image of the same dimensions as RiotX (e.g. a screenshot made on the phone), and then enter the "picture overview" fragment, the transparent toolbar buttons can become very unnoticeable. WhatsApp and Signal both use white icons with dark grey outline, making them more visible on dark and light backgrounds.

Long story, but the short of it is that I'm really happy images can be edited and are sent scaled/compressed and everything! 🎉

vector/src/main/java/im/vector/riotx/features/share/IncomingShareFragment.kt
```diff
@@ -0,0 +1,239 @@
...
+        }
+        super.onViewCreated(view, savedInstanceState)
+        setupRecyclerView()
+        setupToolbar(incomingShareToolbar)
+        attachmentsHelper = AttachmentsHelper.create(this, this).register()
+
+        val intent = vectorBaseActivity.intent
+        if (intent?.action == Intent.ACTION_SEND || intent?.action == Intent.ACTION_SEND_MULTIPLE) {
+            var isShareManaged = attachmentsHelper.handleShareIntent(
+                    IntentUtils.getPickerIntentForSharing(intent)
```
steef435: This seems to break the intent.resolveType() at handleShareIntent (AttachmentsHelper.kt), thus making it impossible to share multiple files by share intent. (Just passing the intent as-is allows for multiple files to be shared.) Why is this function called?

onurays:
https://github.com/vector-im/riotX-android/blob/88b8ceaeb5edbfd12a7ed7aacd2e1a321c8aab1e/vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/action/MessageActionsViewModel.kt#L285

"Share" option is not appearing in encrypted rooms for images. Because messageContent.url is null. Instead, should we check messageContent.encryptedFileInfo.url for encrypted rooms?

onurays:
Screenshot_2020-02-17-12-19-38-253_im vector riotx debug

Room_1 shares an image with Room_2. None of the rooms are encrypted. Room_2 always gets the error message:

**Upload request missing 'Content-Type'**

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/DefaultSendService.kt
```diff
@@ -277,18 +300,33 @@ internal class DefaultSendService @AssistedInject constructor(
...
         val uploadWorkData = WorkerParamsFactory.toData(uploadMediaWorkerParams)
 
         return workManagerProvider.matrixOneTimeWorkRequestBuilder<UploadContentWorker>()
                 .setConstraints(WorkManagerProvider.workConstraints)
                 .startChain(startChain)
                 .setInputData(uploadWorkData)
-                .setBackoffCriteria(BackoffPolicy.LINEAR, BACKOFF_DELAY, TimeUnit.MILLISECONDS)
+                .setBackoffCriteria(BackoffPolicy.LINEAR, WorkManagerProvider.BACKOFF_DELAY, TimeUnit.MILLISECONDS)
+                .build()
+    }
```
onurays: We may simplify this flow and reduce code duplication like:

```suggestion
    val cancelableBag = CancelableBag()
        allLocalEchoes
                .groupBy { cryptoService.isRoomEncrypted(it.roomId!!) }
                .apply {
                    keys.forEach { isRoomEncrypted ->
                        val uploadWork = createUploadMediaWork(get(isRoomEncrypted)!!, attachment, isRoomEncrypted, compressBeforeSending, startChain = true)
                        val dispatcherWork = createMultipleEventDispatcherWork(isRoomEncrypted)
                       
                        workManagerProvider.workManager
                                .beginUniqueWork(buildWorkName(UPLOAD_WORK), ExistingWorkPolicy.APPEND, uploadWork)
                                .then(dispatcherWork)
                                .enqueue()

                        cancelableBag.add(CancelableWork(workManagerProvider.workManager, dispatcherWork.id))
                    }
                }
```

vector/src/main/java/im/vector/riotx/features/share/IncomingShareFragment.kt
bmarty: Note sure about this code, but I fixed it be splitting logic for ACTION_SEND and ACTION_SEND_MULTIPLE

bmarty:
> https://github.com/vector-im/riotX-android/blob/88b8ceaeb5edbfd12a7ed7aacd2e1a321c8aab1e/vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/action/MessageActionsViewModel.kt#L285
> 
> "Share" option is not appearing in encrypted rooms for images. Because messageContent.url is null. Instead, should we check messageContent.encryptedFileInfo.url for encrypted rooms?

Nice catch, I create #1031 to handle this case

bmarty:
> Screenshot_2020-02-17-12-19-38-253_im vector riotx debug
> 
> Screenshot_2020-02-17-12-44-01-924_im vector riotx debug
> 
> Room_1 shares an image with Room_2. None of the rooms are encrypted. Room_2 gets the error message:
> 
> **Upload request missing 'Content-Type'**
> 
> Second and third attempts stuck in **'Waiting...'** status (Rageshaked).

I let you create a dedicated issue for this one please

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/DefaultSendService.kt
bmarty: ok, done

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/EncryptEventWorker.kt
```diff
@@ -80,7 +78,7 @@ internal class EncryptEventWorker(context: Context, params: WorkerParameters)
         var result: MXEncryptEventContentResult? = null
         try {
             result = awaitCallback {
-                crypto.encryptEventContent(localMutableContent, localEvent.type, params.roomId, it)
+                crypto.encryptEventContent(localMutableContent, localEvent.type, localEvent.roomId!!, it)
```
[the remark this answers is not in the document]
bmarty: we are pretty sure it will never be null. It is nullable because Event data class is also use to parse what come form the server, and we do not trust what the server sends.  But in this case, we have built the local echo and have set a roomId

vector/src/main/java/im/vector/riotx/features/attachments/preview/AttachmentsPreviewFragment.kt
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    private fun handleCropResult(result: Intent) {
+        val resultPath = UCrop.getOutput(result)?.path
+        if (resultPath != null) {
+            viewModel.handle(AttachmentsPreviewAction.UpdatePathOfCurrentAttachment(resultPath))
+        } else {
+            Toast.makeText(requireContext(), "Cannot retrieve cropped value", Toast.LENGTH_SHORT).show()
+        }
+    }
```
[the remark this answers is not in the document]
bmarty: Your suggestion is actually a "dangerous" practice, because if the result of the `let` block is null, the `run` block will also be executed, and it's generally not expected.

Sometimes a good old simple if/else pattern is better :)

vector/src/main/java/im/vector/riotx/features/attachments/preview/AttachmentsPreviewFragment.kt
```diff
@@ -0,0 +1,218 @@
...
+                SnapOnScrollListener.Behavior.NOTIFY_ON_SCROLL_STATE_IDLE,
+                object : OnSnapPositionChangeListener {
+                    override fun onSnapPositionChange(position: Int) {
+                        viewModel.handle(AttachmentsPreviewAction.SetCurrentAttachment(position))
+                    }
+                })
+        attachmentPreviewerBigList.setHasFixedSize(true)
+        attachmentPreviewerBigList.adapter = attachmentBigPreviewController.adapter
+    }
+}
```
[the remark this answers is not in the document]
bmarty: I change to use `let`
Be careful when using `apply`, the common mistake is that in the apply scope you get access to the parent class methods which can lead to unexpected behavior.

bmarty:
Thanks @onurays and @steef435 for your review.
I've updated the PR.

PR #1037, opened 2020-02-19
Migrate to binary QR code verification (#994)

### LINKED ISSUE #994: Iterate on verification QR code content
The [spec](https://github.com/uhoreg/matrix-doc/blob/qr_key_verification/proposals/1543-qr_code_key_verification.md#qr-code-format) as been updated to reduce the size of the data encoded in the QR code and so make it look less scary and easier to scan.

Here is the current text, but be sure to refer to the spec when implementing this:

### QR code format

The QR codes to be displayed and scanned using this format will encode binary
strings in the general form:

- the ASCII string "MATRIX"
- one byte indicating the QR code version (must be `0x02`)
- one byte indicating the QR code verification mode.  May be one of the
  following values:
  - `0x00` verifying another user with cross-signing
  - `0x01` self-verifying in which the current device does trust the master key
  - `0x02` self-verifying in which the current device does not yet trust the
    master key
- the event ID or `transaction_id` of the associated verification
  request event, encoded as:
  - two bytes in network byte order (big-endian) indicating the length of the
    ID
  - the ID as an ASCII string
- the first key, as 32 bytes.  The key to use depends on the mode field:
  - if `0x00` or `0x01`, then the user's own master cross-signing public key
  - if `0x02`, then the current device's device key
- the second key, as 32 bytes.  The key to use depends on the mode field:
  - if `0x00`, then what the device thinks the other user's master
    cross-signing key is
  - if `0x01`, then what the device thinks the other device's device key is
  - if `0x02`, then what the device thinks the user's master cross-signing key
    is
- a random shared secret, as a byte string.  It is suggested to use a secret
  that is about 8 bytes long.  Note: as we do not share the length of the
  secret, and it is not a fixed size, clients will just use the remainder of
  binary string as the shared secret.

For example, if Alice displays a QR code encoding the following binary string:

```
      "MATRIX"    |ver|mode| len   | event ID
 4D 41 54 52 49 58  02  00   00 2D   21 41 42 43 44 ...
| user's cross-signing key    | other user's cross-signing key | shared secret
  00 01 02 03 04 05 06 07 ...   10 11 12 13 14 15 16 17 ...      20 21 22 23 24 25 26 27
```

this indicates that Alice is verifying another user (say Bob), in response to
the request from event "!ABCD...", her cross-signing key is
`0001020304050607...` (which is "AAECAwQFBg..." in base64), she thinks that
Bob's cross-signing key is `1011121314151617...` (which is "EBESExQVFh..." in
base64), and the shared secret is `2021222324252627` (which is "ICEiIyQlJic" in
base64).

bmarty:
Fixes #994

Self verification has not been tested yet.

To reviewers: I'm not sure about the changes in DefaultQrCodeVerificationTransaction.kt and DeviceVerificationservice.kt


matrix-sdk-android/src/androidTest/java/im/vector/matrix/android/internal/crypto/verification/qrcode/QrCodeTest.kt
```diff
@@ -0,0 +1,321 @@
...
+            deviceKey = "TXluZKTZLvSRWOTPlOqLq534bA+/K4zLFKSu9cGLQaU",
+            userMasterCrossSigningPublicKey = "ktEwcUP6su1xh+GuE+CYkQ3H6W/DIl+ybHFdaEOrolU",
+            sharedSecret = "MTIzNDU2Nzg"
+    )
+
+    private val value3 = "MATRIX\u0002\u0002\u0000\u000DMaTransactionMynd¤Ù.ô\u0091XäÏ\u0094ê\u008B«\u009Døl\u000F¿+\u008CË\u0014¤®õÁ\u008BA¥\u0092Ñ0qCú²íq\u0087á®\u0013à\u0098\u0091\u000DÇéoÃ\"_²lq]hC«¢U12345678"
+
+    private val sharedSecretByteArray = "12345678".toByteArray(Charsets.ISO_8859_1)
+
+    // 4d 79 6e 64 a4 d9 2e f4 91 58 e4 cf 94 ea 8b ab 9d f8 6c 0f bf 2b 8c cb 14 a4 ae f5 c1 8b 41 a5
```
BillCarsonFr: we don't have a hexToByte somewhere?

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/util/Byte.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.matrix.android.internal.util
+
+fun Byte.toPositiveInt(): Int {
```
BillCarsonFr: fun Byte.toUnsignedInt() = toInt() and 0xff 
?

matrix-sdk-android/src/androidTest/java/im/vector/matrix/android/internal/crypto/verification/qrcode/QrCodeTest.kt
bmarty: done

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/util/Byte.kt
bmarty: done

PR #1044, opened 2020-02-20
Use exif info of the image for correct rotation.

### LINKED ISSUE #647: Image rotation not respected in full-size image, but is in preview
Image rotation in EXIF data is respected in previews, both in the channel and the full-size progressive image load, but not when viewing the full-size image.

Screenshot_20191028-072308

onurays:
Fixes #647


vector/src/main/java/im/vector/riotx/features/media/ImageContentRenderer.kt
```diff
@@ -142,6 +144,13 @@ class ImageContentRenderer @Inject constructor(private val activeSessionHolder:
             return
         }
 
+        imageView.setImageLoaderCallback(object: DefaultImageLoaderCallback {
+            override fun onSuccess(image: File?) {
+                super.onSuccess(image)
```
bmarty: No need to call the super method here

PR #1050, opened 2020-02-21
Feature/fix some perf

ganfra:
Fix some performance issues with crypto


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/DefaultSession.kt
```diff
@@ -163,6 +166,7 @@ internal class DefaultSession @Inject constructor(
     override fun close() {
         assert(isOpen)
         stopSync()
+        timelineEventDecryptor.destroy()
```
bmarty: `dispose()` instead of `destroy()`

ganfra:
> LGTM. Maybe add somewhere the limitation introduced by this PR (sometimes (when?) the last event in the room list will not be decrypted without entering in the room detail screen)

This should not changed from before, we decrypt when we get the keys. Doesn't work on your side?

PR #1069, opened 2020-02-25
Fix dm chips

bmarty:
A race condition was sometimes adding twice the first chip (#1032) 
This PR also fixes the fact that adding from userId was not working well if there was already a user selected.


ganfra:
LGTM

Issue #1014, opened 2020-02-14
WorkManager: iterate on the initialization process


bmarty:
A warning in the documentation of the library is:

> Note that on-demand initialization may delay some useful features of WorkManager such as automatic rescheduling of work following a crash and recovery from the application being force-stopped by the user or device.

```
    /**
     * A class that can provide the {@link Configuration} for WorkManager and allow for on-demand
     * initialization of WorkManager.  To do this:
     * <p><ul>
     *   <li>Disable {@code androidx.work.impl.WorkManagerInitializer} in your manifest</li>
     *   <li>Implement the {@link Configuration.Provider} interface on your
     *   {@link android.app.Application} class</li>
     *   <li>Use {@link WorkManager#getInstance(Context)} when accessing WorkManger (NOT
     *   {@link WorkManager#getInstance()})</li>
     * </ul></p>
     * <p>
     * Note that on-demand initialization may delay some useful features of WorkManager such as
     * automatic rescheduling of work following a crash and recovery from the application being
     * force-stopped by the user or device.
     *
     * @see WorkManager#initialize(Context, Configuration) for manual initialization.
     */
    public interface Provider {

        /**
         * @return The {@link Configuration} used to initialize WorkManager
         */
        @NonNull Configuration getWorkManagerConfiguration();
    }
```

It makes me think that we should not initialize the work manager, but let the lib do it.

Also #1012 is a thing

Issue #1012, opened 2020-02-14
Remove androidx.work.impl.WorkManagerInitializer from your AndroidManifest.xml


Bubu:
```
> Task :vector:lintVitalFdroidRelease
/home/bubu/projects/oss/fdroiddata/build/im.vector.riotx/vector: Error: Remove androidx.work.impl.WorkManagerInitializer from your AndroidManifest.xml when using on-demand initialization. [RemoveWorkManagerInitializer]

   Explanation for issues of type "RemoveWorkManagerInitializer":
   If an android.app.Application implements
   androidx.work.Configuration.Provider,
   the default androidx.work.impl.WorkManagerInitializer needs to be removed
   from the
   AndroidManifest.xml file.

1 errors, 0 warnings

> Task :vector:lintVitalFdroidRelease FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':vector:lintVitalFdroidRelease'.
> Lint found fatal errors while assembling a release target.
  
  To proceed, either fix the issues identified by lint, or modify your build script as follows:
  ...
  android {
      lintOptions {
          checkReleaseBuilds false
          // Or, if you prefer, you can continue to check for errors in release builds,
          // but continue the build even when errors are found:
          abortOnError false
      }
  }
  ...

```

bmarty:
The new lint check has been added in [work 2.3.0-beta02](https://developer.android.com/jetpack/androidx/releases/work#2.3.0-beta02)

And the lib has been updated to this version in RiotX since the [6th of January](99c523b710bec38d3ce80515530925761cb1c6e5), which can explain why you see the lint error now (but why not in the previous release?).

`./gradlew assembleFdroidRelease` does not fail for me, and the task `lintVitalFdroidRelease` is ok on my side.
The issue reported by `lintVitalFdroidRelease` in your case seems like a false positive because the code in the SDK ensure that the node is removed from the built application (see https://github.com/vector-im/riotX-android/blob/859c75df983b085825cab9c1e6f00032abb06813/matrix-sdk-android/src/main/AndroidManifest.xml#L15)

Bubu:
> but why not in the previous release?

I was on vacation :-)

I'll disable the lint check for this version then, I've no idea why you don't see it.
One thing that f-droid does differently then when you build normally is that it starts from a clean git repo checkout (`git clean -df`) which I think is far more thorough than `gradle clean`.

bmarty:
[...]

PR #1074, opened 2020-02-26
gradle and build tools update - DO NOT MERGE

bmarty:
Do not merge it, because we will have the same issue than @Bubu about lint error

(see #1014 and #1012)


bmarty:
Will be done in a new PR

PR #1081, opened 2020-02-26
Fix redacted event format on room overview.

onurays:
redaction_reason


vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/format/NoticeEventFormatter.kt
```diff
@@ -317,4 +317,28 @@ class NoticeEventFormatter @Inject constructor(private val sessionHolder: Active
             else                 -> null
         }
     }
+
+    fun getRedactionReason(event: Event): String {
```
bmarty: Maybe rename the method to `formatRedactedEvent` to respect the naming in this class

PR #1082, opened 2020-02-26
Fix crashes reported by rageshakes

bmarty:
One commit per fix

### DIFF
--- a/vector/src/fdroid/java/im/vector/riotx/fdroid/receiver/AlarmSyncBroadcastReceiver.kt
+++ b/vector/src/fdroid/java/im/vector/riotx/fdroid/receiver/AlarmSyncBroadcastReceiver.kt
@@ -48,7 +48,7 @@ class AlarmSyncBroadcastReceiver : BroadcastReceiver() {
             }
         }
 
-        val sessionId = intent.getStringExtra(SyncService.EXTRA_SESSION_ID)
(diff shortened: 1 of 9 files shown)


vector/src/fdroid/java/im/vector/riotx/fdroid/receiver/AlarmSyncBroadcastReceiver.kt
```diff
@@ -48,7 +48,7 @@ class AlarmSyncBroadcastReceiver : BroadcastReceiver() {
             }
         }
 
-        val sessionId = intent.getStringExtra(SyncService.EXTRA_SESSION_ID)
+        val sessionId = intent.getStringExtra(SyncService.EXTRA_SESSION_ID) ?: return
```
BillCarsonFr: If we return, the alarm will never be launched again?

bmarty: I think this was a temporary crash, a scheduled alarm with a previous version of RiotX. Not sure though... The rageshake was quite old

ganfra: It had been fixed 2 months ago

Issue #12, opened 2018-12-17
Editor: slash commands


manuroe:
The editor must manage slash command:
- [ ] manage completion
- [ ] display help
- [ ] convert the command into action

vc7ee:
/rainbow
/shrug
/rainbowshrug
;)
There should also be better markdown handling, if there isn't an issue for it.

fbruetting:
May it be useful to create pills out of for example `/me`?

Currently (v0.10.0) it also makes no difference for the room preview in the room overview. So `/me is going` and just `is going` are both rendered as **`Frank`**` – is going`. Shouldn’t the first be shown as **`Frank`**` – Frank is going` (or maybe just **`Frank`**` is going`)?

bpcurse:
What about /plain?
It is quite helpful if you want to demonstrate the use of markdown and it's only temporary.

bmarty:
The slash commands list on RiotWeb is growing too fast :)

bmarty:
Support for `/plain` command is in #1093 :)

Fidgetcetera:
also chiming in for `/myroomnick` and `/myroomavatar`, which are very nice to have.

bmarty:
I updated the list

Prog1563:
Glad I found this, I have been trying to make /myroomnick work on mobile for days and was about to make a bug report. Glad to see it is already being worked on :)

lambroisie:
`/ignore` is also missing @bmarty

PR #1093, opened 2020-03-01
Add support for `/plain` command (#12)

bmarty:
Small, easy PR for a Sunday morning


bmarty:
> Maybe show it only when auto markdown is ON?

Slash command are always all displayed. I'm afraid user ill complain that they do not see the new command, since markdown is off by default now.

PR #1099, opened 2020-03-03
Share images from clear and encrypted rooms.

onurays:


vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
```diff
@@ -261,12 +261,9 @@ class MessageActionsViewModel @AssistedInject constructor(@Assisted
                     }
 
                     if (canShare(msgType)) {
-                        if (messageContent is MessageImageContent) {
-                            session.contentUrlResolver().resolveFullSize(messageContent.url)?.let { url ->
-                                add(EventSharedAction.Share(url))
-                            }
+                        if (messageContent is MessageWithAttachmentContent) {
```
bmarty: Maybe this test should be done in canShare() method (and maybe replace the current canShare() method body

PR #1103, opened 2020-03-04
Detect spaces in password if user fails to login (#1038)

### LINKED ISSUE #1038: Whitespace at the end of the password is not removed
If you paste the backup key password into riotx and there is a space at the end, it will be kept. As c&p from other apps might cause this, this causes unexpected user experience.

Tested on 0.16.0

bmarty:
Fixes #1038


telmich:
Nice solution, good compromise!

PR #1116, opened 2020-03-06
Add some documentation on Workers

bmarty:
And do some cleanup

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/worker/SessionWorkerParams.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/worker/SessionWorkerParams.kt
@@ -16,9 +16,16 @@
 
 package im.vector.matrix.android.internal.worker
 
+/**
+ * Note about the Worker usage:
+ * The workers we chain, or when using the append strategy, should never return Result.Failure(), else the chain will be broken forever
+ */
 interface SessionWorkerParams {
     val sessionId: String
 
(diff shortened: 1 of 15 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/worker/SessionWorkerParams.kt
```diff
@@ -16,9 +16,16 @@
...
+ * Note about the Worker usage:
+ * The workers we chain, or when using the append strategy, should never return Result.Failure(), else the chain will be broken forever
+ */
 interface SessionWorkerParams {
     val sessionId: String
 
-    // Null is no error occurs. When chaining Workers, first step is to check that there is no lastFailureMessage from the previous workers
+    /**
+     * Null when no error occurs. When chaining Workers, first step is to check that there is no lastFailureMessage from the previous workers
+     * If it is the case, the worker should just transmit the error and shouldn't do anything else
```
BillCarsonFr: i think you should use it and react gracefully. E.g the SentWorker might check for the EncryptWork result, and if there is an error just mark the message as failed to sent

bmarty: yes, will do that in another pr

bmarty: After double check , it seems that every cases of possible error are corretly managed, with local echo updated. I let you double check

PR #1128, opened 2020-03-10
Remove dangerous runBlocking (and un-needed)

BillCarsonFr:
un needed runblocking, and dangerous (should not be used from a coroutine):

````
 * Runs a new coroutine and **blocks** the current thread _interruptibly_ until its completion.
 * This function should not be used from a coroutine. It is designed to bridge regular blocking code
 * to libraries that are written in suspending style, to be used in `main` functions and in tests.


bmarty:
Late approved :)

PR #1131, opened 2020-03-11
Fix / Self verification mode 1 was not working

### LINKED ISSUE #1130: Self verification via QR code is failing on 0.17
Step to reproduce:

Log an new session on RiotX, and try to verify by scanning the QRCode displayed on web
=> RiotX cancels for mismatch keys

BillCarsonFr:
Fixes #1130


bmarty:
Late approval :)

PR #1133, opened 2020-03-11
Fix Message transitions in encrypted rooms

### LINKED ISSUE #518: Message transitions in encrypted rooms are jarring
Sending a message in an encrypted room can be quite a ride:

1. After typing the message and hitting send, nothing happens for a second (or five)
2. Then, the message is inserted in the timeline in green (indicating encryption in progress)
3. The message transitions to grey (indicating sending)
4. The message transitions to black (local echo)
5. The message disappears, leaving empty white space
6. A new grey message with text *encrypted message* appears
7. *encrypted message* transitions to the correct text

Every step can take noticeable time, sometimes multiple seconds. During all of this, the timeline can wildly jump around, especially if someone else is sending messages concurrently into the same room.

BillCarsonFr:
Fixes #518

Immediatly decrypt remote echo of encrypted messages to avoid annoying transition

# Before 
image
image
image

# After
image
image

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/LocalEchoUpdater.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/LocalEchoUpdater.kt
@@ -17,7 +17,11 @@
 package im.vector.matrix.android.internal.session.room.send
 
 import com.zhuinden.monarchy.Monarchy
+import im.vector.matrix.android.api.session.events.model.Content
+import im.vector.matrix.android.api.session.events.model.EventType
 import im.vector.matrix.android.api.session.room.send.SendState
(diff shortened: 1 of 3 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/LocalEchoUpdater.kt
```diff
@@ -38,4 +42,15 @@ internal class LocalEchoUpdater @Inject constructor(private val monarchy: Monarc
             }
         }
     }
+
+    fun updateEncryptedEcho(eventId: String, encryptedContent: Content, mxEventDecryptionResult: MXEventDecryptionResult) {
+        monarchy.writeAsync { realm ->
```
ganfra: Maybe use awaitTransaction with TaskExecutor.executorScope like we do in LocalEchoEventFactory

PR #1140, opened 2020-03-13
Feature/gossiping work

### LINKED ISSUE #944: XSigning | Support SSSS secret sharing
https://github.com/uhoreg/matrix-doc/blob/ssss/proposals/1946-secure_server-side_storage.md

Rather than (or in addition to) storing secrets on the server encrypted by a shared key, devices can send secrets to each other, encrypted using olm.

To request a secret, a client sends a m.secret.request device event with action set to request to other devices, and name set to the name of the secret that it wishes to retrieve. A device that wishes to share the secret will reply with a m.secret.send event, encrypted using olm. When the original client obtains the secret, it sends a m.secret.request event with action set to request_cancellation to all devices other than the one that it received the secret from. Clients should ignore m.secret.send events received from devices that it did not send an m.secret.request event to.

BillCarsonFr:
Fixes #944

- Added support for gossiping SSSS secrets.
- SDK automatically handles SSK/USK gossiping
- Now after a successful verification, riotX is not anymore in read only
- In developer mode , the action menu of a failed to decrypt message will offer option to re-request keys

Refactoring of KeyRequest to more generic GossipingRequest
-> improove auditability (new devtool screen, saving a trail of gossiping events in crypto database)

=> This PR will migrate the Crypto database


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/verification/VerificationTransportRoomMessage.kt
```diff
@@ -244,7 +245,26 @@ internal class VerificationTransportRoomMessage(
...
+                        ?.firstOrNull { it.id == enqueueInfo.second }
+                        ?.let { _ ->
+                            onDone?.invoke()
+                            workLiveData.removeObserver(this)
+                        }
+            }
+        }
+
+        // TODO listen to DB to get synced info
+        GlobalScope.launch(Dispatchers.Main) {
```
ganfra: Why using coroutines here? if it's just for dispatching to ui, prefer posting to handler directly.

PR #1142, opened 2020-03-13
Fix / correctly update local echo failed state

### LINKED ISSUE #1145: Images that failed to send are waiting to be sent 3 days later
https://github.com/matrix-org/riot-android-rageshakes/issues/9650

BillCarsonFr:
Fixes #1145 
+ ensure Workmanager queues are reset after a clear cache


fearedbliss:
Just to note, the PR may have fixed new failed image messages from being stuck in a sending forever state, but existing people in this state will still stay with those failed messages being sent. Only workaround was to log out of the account and log back in, and then re import key backups.

bmarty:
Late LGTM :)

PR #1149, opened 2020-03-17
QuickFix / key banner not clickable

BillCarsonFr:
Fix / When there are locally unknown keys available on the backup the banner is not clickable


bmarty:
Late LGTM :)

PR #1152, opened 2020-03-18
Verif / handle concurrent start

### LINKED ISSUE #794: Verification DM /  Handle concurrent .start after .ready
After the `m.key.verification.ready` event is sent, either party can send an
`m.key.verification.start` event to begin the verification.  If both parties
send an `m.key.verification.start` event, and they both specify the same
verification method, then the event sent by the user whose user ID is the
smallest is used, and the other `m.key.verification.start` event is ignored.
In the case of a single user verifying two of their devices, the device ID is
compared instead.  If both parties send an `m.key.verification.start` event,
but they specify different verification methods, the verification should be
cancelled with a `code` of `m.unexpected_message`.

BillCarsonFr:
Fixes #794
(Verification DM / Handle concurrent .start after .ready)

### DIFF
--- a/matrix-sdk-android/src/androidTest/java/im/vector/matrix/android/internal/crypto/gossiping/KeyShareTests.kt
+++ b/matrix-sdk-android/src/androidTest/java/im/vector/matrix/android/internal/crypto/gossiping/KeyShareTests.kt
@@ -32,9 +32,6 @@ import im.vector.matrix.android.internal.crypto.model.event.EncryptedEventConten
 import junit.framework.TestCase.assertNotNull
 import junit.framework.TestCase.assertTrue
 import junit.framework.TestCase.fail
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.delay
-import kotlinx.coroutines.launch
 import org.junit.Assert
 import org.junit.FixMethodOrder
 import org.junit.Test
@@ -90,10 +87,10 @@ class KeyShareTests : InstrumentedTest {
 
         var outGoingRequestId: String? = null
 
-        retryPeriodicallyWithLatch(waitLatch) {
+        mTestHelper.retryPeriodicallyWithLatch(waitLatch) {
             aliceSession2.cryptoService().getOutgoingRoomKeyRequest()
                     .filter { req ->
-                        // filter out request that was known before
+                        // filter out request thwat was known before
                         !outgoingRequestBefore.any { req.requestId == it.requestId }
                     }
                     .let {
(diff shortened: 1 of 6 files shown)


matrix-sdk-android/src/androidTest/java/im/vector/matrix/android/internal/crypto/gossiping/KeyShareTests.kt
```diff
@@ -90,10 +87,10 @@ class KeyShareTests : InstrumentedTest {
 
         var outGoingRequestId: String? = null
 
-        retryPeriodicallyWithLatch(waitLatch) {
+        mTestHelper.retryPeriodicallyWithLatch(waitLatch) {
             aliceSession2.cryptoService().getOutgoingRoomKeyRequest()
                     .filter { req ->
-                        // filter out request that was known before
+                        // filter out request thwat was known before
```
onurays: Typo?

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/verification/DefaultVerificationService.kt
```diff
@@ -457,7 +457,29 @@ internal class DefaultVerificationService @Inject constructor(
         Timber.d("## SAS onStartRequestReceived ${startReq.transactionId}")
         if (checkKeysAreDownloaded(otherUserId!!, startReq.fromDevice) != null) {
```
onurays: It seems dangerous to force a nullable field.

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/verification/DefaultVerificationService.kt
```diff
@@ -457,7 +457,29 @@ internal class DefaultVerificationService @Inject constructor(
...
+            var existing = getExistingTransaction(otherUserId, tid)
+
+            // After the m.key.verification.ready event is sent, either party can send an
+            // m.key.verification.start event to begin the verification. If both parties
+            // send an m.key.verification.start event, and they both specify the same
+            // verification method, then the event sent by the user whose user ID is the
+            // smallest is used, and the other m.key.verification.start event is ignored.
+            // In the case of a single user verifying two of their devices, the device ID is
+            // compared instead .
+            if (existing != null && !existing.isIncoming) {
```
onurays: I would prefer to use `existing?.isIncoming == false`

BillCarsonFr: lint asks then to simplify to !

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/verification/DefaultVerificationService.kt
BillCarsonFr: done

matrix-sdk-android/src/androidTest/java/im/vector/matrix/android/internal/crypto/gossiping/KeyShareTests.kt
BillCarsonFr: yep, fixed

bmarty:
Late LGTM :)

PR #1153, opened 2020-03-18
Fixes #841

### LINKED ISSUE #841: Missing avatar/displayname after verification request message
After a verification request the avatar of the next message should be displayed, otherwise if the message before the .request is form another user, there will be no ways to know who posted the message after the request

BillCarsonFr:
Fixes #841
 Missing avatar/displayname after verification request message 

After a verificati

Before:
image

After:
image


bmarty:
Late LGTM :)

PR #1175, opened 2020-03-26
Fixes #1174

### LINKED ISSUE #1174: RiotX can't restore cross signing keys saved by web in SSSS
Probably incompatibility in the way we save the private keys in SSS

BillCarsonFr:
Fix / Since migration to symetric SSSS RiotX can't restore cross signing keys saved by web in SSSS


bmarty:
Late LGTM :)

PR #1194, opened 2020-04-01
Clarify UTD notice about no messages to unverified devices

tuxayo:
It's a copy of the message from Riot Web which is much clearer.

### Pull Request Checklist

Is that needed for a string change?

Is that needed for a string change?

Is that needed for a string change?

Is that needed if the string is shorter?

### DIFF
--- a/matrix-sdk-android/src/main/res/values/strings.xml
+++ b/matrix-sdk-android/src/main/res/values/strings.xml
@@ -50,7 +50,7 @@
     <string name="notice_room_third_party_registered_invite">%1$s accepted the invitation for %2$s</string>
 
     <string name="notice_crypto_unable_to_decrypt">** Unable to decrypt: %s **</string>
-    <string name="notice_crypto_error_unkwown_inbound_session_id">The sender\'s device has not sent us the keys for this message.</string>
+    <string name="notice_crypto_error_unkwown_inbound_session_id">The sender has disabled encrypting to unverified devices.</string>
 
     <!-- Messages -->
     <string name="message_reply_to_prefix">In reply to</string>


BillCarsonFr:
Thanks for the PR

Actually it's more than just the copy, we need to implement MSC2399 https://github.com/matrix-org/matrix-doc/pull/2399
The sender will now give more information, it'll be then possible to say if the message cannot be decypted because you are black-listed, or your device is not verified, or if it's a real issue with key share

It's tracked here -> https://github.com/vector-im/riotX-android/issues/797

and coming very soon

tuxayo:
> Actually it's more than just the copy, we need to implement MSC2399 matrix-org/matrix-doc#2399
The sender will now give more information, it'll be then possible to say if the message cannot be decypted because you are black-listed, or your device is not verified, or if it's a real issue with key share

How does Riot Web work currently? Does it assume «The sender has disabled encrypting to unverified devices»?
And RiotX Android for now doesn't make that assumption.

BillCarsonFr:
> > Actually it's more than just the copy, we need to implement MSC2399 [matrix-org/matrix-doc#2399](https://github.com/matrix-org/matrix-doc/pull/2399)
> > The sender will now give more information, it'll be then possible to say if the message cannot be decypted because you are black-listed, or your device is not verified, or if it's a real issue with key share
> 
> How does Riot Web work currently? Does it assume «The sender has disabled encrypting to unverified devices»?
> And RiotX Android for now doesn't make that assumption.

according to MSC the client is sending a to_device message `m.room_key.withheld` to the device that will not receive the keys

PR #1215, opened 2020-04-08
Fixes #1214

### LINKED ISSUE #1214: Cross-Signing | Restore history after recover from passphrase
After verifying a new session with passphrase, I should get the key backup passphrase from SSSS if available and recover history

BillCarsonFr:
Fixes #1214

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecureStorageViewModel.kt
+++ b/vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecureStorageViewModel.kt
@@ -37,6 +37,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import timber.log.Timber
 import java.io.ByteArrayOutputStream
 
 data class SharedSecureStorageViewState(
@@ -117,11 +118,15 @@ class SharedSecureStorageViewModel @AssistedInject constructor(
                 withContext(Dispatchers.IO) {
                     args.requestedSecrets.forEach {
                         val res = awaitCallback<String> { callback ->
-                            session.sharedSecretStorageService.getSecret(
-                                    name = it,
-                                    keyId = keyInfo.id,
-                                    secretKey = keySpec,
-                                    callback = callback)
+                            if (session.getAccountDataEvent(it) != null) {
+                                session.sharedSecretStorageService.getSecret(
+                                        name = it,
+                                        keyId = keyInfo.id,
+                                        secretKey = keySpec,
+                                        callback = callback)
+                            } else {
+                                Timber.w("## Cannot find secret $it in SSSS, skip")
+                            }
                         }
                         decryptedSecretMap[it] = res
                     }
--- a/vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecureStorageViewModel.kt
+++ b/vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecureStorageViewModel.kt
@@ -16,6 +16,7 @@
 
 package im.vector.riotx.features.crypto.quads
 
+import androidx.lifecycle.viewModelScope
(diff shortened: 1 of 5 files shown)


vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecureStorageViewModel.kt
```diff
@@ -117,11 +118,15 @@ class SharedSecureStorageViewModel @AssistedInject constructor(
...
+                                session.sharedSecretStorageService.getSecret(
+                                        name = it,
+                                        keyId = keyInfo.id,
+                                        secretKey = keySpec,
+                                        callback = callback)
+                            } else {
+                                Timber.w("## Cannot find secret $it in SSSS, skip")
+                            }
                         }
                         decryptedSecretMap[it] = res
```
Dominaezzz, line 131: If the else branch is hit, `awaitCallback` will suspend forever! Since `callback` is not being called in that branch.
This is probably what you wanted to do. (Unless you wanted to store null in the map)
```suggestion
                        if (session.getAccountDataEvent(it) != null) {
                            val res = awaitCallback<String> { callback ->
                                session.sharedSecretStorageService.getSecret(
                                    name = it,
                                    keyId = keyInfo.id,
                                    secretKey = keySpec,
                                    callback = callback)
                            }
                            decryptedSecretMap[it] = res
                        } else {
                            Timber.w("## Cannot find secret $it in SSSS, skip")
                        }
```

BillCarsonFr: thx good catch!

vector/src/main/java/im/vector/riotx/features/crypto/verification/VerificationBottomSheetViewModel.kt
```diff
@@ -334,38 +341,80 @@ class VerificationBottomSheetViewModel @AssistedInject constructor(
...
+                        res?.get(USER_SIGNING_KEY_SSSS_NAME),
+                        res?.get(SELF_SIGNING_KEY_SSSS_NAME)
+                )
+                if (trustResult.isVerified()) {
+                    // Sign this device and upload the signature
+                    session.sessionParams.credentials.deviceId?.let { deviceId ->
+                        session.cryptoService()
+                                .crossSigningService().trustDevice(deviceId, object : MatrixCallback<Unit> {
+                                    override fun onFailure(failure: Throwable) {
+                                        Timber.w(failure, "Failed to sign my device after recovery")
```
onurays, line 364: Should we notify the user if this fails?

vector/src/main/java/im/vector/riotx/features/crypto/verification/VerificationBottomSheetViewModel.kt
```diff
@@ -334,38 +341,80 @@ class VerificationBottomSheetViewModel @AssistedInject constructor(
...
-                    _viewEvents.post(VerificationBottomSheetViewEvents.ModalError(failure.localizedMessage))
+
+                    setState {
+                        copy(verifiedFromPrivateKeys = true)
+                    }
+
+                    // try to get keybackup key
+                } else {
+                    // POP UP something
+                    _viewEvents.post(VerificationBottomSheetViewEvents.ModalError("Failed to import keys"))
```
onurays, line 376: I think we should use the StringProvider.

vector/src/main/java/im/vector/riotx/features/crypto/verification/VerificationBottomSheetViewModel.kt
BillCarsonFr: It should not fail :D 
Not sure how to present that to user, for now I just log it to help debugging

vector/src/main/java/im/vector/riotx/features/crypto/verification/VerificationBottomSheetViewModel.kt
BillCarsonFr: done, thx

PR #1222, opened 2020-04-10
Uniform emoji representation for all platform

### LINKED ISSUE #1220: Emoji Verification | It's not the same butterfly !
Depending on android device, the emoji for verification can be very different.
Make it hard to tap the 'they match' button :D

image

BillCarsonFr:
Fixes #1220

Added vector assets matching web for verification (tweemoji assets, Creative Commons)
As a consequence even older android now supports emoji instead of only decimal.

## Web <-> RiotX with same emojis!
image

## previously
image

image


daviddias:
This is really great!! Now folks will be confident that they are indeed looking at the same emojis. Thanks for fixing it!

bmarty:
Late LGTM :)

PR #1226, opened 2020-04-10
Set `tickerText` to improve accessibility of notifications.

ndarilek:


vector/src/main/java/im/vector/riotx/features/notifications/NotificationDrawerManager.kt
```diff
@@ -343,12 +343,19 @@ class NotificationDrawerManager @Inject constructor(private val context: Context
                         globalLastMessageTimestamp = lastMessageTimestamp
                     }
 
+                    val tickerText = if (roomEventGroupInfo.isDirect) {
+                        String.format("%s: %s", events[0].senderName, events[0].description)
```
bmarty, line 347: We should use string resource for that.
What about defining 2 strings like "%1$s says: %2$s" for the first case and "In room %1$s, %2$s says: %3$s"?

bmarty, line 347: Why using the first event? I think this will be the oldest one. We should maybe use the latest one. I let you decide what is best

ndarilek: As a screen reader user, I dislike when chatroom messages read out as something like "Nolan says: Hey. or "In Matrix HQ: Nolan says What's up!" It only adds a few syllables, but over time slows down the experience. Further, I can usually figure out what's happening after the first message or two, after which point the extra syllables waste time. Riot Web did this for a bit in places, and I recall many of us requested succinctness.

ndarilek: Let me think about that for a bit. You're probably right.

ndarilek:
How do I sign off on this?

bmarty:
> How do I sign off on this?

see https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off

vector/src/main/java/im/vector/riotx/features/notifications/NotificationDrawerManager.kt
bmarty: ok, I get the point. Anyway we need to use string resources to handle RTL correctly. You get use the same mechanism than https://github.com/vector-im/riotX-android/blob/5081361c2d17fa3a46b3f455b65a5c25dff58641/vector/src/main/res/values/strings.xml#L1310.

ndarilek:
OK, makes sense. Will do that later today.

ndarilek:
I think this should address everything. Please let me know if I've missed something. I'd like to work on a few more RiotX accessibility improvements under Android.

Thanks!

PR #1230, opened 2020-04-13
Catch network errors during file downloading.

onurays:


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/DefaultFileService.kt
```diff
@@ -77,9 +78,10 @@ internal class DefaultFileService @Inject constructor(
                                 .url(resolvedUrl)
                                 .build()
 
-                        val response = okHttpClient.newCall(request).execute()
+                        val response = tryThis { okHttpClient.newCall(request).execute() } ?: return@flatMap Try.Failure(IOException())
```
bmarty, line 81: The pb with using `tryThis` is that we lose the exception. It has been creating by Valere when we do not care about the exception. But is this case you have to create a new Exception. Why not using a regular try catch block?

onurays: Fixed and tested.

bmarty:
Thanks for the update

PR #1239, opened 2020-04-16
Use a bigger thread pool

### LINKED ISSUE #1221: Messages take a long time to send
Messages in RiotX consistently take about 30-45 seconds to send. This is particularly annoying when you're trying to chime into a conversation and your posts end up getting sent way too late to make sense.

While this is happening the text of the message you just sent is grayed out. Eventually, if you wait long enough, the message sends. Or, if you exit to the main menu and come back into the room the message is now black, it's not just gray anymore, but it still hasn't sent. It takes about the same amount of time to send no matter what I do.

It's also worth noting that this was a regression at some point. For a while RiotX had great performance, this only started happening a couple of months ago. Unfortunately I didn't write down what specific day it was :(

This is reproducible too, it happens to everyone I know using RiotX. We're split up across three small self-hosted homeservers and the official homeserver, and I know that it's not just the homeserver because webriot works fine. I'm on the dev build of RiotX from fdroid.

Do you guys want any logs or anything? I sent a rageshake bug report a few hours ago but figured I'd put something on here too so I could track it.

waylon531:
This patch makes RiotX use an unbounded thread pool to handle connections. The
default thread pool for the android WorkManager has a fairly anemic
number of threads and I suspect this was causing performance issues
especially because of all the long-running jobs that happen whenever you
sync.

I tested this out on my phone and all the sluggishness of RiotX appears to
have gone away. I tested both the debug and release builds to make sure
it wasn't just some release optimization. RiotX is so much snappier now!

This fixes #1221

Signed-off-by: Waylon Cude <waylon.cude@finzdani.net>

### Pull Request Checklist

- [x] Changes has been tested on an Android device or Android emulator with API 16


MatMaul:
Kuddos for finding that, it's a night and day difference indeed.

Half-Shot:
Hey, myself and @anoadragon453 have been running this since yesterday and the performance is a definite improvement! For myself, it's been very quick to read history and generally swap between views (e.g. opening the app is <1s now).

I don't know how much this will negatively affect battery life though, perhaps one of the RiotX maintainers or @waylon531 can find out?

waylon531:
I suspect that it does increase battery usage but it's not a noticeable amount for me. I've been running my copy of the code for almost a week and I haven't noticed my phone dying any faster. The battery usage chart still shows a fairly low amount of battery used for RiotX, and it's also less than other communications apps on my phone like discord so I think the battery usage isn't significant.

I'll ask around though, I've got a few friends who are running my copy of the code.

hendersonreed:
I've been using the fork and haven't noticed a real difference. The battery stats for my phone show a 2% drop over 32 minutes for RiotX. For comparison, Youtube I used for 45 minutes and saw a 10% drop, and 30 minutes of FireFox saw an 11% drop.

waylon531:
It might make sense to use a larger fixed thread pool instead of the cached thread pool, the cached thread pool that I used can scale infinitely which might be causing the notifications that reed is seeing. 

For now though I think the tradeoff is worth it, notifications like that are bad UX but laggy interfaces are even worse.

ganfra:
Hello, looking at the implementation, it shouldn't change anything to our issues.
Indeed, we are using CoroutineWorker which are not running on the Executor threads at all, looking at the documentation also indicates in that sense: 
"Note that CoroutineWorker.doWork() is a suspending function. Unlike Worker, this code does not run on the Executor specified in your Configuration. Instead, it defaults to Dispatchers.Default.[...]"

bmarty:
Yes, thanks @waylon531!

Looking at the documentation of the work manager, it seems that the change you've maid could improve things, even if we do not really understand why, as @ganfra said. We are just a bit concerned about using too much threads in the application, if there is no reason to do so. 

According to the improvement reported by users, let's merge the PR and see what happen next.

PR #1256, opened 2020-04-20
Increase default timeout to match old riot

BillCarsonFr:
Increase default timeout to match old riot
https://github.com/matrix-org/matrix-android-sdk/blob/3427bf4d96fb707e4d5429b8739af6aa4cbcc179/matrix-sdk-core/src/main/java/org/matrix/androidsdk/RestClientHttpClientFactory.kt#L35

-> helps for verification with toDevice on mx

Ideally we should do dynamic timeouts based on connection type like former SDK was doing


bmarty:
We sometimes (ex: login request) set the timeout to 60 seconds... So this is now a bit useless... But we keep the code for the moment

PR #1259, opened 2020-04-20
KeyBackup / Use 4S if key in quadS

### LINKED ISSUE #1163: KeyBackup / SSSS | Should get the key from SSSS instead of asking recovery Key
Currently if SSSS is configured with key backup, the keybackup key is in SSSS.
So the keybackup restore screen should ask for SSSS key instead of keybackup recovery key (that the user never had access to)

image

BillCarsonFr:
Fixes #1163
Keybackup created by bootstrapping could not be restored from riotX.

With this PR, on tapping restore backup, we will check:
- If the secret has been gossiped, and in this case automatically load keys
- If the secret is in SSSS, in this case ask for the 4S passphrase to get backup key and restore
- if none of above, ask directly for the backup passphrase/key

=> We need to open issues for when user tries to delete backup (to clean 4S), as well as when user creates a backup when SSSS is enabled

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/crypto/keysbackup/restore/KeysBackupRestoreFromKeyViewModel.kt
+++ b/vector/src/main/java/im/vector/riotx/features/crypto/keysbackup/restore/KeysBackupRestoreFromKeyViewModel.kt
@@ -15,21 +15,19 @@
  */
 package im.vector.riotx.features.crypto.keysbackup.restore
 
-import android.content.Context
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
-import im.vector.matrix.android.api.MatrixCallback
-import im.vector.matrix.android.api.listeners.StepProgressListener
-import im.vector.matrix.android.api.session.crypto.keysbackup.KeysBackupService
-import im.vector.matrix.android.internal.crypto.keysbackup.model.rest.KeysVersionResult
-import im.vector.matrix.android.internal.crypto.model.ImportRoomKeysResult
+import androidx.lifecycle.viewModelScope
 import im.vector.riotx.R
 import im.vector.riotx.core.platform.WaitingViewData
-import im.vector.riotx.core.ui.views.KeysBackupBanner
-import timber.log.Timber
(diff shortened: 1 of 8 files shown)


vector/src/main/java/im/vector/riotx/features/crypto/keysbackup/restore/KeysBackupRestoreFromKeyViewModel.kt
```diff
@@ -15,21 +15,19 @@
...
-import im.vector.riotx.core.ui.views.KeysBackupBanner
-import timber.log.Timber
+import im.vector.riotx.core.resources.StringProvider
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import javax.inject.Inject
 
-class KeysBackupRestoreFromKeyViewModel @Inject constructor() : ViewModel() {
+class KeysBackupRestoreFromKeyViewModel @Inject constructor(
+        private val stringProvider: StringProvider
```
bmarty, line 29: thx for using stringProvider instead of passing context to viewModel :)

vector/src/main/java/im/vector/riotx/features/crypto/keysbackup/restore/KeysBackupRestoreSharedViewModel.kt
```diff
@@ -62,30 +84,187 @@ class KeysBackupRestoreSharedViewModel @Inject constructor() : ViewModel() {
...
+                        progressObserver,
+                        it
+                )
             }
-        })
+            loadingEvent.postValue(null)
+            didRecoverSucceed(result)
+            trustOnDecrypt(keysBackup, keyVersion)
+        } catch (failure: Throwable) {
+            throw failure
```
bmarty, line 221: the try catch seems a bit useless here, no?

BillCarsonFr: good catch, I should  loadingEvent.postValue(null) in the catch. thx

PR #1261, opened 2020-04-20
Feature/unwedging

bmarty:
Implement [MSC1719](https://github.com/matrix-org/matrix-doc/pull/1719)

I'm a bit stuck now. The test is passing, but 

When I implementing it a month ago, the test `testUnwedging()` was:
- passing with 0cb43ee
- failing with 5900245 (so this is ok)

but now (after rebasing the branch), it is passing, even with 5900245

The last commit (a6368c4) contains the implementation, and the test is passing too, but we do not send any `m.dummy` message, `markOlmSessionForUnwedging()` is never called.

I think a fresh look on the code will be help full.
@BillCarsonFr could you help please?


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/DefaultCryptoService.kt
```diff
@@ -670,61 +682,73 @@ internal class DefaultCryptoService @Inject constructor(
...
+//        val deviceKey = encryptedMessage.senderKey ?: return
+//        encryptedMessage.algorithm?.takeIf { it == MXCRYPTO_ALGORITHM_MEGOLM } ?: return
+//
+//        if (senderId == userId
+//                && deviceKey == olmDevice.deviceCurve25519Key) {
+//            Timber.d("## CRYPTO |  markOlmSessionForUnwedging: Do not unwedge ourselves")
+//            return
+//        }
+//
+//        val lastForcedDate = lastNewSessionForcedDates.getObject(senderId, deviceKey) ?: 0
```
bmarty, line 734: the date management is not used anymore then?

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/crypto/CryptoService.kt
```diff
@@ -111,6 +111,8 @@ interface CryptoService {
                             roomId: String,
                             callback: MatrixCallback<MXEncryptEventContentResult>)
 
+    fun discardOutbundSession(roomId: String)
```
bmarty, line 114: typo, a `o` is missing

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/DefaultCryptoService.kt
BillCarsonFr: good catch, didn't know it was implemented

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/crypto/CryptoService.kt
BillCarsonFr: fixed

matrix-sdk-android/src/androidTest/java/im/vector/matrix/android/common/CommonTestHelper.kt
```diff
@@ -58,6 +59,8 @@ class CommonTestHelper(context: Context) {
     val matrix: Matrix
 
     init {
+        Timber.plant(Timber.DebugTree())
```
BillCarsonFr, line 62: \o/

Issue #1264, opened 2020-04-22
Deactivate account when logged in with SSO


bmarty:
Account deactivation is supported for account with password.
Bootstrapping has also been disabled for SSO, and deleting a device is not possible too.

Related to https://github.com/vector-im/riot-web/issues/13310

bmarty:
https://github.com/vector-im/riot-web/issues/12692 is a well summarize to the same problem

bmarty:
It's implemented since 1.0.15, but is crashing at the moment, will be fixed in the next release (1.1.0)

PR #1265, opened 2020-04-22
Deactivate account using password

bmarty:
Deactivating account for SSO will be handled later (see #1264 )

Wording from Riot-Android are used. The screen has been improved to handle password reveal, and inline error.

This PR also contains integration test for password change, and minor improvement on login screen.

Screenshots:
image

image


bmarty:
FTR When trying to delete an account using a SSO account, entering any password, we get this:

image

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/account/DeactivateAccountParams.kt
```diff
@@ -0,0 +1,40 @@
...
+import im.vector.matrix.android.internal.crypto.model.rest.UserPasswordAuth
+
+@JsonClass(generateAdapter = true)
+internal data class DeactivateAccountParams(
+        @Json(name = "auth")
+        val auth: UserPasswordAuth? = null,
+
+        // Set to true to erase all data of the account
+        @Json(name = "erase")
+        val erase: Boolean
```
onurays, line 30: I think "erase" is not self-explaining, although we have comment on top of it. What about using "eraseAllData" or "eraseAllAccountData"?

bmarty: Yes, I know... That's an internal class, so not part of the API. I used a more explicit name for the SDK API.
Also it's worth noting that this parameter is not documented at all on the [CS API documentation](https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-account-deactivate)...

vector/src/main/java/im/vector/riotx/features/settings/account/deactivation/DeactivateAccountFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+    private val viewModel: DeactivateAccountViewModel by fragmentViewModel()
+
+    override fun getLayoutResId() = R.layout.fragment_deactivate_account
+
+    override fun onResume() {
+        super.onResume()
+        (activity as? VectorBaseActivity)?.supportActionBar?.setTitle(R.string.deactivate_account_title)
+    }
+
+    private var settingsActivity: VectorSettingsActivity? = null
```
onurays, line 49: I am not sure if it worths creating a shared ViewModel instead of directly accessing Activity...

bmarty: Yes, consider it as a shortcut... This Fragment should never be used anywhere else...

PR #1287, opened 2020-04-27
Remember ignored unknown sessions

BillCarsonFr:
Currently 'unknown device security notice' will keep poping up even if you dismissed it.
This PR will fix that, once dismissed it wont show + it will remember it also for next login.

-> Now users have to go to Security&Privacy to verify a session if the alert was dismissed

-- This will need some re-work, and a cross platform strategy. This PR is a quick change to avoid annoying users indefinitly (also given that some clients -old riot- cannot complete this flow)

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/home/HomeDetailFragment.kt
+++ b/vector/src/main/java/im/vector/riotx/features/home/HomeDetailFragment.kt
@@ -1,4 +1,3 @@
-
 /*
  * Copyright 2019 New Vector Ltd
  *
@@ -61,7 +60,7 @@ class HomeDetailFragment @Inject constructor(
     private val unreadCounterBadgeViews = arrayListOf<UnreadCounterBadgeView>()
 
     private val viewModel: HomeDetailViewModel by fragmentViewModel()
-    private val unknownDeviceDetectorSharedViewModel : UnknownDeviceDetectorSharedViewModel by activityViewModel()
+    private val unknownDeviceDetectorSharedViewModel: UnknownDeviceDetectorSharedViewModel by activityViewModel()
 
     private lateinit var sharedActionViewModel: HomeSharedActionViewModel
 
@@ -112,7 +111,11 @@ class HomeDetailFragment @Inject constructor(
                                             ?.navigator
                                             ?.requestSessionVerification(requireContext(), newest.deviceId ?: "")
                                 }
-                                dismissedAction = Runnable {}
+                                dismissedAction = Runnable {
+                                    unknownDeviceDetectorSharedViewModel.handle(
+                                            UnknownDeviceDetectorSharedViewModel.Action.IgnoreDevice(newest.deviceId ?: "")
+                                    )
+                                }
                             }
                     )
                 }
--- a/vector/src/main/java/im/vector/riotx/features/home/UnknwonDeviceDetectorSharedViewModel.kt
+++ b/vector/src/main/java/im/vector/riotx/features/home/UnknwonDeviceDetectorSharedViewModel.kt
@@ -19,6 +19,7 @@ package im.vector.riotx.features.home
 import com.airbnb.mvrx.Async
 import com.airbnb.mvrx.MvRxState
 import com.airbnb.mvrx.MvRxViewModelFactory
+import com.airbnb.mvrx.Success
 import com.airbnb.mvrx.Uninitialized
 import com.airbnb.mvrx.ViewModelContext
 import im.vector.matrix.android.api.session.Session
@@ -30,9 +31,11 @@ import im.vector.matrix.android.internal.crypto.model.rest.DevicesListResponse
 import im.vector.matrix.rx.rx
 import im.vector.matrix.rx.singleBuilder
 import im.vector.riotx.core.di.HasScreenInjector
-import im.vector.riotx.core.platform.EmptyAction
 import im.vector.riotx.core.platform.EmptyViewEvents
 import im.vector.riotx.core.platform.VectorViewModel
+import im.vector.riotx.core.platform.VectorViewModelAction
+import im.vector.riotx.features.settings.VectorPreferences
+import timber.log.Timber
 import java.util.concurrent.TimeUnit
 
 data class UnknownDevicesState(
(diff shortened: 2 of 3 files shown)


vector/src/main/java/im/vector/riotx/features/home/UnknwonDeviceDetectorSharedViewModel.kt
```diff
@@ -76,12 +97,28 @@ class UnknownDeviceDetectorSharedViewModel(session: Session, initialState: Unkno
...
+                    state.unknownSessions.invoke()?.let {
+                        val updated = it.filter { it.second.deviceId != action.deviceId  }
+                        setState {
+                            copy(unknownSessions = Success(updated))
+                        }
+                    }
+                }
+                ignoredDeviceList.add(action.deviceId)
+                vectorPreferences.storeUnknownDeviceDismissedList(ignoredDeviceList)
+            }
```
bmarty, line 114: please add `.exhaustive`

vector/src/main/java/im/vector/riotx/features/home/UnknwonDeviceDetectorSharedViewModel.kt
```diff
@@ -76,12 +97,28 @@ class UnknownDeviceDetectorSharedViewModel(session: Session, initialState: Unkno
...
+        }
+    }
 
     companion object : MvRxViewModelFactory<UnknownDeviceDetectorSharedViewModel, UnknownDevicesState> {
         override fun create(viewModelContext: ViewModelContext, state: UnknownDevicesState): UnknownDeviceDetectorSharedViewModel? {
             val session = (viewModelContext.activity as HasScreenInjector).injector().activeSessionHolder().getActiveSession()
-            return UnknownDeviceDetectorSharedViewModel(session, state)
+            return UnknownDeviceDetectorSharedViewModel(session, VectorPreferences(viewModelContext.activity()), state)
```
bmarty, line 121: I think this is not the correct way to pass the VectorPreferences here. It should be injected in the ViewModel's constructor. See for instance RoomDetailViewModel

vector/src/main/java/im/vector/riotx/features/settings/VectorPreferences.kt
```diff
@@ -364,6 +367,18 @@ class VectorPreferences @Inject constructor(private val context: Context) {
         return defaultPrefs.getBoolean(SETTINGS_PLAY_SHUTTER_SOUND_KEY, true)
     }
 
+    fun storeUnknownDeviceDismissedList(deviceIds: List<String>) {
+        defaultPrefs.edit(true) {
```
bmarty, line 371: I do not see any reason to commit instead of applying. Please use `.edit {}`

vector/src/main/java/im/vector/riotx/features/settings/VectorPreferences.kt
```diff
@@ -364,6 +367,18 @@ class VectorPreferences @Inject constructor(private val context: Context) {
...
+            putStringSet(SETTINGS_UNKNOWN_DEVICE_DISMISSED_LIST, deviceIds.toSet())
+        }
+    }
+
+    fun getUnknownDeviceDismissedList(): List<String> {
+        return tryThis {
```
bmarty, line 377: Why using tryThis?

vector/src/main/java/im/vector/riotx/features/home/HomeDetailFragment.kt
```diff
@@ -112,7 +111,11 @@ class HomeDetailFragment @Inject constructor(
                                             ?.navigator
                                             ?.requestSessionVerification(requireContext(), newest.deviceId ?: "")
                                 }
-                                dismissedAction = Runnable {}
+                                dismissedAction = Runnable {
+                                    unknownDeviceDetectorSharedViewModel.handle(
+                                            UnknownDeviceDetectorSharedViewModel.Action.IgnoreDevice(newest.deviceId ?: "")
```
bmarty, line 116: Maybe do nothing if newest.deviceId is null or empty (which should not happen anyway)

vector/src/main/java/im/vector/riotx/features/home/UnknwonDeviceDetectorSharedViewModel.kt
```diff
@@ -30,26 +31,43 @@ import im.vector.matrix.android.internal.crypto.model.rest.DevicesListResponse
...
+        session: Session,
+        private val vectorPreferences: VectorPreferences,
+        initialState: UnknownDevicesState)
+    : VectorViewModel<UnknownDevicesState, UnknownDeviceDetectorSharedViewModel.Action, EmptyViewEvents>(initialState) {
+
+    sealed class Action : VectorViewModelAction {
+        data class IgnoreDevice(val deviceId: String) : Action()
+    }
+
+    val ignoredDeviceList = ArrayList<String>()
```
bmarty, line 56: Maybe a bit useless to keep a local cach of this list. We could only used what's in the sharedPref, there will be no diff regarding perf because there is a cache in the shared pref.

vector/src/main/java/im/vector/riotx/features/settings/VectorPreferences.kt
BillCarsonFr: sometimes pref got corrupted. Defensive coding

vector/src/main/java/im/vector/riotx/features/home/UnknwonDeviceDetectorSharedViewModel.kt
BillCarsonFr: I didn't wanted to add the full framework, just to create a view model
Will see in next PR, as it will be reworked (no more one popup per device)

vector/src/main/java/im/vector/riotx/features/home/UnknwonDeviceDetectorSharedViewModel.kt
BillCarsonFr: done

vector/src/main/java/im/vector/riotx/features/home/HomeDetailFragment.kt
BillCarsonFr: Will be reworked in next RP

BillCarsonFr:
This will be modified a lot by aggregated toast, will update when implementing it https://github.com/vector-im/riotX-android/issues/1293

PR #1288, opened 2020-04-27
Fix / backup key was not save in 4S after bootstrap

BillCarsonFr:
Fix / backup key was not saved in QuadS after bootstrap


jluebbe:
@BillCarsonFr I ran into this issue today. Will the fixed version trigger the upload automatically, or is there some manual step I need to do? Alternatively, should I just restart the cross-signing bootstrap from Riot Web?

PR #1291, opened 2020-04-27
Add hint to translators

bmarty:


bmarty:
(Note: this string is not used anymore)

PR #1308, opened 2020-04-30
Fix / Move DM shield rules to task

### LINKED ISSUE #963: CrossSigning / Update Shield Logic for DM
#see https://github.com/vector-im/riot-web/issues/12214

There are some cases where this doesn't happen today, e.g. if you have not verified one of your own devices, all rooms would go red (even large public rooms where you have no intention verifying users).

To make the experience more straightforward, we decided after chatting with @dbkr and @nadonomy to add the additional check that green / red states are only shown if you have verified at least one user in the room. If you haven't, the room is black.

Summary doc on shields:  https://docs.google.com/document/d/1RFUN0YRrOs8QZwiblaUbs4wWR0BjA0hBJ27rKJ57Zvk/edit

BillCarsonFr:
Fixes #963 

The issue has been re-open because sometimes the shield for DM was not following the right rules.
ComputeTrustTask is called at two points, on room membership change, and on crypto device changes (user adding/removing devices), but the DM rules was only applied in one case.
On room member change, the correct rule was applied (ignore yourself in shield trust computation), but on crypto device change it was not done.

=> Refactored so that the specific rule is applied directly by the task

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/crosssigning/ComputeTrustTask.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/crosssigning/ComputeTrustTask.kt
@@ -19,6 +19,7 @@ import im.vector.matrix.android.api.crypto.RoomEncryptionTrustLevel
 import im.vector.matrix.android.api.extensions.orFalse
 import im.vector.matrix.android.api.session.crypto.crosssigning.MXCrossSigningInfo
 import im.vector.matrix.android.internal.crypto.store.IMXCryptoStore
+import im.vector.matrix.android.internal.di.UserId
 import im.vector.matrix.android.internal.task.Task
 import im.vector.matrix.android.internal.util.MatrixCoroutineDispatchers
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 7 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/crosssigning/ComputeTrustTask.kt
```diff
@@ -19,24 +19,37 @@ import im.vector.matrix.android.api.crypto.RoomEncryptionTrustLevel
...
     data class Params(
-            val userIds: List<String>
+            val activeMemberUserIds: List<String>,
+            val isDirectRoom: Boolean
     )
 }
 
 internal class DefaultComputeTrustTask @Inject constructor(
         private val cryptoStore: IMXCryptoStore,
+        @UserId private val myUserId: String,
```
bmarty, line 37: convention is to use `userId` for myself and `otherUserId` for others'

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/RoomSummaryUpdater.kt
```diff
@@ -161,15 +161,7 @@ internal class RoomSummaryUpdater @Inject constructor(
...
-                // The set of “all users” depends on the type of room:
-                // For regular / topic rooms, all users including yourself, are considered when decorating a room
-                // For 1:1 and group DM rooms, all other users (i.e. excluding yourself) are considered when decorating a room
-                val listToCheck = if (roomSummaryEntity.isDirect) {
-                    roomSummaryEntity.otherMemberIds.toList()
-                } else {
-                    roomSummaryEntity.otherMemberIds.toList() + userId
-                }
-                eventBus.post(SessionToCryptoRoomMembersUpdate(roomId, listToCheck))
+                eventBus.post(SessionToCryptoRoomMembersUpdate(roomId, roomSummaryEntity.isDirect, roomSummaryEntity.otherMemberIds.toList() + userId))
```
bmarty, line 164: It's strange to add `userId` to the list here, because, IIUC, after it maybe filtered out. Why not adding a new member to `SessionToCryptoRoomMembersUpdate`, and rename `SessionToCryptoRoomMembersUpdate.userIds` to  `SessionToCryptoRoomMembersUpdate.otherUserIds`?

PR #1315, opened 2020-04-30
Feature/forward pagination

### LINKED ISSUE #1008: After jump to unread, newer messages are never loaded
When hitting the "jump to unread messages" banner and there are lots of them, it fails to load messages when you scroll down. I have to switch to a different channel and switch back and then manually scroll back to find where I left off in order to see the stuff in between.
I've had that happen to me 2 or 3 times the last couple days
It just sits and spins like in the screenshot forever
Screenshot_20200213-131533_RiotX

bmarty:
This PR fixes #1008 

It also display the room creation event, if all events have to be displayed in the timeline (in developer mode).

Other cleanup.

All 3 tests added are passing
image


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/timeline/DefaultTimeline.kt
```diff
@@ -516,11 +516,17 @@ internal class DefaultTimeline(
...
+     * This has to be called on TimelineThread as it accesses realm live results
      */
     private fun executePaginationTask(direction: Timeline.Direction, limit: Int) {
         val token = getTokenLive(direction)
         if (token == null) {
+            val currentChunk = getLiveChunk()
+            if (direction == Timeline.Direction.FORWARDS && currentChunk?.hasBeenALastForwardChunk() == true) {
+                // We are in the case that next event exists, but we do not know the next token.
+                // Fetch (again) the last event to get a nextToken
+                currentChunk.timelineEvents.lastOrNull()?.eventId?.let { fetchEvent(it) }
```
ganfra, line 528: Don't get lastOrNull on chunk as it's not sorted. I think you should use the nonFilteredEvents instead and take the first which is the most recent.

PR #1345, opened 2020-05-09
Add direct shortcuts

### LINKED ISSUE #652: Use DirectShare API
As @bbigras says in #58 , it should be a good idea to use the [DirectShare](https://developer.android.com/about/versions/marshmallow/android-6.0#direct-share) or the new [Sharing Shortcuts API](https://developer.android.com/preview/features/sharing).

### LINKED ISSUE #659: Use new androidx.fragment 1.2.0
See https://www.amryousef.me/new-in-androidx-fragment
FragmentContainerView replace FrameLayout
We also have a FragmentFactory so we can replace Field injection by constructor injection!

emmaguy:
### Pull Request Checklist

👋 Hello! I started using Riot recently and found myself missing having direct shortcuts, so I added them. Fixes #652.

I kept it simple and just allow creating shortcuts from rooms which have been favourited. Android restricts the number of shortcuts you can have to 4, so this seemed like a reasonable thing to do!

I wasn't sure the best place to trigger the creation of the shortcuts - so I decided to do so in home. So when the user opens the app, we'll subscribe and start listening for room changes, and update the available shortcuts as needed. 

Tested on a Pixel 1 running Android 9.x. I've tried to get it going on an Android 6.x or 7.x emulators, but I couldn't get the app installing. Hardware acceleration isn't available on the older emulators, and installing the app just times out 😭. I've used the AndroidX backwards compatible implementation, so that should take care of the earlier versions.

Here's how it looks to set some shortcuts up - I've tested with rooms which do and don't have images:

| add shortcut dialog | example shortcuts  |
|---|---|
| device-2020-05-09-205940 | device-2020-05-09-205607 |

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@ Changes in RiotX 0.20.0 (2020-XX-XX)
 ===================================================
 
 Features ✨:
- -
+ - Add Direct Shortcuts
 
 Improvements 🙌:
  -
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@ Changes in RiotX 0.20.0 (2020-XX-XX)
 ===================================================
 
 Features ✨:
- - Add Direct Shortcuts
+ - Add Direct Shortcuts (#652)
 
 Improvements 🙌:
  -
--- a/vector/src/main/java/im/vector/riotx/features/home/HomeActivity.kt
+++ b/vector/src/main/java/im/vector/riotx/features/home/HomeActivity.kt
@@ -65,6 +65,7 @@ class HomeActivity : VectorBaseActivity(), ToolbarConfigurable {
     @Inject lateinit var notificationDrawerManager: NotificationDrawerManager
     @Inject lateinit var vectorPreferences: VectorPreferences
     @Inject lateinit var popupAlertManager: PopupAlertManager
+    @Inject lateinit var shortcutsHandler: ShortcutsHandler
 
     private val drawerListener = object : DrawerLayout.SimpleDrawerListener() {
         override fun onDrawerStateChanged(newState: Int) {
@@ -144,6 +145,8 @@ class HomeActivity : VectorBaseActivity(), ToolbarConfigurable {
                 && activeSessionHolder.getSafeActiveSession()?.hasAlreadySynced() == true) {
             promptCompleteSecurityIfNeeded()
         }
+
+        shortcutsHandler.observeRoomsAndBuildShortcuts(context = this)
     }
 
     private fun promptCompleteSecurityIfNeeded() {
--- a/vector/src/main/java/im/vector/riotx/features/home/HomeActivity.kt
+++ b/vector/src/main/java/im/vector/riotx/features/home/HomeActivity.kt
@@ -146,7 +146,8 @@ class HomeActivity : VectorBaseActivity(), ToolbarConfigurable {
             promptCompleteSecurityIfNeeded()
         }
 
-        shortcutsHandler.observeRoomsAndBuildShortcuts(context = this)
+        shortcutsHandler.observeRoomsAndBuildShortcuts()
+                .disposeOnDestroy()
     }
 
     private fun promptCompleteSecurityIfNeeded() {
--- a/vector/src/main/java/im/vector/riotx/features/home/ShortcutsHandler.kt
+++ b/vector/src/main/java/im/vector/riotx/features/home/ShortcutsHandler.kt
@@ -16,7 +16,6 @@
 
 package im.vector.riotx.features.home
 
-import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.Bitmap
 import android.os.Build
@@ -28,6 +27,7 @@ import im.vector.matrix.android.api.util.toMatrixItem
 import im.vector.riotx.core.glide.GlideApp
 import im.vector.riotx.core.utils.DimensionConverter
 import im.vector.riotx.features.home.room.detail.RoomDetailActivity
+import io.reactivex.disposables.Disposable
 import io.reactivex.schedulers.Schedulers
 import javax.inject.Inject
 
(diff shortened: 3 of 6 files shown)


vector/src/main/java/im/vector/riotx/features/home/ShortcutsHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+import im.vector.riotx.core.utils.DimensionConverter
+import im.vector.riotx.features.home.room.detail.RoomDetailActivity
+import io.reactivex.schedulers.Schedulers
+import javax.inject.Inject
+
+private val useAdaptiveIcon = Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
+private const val adaptiveIconSizeDp = 108
+private const val adaptiveIconOuterSidesDp = 18
+
+class ShortcutsHandler @Inject constructor(
```
emmaguy, line 38: Named this class to mirror `AppStateHandler` which is observing some room-related state.

vector/src/main/java/im/vector/riotx/features/home/room/detail/RoomDetailActivity.kt
```diff
@@ -110,11 +116,20 @@ class RoomDetailActivity : VectorBaseActivity(), ToolbarConfigurable {
...
+        const val EXTRA_ROOM_ID = "EXTRA_ROOM_ID"
+        const val ACTION_ROOM_DETAILS_FROM_SHORTCUT = "ROOM_DETAILS_FROM_SHORTCUT"
 
         fun newIntent(context: Context, roomDetailArgs: RoomDetailArgs): Intent {
             return Intent(context, RoomDetailActivity::class.java).apply {
                 putExtra(EXTRA_ROOM_DETAIL_ARGS, roomDetailArgs)
             }
         }
+
+        fun shortcutIntent(context: Context, roomId: String): Intent {
```
emmaguy, line 128: It's annoying that this is required, but shortcuts can't have intents with parcelables :(

vector/src/main/java/im/vector/riotx/features/home/ShortcutsHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+        private val homeRoomListStore: HomeRoomListDataSource,
+        private val avatarRenderer: AvatarRenderer,
+        private val dimensionConverter: DimensionConverter
```
bmarty, line 41: The application context can be injected here with `private val context: Context`. It's better than passing the Activity in parameter of observeRoomsAndBuildShortcuts()

vector/src/main/java/im/vector/riotx/features/home/ShortcutsHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+) {
+
+    @SuppressLint("CheckResult")
+    fun observeRoomsAndBuildShortcuts(context: Context) {
```
bmarty, line 45: Remove context parameter and return the disposable. The stream has to be disposed to avoid memory leak. It will be disposed once the Activity is destroyed (see other comment)

vector/src/main/java/im/vector/riotx/features/home/ShortcutsHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+                            .map { room ->
+                                val intent = RoomDetailActivity.shortcutIntent(context, room.roomId)
+
+                                val adaptiveIconSize = dimensionConverter.dpToPx(adaptiveIconSizeDp)
+                                val adaptiveIconOuterSides = dimensionConverter.dpToPx(adaptiveIconOuterSidesDp)
+                                val size = if (useAdaptiveIcon) {
+                                    adaptiveIconSize - adaptiveIconOuterSides
+                                } else {
+                                    dimensionConverter.dpToPx(72)
+                                }
```
bmarty, line 62: It's a shame to compute the same size for each room, and each time the room list change. Instead this computation should be done in the init {} block of this class.

vector/src/main/java/im/vector/riotx/features/home/room/detail/RoomDetailActivity.kt
bmarty: Ok

vector/src/main/java/im/vector/riotx/features/home/HomeActivity.kt
```diff
@@ -144,6 +145,8 @@ class HomeActivity : VectorBaseActivity(), ToolbarConfigurable {
                 && activeSessionHolder.getSafeActiveSession()?.hasAlreadySynced() == true) {
             promptCompleteSecurityIfNeeded()
         }
+
+        shortcutsHandler.observeRoomsAndBuildShortcuts(context = this)
```
bmarty, line 149: Remove the context parameter and add a call to `disposeOnDestroy()` on the disposable

vector/src/main/java/im/vector/riotx/features/home/ShortcutsHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+
+                                ShortcutInfoCompat.Builder(context, room.roomId)
+                                        .setShortLabel(room.displayName)
+                                        .setIcon(bitmap.toProfileImageIcon())
+                                        .setIntent(intent)
+                                        .build()
+                            }
+
+                    ShortcutManagerCompat.removeAllDynamicShortcuts(context)
+                    ShortcutManagerCompat.addDynamicShortcuts(context, shortcuts)
```
bmarty, line 74: The shrotcuts do not need to be updated if the list is the same. You should add a distinct() call somewhere, either in th Rx stream, or by keeping a memory copy of the current shortcut. This API is rate limited, because it can consume system resource (just read that from the doc)

CHANGES.md
```diff
@@ -2,7 +2,7 @@ Changes in RiotX 0.20.0 (2020-XX-XX)
 ===================================================
 
 Features ✨:
- -
+ - Add Direct Shortcuts
```
bmarty, line 5:
```suggestion
 - Add Direct Shortcuts (#652)
```

vector/src/main/java/im/vector/riotx/features/home/ShortcutsHandler.kt
emmaguy: Ah, great idea 👌 this is much better

emmaguy:
That should be everything addressed - thanks for the feedback! 🚀

bmarty:
Hi @emmaguy , thanks for the update!
I merge the PR

PR #1368, opened 2020-05-14
Feature/switch language

### LINKED ISSUE #41: Switch language support
- [ ] Import String resource from Riot

See #40

### LINKED ISSUE #69: Switch font size support
Related to #40

### LINKED ISSUE #645: RiotX forgets configured font size
Since the latest update (0.7.0) from the play store, RiotX sometimes reverts to using "normal" as font size, even though I have set it to "small". This only happens when switching to the app, not while it is in the foreground.

When this happens, the font size in settings still shows "small" and resetting it to actually be "small" entails switching to "normal", leaving settings, entering settings again and then setting it to "small".

bmarty:
Switch language support (Fixes #41)

Also fix issue with font setting not always working (Fixes #69, Fixes #645)

There is also a little addition to the template

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/settings/locale/LocalePickerViewModel.kt
+++ b/vector/src/main/java/im/vector/riotx/features/settings/locale/LocalePickerViewModel.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2020 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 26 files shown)


vector/src/main/java/im/vector/riotx/features/settings/locale/LocalePickerViewModel.kt
```diff
@@ -0,0 +1,75 @@
...
+) : VectorViewModel<LocalePickerViewState, LocalePickerAction, LocalePickerViewEvents>(initialState) {
+
+    @AssistedInject.Factory
+    interface Factory {
+        fun create(initialState: LocalePickerViewState): LocalePickerViewModel
+    }
+
+    init {
+        viewModelScope.launch {
+            val result = VectorLocale.getSupportedLocales()
```
ganfra, line 43: Should be injected instead of being an object I guess

bmarty: Yes, I tried, but it's a mess with VectorConfiguration, which is injected AND created using constructor

PR #1374, opened 2020-05-15
support new key agreement method for SAS

BillCarsonFr:
See https://github.com/matrix-org/matrix-js-sdk/pull/1376


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/verification/DefaultVerificationService.kt
```diff
@@ -610,7 +610,7 @@ internal class DefaultVerificationService @Inject constructor(
 
         if (validCancelReq == null) {
             // ignore
-            Timber.e("## SAS Received invalid key request")
+            Timber.e("## SAS Received invalid cancelt request")
```
uhoreg, line 613:
```suggestion
            Timber.e("## SAS Received invalid cancel request")
```

PR #1376, opened 2020-05-17
Fix gallery intent for certain apps

dkanada:
Let me know if you'd be interested in merging this and I'll add the rest of the required items. You can see this ancient [link](https://github.com/vector-im/riot-android/pull/2583/files/0fea24848583ebad93c613689aa155be8603e1f9..fe363436e446813061d13f6796bd9619b8620e94#r217405702) for the reason I made the change. Lots of gallery apps don't like the open document intent and won't show up properly.


dkanada:
@bmarty do you have any thoughts about this change? I'll close it if you want to keep the current intent but it seems like the other one might have better support to me.

dkanada:
You can test this change out with the Camera Roll app on F-Droid, which will only show up in the selection menu with the intent I added here.

dkanada:
In another pull request to Riot, I had applied looping to video content in the image view fragment but it was rejected. I still really enjoy looping video content so would an option in the settings be sufficient to apply that change?

MattDiesel:
@dkanada this sounds like it could be a possible fix for #1413 ? Possibly with the addition of FLAG_GRANT_READ_URI_PERMISSION as mentioned in this ucrop bug report: https://github.com/Yalantis/uCrop/issues/669 

Or just ignore me if I'm talking nonsense.

dkanada:
I don't believe this would resolve that issue, since it's an internal activity.

dkanada:
@bmarty if possible, I would also like to address the inconsistent capitalization of strings in this pull request. There seems to be no pattern between `olm version`, `Terms & conditions`, `Help & About`, `Advanced settings`, and similar strings. There is also inconsistent usage of punctuation for descriptions and unexplained newlines. I prefer title casing but sentence casing would also be fine.

dkanada:
It's good to know my attempt to build yesterday wasn't a fluke and the CI is also unable to find the OLM library.

bmarty:
According to https://developer.android.com/guide/topics/providers/document-provider and specifically

image

it seems that you are right and it's better to use `ACTION_GET_CONTENT` in our use case.

@onurays WDYT?

dkanada:
That would explain why my gallery app only supports the latter option. I'm not sure many third party galleries would include an intent filter for long term, persistent access to documents. I'll open a new PR for the casing when I have some free time.

onurays:
I just tested with `ACTION_GET_CONTENT` and it is working well. And now we can see other providers like Google Photos. Thanks for the PR, we can quickly test and merge after improvements on other file types.

WhatsApp Image 2020-09-04 at 14 49 39

bmarty:
@dkanada thanks for the rebase. Is this change also applicable to the 3 other Pickers?

dkanada:
I believe so. Changing it for the audio picker should display the sound recorder as an option on LineageOS for example, but I haven't actually tested anything besides the image picker.

dkanada:
Should I fix the merge conflicts?

bmarty:
I will handle the conflicts, thanks

PR #1380, opened 2020-05-17
Fixing french typo

Simounet:
Hi there,
I'm just fixing a french typo.
Have a nice day and stay safe.


bmarty:
Thanks. I've reported the change on [weblate](https://translate.riot.im/projects/riot-android/), which is were we managed translations, so it will be applied at the next weblate sync. This PR does not need to be merged.

PR #1383, opened 2020-05-18
Remove dependency to event bus

bmarty:
Using an EventBus library is maybe a workaround to handle things not really properly.
As we use dependency injection, it is not necessary to use an EventBus
This PR removes the dependency, one commit per step


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/crypto/crosssigning/ShieldTrustUpdaterInput.kt
```diff
@@ -13,19 +13,18 @@
...
-package im.vector.matrix.android.internal.eventbus
+import im.vector.matrix.android.internal.session.SessionScope
+import javax.inject.Inject
 
-import org.greenrobot.eventbus.Logger
-import timber.log.Timber
-import java.util.logging.Level
+@SessionScope
+internal class ShieldTrustUpdaterInput @Inject constructor() {
+    var listener: Listener? = null
```
ganfra, line 23: list of listeners?

bmarty:
I have to work again on that old PR...

PR #1411, opened 2020-05-26
Feature/attachment e2e

### LINKED ISSUE #1405: attachments list doesn't work in e2ee rooms
It just displays

> There are no media in this room
> There are no files in this room

bmarty:
Fixes #1405 

The added code displays the list of already synced events with attachment in the upload section of a e2e room. It may be improve later, because the list of files should not be exhaustive.

The PR also contain a fix to avoid duplicated event in DB. I have done some tests and it seems ok, but please review carefully...

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/helper/TimelineEventEntityHelper.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/helper/TimelineEventEntityHelper.kt
@@ -18,6 +18,7 @@ package im.vector.matrix.android.internal.database.helper
 
 import im.vector.matrix.android.internal.database.model.TimelineEventEntity
 import im.vector.matrix.android.internal.database.model.TimelineEventEntityFields
+import im.vector.matrix.android.internal.extensions.assertIsManaged
 import io.realm.Realm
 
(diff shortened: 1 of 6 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/query/EventEntityQueries.kt
```diff
@@ -25,9 +25,12 @@ import io.realm.kotlin.where
...
     return realm.where<EventEntity>()
-        .equalTo(EventEntityFields.EVENT_ID, eventId)
-        .equalTo(EventEntityFields.ROOM_ID, roomId)
-        .findFirst() ?: realm.copyToRealm(this)
+            .equalTo(EventEntityFields.EVENT_ID, eventId)
+            .equalTo(EventEntityFields.ROOM_ID, roomId)
+            .findFirst()
+            // If Event already exist, ensure state is up to date
+            ?.also { it.sendState = sendState }
+            ?: realm.copyToRealm(this)
```
bmarty, line 33: Actually there is maybe a pb here: synced event content will not replace what we had for the local echo anymore. Synced event may contain other field (in unsigned data?)

PR #1423, opened 2020-05-29
Yet another cleanup PR

bmarty:
- use invalidate() from MvRx
- use throttleFirst instead of debounce
- remove useless observe on main thread
- remove useless calls to super.invalidate()

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecuredStorageKeyFragment.kt
+++ b/vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecuredStorageKeyFragment.kt
@@ -28,6 +28,7 @@ import im.vector.matrix.android.api.extensions.tryThis
 import im.vector.riotx.R
 import im.vector.riotx.core.platform.VectorBaseFragment
 import im.vector.riotx.core.utils.startImportTextFromFileIntent
+import io.reactivex.android.schedulers.AndroidSchedulers
 import kotlinx.android.synthetic.main.fragment_ssss_access_from_key.*
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
@@ -44,6 +45,7 @@ class SharedSecuredStorageKeyFragment @Inject constructor() : VectorBaseFragment
 
         ssss_key_enter_edittext.editorActionEvents()
                 .throttleFirst(300, TimeUnit.MILLISECONDS)
+                .observeOn(AndroidSchedulers.mainThread())
                 .subscribe {
                     if (it.actionId == EditorInfo.IME_ACTION_DONE) {
                         submit()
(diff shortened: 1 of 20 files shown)


vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecuredStorageKeyFragment.kt
```diff
@@ -48,8 +43,7 @@ class SharedSecuredStorageKeyFragment @Inject constructor(
         ssss_restore_with_key_text.text = getString(R.string.enter_secret_storage_input_key)
 
         ssss_key_enter_edittext.editorActionEvents()
-                .debounce(300, TimeUnit.MILLISECONDS)
-                .observeOn(AndroidSchedulers.mainThread())
+                .throttleFirst(300, TimeUnit.MILLISECONDS)
```
ganfra, line 46: Why do you replace debounce by throttle? In the context of an edittext it's better to keep with debounce

vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecuredStorageKeyFragment.kt
```diff
@@ -48,8 +43,7 @@ class SharedSecuredStorageKeyFragment @Inject constructor(
         ssss_restore_with_key_text.text = getString(R.string.enter_secret_storage_input_key)
 
         ssss_key_enter_edittext.editorActionEvents()
-                .debounce(300, TimeUnit.MILLISECONDS)
-                .observeOn(AndroidSchedulers.mainThread())
```
ganfra, line 52: Also, why removing mainThread? the submit method is changing some ui, we might get into trouble if we are not switching back to the ui thread

vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecuredStorageKeyFragment.kt
bmarty: We do not observe textChange here, but editorActionEvents(), to get EditorInfo.IME_ACTION_DONE event. So I think throttleFirst is better.

vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecuredStorageKeyFragment.kt
bmarty: we are in the main thread, editorActionEvents() is supposed to emit only on main thread (no?) and debounce does not change the thread, so I think observeOn(AndroidSchedulers.mainThread()) is not necesarry. But maybe I'm wrong?

ganfra: Deboune and throttle does change thread by default

vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecuredStorageKeyFragment.kt
ganfra: Oh yes, didn't see that

vector/src/main/java/im/vector/riotx/features/crypto/quads/SharedSecuredStorageKeyFragment.kt
bmarty: Ok, I reverted this change

PR #1436, opened 2020-06-02
Hide "X made no changes" event by default in timeline (#1430)

### LINKED ISSUE #1430: X made no changes event in timeline
I sometimes see `X made no changes` in the timeline in riotX. Riot-desktop filters such empty state events, I guess these shouldn't be shown in the timeline except when "Show hidden events" is selected from developer settings.

This is the event in question.
image

bmarty:
Fixes #1430 

Before PR:

image

After PR:

image


matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/timeline/TimelineSettings.kt
```diff
@@ -32,6 +32,10 @@ data class TimelineSettings(
          * A flag to filter redacted events
          */
         val filterRedacted: Boolean = false,
+        /**
+         * A flag to filter useless events, such as membership events without any change
+         */
+        val filterUseless: Boolean = false,
```
ganfra, line 38: isUseless seems a bit hard for a sdk naming

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/mapper/EventMapper.kt
```diff
@@ -38,6 +39,7 @@ internal object EventMapper {
         eventEntity.content = ContentMapper.map(event.content)
         val resolvedPrevContent = event.prevContent ?: event.unsignedData?.prevContent
         eventEntity.prevContent = ContentMapper.map(resolvedPrevContent)
+        eventEntity.isUseless = event.type == EventType.STATE_ROOM_MEMBER && eventEntity.content == eventEntity.prevContent
```
ganfra, line 42: I don't think it should be saved in db?

bmarty: It's saved in DB to be able to do query on it. Maybe you known a better way to do query which check if two value in the DB are identical?

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/timeline/TimelineSettings.kt
bmarty: It's true anyway, it's useless...

ganfra: Yes but it sounds a bit weird to use as a sdk user... but ok for me anyway

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/mapper/EventMapper.kt
ganfra: Oh, right, thanks to realm for not being flexible enough...
Could you create a class like IsUselessResolver to handle this?

bmarty: I thought about it, but, it will call again what is called in the three line above, so for the perf, as we are inserting event in DB here, also from initial sync, I do not know if it's a good idea... What do you think?

bmarty: Ok done, I let you check

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/mapper/IsUselessResolver.kt
```diff
@@ -0,0 +1,37 @@
...
+internal object IsUselessResolver {
+
+    /**
+     * @return true if the event is useless
+     */
+    fun isUseless(event: Event): Boolean {
+        return when (event.type) {
+            EventType.STATE_ROOM_MEMBER -> {
+                // Call toContent(), to filter out null value
+                event.content?.toContent() == event.resolvedPrevContent()?.toContent()
```
ganfra, line 32: I don't think we can have both content and prevContent as null. And if they are, they are usesell as well, can't you juste compare the contents directly?

Issue #766, opened 2019-12-11
Left rooms are still displayed in breadcrumbs


bmarty:

BillCarsonFr:
See https://github.com/vector-im/riot-web/issues/11644

bmarty:
So it's a feature, because you can eventually re-join the left room :)

steef435:
Currently, the breadcrumb items of left rooms also get a green "new notifications" bar. Even though a design decision has to be made on whether they should stick or not after leaving the room, could those highlights be removed in any case?

bmarty:
Fixed in #1446

PR #1446, opened 2020-06-04
Hide left rooms in breadcrumbs (#766)

bmarty:
User get confused to see left rooms in the breadcrumbs.
I decided to filter them out.
Note that they are still present in the breadcrumbs account data, but should naturally be excluded once the user has navigated to any 20 other rooms.
If the user joins the room again, it may appear in its current breadcrumb index, which is OK I think.


matrix-sdk-android-rx/src/main/java/im/vector/matrix/rx/RxSession.kt
```diff
@@ -56,10 +56,10 @@ class RxSession(private val session: Session) {
                 }
     }
 
-    fun liveBreadcrumbs(): Observable<List<RoomSummary>> {
-        return session.getBreadcrumbsLive().asObservable()
+    fun liveBreadcrumbs(onlyJoinedRooms: Boolean): Observable<List<RoomSummary>> {
```
ganfra, line 59: As we are returning RoomSummary, shouldn't we use the RoomSummaryQueryParams?

bmarty: Yrs, I've thought about it, will do the change

PR #1447, opened 2020-06-04
Feature/fix version

### LINKED ISSUE #1442: Error when handling /version responses
Response to /versions:
1.  [r0.5.0, r0.6.0] works
2.  [r0.5.0, r0.6.1] doesn't work
3.  [r0.6.0] doesn't work

If 1. works, I would expect 2. to work too, because it's just a minor version upgrade. [Servers have to send only the latest minor version.](https://matrix.org/docs/spec/client_server/unstable#get-matrix-client-versions).

bmarty:
Fixes #1442 

With this change, the fact that homeserver sends a list of versions is reduced to checking the latest version it supports.

So maybe this is not right to do that, I do not know.

Anyway I think it is OK.

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/auth/version/HomeServerVersion.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/auth/version/HomeServerVersion.kt
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2020 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/auth/version/HomeServerVersion.kt
```diff
@@ -0,0 +1,60 @@
...
+        internal fun parse(value: String): HomeServerVersion? {
+            val result = pattern.matchEntire(value) ?: return null
+            return HomeServerVersion(
+                    major = result.groupValues[1].toInt(),
+                    minor = result.groupValues[2].toInt(),
+                    patch = result.groupValues[3].toInt()
+            )
+        }
+
+        val r0_0_1 = parse("r0.0.1")!!
```
ganfra, line 52: Is it really necessary to use parsing here?

bmarty: No, I've change it, thanks. Also mini version is now `r0.0.0` (instead of `r0.0.1`)

PR #1448, opened 2020-06-04
Feature/fix

### LINKED ISSUE #1443: /register response handling is outdated
RiotX seems to expect that the server responds with a `home_server` field, but that's deprecated in the spec.

bmarty:
Fixes #1443


matrix-sdk-android/src/main/java/im/vector/matrix/android/api/auth/data/Credentials.kt
```diff
@@ -45,7 +45,7 @@ data class Credentials(
          * @Deprecated. Clients should extract the server_name from user_id (by splitting at the first colon)
          * if they require it. Note also that homeserver is not spelt this way.
          */
-        @Json(name = "home_server") val homeServer: String,
+        @Json(name = "home_server") val homeServer: String?,
```
ganfra, line 48: Is the server returning null value? Or is it just absent from response?

bmarty: I do not know.

PR #1478, opened 2020-06-11
Feature/proguard

ganfra:
This PR activates proguard in release to shrink and optimize code.
It doesn't activate obfuscation as it makes it hard then to read crash reports. 
Please try with care.

### DIFF
--- a/matrix-sdk-android/proguard-rules.pro
+++ b/matrix-sdk-android/proguard-rules.pro
@@ -19,3 +19,45 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+
+### EVENT BUS ###
+
+-keepattributes *Annotation*
+-keepclassmembers class * {
+    @org.greenrobot.eventbus.Subscribe <methods>;
+}
+-keep enum org.greenrobot.eventbus.ThreadMode { *; }
+
+### MOSHI ###
+
+# JSR 305 annotations are for embedding nullability information.
+
+-dontwarn javax.annotation.**
+
+-keepclasseswithmembers class * {
+    @com.squareup.moshi.* <methods>;
+}
+
+-keep @com.squareup.moshi.JsonQualifier interface *
+
+# Enum field names are used by the integrated EnumJsonAdapter.
+# values() is synthesized by the Kotlin compiler and is used by EnumJsonAdapter indirectly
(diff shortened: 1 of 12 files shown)


matrix-sdk-android/proguard-rules.pro
```diff
@@ -19,3 +19,45 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+
+### EVENT BUS ###
```
bmarty, line 24: I like when config is split into several files

bmarty:
Please add a line in CHANGE, enabling proguard can lead to hard-to-find bug (because not reproduced on debug mode), it will help to get some clue about it :)

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/model/Membership.kt
```diff
@@ -17,10 +17,12 @@
 package im.vector.matrix.android.api.session.room.model
 
 import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
 
 /**
  * Represents the membership of a user on a room
  */
+@JsonClass(generateAdapter = false)
```
bmarty, line 25: I assume it's necessary for PG, but why `false`? This is the first time we use `false` :)

vector/build.gradle
```diff
@@ -192,8 +192,14 @@ android {
             resValue "bool", "debug_mode", "false"
             buildConfigField "boolean", "LOW_PRIVACY_LOG_ENABLE", "false"
 
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+
+            postprocessing {
+                removeUnusedCode true
+                removeUnusedResources true
+                obfuscate false
```
bmarty, line 199: Maybe comment here why we do not obfuscate

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/room/model/Membership.kt
ganfra: It's the rule for proguard recommanded by Moshi devs. It's to make it clear it shouldn't be removed but we don't need adapter as it use the EnumJsonAdapter from Moshi.

vector/build.gradle
bmarty: I do it

PR #1479, opened 2020-06-11
Feature/memory leaks

ganfra:
This PR looks at some memory leaks discovered with LeakCanary.
The dependency has been commented as we have users using dbg builds...
What do you think about it?


vector/src/main/java/im/vector/riotx/core/epoxy/LayoutManagerStateRestorer.kt
```diff
@@ -25,12 +25,17 @@ import java.util.concurrent.atomic.AtomicReference
 
 private const val LAYOUT_MANAGER_STATE = "LAYOUT_MANAGER_STATE"
 
-class LayoutManagerStateRestorer(private val layoutManager: RecyclerView.LayoutManager) : Restorable, DefaultListUpdateCallback {
+class LayoutManagerStateRestorer(layoutManager: RecyclerView.LayoutManager) : Restorable, DefaultListUpdateCallback {
 
+    var layoutManager: RecyclerView.LayoutManager? = null
```
bmarty, line 30: Maybe keep it private and add a `clear()` method?

vector/src/main/java/im/vector/riotx/features/autocomplete/command/AutocompleteCommandPresenter.kt
```diff
@@ -32,8 +32,6 @@ class AutocompleteCommandPresenter @Inject constructor(context: Context,
     }
 
     override fun instantiateAdapter(): RecyclerView.Adapter<*> {
-        // Also remove animation
-        recyclerView?.itemAnimator = null
```
bmarty, line 36: Thanks for handling correctly this :)

PR #1520, opened 2020-06-21
Random housekeeping indicated by Lint

johnjohndoe:
# Description
- Here is few improvements which Lint indicated when running it on the project.
- The last commit contains a fix where a self-assignment looks wrong.

Signed-off-by: Tobias Preuss <tobias.preuss@googlemail.com>

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/notification/DefaultRoomPushRuleService.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/notification/DefaultRoomPushRuleService.kt
@@ -51,7 +51,7 @@ internal class DefaultRoomPushRuleService @AssistedInject constructor(@Assisted 
     override fun setRoomNotificationState(roomNotificationState: RoomNotificationState, matrixCallback: MatrixCallback<Unit>): Cancelable {
         return setRoomNotificationStateTask
                 .configureWith(SetRoomNotificationStateTask.Params(roomId, roomNotificationState)) {
-                    this.callback = callback
+                    this.callback = matrixCallback
                 }
                 .executeBy(taskExecutor)
     }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,7 +22,8 @@ Build 🧱:
  - SDK is now API level 21 minimum, and so RiotX (#405)
 
 Other changes:
- -
+ - Fix self-assignment of callback in `DefaultRoomPushRuleService#setRoomNotificationState` (#1520)
+ - Random housekeeping clean-ups indicated by Lint (#1520)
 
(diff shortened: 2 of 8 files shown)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/notification/DefaultRoomPushRuleService.kt
```diff
@@ -51,7 +51,7 @@ internal class DefaultRoomPushRuleService @AssistedInject constructor(@Assisted
     override fun setRoomNotificationState(roomNotificationState: RoomNotificationState, matrixCallback: MatrixCallback<Unit>): Cancelable {
         return setRoomNotificationStateTask
                 .configureWith(SetRoomNotificationStateTask.Params(roomId, roomNotificationState)) {
-                    this.callback = callback
```
bmarty, line 54: This should be treated as an error by the compiler...

johnjohndoe:
> Is it in the Draft state because you want to add some commits?

No, I just forgot to change the state.
I rebased the branch to ease integration and facilitate linear history. Feel free to rebase it again before you merged.

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/notification/DefaultRoomPushRuleService.kt
johnjohndoe: Very true.

Dominaezzz: Wouldn't happen with suspend functions. :stuck_out_tongue:

johnjohndoe: Why not?

Dominaezzz: With suspend functions the compiler passes the callbacks for you, so you don't get it wrong. (Unless you touch the unsafe low-level utilities)

johnjohndoe: Ah, you mean the `Continuation` object will encapsulate callback. Interesting. Maybe, this would fit here. - Still, I agree with @bmarty that the compiler should warn us a bit harder.

bmarty: Indeed. One day we will probalby use coroutines every where in the SDK API

PR #1530, opened 2020-06-22
Add capability to change the display name (#1529)

### LINKED ISSUE #1529: Setting user (global) display name
Along with #1054.

It should not be long to implement.

bmarty:
Fixes #1529


vector/src/main/java/im/vector/riotx/features/settings/VectorSettingsGeneralFragment.kt
```diff
@@ -858,44 +858,23 @@ class VectorSettingsGeneralFragment : VectorSettingsBaseFragment() {
      * Update the displayname.
      */
     private fun onDisplayNameClick(value: String?) {
-        notImplemented()
-        /* TODO
-        if (!TextUtils.equals(session.myUser.displayname, value)) {
+        value ?: return
```
onurays, line 861: We don't need that "ugly" check, because currentDisplayName cannot be null (2 lines below ?: ""). But we can keep it if getting user from session is an expensive operation.

bmarty: I change that, thanks

PR #1534, opened 2020-06-23
Pagination do not set ageLocalTs

BillCarsonFr:
Call events from back pagination can trigger the call UX (notification, then call activity)
Looks like the ageLocalTS check is failing; because ageLocalTS was not correctly set when paginating

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/timeline/TokenChunkEventPersistor.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/timeline/TokenChunkEventPersistor.kt
@@ -207,13 +207,15 @@ internal class TokenChunkEventPersistor @Inject constructor(@SessionDatabase pri
                 roomMemberContentsByUser[stateEvent.stateKey] = stateEvent.content.toModel<RoomMemberContent>()
             }
         }
+        val now = System.currentTimeMillis()
         val eventIds = ArrayList<String>(eventList.size)
         for (event in eventList) {
             if (event.eventId == null || event.senderId == null) {
                 continue
             }
+            val ageLocalTs = event.unsignedData?.age?.let { now - it }
             eventIds.add(event.eventId)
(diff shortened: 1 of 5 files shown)


ganfra:
LGTM

CHANGES.md
```diff
@@ -10,6 +10,7 @@ Improvements 🙌:
 Bugfix 🐛:
  - Fix dark theme issue on login screen (#1097)
  - Incomplete predicate in RealmCryptoStore#getOutgoingRoomKeyRequest (#1519)
+ - Fix / Old call invite event from pagination can trigger call UX
```
bmarty, line 13: Maybe not useful to add something here, because it's part of VoIP which is not release yet. But I can do the cleanup later

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/timeline/TokenChunkEventPersistor.kt
```diff
@@ -206,13 +206,15 @@ internal class TokenChunkEventPersistor @Inject constructor(private val monarchy
...
+        val now = System.currentTimeMillis()
         val eventIds = ArrayList<String>(eventList.size)
         for (event in eventList) {
             if (event.eventId == null || event.senderId == null) {
                 continue
             }
+            val ageLocalTs = event.unsignedData?.age?.let { now - it }
             eventIds.add(event.eventId)
-            val eventEntity = event.toEntity(roomId, SendState.SYNCED).copyToRealmOrIgnore(realm)
+            val eventEntity = event.toEntity(roomId, SendState.SYNCED, ageLocalTs).copyToRealmOrIgnore(realm)
```
bmarty, line 217: Another example to prove that using default value for parameter can lead to bug

bmarty:
Is the chang applicable to other call to Event.toEntity?

ganfra:
Yes maybe we should just put that code directly into the mapper ? So we don't have to handle it everywhere?

PR #1547, opened 2020-06-25
Compress images before sending

onurays:


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/content/UploadContentWorker.kt
```diff
@@ -154,18 +161,49 @@ internal class UploadContentWorker(val context: Context, params: WorkerParameter
                 var uploadedFileEncryptedFileInfo: EncryptedFileInfo? = null
 
                 return try {
+                    var cacheFile = File.createTempFile(attachment.name ?: UUID.randomUUID().toString(), ".jpg", context.cacheDir)
```
bmarty, line 164: Please document why you are doing this. Also, maybe do it on ly on Android 10 and above?

PR #1549, opened 2020-06-25
Implementation of updating user avatar.

### LINKED ISSUE #1054: Update user avatar
Not implemented yet… :)

onurays:
Fixes #1054


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/profile/DefaultProfileService.kt
```diff
@@ -63,6 +81,41 @@ internal class DefaultProfileService @Inject constructor(private val taskExecuto
                 .executeBy(taskExecutor)
     }
 
+    override fun updateAvatar(userId: String, newAvatarUri: Uri, fileName: String, matrixCallback: MatrixCallback<Unit>): Cancelable {
+        val cancelableBag = CancelableBag()
+        val workerParams = UploadAvatarWorker.Params(sessionId, newAvatarUri, fileName)
+        val workerData = WorkerParamsFactory.toData(workerParams)
+
+        val uploadAvatarWork = workManagerProvider.matrixOneTimeWorkRequestBuilder<UploadAvatarWorker>()
```
ganfra, line 89: I don't think we need to make uploading avatar uses a worker. It's alright if it fails, we'll get an error and can retry later. What do you think about that @bmarty ?

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/profile/ProfileService.kt
```diff
@@ -48,6 +49,14 @@ interface ProfileService {
      */
     fun setDisplayName(userId: String, newDisplayName: String, matrixCallback: MatrixCallback<Unit>): Cancelable
 
+    /**
+     * Update the avatar for this user
+     * @param userId the userId to update the avatar of
+     * @param newAvatarUri the new avatar uri of the user
+     * @param fileName the fileName of selected image
+     */
+    fun updateAvatar(userId: String, newAvatarUri: Uri, fileName: String, matrixCallback: MatrixCallback<Unit>): Cancelable
```
ganfra, line 58: Can we update avatar of other users? otherwise, we should not offer the userId param and inject @UserId instead

vector/src/main/java/im/vector/riotx/features/settings/VectorSettingsGeneralFragment.kt
```diff
@@ -370,26 +405,88 @@ class VectorSettingsGeneralFragment : VectorSettingsBaseFragment() {
...
-        val intent = Intent(activity, VectorMediaPickerActivity::class.java)
-        intent.putExtra(VectorMediaPickerActivity.EXTRA_AVATAR_MODE, true)
-        startActivityForResult(intent, VectorUtils.TAKE_IMAGE)
-        */
+    private fun onAvatarSelected(image: MultiPickerImageType) {
+        val destinationFile = File(requireContext().cacheDir, "${image.displayName}_edited_image_${System.currentTimeMillis()}")
+        val uri = image.contentUri
+        UCrop.of(uri, destinationFile.toUri())
+                .withOptions(
+                        UCrop.Options()
```
ganfra, line 434: Didn't check, but if Ucrop options are the same than for attachment, we should extract this in a helper

matrix-sdk-android/src/main/java/im/vector/matrix/android/api/session/profile/ProfileService.kt
bmarty: The Matrix API take a userId, that's why it's exposed here... But yes in the app, it will be only possible to change avatar of the current user

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/profile/DefaultProfileService.kt
bmarty: I think this code is more complicated that I'd expect it to be. We should use the uploadService facilities. I have not checked in the code if it is possible.

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/profile/DefaultProfileService.kt
```diff
@@ -63,6 +81,41 @@ internal class DefaultProfileService @Inject constructor(private val taskExecuto
...
+        cancelableBag.add(CancelableWork(workManagerProvider.workManager, uploadAvatarWork.id))
+
+        taskExecutor.executorScope.launch(coroutineDispatchers.main) {
+            workManagerProvider.workManager.getWorkInfoByIdLiveData(uploadAvatarWork.id)
+                    .observeForever { info ->
+                        if (info != null && info.state.isFinished) {
+                            val result = WorkerParamsFactory.fromData<UploadAvatarWorker.OutputParams>(info.outputData)
+                            cancelableBag.add(
+                                    setAvatarUrlTask
+                                            .configureWith(SetAvatarUrlTask.Params(userId = userId, newAvatarUrl = result?.imageUrl!!)) {
```
bmarty, line 108: I would put the `!!` to `result` instead

vector/src/main/java/im/vector/riotx/features/settings/VectorSettingsGeneralFragment.kt
bmarty: Maybe not the same that for attachment. We should limit to a square image, or if possible a round image

vector/src/main/java/im/vector/riotx/features/settings/VectorSettingsGeneralFragment.kt
```diff
@@ -370,26 +405,88 @@ class VectorSettingsGeneralFragment : VectorSettingsBaseFragment() {
...
+                if (!isAdded) return
+                onCommonDone(failure.localizedMessage)
+            }
+        })
     }
 
-    // ==============================================================================================================
-    // contacts management
-    // ==============================================================================================================
+// ==============================================================================================================
```
bmarty, line 487: Please revert those fomartting change

Issue #174, opened 2019-06-10
reply message ui design


violoncelloCH:
some ideas for displaying the in reply to `<mx-reply>` description and content in a (probably) nicer way:

- link the whole `<mx-reply>` to the original message so a tap on it scrolls up the message history (if necessary) and highlights the original message
- don't display nested replys (in longer discussions I don't wan't to have everything from above again in the new message, but just the last message the reply refers to; when I wan't to read preceding messages content I can scroll up / click on the reply to be scrolled up to the corresponding message)
- only display the first part of the quoted text (maybe also in a little bit smaller font size) to prevent reply messages to fill up a huge amount of space, only because the original message was a long one

BillCarsonFr:
Thanks, yes RiotX need to implement 'Rich Reply'  
https://matrix.org/docs/spec/client_server/r0.4.0.html#rich-replies

tulir:
To be specific, the `<mx-reply>` shouldn't be rendered at all, it needs to be stripped away completely. The original message should be fetched from the server using the metadata in the event (`m.relates_to`) and that event should be rendered above the reply appropriately.

fbruetting:
I'm pro shortening the referred-to message via ellipsis, with the possibility to tap and show it fully, instead of shortening the message.

tulir:
> tap and show it fully

Isn't it better to have tapping jump to message? You see the whole message and the context, rather than just seeing the message and not having a convenient way to jump to context.

pioneer:
I'd add the following improvement: if the original message contains an image, either show a small image preview instead of a full-sized image or show a small image placeholder, as Facebook Messenger does.

bnainar:
Just use the Telegram's approach

pioneer:
The previous message is not redundant, it helps keeping and following the context. In contrast, showing only the name (@ Anna in the above example), will not add anything about how the reply is related to the original message. Most of the messengers show the original text, or a part of that text using a smaller font, which is usually enough to follow the conversation (given that the original message can be a couple of screens above, so otherwise you must scroll back to understand why someone replied).

bilfri:
@pioneer I actually agree with you, but I think you misunderstand my suggestion: It is only about the case where the reply is to the latest message in the chat. Such that the message which is replied to is visible just above the new message anyway. In cases where there have been other messages since the one that is being replied to I think that yes, the small quote/summary makes sense (which I tried to say in my suggestion as well). I will try to make this clear by editing my message above

[...]

PR #1610, opened 2020-07-03
Fix reply fallback leaking sender locale

tulir:
This pull request fixes the privacy issue where the sender's locale leaks whenever they reply to a message. It is also more conforming to the spec, as the spec does not state that clients should translate the fallback.

Receiver-side translations should be added in #174

Signed-off-by: Tulir Asokan &lt;tulir@maunium.net&gt;

### Pull Request Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/LocalEchoEventFactory.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/LocalEchoEventFactory.kt
@@ -177,7 +177,6 @@ internal class LocalEchoEventFactory @Inject constructor(
         val body = bodyForReply(originalEvent.getLastMessageContent(), originalEvent.root.getClearContent().toModel())
         val replyFormatted = REPLY_PATTERN.format(
                 permalink,
-                stringProvider.getString(R.string.message_reply_to_prefix),
                 userLink,
                 originalEvent.senderInfo.disambiguatedDisplayName,
                 body.takeFormatted(),
@@ -372,7 +371,6 @@ internal class LocalEchoEventFactory @Inject constructor(
         val body = bodyForReply(eventReplied.getLastMessageContent(), eventReplied.root.getClearContent().toModel())
         val replyFormatted = REPLY_PATTERN.format(
                 permalink,
-                stringProvider.getString(R.string.message_reply_to_prefix),
                 userLink,
                 userId,
                 body.takeFormatted(),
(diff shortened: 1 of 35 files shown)


matrix-sdk-android/src/main/res/values/strings.xml
```diff
@@ -139,12 +138,6 @@
     <string name="medium_email">Email address</string>
     <string name="medium_phone_number">Phone number</string>
 
-    <!-- Reply to -->
-    <string name="reply_to_an_image">sent an image.</string>
-    <string name="reply_to_a_video">sent a video.</string>
-    <string name="reply_to_an_audio_file">sent an audio file.</string>
-    <string name="reply_to_a_file">sent a file.</string>
```
bmarty, line 146: Should be also removed from the translations file. I'm surprised that lint does not complain...

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/LocalEchoEventFactory.kt
```diff
@@ -489,6 +487,6 @@ internal class LocalEchoEventFactory @Inject constructor(
         //     </blockquote>
         // </mx-reply>
         // No whitespace because currently breaks temporary formatted text to Span
-        const val REPLY_PATTERN = """<mx-reply><blockquote><a href="%s">%s</a><a href="%s">%s</a><br />%s</blockquote></mx-reply>%s"""
+        const val REPLY_PATTERN = """<mx-reply><blockquote><a href="%s">In reply to</a><a href="%s">%s</a><br />%s</blockquote></mx-reply>%s"""
```
bmarty, line 490: This is not new but I think a space is missing between the two links.

matrix-sdk-android/src/main/res/values/strings.xml
bmarty: But maybe for the moment we should keep the strings, even if they are not used anymore.

tulir: I changed the commits so all the strings are removed in a single commit that can be easily reverted if needed

matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/send/LocalEchoEventFactory.kt
tulir: I added the space, but is the

> No whitespace because currently breaks temporary formatted text to Span

comment referring to that or something else?

bmarty: No, looking at the history, it seems that before the string was defined on several lines. But I think it's ok like that.

bmarty: (thanks)

PR #1611, opened 2020-07-03
Feature/okhttp for glide

bmarty:
I create the draft PR to expose what has been done so far.

This PR tries to set the OkHttpClient with log and certificate handling to Glide.

It works fine when viewing image in the ImageViewer, but:

- [ ] Glide does not use correct OkHttpClient when loading Avatar
- [ ] Glide do not use correct OkHttpClient when loading Image in the timeline

This PR also bring some cleanup in the code, see each commit for details.

### DIFF
--- a/vector/src/main/java/im/vector/riotx/core/di/ActiveSessionHolder.kt
+++ b/vector/src/main/java/im/vector/riotx/core/di/ActiveSessionHolder.kt
@@ -37,7 +37,8 @@ class ActiveSessionHolder @Inject constructor(private val authenticationService:
                                               private val incomingVerificationRequestHandler: IncomingVerificationRequestHandler,
                                               private val webRtcPeerConnectionManager: WebRtcPeerConnectionManager,
                                               private val pushRuleTriggerListener: PushRuleTriggerListener,
-                                              private val sessionListener: SessionListener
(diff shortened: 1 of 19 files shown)


vector/src/main/java/im/vector/riotx/core/di/ActiveSessionHolder.kt
```diff
@@ -30,23 +34,40 @@ import javax.inject.Singleton
...
-                                              private val incomingVerificationRequestHandler: IncomingVerificationRequestHandler
+                                              private val incomingVerificationRequestHandler: IncomingVerificationRequestHandler,
+                                              private val webRtcPeerConnectionManager: WebRtcPeerConnectionManager,
+                                              private val pushRuleTriggerListener: PushRuleTriggerListener,
+                                              private val sessionListener: SessionListener
 ) {
 
     private var activeSession: AtomicReference<Session?> = AtomicReference()
 
     fun setActiveSession(session: Session) {
```
ganfra, line 46: Maybe we could event extract the whole configureAndStart extension method over there?

bmarty: Yes, let's keep this idea for later

Issue #1617, opened 2020-07-04
Not available on PlayStore with a chromebook


Wandelstock:
With the new Riot.IMx an old bug reappeared (#932), app isnt as beta available at the Google playstore for my Chromebook

bmarty:
Ok, I will check that, maybe due to the permissions we have added related to VoIP.

Wandelstock:
Could it be the camera permission for scanning qr-codes?

bmarty:
This is normally already handled (in the fix for #932)

Wandelstock:
Works now after the update

bmarty:
Ok, thanks for the update, so I reject #1641

PR #1641, opened 2020-07-07
Try to make the app compatible with Chromebook (#1617)

bmarty:


bmarty:
#1617 is not a pb anymore

Issue #548, opened 2019-09-06
[Feature] 3pid invite


bmarty:

bmarty:
Unfortunately it requires to implement the identity server part (#607), which has not been identified as a showstopper.

henry-nicolas:
Are there any plans for the 3pid invite by phone feature?

It feels weird when one want to invite another user by doing a contact lookup and no result pops up.
On Android (phones, arguably not on tablets form factors), contacts would have their phone number fields filed in, making this feature very relevant imho.

PR #1658, opened 2020-07-10
3pid invite

bmarty:
Handle first part of #548:

A configured identity server is mandatory to use this feature.

Display contacts of the user with email to be able to create a DM or the invite them to a room.

Some screenshots:

image

image

image

image

image


vector/src/main/java/im/vector/riotx/core/contacts/ContactsDataSource.kt
```diff
@@ -0,0 +1,155 @@
...
+import android.content.Context
+import android.database.Cursor
+import android.net.Uri
+import android.provider.ContactsContract
+import androidx.annotation.WorkerThread
+import timber.log.Timber
+import javax.inject.Inject
+import kotlin.system.measureTimeMillis
+
+class ContactsDataSource @Inject constructor(
```
onurays, line 28: We may move this to ContactPicker in multipicker module later.

PR #1666, opened 2020-07-11
Feature/tab notification labs

BillCarsonFr:
Proposal for a lab option to add a third tab that displays all rooms with unread notifications

False by default, users with lots of room can enable it;

+ Refactored bottom navigation bar to use the Compat Badges instead of adding custom views

image

image

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/settings/VectorSettingsLabsFragment.kt
+++ b/vector/src/main/java/im/vector/riotx/features/settings/VectorSettingsLabsFragment.kt
@@ -34,6 +34,10 @@ class VectorSettingsLabsFragment @Inject constructor(
             it.isChecked = vectorPreferences.labAllowedExtendedLogging()
         }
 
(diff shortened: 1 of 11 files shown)


vector/src/main/java/im/vector/riotx/features/settings/VectorSettingsLabsFragment.kt
```diff
@@ -34,6 +34,10 @@ class VectorSettingsLabsFragment @Inject constructor(
             it.isChecked = vectorPreferences.labAllowedExtendedLogging()
         }
 
+        findPreference<VectorSwitchPreference>(VectorPreferences.SETTINGS_LABS_UNREAD_NOTIFICATIONS_AS_TAB)?.let {
+            it.isChecked = vectorPreferences.labAddNotificationTab()
```
bmarty, line 38: I think this is not necessary, it is done automatically

PR #1701, opened 2020-07-16
RiotX -> Element

bmarty:
Change all visible in the app mention of "RiotX" to "Element".

There are still lots of mention of RiotX in comments, and internal files, as well as in some of .md file, which will be updated later.

Login with SSO tested OK

### DIFF
--- a/vector/src/main/java/im/vector/riotx/features/login/LoginViewState.kt
+++ b/vector/src/main/java/im/vector/riotx/features/login/LoginViewState.kt
@@ -73,7 +73,7 @@ data class LoginViewState(
             append(homeServerUrl?.trim { it == '/' })
             append(SSO_REDIRECT_PATH)
             // Set a redirect url we will intercept later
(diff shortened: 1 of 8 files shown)


vector/src/main/java/im/vector/riotx/features/login/LoginViewState.kt
```diff
@@ -73,7 +73,7 @@ data class LoginViewState(
             append(homeServerUrl?.trim { it == '/' })
             append(SSO_REDIRECT_PATH)
             // Set a redirect url we will intercept later
-            appendParamToUrl(SSO_REDIRECT_URL_PARAM, RIOTX_REDIRECT_URL)
+            appendParamToUrl(SSO_REDIRECT_URL_PARAM, VECTOR_REDIRECT_URL)
```
ilmari, line 76: Any particular reason this is `VECTOR` rather than `ELEMENT`?

bmarty: I prefer to use Vector internally

PR #1742, opened 2020-07-19
Fix build on encyrpted file systems

czeidler:
Some generated class names were too long and the build failed with
errors like: "PollResponseAggregatedSummaryEntityColumnInfo.class (File
name too long)". Shorten the problematic classes a little.

Rename:
PollResponseAggregatedSummaryEntity -> PollResAggregatedSummaryEntity
WellknownIntegrationManagerConfigEntity ->
WellknownIntegrationMgrConfEntity

### Pull Request Checklist

- [x] Changes has been tested on an Android device or Android emulator with API 16

### DIFF
--- a/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/mapper/PollResponseAggregatedSummaryEntityMapper.kt
+++ b/matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/database/mapper/PollResponseAggregatedSummaryEntityMapper.kt
@@ -19,12 +19,12 @@ package im.vector.matrix.android.internal.database.mapper
 import im.vector.matrix.android.api.session.events.model.toContent
 import im.vector.matrix.android.api.session.events.model.toModel
 import im.vector.matrix.android.api.session.room.model.PollResponseAggregatedSummary
-import im.vector.matrix.android.internal.database.model.PollResponseAggregatedSummaryEntity
+import im.vector.matrix.android.internal.database.model.PollResAggregatedSummaryEntity
 import io.realm.RealmList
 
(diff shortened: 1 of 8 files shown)


bmarty:
Hello, on which configuration the build fails for you? In our machines and on ci it does not fail. I want to make sure the problem will not come back later.
Also before merging, we need to decide internally if we want to handle db migration for the session.

czeidler:
Sorry forgot to mention, I am running on Ubuntu 20.04. The filesystem is BTRFS and is encrypted. I just did some more research and the encryption might be the source of the problem (https://stackoverflow.com/questions/49660310/error-error-while-writing-file-name-too-long). As I understand, encrypting the filenames results in longer filenames in the backing fs and reduces the max filename length to ~140 bytes.

What do you mean by db migration?

czeidler:
I copied the source files to a plain directory without encryption and the build succeeds. However, I think it would still good to be able to build riotx from an encrypted directory.

bmarty:
DB migration: you are renaming Entities, which are classes used as DB tables (by Realm). So if you rename them, you have to handle a DB migration.

czeidler:
Ok thanks, feared that it was used for something like that. Is there any docs for how Realm is used in Element/Matrix? e.g. is it only used for RiotX? or is it matrix.org related? Is it the primary way to sync data with a matrix server?

bkil:
Is there any reason not to build in a tmpfs or zram ramdisk instead?

bmarty:
Sorry, we cannot merge this PR without doing a proper Realm migration.
As suggested by @bkil , there is maybe another solution to build Element Android on an encrypted filesystem.

Issue #1761, opened 2020-07-20
404 on EMS


manuroe:
https://twitter.com/ricardosalta/status/1285030317435256832

![image](https://user-images.githubusercontent.com/8418515/87949886-e5fa7600-caa6-11ea-99c0-f4c67ec431e1.png)

bmarty:
Should be fixed by #1766

Issue #1725, opened 2020-07-18
Requesting avatar thumbnails in Element uses wrong http "user-agent" string


andyg1001:
Since the migration to Element, user avatars are now no longer shown at all, just the coloured circle with a letter in it.  Changing my own avatar from the general settings screen works in the sense that the chosen picture is updated on my parallel Linux element-desktop installation (1.7.1) but I still don't get to see it in the Android client.  I don't know if this is also true of avatars for communities/group chats, but it may be.

This is definitely a regression from the previous Riot.im client on Android - which worked perfectly in this regard.

I am connecting to our own home-server, and in case this has an impact, the home-server is not on port 443 because this is assigned to another server in our NAT.

andyg1001:
Ok, done some research.  The issue is not the different https port, but that when the client requests "GET /_matrix/media/r0/thumbnail/..." it doesn't use the normal http-agent "Element (Riot.im)/1.0.0 (Linux; U; Android 8.0.0 ...)" but a totally different one "Dalvik/2.1.0 (Linux; U; Android 8.0.0 ...)" and this is what the firewall is blocking.

I have checked all the way through the server logs.  This is the only message that uses this alternative http-agent.  All other messages use the normal one.

Can this be changed so that it uses the normal (correct?) http-agent for this message too?

bmarty:
We should give the user agent to Glide

andyg1001:
Just popping by to say this has been fixed in the latest update (v1.0.2).  Thank you!

bmarty:
Fixed by #1766, thanks for the follow up @andyg1001 !

Issue #1709, opened 2020-07-17
Element crashes in loop after initial sync


lub:
It just brings a popup "Element (Riot.im) keeps stopping" directly after starting it

I can't even send a rageshake

It happens after login and the first-sync-progress-bar

ftr 1.0.1-dev [205646062] (F-b3287) still syncs without issues after clearinf cache

lub:
I'm happy to gather logs via other ways, I just don't know how

lub:
I was able to send a rageshake :)

kannanvm:
I had issues with v1 but updating to element-dbg 1.0.1 solved it.

ganfra:
Should be fixed in next release, it's a proguard/realm issue. Thanks for pointing!

bmarty:
Fixed by #1766, thanks for the follow up @lub !

Issue #1699, opened 2020-07-15
Infinite loop at startup when migrating account from Riot


bjonnh:
I cannot use RiotX (installed from the Beta Program PlayStore), it just keeps restarting the initial setup (the progress bar resets to zero and it start again getting account data, crypto keys, joined rooms etc)

I tried to remove (multiple times) cache and storage, reinstall… nothing worked. it just keeps doing that over and over (I left it for 30 min, it was still doing that)

OS is Android 9

bjonnh:
I reverted back to Riot, and it was able to sync my account  in less than 5 seconds…

bmarty:
What is your homeserver?
Have you sent rageshakes?

bjonnh:
I can't send rageshakes it does nothing
I'm using the matrix server

bjonnh:
Still happening with Element from Google Play Store.

bjonnh:
I finally was able to do a rageshake and gave this issue number in the title. no idea if it went out as I got no confirmation…

bjonnh:
I can make a video of the infinite loop if that help…

ovitters:
I'm facing the same with both RiotX and Element. No idea what a rageshake is. Deleting data doesn't help. OS is Android 10.

I seldom use Riot. I have a Matrix account that's logged into a few IRC bridges (GimpNet, Freenode) and related channels.

bjonnh:
Rageshake is when you shake your phone inside Riot/Element, it allows you to send a bug report that could help the developers find out what is wrong. Try shaking it a few times.

bjonnh:
I uninstalled the google play one, rebooted, installed the f-droid one, same problem.

bjonnh:
I just did a rageshake from the f-droid version with the github ID in the title.

bjonnh:
If that can help, it keeps showing me a verification screen, and sometimes I can validate (from another session), but it still continues to display it… Even if the login was verified… (got the shields and okayed)

bjonnh:
@ganfra  found the issue, I went to the settings and removed all the ignored users I had and it solved the issue. There was likely a corruption in the JSON data there.

I suppose you may want to still have a way to skip that JSON parsing and notify the user they have to check their account data…

ganfra:
Thanks for checking, will be fixed in next release!

ehuelsmann:
Same behaviour here. Care to elaborate on the steps to take?

bjonnh:
[...]

PR #1766, opened 2020-07-20
Fix post Element release issues

ganfra:
This PR should (hopefully) fix the following issues: 

 - Fix uploads don't work with Room v6 (#1558)
 - Fix Requesting avatar thumbnails in Element uses wrong http "user-agent" string (#1725)
 - Fix 404 on EMS (#1761)
 - Fix Infinite loop at startup when migrating account from Riot (#1699)
 - Fix Element crashes in loop after initial sync (#1709)


matrix-sdk-android/src/main/java/im/vector/matrix/android/internal/session/room/RoomAPI.kt
```diff
@@ -129,6 +129,21 @@ internal interface RoomAPI {
...
+     * @param txId      the transaction Id
+     * @param roomId    the room id
+     * @param eventType the event type
+     * @param content   the event content as string
+     */
+    @PUT(NetworkConstants.URI_API_PREFIX_PATH_R0 + "rooms/{roomId}/send/{eventType}/{txId}")
+    fun send(@Path("txId") txId: String,
+             @Path("roomId") roomId: String,
+             @Path("eventType") eventType: String,
+             @Body content: String?
```
bmarty, line 144: Nullable?

ganfra: I kept the same than with plain object, but in practice it won't be null

bmarty: ok

PR #1783, opened 2020-07-22
Feature/sending retry

ganfra:
This PR brings this behavior for sending: 
 - Sending events is now tried only 3 times, so we avoid blocking the sending queue too long.
 - Display warning when fail to send events in room list so user know there is something wrong.

There are lots of things to improve on sending, but it requires UI/UX work with the team.


lub:
Honestly I don't like this change. I already disliked the same behaviour in legacy riot-android (see e.g. https://github.com/vector-im/riot-android/issues/882).

When I tap on send I expect the message to send and a especially a mobile phone app should imo be able to handle shitty network conditions.

I often use element in bus/train in rural areas and the connection is very bad, so I have to send many messages two times (once initially and then after I noticed it failed a couple of minutes later I have to tap on 'retry').

PR #1821, opened 2020-07-27
Check if migration disabled notif

### LINKED ISSUE #1723: Migration from old client does not enable notifications
When upgrading from old Riot-Android to Element today, I had to manually check "Enable notifications for this session" under notification settings in the app. I gather that I would have been missing notifications had I not checked all the app settings (or if I hadn't understood the potentially confusing setting text and ignored it). If this requires intervention by the user for some reason, they should be prompted about it on first startup after migration.

BillCarsonFr:
Fixes #1723

Root cause not found yet; so has a work around a check is performed (once) to see if notifications are disabled for this session and prompt user to review settings if needed.

image

review


ganfra:
Just fix code quality before merging

PR #1844, opened 2020-07-30
Feature/pin code

ganfra:
This PR implements Pin code as https://github.com/vector-im/riotX-android/issues/1700

.


tools/check/forbidden_strings_in_code.txt
```diff
@@ -164,7 +164,7 @@ Formatter\.formatShortFileSize===1
 # android\.text\.TextUtils
 
 ### This is not a rule, but a warning: the number of "enum class" has changed. For Json classes, it is mandatory that they have `@JsonClass(generateAdapter = false)`. If it is ok, change the value in file forbidden_strings_in_code.txt
-enum class===74
+enum class===76
```
BillCarsonFr, line 167: ?

vector/src/main/java/im/vector/riotx/features/pin/PinLocker.kt
```diff
@@ -0,0 +1,103 @@
...
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.OnLifecycleEvent
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import timber.log.Timber
+import javax.inject.Inject
+import javax.inject.Singleton
+
+private const val PERIOD_OF_GRACE_IN_MS = 2 * 60 * 1000L
```
BillCarsonFr, line 31: This could probably be a user setting?

vector/src/main/java/im/vector/riotx/features/pin/UnlockedActivity.kt
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.riotx.features.pin
+
```
BillCarsonFr, line 18: Add a comment : 'Tag class for activities that should not be protected by PIN code'

tools/check/forbidden_strings_in_code.txt
ganfra: It's Benoit who added a check to be sure we don't forget to set generateAdapter = false for enums we need to keep safe from proguard

vector/src/main/java/im/vector/riotx/features/pin/PinLocker.kt
ganfra: Don't know, for now I think it's ok...

vector/src/main/java/im/vector/riotx/features/pin/UnlockedActivity.kt
ganfra: Done

TR-SLimey:
Not sure if this is the case already, but please make it so that pins can be over 4 chars long, for security.

PR #1878, opened 2020-08-05
Kill RiotX (#1727)

### LINKED ISSUE #1727: Kill RiotX
After the release of Element as an upgrade of Riot-Android, we should properly kiil the beta of RiotX on the PlayStore

## PlayStore

- [x] Update the PlayStore description to clearly tell that this app should not be installed. We keep it on the PlayStore for the latest update with the change below

## RiotX code

On a branch created before the re-branding is merged, and which will be of course never merged, we should:

- [x] Always display the "Welcome to the Beta" popup with new wording: title "The Beta is over", content:

> RiotX will not be updated anymore.
 Please install Element!

Description may also contain advice to properly sign out or to keep the session open to verify the new session with RiotX, but not sure it's very useful/it worth it.

2 buttons:
"Later" -> dismiss the popup
"Get Element" -> Open the PlayStore on the Element page

- [x] Prevent new login with RiotX

- [x] Publish the RiotX app on the beta channel of the PlayStore

## Finally

- [x] Say 1 month after the update: remove RiotX from the PlayStore

bmarty:
Fixes #1727

- Restore RiotX appId and App name
- Prevent share to RiotX
- Show the warning popup each time the app is launch
- Prevent any new login and propose to uninstall the app

Popup:

image

With "Later" and "Get Element" option

First screen (when not logged in/after logging out)

image

This PR is draft, to not being merged by mistake :)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,27 +1,16 @@
-Changes in Riot.imX 0.91.6 (2020-XX-XX)
+Changes in RiotX 0.23.0 (2020-08-05)
 ===================================================
 
-Features ✨:
- -
+This is the very last version of RiotX, published on the PlayStore.
 
-Improvements 🙌:
- -
+This branch will never be merged on develop.
+
+Features ✨:
+ - Inform user that the app will not be updated anymore (#1727)
 
 Bugfix 🐛:
  - Video calls are shown as a voice ones in the timeline (#1676)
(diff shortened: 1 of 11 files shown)


bmarty:
RiotX 0.23.0 is on its way to the PlayStore (beta channel of RiotX)

bmarty:
RiotX v0.23.0 is now live on the PlayStore

PR #1902, opened 2020-08-12
Feature/verify one session

### LINKED ISSUE #1901: Do not propose to verify session if there is only one session
From the sessions list:

image

bmarty:
First commits are some cleanup.

Last commit fixes #1901: If there is no other session available, we do not propose to verify it:

## When other sessions are available (so no change by this PR)

image

## When no other sessions are available

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ Features ✨:
 Improvements 🙌:
  - You can now join room through permalink and within room directory search
  - Add long click gesture to copy userId, user display name, room name, room topic and room alias (#1774)
+ - Do not propose to verify session if there is only one session (#1901)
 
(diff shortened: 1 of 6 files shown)


BillCarsonFr (changes requested):
As discussed we should show option to enter 4S passphrase if available (even if no other session, you can still verify). The existing verify button is showing this option, we will lose it of we remote it

bmarty:
User can now verify the session if there is no other session, but 4S is configured

BillCarsonFr (approved):
LGTM. 
We need a new issue, to allow user to reset cross signing if he has no way to recover

bmarty:
(#2052)

PR #1914, opened 2020-08-13
Feature/jitsi native

### LINKED ISSUE #43: Conference with Jitsi support
- [ ] import last Jitsi lib

BillCarsonFr:
Fixes #43 

Add support for group video/audio conference using native Jitsi SDK.

image

As a consequence of the support of jitsi call the call icons of the toolbars 📞  and 🎥  will always be present in rooms (for better dicoverablity).

/!\ Session database migration (adding preferredJitsiDomain)

## Usage of jitsi SDK

Build is modified and is using 'org.jitsi.react:jitsi-meet-sdk:2.2.2'
As a consequence the dependency to 'org.webrtc:google-webrtc' is removed (not suitable for production use), and the one embedded by jitsi will be used

-> Proguard rules have been updated

## Change in video/audio call button behavior

Depending on the number of joined members the behavior will change:

- If you are alone, you will have a message saying that you cannot call yourself
image

- if you are alone, and an inviation is pending (the other has not joined the essage is a bit different)
image

- If there are exactly 2 participants, the call icons will start a webrtc P2P encrypted call

- If there are more than 2 participants, the actions will create a jitsi widget
image

If you don't have permission to add widgets, you will have this message
image

## Changes in Widget banner in the room screen

As per discussion with design team, the intrusive active widget banner as been removed (taking too much space, and overlaps)
image

It has been replaced with a solution closer to the ios client: We will use the integration icon in tool bar to notify for active widgets.
If there are active widgets, the integration icon will appear in top tool bar, with a badge indicating the number of active widgets:
image

If there are no active widgets, the icon will not be visible in the bar, and will stay under the kebab menu
image

When selected the integration option menu will open the active widgets bottomsheet if there are some widgets, or directly open the integration manager if not.

Some specific widgets could still display a banner on top of the room, it's the case for jitsi widgets
image

## Changes to the active widgets bottom sheet

The bottomsheet now display a bit more informations about the widget by adding the domain (instead of having x Custom Widgets...)

image

## Introducing Native widgets permission

Regular web widgets requires that the user gives permission to share data with the widget domain.
Native widgets are different, as they are not loaded in a web browser (no cookies, ..).
They have a different permission type (native widgets)

For example jitsi widgets, stores 'consent' in account data and is only asked once per jitsi domain 
````
{
  "type": "im.vector.setting.allowed_widgets",
  "content": {
    "widgets": {
      "$1574156694949436BgVoR:matrix.org": true,
      "$1574441023202692UNnpv:matrix.org": true,
      "$157383232735007JXbTi:matrix.org": true,
      "$1597325767791530jITTm:matrix.org": true,
      "$1597331391801189FuhNE:matrix.org": true
    },
  
## NATIVE WIDGETS
    "native_widgets": {
      "jitsi": {
        "jitsi.riot.im": true
      },
      "m.jitsi": {
        "jitsi.riot.im": true
      }
    }
  }
}
````
image

## Preferred Widget domain support

Home server admins can define their own jitsi domain
https://github.com/vector-im/element-web/blob/develop/docs/jitsi.md

When creating jitsi widget (via toolbar icons), element android will use that domain if available and if not will fallback to the default jitsi domain. The default domain is in `config.xml`as `preferred_jitsi_domain`

### Known limitations

-  You will leave the jitsi conf when returning to chat (need to open new issue for that)

### CODE REVIEW

File: vector/src/main/res/values/strings.xml [hunk 1]
```diff
@@ -89,8 +89,17 @@
     <string name="missing_permissions_warning">"Due to missing permissions, some features may be missing…</string>
     <string name="missing_permissions_error">"Due to missing permissions, this action is not possible.</string>
     <string name="missing_permissions_to_start_conf_call">You need permission to invite to start a conference in this room</string>
+    <string name="no_permissions_to_start_conf_call">You do not have permission to start a conference call in this room</string>
+    <string name="conference_call_in_progress">A conference is already in progress!</string>
+    <string name="video_meeting">Start video meeting</string>
+    <string name="audio_meeting">Start audio meeting</string>
+    <string name="audio_video_meeting_description">Meetings use Jisti security and permission policies.  All people currently in the room will see an invite to join while your meeting is happening.</string>
```
steef435, line 96: Typo (Jisti)

bmarty:
I have added 2 commits.
Also 2.2.2 is quite old, we need to get the script to build a more recent version from Riot-Android.
It can be done later though.

All my tests are successful so far, good work!

PR #1946, opened 2020-08-17
Stop background sync when network connection is lost

### LINKED ISSUE #1915: [F-Droid] Excessive offline battery usage
**Description**
When running in the background with no network access (mobile data and wi-fi disabled) Element consumes unreasonable amounts of energy. If the device is mainly in standby mode, it takes up more than a third of the total battery capacity. It looks like the program tries to perform a synchronization every 10 seconds.

**To Reproduce**
1. Fully charge the device and disconnect it from the power supply.
2. Disable all network access (i.e. mobile data and wi-fi).
3. Let Element run in the background.
4. Notice that every 10 seconds an icon pops up briefly in the status bar.
5. Put the device into standby mode and leave it in this state for a few hours, at least.
6. Check the battery status and observe excessive energy consumption.

**Expected Behavior**
In the situation described, Element should not have more than one tenth of the current energy requirement.

**Smartphone / Tablet**
 - Device: Samsung Galaxy S5 / Samsung Galaxy Tab S2
 - OS: Lineage OS 14.1 (Android 7.1.2)

**Additional Context**
 - Element 1.0.4 - F-Droid

### LINKED ISSUE #1920: Stop background sync when network connection is lost
The F-Droid version of Element periodically tries to perform a background sync no matter whether a network connection is available or not. To reduce battery usage #536 the background sync should be stopped when no network connection is available. As far as I can see there are at least two ways to achieve this:

1. ~~Using a `PeriodicWorkRequest` with `NetworkType.CONNECTED` as constraint. https://developer.android.com/topic/libraries/architecture/workmanager/how-to/define-work~~ Edit: A `PeriodicWorkRequest` cannot run more often then every 15 minutes.

2. Using a `ConnectivityManager.NetworkCallback` to be informed when a network connection is available or a network connection is lost. https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback

dontub:
Signed-off-by: Dominic Tubach <dnt@luki.org>

### Pull Request Checklist

Fixes #1920, Fixes #1915

With this PR the background sync in the F-Droid version is stopped when no network connection is available. This is done by using a `OneTimeWorkRequest` with a constraint that requires `NetworkType.CONNECTED` and an initial delay. Thus battery is not drained if Element is in the background and there's no network connection.

I've dropped the active session check:
https://github.com/vector-im/element-android/blob/1b9b18851d2389e41268a56229b8097e663c10b2/vector/src/fdroid/java/im/vector/app/fdroid/receiver/AlarmSyncBroadcastReceiver.kt#L39

The worker is only scheduled if there's an active session. By looking at the code I couldn't spot a place where active session is cleared while Element is in the background. (Even though if it would be possible it should be so uncommon that IMHO it isn't worth the extra code.)

The active lock
https://github.com/vector-im/element-android/blob/1b9b18851d2389e41268a56229b8097e663c10b2/vector/src/fdroid/java/im/vector/app/fdroid/receiver/AlarmSyncBroadcastReceiver.kt#L47
is to my knowledge not required when using a foreground service.


bmarty:
We have implemented several F-Droid sync setting on #2080. And we keep the usage of the alarm for "real time" mode

PR #1949, opened 2020-08-17
Official Gradle Wrapper Validation Action

tilosp:
This GitHub Action validates the `gradle-wrapper.jar` on each PR, to make sure that no malicious one gets merged into the code base.

See: https://github.com/gradle/wrapper-validation-action


bmarty:
Thanks for the PR, I was not aware of this possible attack, which seems quite obvious when we think about it... I've checked that the wrapper we used has been committed by the core team, so let's merge your PR and see what will be the effect for the next PRs

bmarty:
Seems to work like a charm

image

PR #1952, opened 2020-08-18
Fix Canvas: trying to draw too large bitmap

onurays:
We can also preserve the quality even in max zoom level:

Screenshot_2020-12-18-16-04-20-529_im vector app debug


vector/src/main/java/im/vector/app/features/media/ImageContentRenderer.kt
```diff
@@ -117,7 +117,7 @@ class ImageContentRenderer @Inject constructor(private val activeSessionHolder:
                     .load(resolvedUrl)
         }
 
-        req.override(Target.SIZE_ORIGINAL, Target.SIZE_ORIGINAL)
+        req
```
bmarty, line 120: Will this change reduce the quality of the image and so even when the image is pinched (zoomed in)?

BillCarsonFr: Yes this is important, we ideally want the original image. This call is used for the image viewer that has the zoom feature. 
But eventually we only need maxZoom * screen size?
Not sure exactly how to use it but try to see sizeMultiplier() ? We don't have target size here, but we could override to screen size.
Maybe it's too much memory for screen of very high density?

vector/src/main/java/im/vector/app/features/media/ImageContentRenderer.kt
```diff
@@ -152,7 +153,11 @@ class ImageContentRenderer @Inject constructor(private val activeSessionHolder:
                     .load(resolvedUrl)
         }
 
-        req.override(Target.SIZE_ORIGINAL, Target.SIZE_ORIGINAL)
+        val maximumScale = (contextView as? PhotoView)?.maximumScale?.toInt() ?: 1
```
ouchadam, line 156: the default max scale is 3 https://github.com/Baseflow/PhotoView/blob/565505d5cb84f5977771b5d2ccb7726338e77224/photoview/src/main/java/com/github/chrisbanes/photoview/PhotoViewAttacher.java#L42

vector/src/main/java/im/vector/app/features/media/ImageContentRenderer.kt
```diff
@@ -152,7 +153,11 @@ class ImageContentRenderer @Inject constructor(private val activeSessionHolder:
...
+        val width = min(contextView.resources.displayMetrics.widthPixels, data.width ?: 0).times(maximumScale)
+        val height = min(contextView.resources.displayMetrics.heightPixels, data.height ?: 0).times(maximumScale)
```
ouchadam, line 158: if I've understood this correctly and this is the image height/width rather than viewport... then we'll break the aspect ratio if we end up using the screen width/height instead of the data width/height (eg causing a landscape photo to become portrait)

it would be safer to always apply the scaling to both dimensions 

```kotlin
// not taking into account any nulls
val width = data.width
val height = data.height

if (data.width > displayMetrics.widthPixels || data.height > displayMetrics.heightPixels) {
  val ratio = data.width / data.height
  width = displayMetrics.widthPixels
  height = (displayMetrics.heightPixels * ratio)
}

req.override(width.times(maxScale), height.times(maxScale)
```






```

SebyakinAndrei:
When this PR is going to be merged? I'm experiencing an issue with big images not displaying.

bmarty:
I close the PR. We will investigate this issue again.

PR #1961, opened 2020-08-19
Fix IllegalArgumentException: Receiver not registered: NetworkInfoReceiver

### LINKED ISSUE #1960: java.lang.IllegalArgumentException: Receiver not registered: NetworkInfoReceiver
SyncThread may call NetworkInfoReceiver.unregister() before registration completed.

onurays:
SyncThread is calling unregister before registration completed.

Fixes #1960.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/NetworkCallbackStrategy.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/NetworkCallbackStrategy.kt
@@ -47,7 +47,12 @@ internal class FallbackNetworkCallbackStrategy @Inject constructor(private val c
 
     override fun unregister() {
         networkInfoReceiver.isConnectedCallback = null
-        context.unregisterReceiver(networkInfoReceiver)
+        // Crashes if the SyncThread calls this before registration completed
+        try {
+            context.unregisterReceiver(networkInfoReceiver)
+        } catch (t: Throwable) {
+            Timber.e(t, "Unable to unregister network info receiver")
+        }
     }
 }
 
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/NetworkCallbackStrategy.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/NetworkCallbackStrategy.kt
@@ -47,12 +47,7 @@ internal class FallbackNetworkCallbackStrategy @Inject constructor(private val c
 
     override fun unregister() {
         networkInfoReceiver.isConnectedCallback = null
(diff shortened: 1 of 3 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/NetworkCallbackStrategy.kt
```diff
@@ -47,7 +47,12 @@ internal class FallbackNetworkCallbackStrategy @Inject constructor(private val c
 
     override fun unregister() {
         networkInfoReceiver.isConnectedCallback = null
-        context.unregisterReceiver(networkInfoReceiver)
+        // Crashes if the SyncThread calls this before registration completed
+        try {
+            context.unregisterReceiver(networkInfoReceiver)
+        } catch (t: Throwable) {
+            Timber.e(t, "Unable to unregister network info receiver")
```
bmarty, line 54: In this case, does it mean that the registration will complete after and so we will have a registered receiver for nothing?

bmarty (commented):
One question, maybe @ganfra can help too

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/NetworkCallbackStrategy.kt
onurays: Good question. In SyncThread.run() if syncstate is "killing", we are immediately trying to unregister the receiver right after registering it. So yes, it is possible to have a registered receiver later. I couldn't reproduce the crash but I will improve this PR by not registering receivers if the initial state is "killing".

ganfra: I think the issue is in NetworkConnectivityChecker, in the register method:
```
 override fun register(listener: NetworkConnectivityChecker.Listener) {
        if (listeners.isEmpty()) {
            if (backgroundDetectionObserver.isInBackground) {
                unbind()
            } else {
                bind()
            }
            backgroundDetectionObserver.register(backgroundDetectionObserverListener)
        }
        listeners.add(listener)
    }
```

I guess unbind() should not be called in this case

onurays: @ganfra you are totally right. I can reproduce the issue on Emulator API 23.

bmarty: Is it ok to keep this try catch then?

bmarty:
I've squashed the 4 commits to remove the bad history

PR #1963, opened 2020-08-19
Improve user avatar setting

bmarty:
Make the avatar live, even if changed from another client.


vector/src/main/java/im/vector/app/features/settings/VectorSettingsGeneralFragment.kt
```diff
@@ -120,10 +124,21 @@ class VectorSettingsGeneralFragment : VectorSettingsBaseFragment() {
         }
     }
 
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        session.rx()
```
ganfra, line 130: Extract in a method?

ganfra (approved):
LGTM, just one small remark

vector/src/main/java/im/vector/app/features/settings/VectorSettingsGeneralFragment.kt
bmarty: Done

PR #1965, opened 2020-08-19
Share button in rooms gives room ID link without via parameters (#1927)

### LINKED ISSUE #1927: Share button in rooms gives room ID link without via parameters
The share button in the top right of the room info view shows a matrix.to URL with the room ID (e.g. https://matrix.to/#/!AZozoWghOYSIAfaZjJ:matrix.org). Such links are not routable, and can't be used for joining rooms*. Sharing links that can't be used is rather pointless.

Element Web solves this by including up to three servers in the `via` query parameter: https://matrix.to/#/!AZozoWghOYSIAfaZjJ:matrix.org?via=matrix.org&via=privacytools.io&via=tchncs.de

The other option is to make the share button create a link with the canonical alias of the room (e.g. https://matrix.to/#/#element-android:matrix.org), but room IDs are more permanent.

<sup>* unless your server happens to know about the room already</sup>

bmarty:
Following the algorithm described in #1927

Fixes #1927


bmarty:
I still have work to do for eventId.
Also I will create a PermalinkService

bmarty:
Now it's ready to be reviewed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkService.kt
```diff
@@ -15,87 +15,64 @@
...
 
-package org.matrix.android.sdk.api.permalinks
+package org.matrix.android.sdk.api.session.permalinks
 
 import org.matrix.android.sdk.api.session.events.model.Event
 
 /**
  * Useful methods to create Matrix permalink (matrix.to links).
  */
-object PermalinkFactory {
```
ganfra, line 25: I would have keep the PermalinkFactory as an internal class with required dependencies, and inject it on the permalink service to delegate. So we can use it internally in SDK without using the service, as we spoke. 
We also have a PermalinkParser, we might want to do the same.

ganfra (changes requested):
See my comment

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkService.kt
bmarty: Yes, it will be better. Are you OK to merge this PR first and then schedule a global rework to remove all the injected services in the SDK?

bmarty: Ok, I've made the change for PermalinkFactory.
PermalinkParser is more static, so I prefer not touch it for the moment

ganfra (approved):
LGTM

PR #1969, opened 2020-08-20
DM creation: handled denied federation error

bmarty:
When creating a DM and inviting a user whose account is on a HomeServer which is not federated, we get an error, but the room is actually created.

This PR inform the user about that and avoid the user to click again on "Create", which will create many rooms.

Here is the displayed dialog:

image

Clicking on OK close the room creation screen.

Note: As we do not know the created roomId, we cannot add it to the DM list.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateRoomTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateRoomTask.kt
@@ -21,6 +21,8 @@ import com.zhuinden.monarchy.Monarchy
 import io.realm.RealmConfiguration
 import kotlinx.coroutines.TimeoutCancellationException
 import org.greenrobot.eventbus.EventBus
+import org.matrix.android.sdk.api.failure.Failure
+import org.matrix.android.sdk.api.failure.MatrixError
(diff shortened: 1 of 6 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateRoomTask.kt
```diff
@@ -55,10 +57,26 @@ internal class DefaultCreateRoomTask @Inject constructor(
...
-            apiCall = roomAPI.createRoom(createRoomBody)
+        val createRoomResponse = try {
+            executeRequest<CreateRoomResponse>(eventBus) {
+                apiCall = roomAPI.createRoom(createRoomBody)
+            }
+        } catch (throwable: Throwable) {
+            if (throwable is Failure.ServerError
+                    && throwable.httpCode == 403
+                    && throwable.error.code == MatrixError.M_FORBIDDEN
+                    && throwable.error.message.startsWith("Federation denied with")) {
```
onurays, line 75: I didn't like this but I guess there isn't any other way to handle...

bmarty: Yes, I suppose. This whole thing is a big workaround which may be improved server side in the future.

PR #1970, opened 2020-08-20
Highlighted Event when opening a permalink from another room

### LINKED ISSUE #1033: Event is not highlighted when opening a permalink from another room
The navigation is correct, but the target event is not highlighted, so is hard to find.
When navigating inside a room, the Event is correctly highlighted.

bmarty:
Fixes #1033


ganfra (approved):
LGTM

PR #1971, opened 2020-08-20
Turn fix

tzeitlho:
Renew turnserver credentials when ttl runs out
    
    The previous implementation caches the turnserver response indefinitely.
    This breaks VoIP calls as soon as the ttl of the received turnserver
    credentials runs out. So, take care to renew the turnserver credentials
    by allowing the cache to expire.

### Pull Request Checklist

- [ X] Changes has been tested on an Android device or Android emulator with API 21
- [ X] Pull request is based on the develop branch
- [ X] Pull request updates [CHANGES.md](https://github.com/vector-im/element-android/blob/develop/CHANGES.md)
- [ X] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/DefaultCallSignalingService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/DefaultCallSignalingService.kt
@@ -58,18 +58,32 @@ internal class DefaultCallSignalingService @Inject constructor(
 
     private val activeCalls = mutableListOf<MxCall>()
 
-    private var cachedTurnServerResponse: TurnServerResponse? = null
+    private var cachedTurnServerResponse = object {
+
+        private val MIN_TTL = 60
+
+        private val now = { System.currentTimeMillis() / 1000 }
+
+        private var expiresAt: Long = 0
+
+        var data: TurnServerResponse? = null
+            get() = if (expiresAt > now()) field else null
+            set(value) {
+                expiresAt = now() + (value?.ttl ?: 0) - MIN_TTL
+                field = value
+            }
+    }
 
     override fun getTurnServer(callback: MatrixCallback<TurnServerResponse>): Cancelable {
-        if (cachedTurnServerResponse != null) {
-            cachedTurnServerResponse?.let { callback.onSuccess(it) }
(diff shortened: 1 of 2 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/DefaultCallSignalingService.kt
```diff
@@ -58,18 +58,32 @@ internal class DefaultCallSignalingService @Inject constructor(
 
     private val activeCalls = mutableListOf<MxCall>()
 
-    private var cachedTurnServerResponse: TurnServerResponse? = null
+    private var cachedTurnServerResponse = object {
```
bmarty, line 61: It should be a `val` now.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/DefaultCallSignalingService.kt
```diff
@@ -58,18 +58,32 @@ internal class DefaultCallSignalingService @Inject constructor(
...
+
+        private val MIN_TTL = 60
+
+        private val now = { System.currentTimeMillis() / 1000 }
```
bmarty, line 65: This is a detail, but I more used to write:
`private fun now() = System.currentTimeMillis() / 1000`
Maybe there is an advantage I do not know to use `val`?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/DefaultCallSignalingService.kt
```diff
@@ -58,18 +58,32 @@ internal class DefaultCallSignalingService @Inject constructor(
...
+
+        private var expiresAt: Long = 0
+
+        var data: TurnServerResponse? = null
+            get() = if (expiresAt > now()) field else null
+            set(value) {
+                expiresAt = now() + (value?.ttl ?: 0) - MIN_TTL
```
bmarty, line 72: Shouldn't it be
`expiresAt = now() + max(value?.ttl ?: 0, MIN_TTL)`

PR #2001, opened 2020-08-25
Fix mention display name

### LINKED ISSUE #1781: Words containing my name should not trigger notifications
I receive push notifications for messages with words that are containing my user-/displayname. Opening the app does not show any highlights though.

I noticed this in multiple public rooms with notification level "mentions only".

### Expected

Only send notifications for words matching my username.

### Example

When both account username and displayname is set to `rda`, I receive notifications if a message contains the word "yesterday".

### Version

Version: Element Android 1.0.0
Android Version: 10

bmarty:
Fixes #1781 . A notification was displayed by mistake when the user name was included in a word.

Actually adding a special treatment for RULE_ID_CONTAIN_USER_NAME.

This PR also includes some refactoring and cleaning which was pending since a few month in a full a dust branch.


ganfra (approved):
LGTM, thanks for cleaning up

PR #2006, opened 2020-08-26
Add long click gesture to copy userId, user display name, room name, …

### LINKED ISSUE #1774: it's not possible to copy an mxid
In old riot-android it was possible to copy an mxid from the member profile by long tapping it. It seems this is not possible anymore. Only way to share it is manually typing it or using the share button, but this generates an matrix.to url which is not always what you want.

bmarty:
…room topic and room alias

Fixes #1774

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@ Features ✨:
  -
 
 Improvements 🙌:
- -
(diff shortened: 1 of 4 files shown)


onurays (approved):
Nice feature!

PR #2014, opened 2020-08-27
Fix: Loudspeaker is always used

### LINKED ISSUE #1685: Loudspeaker is always used
Whether wired headset is connected or not, or which sound device is selected: `Phone`, `Headset` or `Speaker` the loudspeaker is always used.

onurays:
Fixes #1685

### DIFF
--- a/vector/src/main/java/im/vector/app/features/call/CallAudioManager.kt
+++ b/vector/src/main/java/im/vector/app/features/call/CallAudioManager.kt
@@ -125,7 +125,7 @@ class CallAudioManager(
             } else {
                 // if a wired headset is plugged, sound will be directed to it
                 // (can't really force earpiece when headset is plugged)
-                if (isBluetoothHeadsetOn()) {
+                if (isBluetoothHeadsetConnected(audioManager)) {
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/call/CallAudioManager.kt
```diff
@@ -246,7 +246,7 @@ class CallAudioManager(
     }
 
     private fun isHeadsetOn(): Boolean {
-        return isWiredHeadsetOn() || isBluetoothHeadsetOn()
+        return isWiredHeadsetOn() || (audioManager?.let { isBluetoothHeadsetConnected(it) } ?: false)
```
bmarty, line 249: `audioManager` should only be used from a single thread, according to Valere's comment.
The fix LGTM, but let discuss with Valere about it.

BillCarsonFr: Should be ok to read state

bmarty: Thanks

PR #2018, opened 2020-08-27
Feature/moar sonar fixes

bmarty:


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/RoomAPI.kt
```diff
@@ -220,9 +220,9 @@ internal interface RoomAPI {
      */
     @POST(NetworkConstants.URI_API_PREFIX_PATH_R0 + "rooms/{roomId}/send_relation/{parent_id}/{relation_type}/{event_type}")
     fun sendRelation(@Path("roomId") roomId: String,
-                     @Path("parentId") parent_id: String,
+                     @Path("parent_id") parentId: String,
```
onurays, line 223: Good catch.

vector/src/main/java/im/vector/app/features/reactions/widget/DotsView.kt
```diff
@@ -186,12 +186,12 @@ class DotsView @JvmOverloads constructor(context: Context, attrs: AttributeSet?
...
-            override operator fun get(`object`: DotsView): Float? {
-                return `object`.currentProgress
+            override operator fun get(o: DotsView): Float? {
+                return o.currentProgress
             }
 
-            override operator fun set(`object`: DotsView, value: Float?) {
-                `object`.currentProgress = value!!
+            override operator fun set(o: DotsView, value: Float?) {
+                o.currentProgress = value!!
```
onurays, line 194: We should remove the "?" on Float instead of adding !! to the value.

`override operator fun set(`object`: DotsView, value: Float) {
                o.currentProgress = value
            }`

onurays (approved):
LGTM! Just one remark.

vector/src/main/java/im/vector/app/features/reactions/widget/DotsView.kt
bmarty: Yes, probably, but this is out of scope for this PR.

PR #2020, opened 2020-08-28
Add email and phone numbers

bmarty:
## New entry in the settings

image

### Empty lists

image

### Not empty lists

image

### Adding emails

image

### Adding phones

image

### Pending state

image


BillCarsonFr:
There is often a re-send sms / mail in other services, can we do that?

bmarty:
> There is often a re-send sms / mail in other services, can we do that?

Yes, the `send_attempt` exists to support that. We are not using it though (Element Web neither) for the moment. It's easy to restart the whole process in case of error

vector/src/main/res/xml/vector_settings_general.xml
```diff
@@ -22,33 +22,11 @@
...
-
-        <!-- Note: iconTint does not work, it is also done in code -->
         <im.vector.app.core.preference.VectorPreference
-            android:icon="@drawable/ic_material_add"
-            android:key="ADD_PHONE_NUMBER_PREFERENCE_KEY"
-            android:order="200"
-            android:title="@string/settings_add_phone_number"
-            app:iconTint="@color/riotx_accent" />
-
-        <im.vector.app.core.preference.VectorPreference
```
bmarty, line 46: This setting has been removed by mistake, please restore

ganfra: Nice catch!

PR #2028, opened 2020-08-31
Relax IS dicovery

### LINKED ISSUE #2027: Login with mxid | Autodiscovery fails if identity server is invalid and Homeserver ok
If the well-known/client contains a valid HomeServer but an invalid Identity server the app won't let you login
image

for privacy tools config
````
{
    "m.homeserver": {
        "base_url": "https://chat.privacytools.io"
    },
    "m.identity_server": {
        "base_url": "https://chat.privacytools.io"
    }
}
````

As per spec, if m.identity_server is present it should be valid.
But it looks like web as relaxed that rule:
https://github.com/matrix-org/matrix-js-sdk/pull/1062

see https://github.com/vector-im/element-web/issues/11508
see https://github.com/vector-im/element-web/issues/11102

BillCarsonFr:
Fixes #2027

Align with web and relax IS discovery for login with matrix id.
See https://github.com/vector-im/element-web/issues/11102

I hesitated to create a FailPromptInvalidateIS data class (would be cleaner), but maybe it's better to stick with types defined in spec?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@ Bugfix 🐛:
  - Change user or room avatar: when selecting Gallery, I'm not proposed to crop the selected image (#1590)
  - Fix uploads still don't work with room v6 (#1879)
  - Can't handle ongoing call events in background (#1992)
+ - Login with Matrix-Id | Autodiscovery fails if identity server is invalid and Homeserver ok (#2027)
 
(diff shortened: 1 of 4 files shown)


bmarty (approved):
LGTM, let stick with the model from the spec, which is far from perfect though.

PR #1702, opened 2020-07-16
Use Context#getSystemService extension function provided by core-ktx.

johnjohndoe:
- Here the `Context#getSystemService` extension function is used which is provided by the `core-ktx` artifact.
- Sources: https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/core/core-ktx/src/main/java/androidx/core/content/Context.kt#30

Signed-off-by: Tobias Preuss tobias.preuss@googlemail.com

---

_Feel free to rebase the branch before merging it to facilitate a linear history. Let me know if I should do this._

PR #2030, opened 2020-08-31
Turn the screen off while using ear speaker

### LINKED ISSUE #1735: Call screen does not use proximity sensor
In a voice call holding the phone to the ear does not turn off the display so touching it with your ear or cheek is opening up the notification drawer and doing some other unwanted things.

onurays:
Fixes #1735

### DIFF
--- a/vector/src/main/java/im/vector/app/core/services/VectorService.kt
+++ b/vector/src/main/java/im/vector/app/core/services/VectorService.kt
@@ -17,10 +17,8 @@
 package im.vector.app.core.services
 
 import android.app.Service
-import android.content.Context
 import android.content.Intent
 import android.os.IBinder
-import im.vector.app.core.extensions.vectorComponent
 import timber.log.Timber
 
 /**
@@ -32,10 +30,6 @@ abstract class VectorService : Service() {
      * Tells if the service self destroyed.
      */
     private var mIsSelfDestroyed = false
-
-    override fun attachBaseContext(base: Context) {
-        super.attachBaseContext(vectorComponent().vectorConfiguration().getLocalisedContext(base))
-    }
 
     override fun onCreate() {
         super.onCreate()
--- a/vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
+++ b/vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
@@ -23,6 +23,7 @@ import android.hardware.SensorEvent
 import android.hardware.SensorEventListener
 import android.hardware.SensorManager
 import android.os.PowerManager
+import androidx.core.content.ContextCompat.getSystemService
 import im.vector.app.R
 import javax.inject.Inject
 
@@ -33,15 +34,13 @@ class CallProximityManager @Inject constructor(val context: Context) : SensorEve
 
     private val PROXIMITY_WAKE_LOCK_TAG = "PROXIMITY_WAKE_LOCK_TAG"
 
-    private lateinit var powerManager: PowerManager
-    private lateinit var sensorManager: SensorManager
+    private var powerManager: PowerManager = getSystemService(context, PowerManager::class.java)!!
+    private var sensorManager: SensorManager = getSystemService(context, SensorManager::class.java)!!
 
     private var wakeLock: PowerManager.WakeLock? = null
     private var sensor: Sensor? = null
 
     init {
-        powerManager = context.getSystemService(Context.POWER_SERVICE) as PowerManager
-        sensorManager = context.getSystemService(Context.SENSOR_SERVICE) as SensorManager
         sensor = sensorManager.getDefaultSensor(Sensor.TYPE_PROXIMITY)
     }
 
--- a/vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
+++ b/vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
@@ -39,14 +39,16 @@ class CallProximityManager @Inject constructor(val context: Context) : SensorEve
     private var wakeLock: PowerManager.WakeLock? = null
     private var sensor: Sensor? = null
 
+    init {
+        powerManager = context.getSystemService(Context.POWER_SERVICE) as PowerManager
+        sensorManager = context.getSystemService(Context.SENSOR_SERVICE) as SensorManager
+        sensor = sensorManager.getDefaultSensor(Sensor.TYPE_PROXIMITY)
+    }
+
     /**
      * Start listening the proximity sensor. [stop] function should be called to release the sensor and the WakeLock.
      */
     fun start() {
-        powerManager = context.getSystemService(Context.POWER_SERVICE) as PowerManager
-        sensorManager = context.getSystemService(Context.SENSOR_SERVICE) as SensorManager
-        sensor = sensorManager.getDefaultSensor(Sensor.TYPE_PROXIMITY)
-
         if (sensor != null && powerManager.isWakeLockLevelSupported(PowerManager.PROXIMITY_SCREEN_OFF_WAKE_LOCK)) {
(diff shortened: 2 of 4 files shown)


vector/src/main/java/im/vector/app/core/services/VectorService.kt
```diff
@@ -33,10 +31,6 @@ abstract class VectorService : Service() {
      */
     private var mIsSelfDestroyed = false
 
-    override fun attachBaseContext(base: Context) {
```
BillCarsonFr, line 36: Why is this removed?

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Start listening the proximity sensor. [stop] function should be called to release the sensor and the WakeLock.
+     */
+    fun start() {
+        powerManager = context.getSystemService(Context.POWER_SERVICE) as PowerManager
+        sensorManager = context.getSystemService(Context.SENSOR_SERVICE) as SensorManager
+        sensor = sensorManager.getDefaultSensor(Sensor.TYPE_PROXIMITY)
+
+        if (sensor != null && powerManager.isWakeLockLevelSupported(PowerManager.PROXIMITY_SCREEN_OFF_WAKE_LOCK)) {
```
BillCarsonFr, line 50: could it be really null?

Maybe we need some synchronisation?

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
```diff
@@ -0,0 +1,92 @@
...
+            sensorManager.registerListener(this, sensor, SensorManager.SENSOR_DELAY_NORMAL)
+        }
+    }
+
+    /**
+     * Stop listening proximity sensor changes and release the WakeLock.
+     */
+    fun stop() {
+        if (this::sensorManager.isInitialized) {
```
BillCarsonFr, line 59: why not initialize in constructor if it's lazy?

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+        wakeLock?.release()
+    }
+
+    override fun onAccuracyChanged(sensor: Sensor?, accuracy: Int) {
+        // NOOP
+    }
+
+    override fun onSensorChanged(event: SensorEvent) {
+        val distanceInCentimeters = event.values[0]
```
BillCarsonFr, line 70: In the app rtc reference app, they doing a check against sensor type, maybe we should do it too? if (event.sensor.getType() != Sensor.TYPE_PROXIMITY) return;

vector/src/main/java/im/vector/app/features/call/VectorCallViewModel.kt
```diff
@@ -145,10 +146,17 @@ class VectorCallViewModel @AssistedInject constructor(
         }
 
         override fun onAudioDevicesChange(mgr: WebRtcPeerConnectionManager) {
+            val currentSoundDevice = mgr.audioManager.getCurrentSoundDevice()
+            if (currentSoundDevice == CallAudioManager.SoundDevice.PHONE) {
```
BillCarsonFr, line 150: Why only do it if on ear peace?

BillCarsonFr (changes requested):
Please see my comments

vector/src/main/java/im/vector/app/core/services/VectorService.kt
onurays: It is crashing the application. @bmarty suggested to remove it.

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
onurays: Yes, the official documentation says some devices may not have the proximity sensor.
https://developer.android.com/guide/topics/sensors/sensors_position#sensors-pos-prox

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
onurays: It is not lazy, but there is a change that `stop()` may be called before `start()`. But it is ugly, I moved initializations to the constructor.

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
onurays: We are pretty sure we have only proximity sensor here since we don't register any other sensors. I think it is so defensive coding.

vector/src/main/java/im/vector/app/features/call/VectorCallViewModel.kt
onurays: We are trying to protect the screen against undesired touches with the ear. Do you think we have another use case?

vector/src/main/java/im/vector/app/core/services/VectorService.kt
BillCarsonFr: @bmarty any thought?

vector/src/main/java/im/vector/app/features/call/VectorCallViewModel.kt
BillCarsonFr: I think there could be some use to switch from earpiece/speaker depending on proximity, but maybe as an option

vector/src/main/java/im/vector/app/core/services/VectorService.kt
bmarty: This is OK to remove it, I think this is not really necessary for Services. It was added recently (after the previous release) to have a localized Context. It's probably useless for Services.

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    private lateinit var powerManager: PowerManager
+    private lateinit var sensorManager: SensorManager
+
+    private var wakeLock: PowerManager.WakeLock? = null
+    private var sensor: Sensor? = null
+
+    init {
+        powerManager = context.getSystemService(Context.POWER_SERVICE) as PowerManager
+        sensorManager = context.getSystemService(Context.SENSOR_SERVICE) as SensorManager
```
bmarty, line 44: Please use `getSystemService<Class>` (see https://github.com/vector-im/element-android/pull/1702)

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
```diff
@@ -0,0 +1,92 @@
...
+            onProximityNear()
+        } else {
+            onProximityFar()
+        }
+    }
+
+    /**
+     * Recommending naming convention for WakeLock tags is "app:tag"
+     */
+    private fun generateWakeLockTag() = "${context.getString(R.string.app_name)}:$PROXIMITY_WAKE_LOCK_TAG"
```
bmarty, line 81: You should inject and use the StringProvider to get Strings

bmarty (changes requested):
2 remarks for code quality

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
onurays: Done.

vector/src/main/java/im/vector/app/features/call/CallProximityManager.kt
onurays: Done.

bmarty (approved):

Issue #1375, opened 2020-05-16
Strange formatted message handling


ydnax:
riotX shows strange display behavior with formatted messages.

# Expected behavior:
if a message contains `"format":"org.matrix.custom.html" `and a `formatted_body` field, the contents of the `formatted_body` are displayed.
# Current behavior:
if the `body` field contains a code block in markdown, the body field is parsed as markdown and displayed instead of the formatted body(markdown parsing only if it is a `org.matrix.custom.html` message). 

Furthermore, when the html body is used, the pre/code block is not properly formatted(see third example in the screenshots below).

screenshots:

Expected message display(riot desktop):
![2020-05-16-184925_456x240_scrot](https://user-images.githubusercontent.com/676264/82125389-3696df80-9795-11ea-9043-39a05be07087.png)

Riotx display:
![2020-05-16-185301_1074x388_scrot](https://user-images.githubusercontent.com/676264/82125483-bfae1680-9795-11ea-90eb-bbe3676894ed.png)

Message sources:
```
  "content": {
    "msgtype": "m.text",
    "format": "org.matrix.custom.html",
    "body": "```\ncode body 1",
    "formatted_body": "<pre><code>html body 1</code></pre>"
  },
```

```
  "content": {
    "msgtype": "m.text",
    "format": "org.matrix.custom.html",
    "body": "normal body 2",
    "formatted_body": "<pre><code>html body 2</code></pre>"
  },
```

```
  "content": {
    "body": "```\ncode body 3",
    "msgtype": "m.text"
  },
```
@tulir identified the source for the issue [here](https://github.com/vector-im/riotX-android/blob/5fa247a0c5c02e34ea2caf6b68e458b115eb21ce/vector/src/main/java/im/vector/riotx/features/home/room/detail/timeline/factory/MessageItemFactory.kt#L356)

Our troubleshooting discussion in the #riotx room can be found [here](https://matrix.to/#/!gGkYcJLTzcNrfERksd:matrix.org/$a2pBDO4VhlGZyxQyM2KJRGUeMC6aIk5PbYuwSyghwAI?via=matrix.org&via=privacytools.io&via=tchncs.de)

Tested version: 0.19.0 [40019000] (F-be9393fa)

tulir:
Somewhat related: #350

bmarty:
I reopen the issue, we will close it once when we can confirm that we do not reproduce it

Issue #350, opened 2019-07-12
Formatted (HTML) messages with markdown features are rendered incorrectly as markdown


tulir:
When the `formatted_body` field of a `"format": "org.matrix.custom.html"` message contains special markdown characters, they're rendered as markdown instead of as plain text. This could be an actual problem if you escape markdown characters in a message that contains other formatting

<details>
<summary>Exaggerated example</summary>

### Expected rendering
![riot web](https://user-images.githubusercontent.com/4224639/61151780-64fdbb80-a4ef-11e9-90f2-50583bcaa874.png)
![riot android](https://user-images.githubusercontent.com/4224639/61151845-8ced1f00-a4ef-11e9-9216-ac15e4bac826.png)

### Actual rendering
![riot x](https://user-images.githubusercontent.com/4224639/61151872-99717780-a4ef-11e9-8345-1a7567961616.png)

### Event source
```json
{
  "type": "m.room.message",
  "sender": "@tulir:maunium.net",
  "content": {
    "msgtype": "m.text",
    "body": "Blank message (or is it? 🤔)",
    "format": "org.matrix.custom.html",
    "formatted_body": "---\n> * # [`RiotX` renders this message ***completely incorrectly***](https://github.com/vector-im/riotX-android/issues/289#issuecomment-510894293)\n---"
  },
  "event_id": "$156295723912295JSonL:maunium.net",
  "origin_server_ts": 1562957239380,
  "unsigned": {
    "age": 52
  },
  "room_id": "!tTIucwUZLRtKnXeurb:matrix.org"
}
```
</details>

<details>
<summary>Non-exaggerated example</summary>

### Expected rendering
![riot web](https://user-images.githubusercontent.com/4224639/61152222-7abfb080-a4f0-11e9-87d9-fa0d30c7994b.png)
![riot android](https://user-images.githubusercontent.com/4224639/61152269-9f1b8d00-a4f0-11e9-9d15-9664b5010a04.png)

### Actual rendering
![riot x](https://user-images.githubusercontent.com/4224639/61152284-a9d62200-a4f0-11e9-8f01-2fdf9a11753b.png)

### Event source
```json
{
  "type": "m.room.message",
  "sender": "@tulir:maunium.net",
  "content": {
    "msgtype": "m.text",
    "format": "org.matrix.custom.html",
    "body": "*less exaggerated example*: \\*\\*stars!**",
    "formatted_body": "<em>less exaggerated example</em>: **stars!**"
  },
[...]

PR #2043, opened 2020-09-01
Cleanup after #2002

bmarty:
Following what have been done on #2002, tests had to be updated.


bmarty:
@MexHigh I'm not sure that removing `\n` from HTML code was source of bugs. Could you confirm please?

tulir:
Any bugs that appear to be caused by removing `\n` from outgoing HTML are actually caused by whatever breaks when receiving HTML without `\n`. Other clients might remove `\n` even if you don't, so the bug needs to be fixed in the handling, not the sending.

Also as I noted in https://github.com/vector-im/element-android/pull/2002#issuecomment-685113343, at least two of the four mentioned issues were not fixed.

bmarty:
The more I think about it, the more I have the feeling that having a `body` and a `formatted_body` is really a bad solution. The text itself is duplicated (and can even be totally different...), and it's not possible to handle all the possible cases without bug: save draft, edit message, etc.

bmarty:
Anyway @tulir I think you are right, I will revert some change and reopen closed issues by #2002, until we can confirm that they are fixed.

MexHigh:
One of the main problems with removing all `\n` from a message was that if it contains not only code, but normal text as well, the linebreaks in the plaintext would be removed as well. Another solution would be to replace `\n` with `<br>` tags but they are not used in "normal" messages either.
Linebreaks don't get removed on Element Web too, so I've oriented myself towards that.
As @tulir mentioned, the two issues (#350 and #1375) were not fixed, because I misunderstood what the actual problems were. My bad.

bmarty:
@MexHigh thanks.
I will revert the change on line break then (in this PR). And add unit test to cover the case you are talking about.

Issue #1833, opened 2020-07-28
 Update associated site domain


bwindels:
The url to intercept has been changed in https://github.com/vector-im/riot-web/pull/14796, please update the app accordingly.

PR #2047, opened 2020-09-02
Handle room, user, group, email verification links by converting them to permalinks.

onurays:


vector/src/main/java/im/vector/app/features/link/LinkHandlerActivity.kt
```diff
@@ -62,9 +68,45 @@ class LinkHandlerActivity : VectorBaseActivity() {
...
-            finish()
+            if (!sessionHolder.hasActiveSession()) {
+                startLoginActivity(uri)
+                finish()
+            } else {
+                convertUrlToPermalink(uri.toString())?.let { permalink ->
+                    startPermalinkHandler(permalink)
+                } ?: run {
+                    // Other link are not yet handled, but should not comes here (manifest configuration error?)
+                    Timber.w("Unable to handle this uir: $uri")
```
bmarty, line 79: There is a typo in the text: `uir` (I know it is not new :) )

onurays: Fixed.

bmarty:
We should also support `develop.element.io` and `staging.element.io` hosts. Can you handle that in the manifest and in the code please?

vector/src/main/java/im/vector/app/features/link/LinkHandlerActivity.kt
```diff
@@ -55,19 +60,56 @@ class LinkHandlerActivity : VectorBaseActivity() {
...
+            } ?: run {
+                // Host is correct but we do not recognize path
+                Timber.w("Unable to handle this uri: $uri")
+                finish()
+            }
         }
     }
 
+    private fun convertUriToPermalink(uri: Uri): String? {
+        val path = SUPPORTED_PATHS.find { it in uri.toString() } ?: return null
```
bmarty, line 94: You should rather test if `uri.path` starts with one of the supported paths.

bmarty:
@onurays Can you handle https://github.com/vector-im/element-android/issues/1833 is this PR too? (with a dedicated commit please)

vector/src/main/java/im/vector/app/features/link/LinkHandlerActivity.kt
onurays: Unfortunately Android cannot parse path because of "#" in the url.

vector/src/main/AndroidManifest.xml
```diff
@@ -144,8 +144,8 @@
 
                 <data android:scheme="https" />
                 <data android:host="riot.im" />
-                <data android:host="element.io" />
-                <data android:pathPrefix="/config/" />
+                <data android:host="app.element.io" />
```
bmarty, line 147: We should also handle `develop.element.io` and `staging.element.io` links. Please add theme here and in SUPPORTED_HOSTS

bmarty (changes requested):
One more remark

bmarty:
More things to check:

- [ ] I'm doing some test, all `element.io` links work pretty well, but `riot.im` links does not work (ex: https://riot.im/develop/#/room/#element-android:matrix.org). Maybe due to the `/develop` path, I let you check
- [ ] The displayed error is not correct in the case of a element.io link: `<string name="permalink_malformed">Your matrix.to link was malformed</string>`. You should define another string
- [ ] This PR also fixes #1833, please mention it in the Changelog

vector/src/main/java/im/vector/app/features/link/LinkHandlerActivity.kt
```diff
@@ -54,20 +59,59 @@ class LinkHandlerActivity : VectorBaseActivity() {
...
+        // https://riot.im/develop/#/room/#element-android:matrix.org ->  https://matrix.to/#/#element-android:matrix.org
+        // https://app.element.io/#/room/#element-android:matrix.org  ->  https://matrix.to/#/#element-android:matrix.org
+        return uri
+                .toString()
+                .replace(uri.toString().substring(0, uri.toString().indexOf(path) + path.length), "https://$MATRIX_TO_HOST/#")
```
bmarty, line 99: You can simply write
`return "https://$MATRIX_TO_HOST/#" + uri.toString().substringAfter(path)`

vector/src/main/java/im/vector/app/features/link/LinkHandlerActivity.kt
bmarty: Ok

PR #2053, opened 2020-09-03
Feature/fix poll reply

### LINKED ISSUE #1004: Replies to poll appears in timeline as unsupported events during sending
image

image

Event has a relates_to with the type "m.response"

### LINKED ISSUE #1246: JSON for verification events leaks in to the room list
It's ugly that the message previews in the room list show italicised JSON for verification events - surely it should show the most recent m.room.message or possibly m.call.* rather than any old event?

BillCarsonFr:
Fixes #1246
 - JSON for verification events leaks in to the room list ()

Fixes #1004
 - Replies to poll appears in timeline as unsupported events during sending 

Looks like there was a desync between the timeline filter and the sendingEvent/InMemory filter
Also bus event for local echo creation could add local echos that was never removed (no synced event as they were filtered out)

+ Fixed a bug on poll reply not taken into account: Currently aggregation don't support relation put on a sending event (e.g edit a sending event, reply to a sending event); there was no guard against that for poll buttons, added now

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
@@ -17,6 +17,16 @@
 
 package org.matrix.android.sdk.internal.session.room.timeline
 
+import io.realm.OrderedCollectionChangeSet
+import io.realm.OrderedRealmCollectionChangeListener
+import io.realm.Realm
+import io.realm.RealmConfiguration
+import io.realm.RealmQuery
+import io.realm.RealmResults
+import io.realm.Sort
+import org.greenrobot.eventbus.EventBus
+import org.greenrobot.eventbus.Subscribe
+import org.greenrobot.eventbus.ThreadMode
 import org.matrix.android.sdk.api.MatrixCallback
 import org.matrix.android.sdk.api.extensions.orFalse
 import org.matrix.android.sdk.api.session.events.model.EventType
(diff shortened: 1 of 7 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
```diff
@@ -316,12 +316,15 @@ internal class DefaultTimeline(
...
     fun onLocalEchoCreated(onLocalEchoCreated: OnLocalEchoCreated) {
         if (isLive && onLocalEchoCreated.roomId == roomId) {
-            listeners.forEach {
-                it.onNewTimelineEvents(listOf(onLocalEchoCreated.timelineEvent.eventId))
+            // do not add events that would have been filtered
+            if (listOf(onLocalEchoCreated.timelineEvent).filterEventsWithSettings().isNotEmpty()) {
+                listeners.forEach {
+                    it.onNewTimelineEvents(listOf(onLocalEchoCreated.timelineEvent.eventId))
+                }
+                Timber.v("On local echo created: $onLocalEchoCreated")
```
bmarty, line 324: Please be careful, it leaks the content of the event to the logs

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt
```diff
@@ -1194,6 +1195,8 @@ class RoomDetailViewModel @AssistedInject constructor(
     }
 
     private fun handleReplyToOptions(action: RoomDetailAction.ReplyToOptions) {
+        // Do not allow to reply to unsent local echo
+        if (LocalEcho.isLocalEchoId(action.eventId)) return
```
bmarty, line 1199: Buttons should be disabled, and I do not think it is the case.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
bmarty: I handle it

PR #2058, opened 2020-09-04
Start sync when receive m.call.invite.

### LINKED ISSUE #1921: Incoming call continues to ring if call is answered on another device
**Describe the bug**
If an incoming call rings Element-Android and simultaneously rings Element-Web, and the call is answered on Element-Web, Element-Android continues to ring even though the call has been answered. If the call is dismissed from the android application, the call is terminated on the web session.

**To Reproduce**
1. open Element web with account A
2. Log into Element Android with account A
3. Use Account B to call Account A
4. Answer the incoming call to Account A in Element Web, wait to see that Element Android still indicates an incoming call
5. Decline the incoming call in Element Android after accepting the call (step 4) and see that the call is terminated in the Element Web session

**Expected behavior**
The call should be connected through Element Web, and Element Android should stop ringing because the call was answered elsewhere. The call should also continue to function in Element Web even if declined in the Element Android session.

 - Device: Google Pixel
 - OS: Android 10

**Additional context**
 - App version and store [1.0.4 - F-Droid]
 - Homeserver: [jobmachine.org]

onurays:
Fixes #1921


progserega:
Do the sync will stop, when call ended?
In mobile 1.0.7 ring stop ringing after get call on PC, but mobile app start infinity sync-loop: #2163

BillCarsonFr (approved):
Added few commits LGTM

PR #2061, opened 2020-09-04
Fix / verify popup won't show

BillCarsonFr:
Verify popup won't show if alert does not define shouldBeDisplayedIn callback (it should show if this is not set)


bmarty:
Let me propose another fix, with a code easier to understand

bmarty:
Rejected in favor of #2062

PR #2062, opened 2020-09-04
Feature/fix no verif

bmarty:
Other solution for #2061 (in the first commit)

But I do not reproduce the issue with Element 1.0.5, verification popup are shown.

There are also shown with this PR


BillCarsonFr (approved):
LGTM

Issue #2064, opened 2020-09-05
[Crash] App crash when read Contact permission is granted


bmarty:
Reported by [27 rageshakes](https://github.com/matrix-org/element-android-rageshakes/issues?q=is%3Aissue+is%3Aopen+%22Failure+delivering+result+ResultInfo%7Bwho%3D%40android%3ArequestPermissions%22) so far

Not reproduced on my device (Android 10), but the majority of RS refer to Android 6 devices.

```
Thread: main, Exception: java.lang.RuntimeException: Failure delivering result ResultInfo{who=@android:requestPermissions:, request=579, result=-1, data=Intent { act=android.content.pm.action.REQUEST_PERMISSIONS (has extras) }} to activity {im.vector.app.debug/im.vector.riotx.features.createdirect.CreateDirectRoomActivity}: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
```

bmarty:
Fixed by #2067

PR #2067, opened 2020-09-05
Android 6: App crash when read Contact permission is granted (#2064)

bmarty:
Ugly fix for #2064.
Tested ok on Emulator running Android 6 (API 23).

I cannot find a better fix for now. @onurays @BillCarsonFr @ganfra any thoughts?

### DIFF
--- a/vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
+++ b/vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
@@ -84,6 +84,7 @@ import im.vector.app.receivers.DebugReceiver
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.disposables.CompositeDisposable
 import io.reactivex.disposables.Disposable
+import org.matrix.android.sdk.api.extensions.tryThis
 import org.matrix.android.sdk.api.failure.GlobalError
 import timber.log.Timber
 import kotlin.system.measureTimeMillis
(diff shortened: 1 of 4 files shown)


vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
```diff
@@ -345,6 +345,14 @@ abstract class VectorBaseActivity : AppCompatActivity(), HasScreenInjector {
                 }
     }
 
+    protected var postResumeScheduledAction: Runnable? = null
```
ganfra, line 348: Could u create a list instead? so we can handle multiple things

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomActivity.kt
```diff
@@ -113,7 +115,13 @@ class CreateDirectRoomActivity : SimpleFragmentActivity() {
     override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<String>, grantResults: IntArray) {
         if (allGranted(grantResults)) {
             if (requestCode == PERMISSION_REQUEST_CODE_READ_CONTACTS) {
-                addFragmentToBackstack(R.id.container, ContactsBookFragment::class.java)
+                try {
+                    addFragmentToBackstack(R.id.container, ContactsBookFragment::class.java)
```
ganfra, line 119: I wouldn't even try to do this, directly post it to postResume?

ganfra (changes requested):
Some remarks. Also, you might want to check for other fragment transactions

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomActivity.kt
bmarty: the bug is only for Android 6, so I assume it is a workaround... But Ok I will change that

vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
bmarty: No usage for the moment (so no need to do complicated things), but yes, why not...

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomActivity.kt
ganfra: It's more a question than a request, I don't know if it'll work

ganfra (approved):
Thanks for the changes, LGTM now, but I did not test it

bmarty:
I tested emulator API 23 and real device API 29

PR #2069, opened 2020-09-07
Support data-mx-[bg-]color attribute on <font> tags in formatted messages.

etherealvisage:
### Summary

Pretty much what it says on the can :-) As outlined [in the spec](https://matrix.org/docs/spec/client_server/r0.6.1#m-room-message-msgtypes), `<font>` tags can have the `data-mx-bg-color` attribute attached to indicate that text should have a different background colour. While this is somewhat counter to things like good readability in many instances with e.g. dark or light themes, only supporting the foreground colour also renders some text unreadable for some messages that change both.

First time writing Kotlin, so I'm sure the styling could be improved :-)

### Pull Request Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/html/FontTagHandler.kt
+++ b/vector/src/main/java/im/vector/app/features/html/FontTagHandler.kt
@@ -17,6 +17,7 @@ package im.vector.app.features.html
 
 import android.graphics.Color
 import android.text.style.ForegroundColorSpan
+import android.text.style.BackgroundColorSpan
 import io.noties.markwon.MarkwonConfiguration
 import io.noties.markwon.RenderProps
 import io.noties.markwon.html.HtmlTag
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/html/FontTagHandler.kt
```diff
@@ -31,7 +32,13 @@ class FontTagHandler : SimpleTagHandler() {
...
     override fun getSpans(configuration: MarkwonConfiguration, renderProps: RenderProps, tag: HtmlTag): Any? {
         val colorString = tag.attributes()["color"]?.let { parseColor(it) } ?: Color.BLACK
-        return ForegroundColorSpan(colorString)
+        val bgColor = tag.attributes()["data-mx-bg-color"]
+        if(bgColor == null) {
+            return ForegroundColorSpan(colorString)
+        }
+        else {
+            return arrayOf(ForegroundColorSpan(colorString), BackgroundColorSpan(bgColor.let { parseColor(it) }))
+        }
```
bmarty, line 41: `parseColor` will fallback to BLACK in case of error, so it can lead to text rendered in black, on a black background.
Maybe add a parameter to `pasreColor` with the color to use in case of error. I will use TRANSPARENT for default bg color.

Anyway, using BLACK color for default text color is not good too, because text color depends on the current theme. But this is another story, and not related to this PR

bmarty (changes requested):
Thanks for the PR.

See my remark above.

Maybe take the opportunity to also add support for `data-mx-color` ? (Ref: https://matrix.org/docs/spec/client_server/r0.6.1#m-room-message-msgtypes)

vector/src/main/java/im/vector/app/features/html/FontTagHandler.kt
etherealvisage: Done :-)

etherealvisage:
Updated as requested. I decided to give `data-mx-color` precedence over `color` if present, as that is the one that is specified in the spec. :-)

Let me know if there's anything else I can improve.

bmarty (approved):
Thanks for the update!
It looks good to me.

PR #2075, opened 2020-09-08
Feature/strict mode tracking

ganfra:
Enable strict mode to check for long process made on main thread. 
Most are related to realm, so created a RealmSessionProvider to have a long running realm instance when we are on main thread. Remove some small other things.


bmarty:
I've rebased and added a commit.
We can merge if CI is happy

PR #2080, opened 2020-09-09
Feature/polling work

### LINKED ISSUE #1915: [F-Droid] Excessive offline battery usage

### LINKED ISSUE #1920: Stop background sync when network connection is lost

### LINKED ISSUE #2055: [META] Improve F-Droid Notification
## TL;DR

The scope of this issue is to give F-Droid users more control on Background Syncs without needed any changes on other system (sygnal) and only using full syncs. So basically adding back the same options that were available in riot android

## Problem

The F-Droid element app cannot rely on the Google Push solution (FCM/GCM) that is propriatery code and therefore cannot be used.
As a consequence F-Droid apps need to implement an alternative way of getting messages while the app is in background or killed. This is usually achieved by doing some long polling, or keeping alive a websocket connection while the app is not in foreground.
But recent android version introduced several battery saving features (aka [doze](https://developer.android.com/training/monitoring-device-state/doze-standby.html)), that drastically reduced the capacity for apps to perform any kind of work in background.
If you add to that additional [vendors custom limitations](https://dontkillmyapp.com/problem), you get a fragmented environment making it hard to test and would force user to perform some specific manual step to make thing works.

Notice that when using Google FCM, you get a magic out of jail card, that allow your App to sync in background after receiving a push;

### Current Polling Solution used by Element

Element Android implemented the following solution:
 1. When the App enters Background, an **[Alarm](https://developer.android.com/reference/android/app/AlarmManager)** is scheduled, using **setAndAllowWhileIdle**. This should allow the alarm to be scheduled even if the app is in *doze* mode. When fired the Alarm will send a **broadcast** to the SyncBroadcastReceiver

_It is said that under normal operations alarms should not be allowed to run more than every minutes, in doze mode it is 15mn, white listing your app from battery optimization would relax those rules but you could have the joy to see your alarm executed but with network disabled 🙃_

2. Upon wake-up the **SyncBroadcastReceiver** will:
     - Aquire an additional PARTIAL_WAKE_LOCK to give the app more time
     - Starts a foreground service (SyncService), that is required to display a notification to the user in order to be allowed to run (the `listening for events`notification)
    - Then reschedule an **Alarm** in 30s (hard coded for now)

3. The  **SyncService** is then started, it will perform a complete [sync](https://matrix.org/docs/spec/client_server/latest#syncing) with a 0s timeout, that will get all new events for all your rooms

### Difference with the former riot-android app

The former riot-android app had 3 fdroid notification mode:
 - Optimized for battery
 - Optimized for real-time
 - No background sync

The `optimized for battery mode` was using Workers and was letting the system manage frequency of execution depending on system state (so syncs could be heavilly delayed or even stopped in doze mode)
The `optimized for real-time`was using Alarms to force wake up, and required to be whitelisted from battery optimizations, ensuring users to get notifications even if the phone is in doze (e.g standing still on a table for sev 10's of minutes)

Default delay between syncs is 60s, default timeout is 6s, and those are configurable by the user.

So Element Android way of working is the old `optimized for real-time` with 30s delay (instead of 60s), and 0s (instead of 6s) of timeout

=> **So yes Element android is by default lot more aggressive on battery than old Riot Android**

BillCarsonFr:
Fixes #2055
Fixes #1920 
Fixes #1915 

image

image

This PR add backs the fdroid sync modes that were available in the former app.

It offers 3 modes:

- __Optimized for battery__ : In this mode background syncs are scheduled via Workers, meaning that the system will have control on the periodicity of syncs when battery is low or when the phone is idle (sync will occcur in allowed maintenance windows).  After completion the sync work will schedule another one.

- __Optimized for Real Time__ : This mode requires the app to be exempted from battery optimization. Alarms will be launched and will wake up the app in order to perform the background sync as a foreground service. After completion the service will schedule another alarm

- __No background syncs__ : The app won't sync in background


vector/src/main/java/im/vector/app/features/settings/BackgroundSyncModeChooserDialog.kt
```diff
@@ -0,0 +1,172 @@
...
+            options.forEach {
+                it.isSelected = initialMode == it.mode
+                initialModeFound = true
+            }
+            if (!initialModeFound) {
+                options[0].isSelected = true
+            }
+        }
+
+        override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
```
bmarty, line 97: For a list view of 3 static items, it's really too much code...

DoM1niC:
Yeah Notification Bug is now fixed by this :) <3 THX @BillCarsonFr

PR #2081, opened 2020-09-09
Translations: Fix zh-rCN string's typo

chengww5217:
Hi there, I found a typo of the zh-rCN string when using element.

By default(EN), the string value is:
```xml
<string name="login_registration_not_supported">The application is not able to create an account on this homeserver.\n\nDo you want to signup using a web client?</string>
```

The origin value in zh-rCN is:
```xml
    <string name="login_registration_not_supported">应用无法再次服务器上创建账户。
\n
\n您想要通过网页客户端注册吗？</string>
```

Which means 
> The application is not able to create an account (~~on this~~) homeserver <strong>again</strong>...

It doesn't make sense in Chinese.
The right word should be <strong>在此</strong> which means <strong>on this</strong> (homeserver), instand of <strong>再次</strong>

This is an input error due to the similar pronunciation of two words.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Bugfix 🐛:
  -
 
 Translations 🗣:
- -
+ - Fix zh-rCN string's typo
 
 SDK API changes ⚠️:
  - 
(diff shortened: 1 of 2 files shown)


bmarty:
Good catch (I have to trust you :)).

Sorry, translations comes from Weblate. Please do changes on Weblate, as per https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#i-want-to-help-translating-element. If I merge your PR, we could have some trouble when doing the next Weblate sync.

PR #2103, opened 2020-09-14
Route audio device to earspeaker for audio calls

### LINKED ISSUE #1685: Loudspeaker is always used

onurays:
Fixes #1685


DoM1niC:
still not fix here the Loudspeaker still active and is not switchable between Phone / Speakers only if I use the BT Headset so its possible to switch to other one.

Current Dev + this Merge 
Huawei Mate 20 Lite (Android 10)

PR #2108, opened 2020-09-15
output stream not closed

### LINKED ISSUE #2107: Various report of people that cannot play video
> I can play it under iOS as said above. My Android devices are connected to my homeserver.
So i can‘t test it here. But reproducable i have problems like other users with sending - playing - and replaying videos under 1.0.6
FDroid 1.0.5 is still ok

> And i created an unencrypted room and send the video a second time. Same here, iOS, Windows Linux all good,
Android struggles again with the file.
Thought the encryption could cause this behaviour , but seems not to be the root of the prob.
And what is more important: existing videofiles that have been send by elderly versions in the past, can‘t now be played any longer too.

image

BillCarsonFr:
Fixes #2107

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,7 @@ Improvements 🙌:
 Bugfix 🐛:
  - Clear the notification when the event is read elsewhere (#1822)
  - Speakerphone is not used for ringback tone (#1644, #1645)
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, thanks!

PR #2109, opened 2020-09-15
Feature/fragment transaction

ganfra:
Enable usage of commitAllowingStateLoss for fragment transactions to handle cases where host activity/fragment is after onStop


bmarty (approved):
LGTM, thanks!

Issue #1397, opened 2020-05-24
Long message cannot be sent/takes infinite time & blocks other messages


rugk:
v0.2.0
SDK 0.0.1
OLM 3.1.2
Android 9

## STR

1. In an encrypted direct message chat, send a message. A long message! (very long)

**What happens:** It does a "sending" status (grey text etc.) all the time, takes an infinite time to send, but does not actually do so

**What should happen:** Send it.

On the desktop Riot client, sending the same message works without problems.

**Additional problem:** This one message now blocks all later messages. All later messages in that chat also go into the "endless loading cycle" mode.
Is there maybe a workaround for that? (okay new chat, but that is annoying) Can't we delete the old message.
Maybe a "cancel sending" button would be useful. (already tried to kill RiotX and restart it, that did not help)

bmarty:
Please define "very long". How many chars does the message contain (roughly)?

rugk:
19 lines, 651 words, 3830 characters

bmarty:
I have no pb sending a message to an encrypted room with 3830 "a".

If you enable developer option in the application setting (in advanced settings section), you will have an extra menu (top right) on the home to clear the sending queue IIRC

rugk:
Thx that workaround did exactly what I wanted to do.

--

As for the problem: Cannot reproduce it with 3830 a's too either. There have to be more criteria.
However, I can reproduce it with the original text.

Aaand with a text of 30 paragraphs of lorem ipsum:
```
Quidem odio labore fuga beatae laboriosam. Aut suscipit quod nemo itaque voluptate. Temporibus architecto a sunt impedit molestiae. Odio quidem ut tenetur modi aspernatur saepe quia velit. Quibusdam ducimus totam distinctio nobis ut. Assumenda nesciunt blanditiis sed necessitatibus similique deleniti.

Harum veniam vitae doloremque quia ut quia iure. Et magnam molestiae totam. Quidem repellendus maiores quod sed sit hic.

Eos dolor est qui iusto quidem. Cum et officiis nemo tempora. Assumenda ipsam maxime quisquam et et.

Neque numquam consequuntur distinctio praesentium. Sed voluptatibus doloribus voluptas voluptatem quam totam. Voluptatem ad consectetur qui sit hic. Asperiores pariatur incidunt sunt aliquid consequuntur qui reprehenderit. Magnam veniam animi illum labore optio.

[...]

PR #2129, opened 2020-09-18
Feature/fix sending too long

ganfra:
This PR lets sending workers responsible for getting the event in the db instead of passing it through params.
It allows to fix https://github.com/vector-im/element-android/issues/1397

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/RealmCryptoStore.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/RealmCryptoStore.kt
@@ -20,6 +20,10 @@ package org.matrix.android.sdk.internal.crypto.store.db
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.Transformations
 import com.zhuinden.monarchy.Monarchy
+import io.realm.Realm
+import io.realm.RealmConfiguration
+import io.realm.Sort
+import io.realm.kotlin.where
 import org.matrix.android.sdk.api.auth.data.Credentials
 import org.matrix.android.sdk.api.session.crypto.crosssigning.MXCrossSigningInfo
 import org.matrix.android.sdk.api.session.events.model.Event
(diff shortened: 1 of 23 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/RealmCryptoStore.kt
```diff
@@ -1191,7 +1191,7 @@ internal class RealmCryptoStore @Inject constructor(
                     .findAll()
                     .mapNotNull { entity ->
                         when (entity.type) {
-                            GossipRequestType.KEY    -> {
+                            GossipRequestType.KEY -> {
```
bmarty, line 1194: So you have the same pb than me, Android Studio mess up the alignment of `->`! You pretended it was not the case! Loul!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/send/LocalEchoIdentifiers.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package org.matrix.android.sdk.internal.session.room.send
+
+import com.squareup.moshi.JsonClass
+
+/**
+ * This is used as a holder to pass necessary data to some workers params.
+ */
+@JsonClass(generateAdapter = true)
+data class LocalEchoIdentifiers(val roomId: String, val eventId: String)
```
bmarty, line 25: internal?

bmarty:
Is migration of Work DB safe?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/send/LocalEchoIdentifiers.kt
ganfra: yup thanks

ganfra:
> Is migration of Work DB safe?

If there is some unsent events they will never sent, but the queue should be free.

BillCarsonFr (approved):
LGTM

PR #2132, opened 2020-09-18
Add possibility to delete account data and fix bug

bmarty:
- Add possibility to delete account data by type (advanced setting screen)
- Improve import Megolm key dialog (add show password as for other password input)
- After importing Megolm keys from manual backup, upload keys to the key backup if it exists.


BillCarsonFr (approved):
LGTM

PR #2133, opened 2020-09-18
Maths support (MSC2191)

NickHu:
This is a PR to implement maths support as described [here](https://github.com/matrix-org/matrix-doc/pull/2191). (TLDR: inline maths is given by HTML `<span data-mx-maths="latex here">fallback representation here</span>`, and display maths by `div` instead of `span`).

Fortunately, markwon already supports rendering LaTeX; it was just a
matter of enabling it, bumping its version (for inline LaTeX), and doing
some minor processing for MSC2191.

Current state:
- [X] Rendering
  - [X] Renders inline maths `<span data-mx-maths=...>...</span>`
  - [X] Renders display maths `<div data-mx-maths=...>...</div>`
  - [X] Render messages which have been edited
- [ ] Sending
  - [X] `$latex$` → `<span data-mx-maths="latex"><code>latex</code></span>`
  - [ ] `\(latex\)` → `<span data-mx-maths="latex"><code>latex</code></span>`
  - [X] `$$latex$$` → `<div data-mx-maths="latex"><code>latex</code></div>`
  - [ ] `\[latex\]` → `<div data-mx-maths="latex"><code>latex</code></div>`

I only implemented an inline [Delimiter parser](https://github.com/atlassian/commonmark-java/blob/master/commonmark/src/main/java/org/commonmark/parser/delimiter/DelimiterProcessor.java) so things like

```
$$
latex
$$
```
won't be parsed correctly (requires a block parser), and apparently this class only supports having (possibly many
consecutive) single-character parsers, so `\(` and `\[` weren't so straightforward to implement either. It seems like that with
https://github.com/atlassian/commonmark-java/pull/180 inline parsing is getting an overhaul anyway, so it might be best to
wait for that to implement parsing these. It's probably fine to omit this for a first version.

A strange bug is that when rendering a message which only contains maths, it seems to draw the message with very small vertical height to the point at which it appears like a blank bit of vspace (the content is not visible). Sometimes sections of display maths are cropped at the bottom which suggests in general I'm not setting the height of the TextView properly. Any guidance on this would be appreciated.

`Signed-off-by: Nick Hu <me@nickhu.co.uk>`

### Pull Request Checklist

- [X] Pull request includes screenshots or videos if containing UI changes


tobiasBora:
Can't wait for this to be merged ;-) Did you manage to solve the issue with cut texts?

a22sc:
Since matrix has become more and more popular in science and universities, this feature is a must have!
I'm not deep enough into the code to review the pull request myself. Maybe @bmarty or someone else has capacities..

vector/build.gradle
```diff
@@ -293,7 +293,8 @@ dependencies {
     def epoxy_version = '4.6.1'
     def fragment_version = '1.3.3'
     def arrow_version = "0.8.2"
-    def markwon_version = '4.1.2'
+    def coroutines_version = "1.3.8"
```
bmarty, line 296: why this change?

bmarty (changes requested):
Thanks for the PR. Is it possible to:
- [ ] Add a lab flag to this feature, disabled by default? The impacted code is sensible, we do not want to break anything, also we do not want to downgrade the performance
- [ ] Convert the Java classes you added to kotlin.

kylrth:
Where is this at? This feature is really important for me and it'd be good to get all this merged before the work becomes stale compared to develop.

antonis-tsolomitis:
I agree with @kylrth This is Very important for me too as a Mathematician. I hope someone fixes this to be merged soon.

NickHu:
I have ported the Java over to Kotlin and confirmed that things are working as they were before. I will try to put everything behind a labs setting next.

I still haven't managed to fix the clipping issue however. I think what happens is that JLaTeXMath schedules what it calls an [`AsyncDrawableScheduler`](https://github.com/noties/Markwon/blob/2ea148c30a07f91ffa37c0aa36af1cf2670441af/markwon-ext-latex/src/main/java/io/noties/markwon/ext/latex/JLatexMathPlugin.java#L268) over the `messageView`; however, this preserves the dimensions of the `messageView` which is (especially in display maths) smaller. I need to find some way to rerender the widget, but I'm not sure how.

NickHu:
Ok, now it's gated behind Labs

cc @bmarty

NickHu:
So I've determined that the 'wonkiness' with respect to clipping is a bit more subtle:

1. display maths gets clipped (but you can still essentially see it, just not the bottom of the integration symbol e.g.);
2. if a message contains **only** inline maths, it initially gets drawn with near zero height; however, once you scroll until it falls out of view (dropped by the `RecyclerView`), and then scroll back, it renders correctly.

In any case, it's quite usable even in this state (for the impatient, you can download the APK from buildkite once it finishes building), and I don't think I have the capability to fix this. I'm sure someone more experienced with Android development would be able to figure it out pretty quickly.

All told, I think this can be merged if it's going to be behind a labs toggle. People are pretty desperate to use this now, and I'm sure once it's in there it won't be long before someone can take this the rest of the way. Remember to enable the labs setting and then restart the app; also, enable Markdown formatting if you wish to send messages with LaTeX content.

antonis-tsolomitis:
> **for the impatient**: you can download the APK from buildkite once it finishes building

I am unexperienced... can someone write how to download the apk on an android phone?

mr-c:
> > **for the impatient**: you can download the APK from buildkite once it finishes building
> 
> I am unexperienced... can someone write how to download the apk on an android phone?

Go to https://buildkite.com/matrix-dot-org/element-android/builds/3299#68f085bf-2408-4a5c-8375-fbddc1c05399 
Click on `Artifacts` and side load the relevant APK for your phone.

antonis-tsolomitis:
> Go to https://buildkite.com/matrix-dot-org/element-android/builds/3299#68f085bf-2408-4a5c-8375-fbddc1c05399
> Click on `Artifacts` and side load the relevant APK for your phone.

Thanks a lot!

antonis-tsolomitis:
I installed this version on a 10" android tablet. It seems that there is a size bug for the math that may be relates to the clipping problem. The attached 
sizebug
photo from the tablet shows chat in Greek but it is clear that the Mathematics are too large and  not positioned properly. Some above the base line and some below.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -484,6 +484,7 @@ class MessageItemFactory @Inject constructor(
         }
                 .useBigFont(linkifiedBody.length <= MAX_NUMBER_OF_EMOJI_FOR_BIG_FONT * 2 && containsOnlyEmojis(linkifiedBody.toString()))
                 .canUseTextFuture(canUseTextFuture)
+                .markwonPlugins(htmlRenderer.get().plugins)
```
bmarty, line 487: Add this behind the lab flag too?

vector/src/main/java/im/vector/app/features/html/EventHtmlRenderer.kt
```diff
@@ -20,24 +20,51 @@ import android.content.Context
...
+                    }
+                })
+                .usePlugin(MarkwonInlineParserPlugin.create())
+                .usePlugin(JLatexMathPlugin.create(44F) { builder ->
+                    builder.inlinesEnabled(true)
+                    builder.theme().inlinePadding(JLatexMathTheme.Padding.symmetric(24, 8))
+                })
+    } else {
+        builder
+    }.textSetter(PrecomputedFutureTextSetterCompat.create()).build()
```
bmarty, line 65: What is the goal of `textSetter(PrecomputedFutureTextSetterCompat.create())`?

bmarty (changes requested):
Thanks a lot for the update.

- [ ] I've enabled GitHub action (CI), please fix mandatory action(s) in error.
- [ ] Can you add a changelog file please? See https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#changelog for more details

vector/src/main/java/im/vector/app/features/html/EventHtmlRenderer.kt
NickHu: See https://github.com/noties/Markwon/issues/218. Element uses AppCompatTextView everywhere anyway. I thought it might help with the clipping issue, but it didn't. Doesn't hurt to use this anyway.

NickHu:
> Dear @NickHu , an android phone converts the LaTeX code to math when the message is written say from element-desktop. The other way around does not work. If we write $x^2$ on the phone, the desktop app AND the phone app does not convert this to math. It remains verbatim. Can you please verify?

Please confirm you have both
1. enabled Markdown input in preferences
2. enabled the Maths Labs feature and restarted the app (clear it from recent apps and launch it again)

NickHu:
> Dear @NickHu , an android phone converts the LaTeX code to math when the message is written say from element-desktop. The other way around does not work. If we write $x^2$ on the phone, the desktop app AND the phone app does not convert this to math. It remains verbatim. Can you please verify?

Please confirm you have both
1. enabled Markdown input in preferences
2. enabled the Maths Labs feature and restarted the app (clear it from recent apps and launch it again)

antonis-tsolomitis:
Ah, yes I missed the Markdown setting. So now it works for inline math on a phone. There are two problems. One is that displaymath does not work (or I miss one more setting (?)):



Remaining issues include proper vertical spacing (except clipping) as seen below:

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
NickHu: To be honest, I think this should not be gated behind the labs flag. I think you are supposed to use markwon this way. It just so happens that with the existing configuration, it only loads the html rendering plugin and that doesn't have any pre/post hooks so it doesn't matter if you omit this line. In any case, it shouldn't affect performance.

NickHu:
I rebased onto `develop`, but I had to downgrade the markwon version relative to there (it was updated just over a week ago) because of https://github.com/noties/Markwon/issues/264 (closed but never fixed). It's still a version of markwon newer than what the current stable version uses.

Esokrates:
@NickHu Tested this build here: https://buildkite.com/matrix-dot-org/element-android/builds/3581#e9486708-9309-47d5-9810-665fe4d8ab68

Displayed math is not rendered at all here, no matter if its `$$` or $\[$.

bmarty:
I will handle the remaining work on my branch.

PR #2134, opened 2020-09-18
Fix / Verification in DM not working

BillCarsonFr:
CheckNumberType crashing on timestamp number making workdata parsing failing, and then broke the sending queue on verification requests


bmarty (approved):
LGTM, thanks!

PR #2135, opened 2020-09-18
Create parent class for all MatrixWorker

bmarty:
Trying to normalize the way we use worker, and also try to catch all possible Exception to avoid breaking worker chains.

Parsing Exception is dealt in @ganfra 's PR #2129 

This is really dirty, I cannot to parsing and buildErrorResult() in the parent class, so many duplicated code... Not very proud of it.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/content/UploadContentWorker.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/content/UploadContentWorker.kt
@@ -19,7 +19,6 @@ package org.matrix.android.sdk.internal.session.content
 
 import android.content.Context
 import android.graphics.BitmapFactory
-import androidx.work.CoroutineWorker
 import androidx.work.WorkerParameters
 import com.squareup.moshi.JsonClass
 import org.matrix.android.sdk.api.extensions.tryOrNull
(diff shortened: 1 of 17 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/content/UploadContentWorker.kt
```diff
@@ -217,14 +213,7 @@ internal class UploadContentWorker(val context: Context, params: WorkerParameter
             }
         } catch (e: Exception) {
             Timber.e(e, "## FileService: ERROR")
-            notifyTracker(params) { contentUploadStateTracker.setFailure(it, e) }
```
bmarty, line 221: This line has disappeared, is it OK?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionSafeCoroutineWorker.kt
```diff
@@ -42,43 +43,44 @@ internal abstract class SessionSafeCoroutineWorker<PARAM : SessionWorkerParams>(
...
         return try {
-            doSafeWork(sessionComponent, params)
+            val sessionComponent = getSessionComponent(params.sessionId)
+                    ?: return buildErrorResult(params, "No session")
+
+            // Make sure to inject before handling error as you may need some dependencies to process them.
+            injectWith(sessionComponent)
+            if (params.lastFailureMessage != null) {
+                // Forward error to the next workers
+                return doOnError(params)
```
bmarty, line 58: maybe remove `return` statement here and use a `else` block

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionSafeCoroutineWorker.kt
```diff
@@ -87,12 +89,12 @@ internal abstract class SessionSafeCoroutineWorker<PARAM : SessionWorkerParams>(
     open fun doOnError(params: PARAM): Result {
         // Forward the error
         return Result.success(inputData)
-                .also { Timber.e("Work cancelled due to input error from parent") }
+                .also { Timber.e("Work cancelled due to input error from parent: $params") }
```
bmarty, line 92: it may log private stuff, no?

ganfra: Yup sorry was for debugging

BillCarsonFr (approved):

Issue #2094, opened 2020-09-11
[BUG] Auto scrolling to the latest message


bahur142:
Element Android did not scroll automatically to the latest message when you are the sender.

1. Go to Settings --> Preferences
2. Disable "Show removed messages"
3. Send some messages to someone.
4. Screen did not move to the latest message. Instead the button with the arrow appears to go to the latest message.
5. Enable "Show removed messages".
6. Everything is fine again, when you send messages.

App versions: 1.0.6 & 1.0.7
Homeserver: my own (v.1.19.1)
Devices: Android 6.0.1; Android 9.0

scottwallacesh:
I also see this behaviour.

bmarty:
Fixed by #2144

PR #2144, opened 2020-09-21
Feature/fix filtering redacted

ganfra:
There was an issue with the filtering condition of redacted events in timeline. It was causing this issue => https://github.com/vector-im/element-android/issues/2094

Also, filtering was not working for changes on already built events.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
@@ -333,11 +333,19 @@ internal class DefaultTimeline(
 
 // Private methods *****************************************************************************
 
-    private fun rebuildEvent(eventId: String, builder: (TimelineEvent) -> TimelineEvent): Boolean {
+    private fun rebuildEvent(eventId: String, builder: (TimelineEvent) -> TimelineEvent?): Boolean {
         return builtEventsIdMap[eventId]?.let { builtIndex ->
             // Update the relation of existing event
             builtEvents[builtIndex]?.let { te ->
-                builtEvents[builtIndex] = builder(te)
+                val rebuiltEvent = builder(te)
+                // If rebuilt event is filtered its returned as null and should be removed.
+                if (rebuiltEvent == null) {
+                    builtEventsIdMap.remove(eventId)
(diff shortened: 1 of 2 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
```diff
@@ -333,11 +333,19 @@ internal class DefaultTimeline(
...
             builtEvents[builtIndex]?.let { te ->
-                builtEvents[builtIndex] = builder(te)
+                val rebuiltEvent = builder(te)
+                // If rebuilt event is filtered its returned as null and should be removed.
+                if (rebuiltEvent == null) {
+                    builtEventsIdMap.remove(eventId)
+                    builtEventsIdMap.entries.filter { it.value > builtIndex }.forEach { it.setValue(it.value - 1) }
+                    builtEvents.removeAt(builtIndex)
+                } else {
+                    builtEvents[builtIndex] = builder(te)
```
bmarty, line 347: Use `rebuiltEvent`?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
```diff
@@ -775,8 +784,7 @@ internal class DefaultTimeline(
             }
             if (!filterEdits) return@filter false
 
-            val filterRedacted = !settings.filters.filterRedacted || it.root.isRedacted()
-
+            val filterRedacted = !settings.filters.filterRedacted || it.root.isRedacted().not()
```
bmarty, line 787: For clarity and coherency (I had to read it twice) we should write:

```
val filterRedacted = settings.filters.filterRedacted && it.root.isRedacted()

return !filterRedacted
```

ganfra: Yup totally, my brain was off :O

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
ganfra: Yes good catch

Issue #1972, opened 2020-08-20
Can't login to homeserver - "homeserver outdated".


vladimir-ipatov:
Hello! I Recently updated android element client (1.0.4).
I have account on my phone logged in before, it works good.

But after logout I can't login again.
Click "Start"
Click "Another"
typing my homeserver address and get:

"Homeserver outdated
This homeserver using too old version for connection
Ask administator to update homeserver"

And can't continue.

server version: 0.99.2-5~bpo9+1 from debian stretch backports.

Why I can't use new client with my server despite it technically works?
What the reasons for this weird step?
Why it is not documented?

It have me to remember 2008 when ICQ was changed protocols every 2-3 months and I could not able to use it :)

violoncelloCH:
well there are security reasons to update your homeserver...
and there are other things that have been changed/added to the specification which are being used by the current element-android

bmarty:
Just tested with your homeserver and in your case [this method](https://github.com/vector-im/element-android/blob/894f5b9ed3b2b7caafcd3af3aa75c9da97a867a9/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/Versions.kt#L67) will return false, so login and account creation is not possible because Element Android (sadly) only support the latest API.
We have made the choice when writing RiotX (which is now Element Android) to not waste time supporting deprecated API. Sorry for the inconvenience...

vladimir-ipatov:
> Just tested with your homeserver and in your case [this method](https://github.com/vector-im/element-android/blob/894f5b9ed3b2b7caafcd3af3aa75c9da97a867a9/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/Versions.kt#L67) will return false, so login and account creation is not possible because Element Android (sadly) only support the latest API.
> We have made the choice when writing RiotX (which is now Element Android) to not waste time supporting deprecated API. Sorry for the inconvenience...


So, as I cant see in this test:

```
/**
 * Indicate if the `id_server` parameter is required when registering with an 3pid,
 * adding a 3pid or resetting password.
 */
private fun Versions.doesServerRequireIdentityServerParam(): Boolean {
    if (getMaxVersion() >= HomeServerVersion.r0_6_0) return false
    return unstableFeatures?.get(FEATURE_REQUIRE_IDENTITY_SERVER) ?: true
}

```

This test related with 3pid registering and so on.
So, may be normal login (without 3pid) will work?

[...]

PR #2146, opened 2020-09-21
Allow using an outdated homeserver, at user's risk (#1972)

bmarty:
Just warn the user by a non blocking popup

Tested OK using outdated homeserver (from #1972): `m.ruslo.pro`.


ganfra (approved):
LGTM, thanks

PR #2154, opened 2020-09-22
Use cache for user color

bmarty:
Commit restored from an old Ganfra's branch, and updated by me


ganfra:
LGTM, thanks for getting it in track

Issue #2123, opened 2020-09-17
Show timestamp for every messages


bmarty:
Add a setting (restore the existing one from Riot-Android) to let user chooses to see the timestamp for every message in the timeline.

Riot-Android:

<img width="433" alt="image" src="https://user-images.githubusercontent.com/3940906/93504760-36ad1480-f91a-11ea-8c12-3ab232ed7ab9.png">

Element Web:

<img width="597" alt="image" src="https://user-images.githubusercontent.com/3940906/93504801-49274e00-f91a-11ea-942f-c38aa2ce1c6e.png">

bmarty:
Done un #2155

PR #2155, opened 2020-09-22
Add a setting to show timestamp for all messages (#2123)

bmarty:
Implements #2123


ganfra (approved):
Code is ok, but UI is ugly :/

PR #2158, opened 2020-09-23
Feature/fix draft

### LINKED ISSUE #683: Sharing to RiotX delete the current draft
Share some text to a room with a draft.
The shared text is sent to the room (it should maybe just fill in the composer ?)
Observe that the draft is lost.

### LINKED ISSUE #952: draft gets deleted when you start typing directly after sending a reply
Steps to reproduce:

1. send a messageA
2. swipe on messageA to go into reply mode
3. (the reply preview appears)
4. type some text and tap on submit to send messageB
5. (the input box is now cleared but the reply preview above the input box is still displayed)
6. type some text to compose messageC
7. (messageB gets local echoed, the reply preview is removed and messageC is deleted from the input box)

I think it makes sense to completely clear the composer including the reply preview at step 5, so you can start typing a new message right away.

ganfra:
Simplify PR management to allow one. We are not listening to the db anymore to avoid waiting for transaction. 

Fixes #952
Fixes #683

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/draft/DraftRepository.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/draft/DraftRepository.kt
@@ -75,11 +75,7 @@ internal class DraftRepository @Inject constructor(@SessionDatabase private val 
     }
 
     private fun deleteDraftFromDb(realm: Realm, roomId: String) {
-        UserDraftsEntity.where(realm, roomId).findFirst()?.let { userDraftsEntity ->
-            if (userDraftsEntity.userDrafts.isNotEmpty()) {
-                userDraftsEntity.userDrafts.removeAt(userDraftsEntity.userDrafts.size - 1)
-            }
-        }
+        UserDraftsEntity.where(realm, roomId).findFirst()?.userDrafts?.clear()
     }
 
     private fun saveDraftInDb(realm: Realm, draft: UserDraft, roomId: String) {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/draft/DraftRepository.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/draft/DraftRepository.kt
@@ -26,6 +26,7 @@ import org.matrix.android.sdk.BuildConfig
 import org.matrix.android.sdk.api.session.room.send.UserDraft
 import org.matrix.android.sdk.api.util.Optional
 import org.matrix.android.sdk.api.util.toOptional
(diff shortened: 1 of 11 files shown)


DoM1niC:
I have a big bug since your nice PR, if you write some Text and switch between Apps or Home, so the written Text is lost.... The Draft thing works after I write some and go back internal the App.

ganfra:
Thanks for the remark @DoM1niC, found the cause for this issue

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/draft/DraftRepository.kt
```diff
@@ -68,62 +92,15 @@ class DraftRepository @Inject constructor(@SessionDatabase private val monarchy:
...
-                        // Update the top draft
-                        Timber.d("Draft: update the top draft ${privacySafe(draft)}")
-                        topDraft.draftMode = newDraft.draftMode
-                        topDraft.content = newDraft.content
-                        topDraft.linkedEventId = newDraft.linkedEventId
-                    }
-                }
+                Timber.d("Draft: create a new draft ${privacySafe(draft)}")
+                userDraftEntity.userDrafts.clear()
+                userDraftEntity.userDrafts.add(newDraft)
```
bmarty, line 99: 😢

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt
```diff
@@ -226,52 +226,52 @@ class RoomDetailViewModel @AssistedInject constructor(
...
+            is RoomDetailAction.TapOnFailedToDecrypt -> handleTapOnFailedToDecrypt(action)
+            is RoomDetailAction.SelectStickerAttachment -> handleSelectStickerAttachment()
+            is RoomDetailAction.OpenIntegrationManager -> handleOpenIntegrationManager()
+            is RoomDetailAction.StartCall -> handleStartCall(action)
+            is RoomDetailAction.EndCall -> handleEndCall()
+            is RoomDetailAction.ManageIntegrations -> handleManageIntegrations()
+            is RoomDetailAction.AddJitsiWidget -> handleAddJitsiConference(action)
+            is RoomDetailAction.RemoveWidget -> handleDeleteWidget(action.widgetId)
+            is RoomDetailAction.EnsureNativeWidgetAllowed -> handleCheckWidgetAllowed(action)
+            is RoomDetailAction.CancelSend -> handleCancel(action)
```
bmarty, line 274: AS sucks

bmarty (approved):
LGTM

DoM1niC:
If I share text over the Share Feature and send it, so the Input field will not clean up, I know this is not totally a Draft thing but could be recommended to fix this similar to Draft.

PR #2159, opened 2020-09-23
Feature/bma pin settings

### LINKED ISSUE #1985: PIN Code Improvements

I have noticed that the PIN code system has some frustrating flaws. 

- First of all, it only allows exactly 4 characters, which is just begging for years of births to be used. I think the length should be virtually unlimited for greater security.
- Biometrics seem to be enabled by default if they are available which may not be desirable. For example, if someone has access to your phone via biometrics but you want to keep Element private even from them. I believe there should be a separate option to enable each type of biometric auth.
- The time before Element locks is not configurable. For example, to lock Element instantly, I have to clear it from the open apps menu, otherwise it locks automatically after an unknown (to me) amount of time. I think that should also be configurable with the option to lock as soon as the user leaves the app, after the device locks, or after a certain amount of not using the app.
- In order to include all the above options, I believe PIN should have it's own submenu within Settings > Security & Privacy

See above

None

**Additional context**

None

bmarty:
Fix #1985 

Add a new dedicated preference screen to configure PIN code.

image

image

- When enabled, PIN code is requested to enter this screen, and not anymore to disable PIN protection.
- New setting to disable usage of Biometrics (enabled by default)
- New setting to disable grace period (enable by default)
- New setting to hide notification details (room avatar and name, sender avatar and name, message content, invitations). Instead a short summary is displayed:

image

image

image


vector/src/main/java/im/vector/app/features/pin/PinLocker.kt
```diff
@@ -22,22 +22,29 @@ import androidx.lifecycle.LifecycleObserver
...
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 import timber.log.Timber
 import javax.inject.Inject
 import javax.inject.Singleton
 
+// 2 minutes, when enabled
 private const val PERIOD_OF_GRACE_IN_MS = 2 * 60 * 1000L
 
+// 1 seconds, to avoid asking for PIN code when switching between Activities
```
ganfra, line 35: I don't think this is necessary, we don't lock as long as the app stays in foreground

ganfra (commented):
One remark otherwise LGTM.
Also, could you add necessary metadata to the manifest so there is only one PinActivity in the stack?

vector/src/main/java/im/vector/app/features/pin/PinLocker.kt
bmarty: Ok, you are right. I've removed that

bmarty:
Added singleTop, not sure it will fix known issue about double Pin code asking, but let's try

vector/src/main/res/values/strings.xml
```diff
@@ -2566,8 +2572,20 @@
     <string name="auth_pin_reset_title">Reset pin</string>
     <string name="auth_pin_new_pin_action">New pin</string>
     <string name="auth_pin_reset_content">To reset your PIN, you\'ll need to re-login and create a new one.</string>
+    <string name="settings_security_application_protection_title">Protect access to the application.</string>
```
niquewoodhouse, line 2575: **Suggestion 1 (preferred)**
Protect access ~to the application.~

_I personally don't think you need to say the application, given the context of being in the app. The more text, the more a user has to read to understand and I haven't see other applications using this phrasing._

**Suggestion 2**
Protect access to Element ~to the application.~

**Suggestion 3**
Protect access to the application ~remove full stop at the end.~
_There's no punctuation in other titles_

vector/src/main/res/values/strings.xml
```diff
@@ -2566,8 +2572,20 @@
...
+    <string name="settings_security_application_protection_summary">Protect access to the application using pin code and biometrics.</string>
+    <string name="settings_security_application_protection_screen_title">Configure protection</string>
     <string name="settings_security_pin_code_title">Enable PIN</string>
     <string name="settings_security_pin_code_summary">If you want to reset your PIN, tap Forgot PIN to logout and reset.</string>
+    <string name="settings_security_pin_code_use_biometrics_title">Use biometrics</string>
```
niquewoodhouse, line 2580: **Suggestion 1**

**Enable** ~use~ biometrics.  

You have "enable PIN" just above, I don't think there is a need to use a different phrasing when, to the user, I think they're the same.

vector/src/main/res/values/strings.xml
bmarty: I take suggestion 1, thanks

vector/src/main/res/values/strings.xml
```diff
@@ -2566,8 +2572,20 @@
...
+    <string name="settings_security_pin_code_use_biometrics_summary_on">Depending on your device, you will be able to unlock using fingerprints, face recognition, iris recognition, etc.</string>
+    <string name="settings_security_pin_code_use_biometrics_summary_off">PIN code is the only way to unlock the application.</string>
+    <string name="settings_security_pin_code_notifications_title">Display complete notifications</string>
+    <string name="settings_security_pin_code_notifications_summary_on">Show full notification details, including rooms details and message contents.</string>
+    <string name="settings_security_pin_code_notifications_summary_off">Only display number of unread messages in a simple notification.</string>
+    <string name="settings_security_pin_code_grace_period_title">Enable grace period</string>
+    <string name="settings_security_pin_code_grace_period_summary_on">PIN code will be asked after 2 minutes not using the application.</string>
```
niquewoodhouse, line 2587: **Suggestion 1**
PIN code will be asked **for** after 2 minutes **of** not using **Element.** ~the application.~

**Suggestion 2**
PIN code will be required to unlock after 2 minutes of absence

vector/src/main/res/values/strings.xml
bmarty: Done

vector/src/main/res/values/strings.xml
bmarty: Done (1)

vector/src/main/res/values/strings.xml
```diff
@@ -2566,8 +2572,20 @@
     <string name="auth_pin_reset_title">Reset pin</string>
     <string name="auth_pin_new_pin_action">New pin</string>
     <string name="auth_pin_reset_content">To reset your PIN, you\'ll need to re-login and create a new one.</string>
+    <string name="settings_security_application_protection_title">Protect access to the application.</string>
+    <string name="settings_security_application_protection_summary">Protect access to the application using pin code and biometrics.</string>
+    <string name="settings_security_application_protection_screen_title">Configure protection</string>
     <string name="settings_security_pin_code_title">Enable PIN</string>
     <string name="settings_security_pin_code_summary">If you want to reset your PIN, tap Forgot PIN to logout and reset.</string>
+    <string name="settings_security_pin_code_use_biometrics_title">Use biometrics</string>
+    <string name="settings_security_pin_code_use_biometrics_summary_on">Depending on your device, you will be able to unlock using fingerprints, face recognition, iris recognition, etc.</string>
```
niquewoodhouse, line 2581: **Suggestion 1**
Enable device specific biometrics. ~Depending on your device, you will be able to unlock using fingerprints, face recognition, iris recognition, etc.~
_Don't think its helpful to list the ways biometrics might work_

**Suggestion 2**
Enable device specific biometrics, like fingerprints and face recognition ~Depending on your device, you will be able to unlock using fingerprints, face recognition, iris recognition, etc.~
_Another way to say some biometrics without making a long list_

vector/src/main/res/values/strings.xml
```diff
@@ -2566,8 +2572,20 @@
     <string name="auth_pin_reset_title">Reset pin</string>
     <string name="auth_pin_new_pin_action">New pin</string>
     <string name="auth_pin_reset_content">To reset your PIN, you\'ll need to re-login and create a new one.</string>
+    <string name="settings_security_application_protection_title">Protect access to the application.</string>
+    <string name="settings_security_application_protection_summary">Protect access to the application using pin code and biometrics.</string>
```
niquewoodhouse, line 2576: **Suggestion**
Protect access using PIN and biometrics.

vector/src/main/res/values/strings.xml
```diff
@@ -2566,8 +2572,20 @@
...
+    <string name="settings_security_application_protection_screen_title">Configure protection</string>
     <string name="settings_security_pin_code_title">Enable PIN</string>
     <string name="settings_security_pin_code_summary">If you want to reset your PIN, tap Forgot PIN to logout and reset.</string>
+    <string name="settings_security_pin_code_use_biometrics_title">Use biometrics</string>
+    <string name="settings_security_pin_code_use_biometrics_summary_on">Depending on your device, you will be able to unlock using fingerprints, face recognition, iris recognition, etc.</string>
+    <string name="settings_security_pin_code_use_biometrics_summary_off">PIN code is the only way to unlock the application.</string>
+    <string name="settings_security_pin_code_notifications_title">Display complete notifications</string>
```
niquewoodhouse, line 2583: **Suggestion 1**
Show notifications content on lock screen
_More specific on where notifications will appear, and what we mean by "complete"_

vector/src/main/res/values/strings.xml
```diff
@@ -2566,8 +2572,20 @@
...
+    <string name="settings_security_pin_code_notifications_summary_on">Show full notification details, including rooms details and message contents.</string>
```
niquewoodhouse, line 2584: Show details like room names and message content.

vector/src/main/res/values/strings.xml
```diff
@@ -2566,8 +2572,20 @@
...
+    <string name="settings_security_pin_code_notifications_summary_off">Only display number of unread messages in a simple notification.</string>
+    <string name="settings_security_pin_code_grace_period_title">Enable grace period</string>
```
niquewoodhouse, line 2586: I don't know if this is a well known phrase but I'm not super confident in the following suggestions:

Require PIN after 2 minutes

vector/src/main/res/values/strings.xml
bmarty: I take suggestion 2 :)

vector/src/main/res/values/strings.xml
bmarty: Done

vector/src/main/res/values/strings.xml
bmarty: @niquewoodhouse this is not only on lock screen. _Complete_ means with full details.

vector/src/main/res/values/strings.xml
bmarty: Done

vector/src/main/res/values/strings.xml
bmarty: I replace "Enable grace period" with "Require PIN after 2 minutes"

vector/src/main/res/values/strings.xml
niquewoodhouse: @bmarty thanks for explaining :) My only suggestion left is "Show content in notifications" or "Show notification contents" because I think its clearer to a user that this means the content of a notification.

bmarty: Ok, that's far better from my wording, thanks!

niquewoodhouse (approved):
Looks good to me

PR #2165, opened 2020-09-24
Feature/timeline scroll opti

ganfra:
This is a small optimisation of scrolling experience in timeline

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ Features ✨:
 
 Improvements 🙌:
  - PIN code: request PIN code if phone has been locked
+ - Small optimisation of scrolling experience in timeline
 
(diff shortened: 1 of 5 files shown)


DoM1niC:
I have a big bug since your nice PR, if you write some Text and switch between Apps or Home, so the written Text is lost.... The Draft thing works after I write some and go back into the App.

DoM1niC:
I have no idea, there is a bug from this Merge or Draft one, so this postet content is every time on the bottom of the timeline and couldn't find a way to remove it, very strange.... 

But it could be a other unknown Bug, so I hope you could tag it if I wrong with my assumption....

bmarty (approved):
The timeline backward scroll is way more efficient now, thanks!

PR #2166, opened 2020-09-25
Fix Splash screen layout, especially on small screens

bmarty:
Rely only on ConstraintLayout (no more scroll view) for the first displayed screen of the app.

## On small screen (Nexus One) -> BUG FIX, no need to scroll anymore

Before Screenshot 2020-09-25 at 06 29 03 After image

## On big screen (S9+) -> Nearly no impact (sorry screenshot not in the same language)

Before Screenshot 2020-09-25 at 07 28 29 After image

### Also rendering OK in landscape, for tablet it could be better, but it's not worth that we we have now

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ Improvements 🙌:
  -
 
 Bugfix 🐛:
- -
+ - Fix Splash layout on small screens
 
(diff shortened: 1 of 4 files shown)


ganfra (approved):
LGTM

PR #2181, opened 2020-09-29
PIN code: request PIN code if phone has been locked

bmarty:
PIN code: request PIN code if phone has been locked. No grace period in this case.


BillCarsonFr (approved):
LGTM

PR #2190, opened 2020-09-30
Feature/utd pagination

ganfra:
This PR blocks pagination before invite event in encrypted rooms.
You can still re-enable this in developer settings.


bmarty (approved):
LGTM, thanks!

PR #2196, opened 2020-10-01
Add SearchView to filter room members

### LINKED ISSUE #2184: Filter room member by name
In the room member list screen, add a filter to filter by name. Can be useful in room with lots of users. Can be useful also for moderator

onurays:
Fixes #2184 
 
filter_room_members


bmarty:
Can we use the same hint than on Element Web?

image

bmarty (approved):
Approved after my rework

PR #2199, opened 2020-10-01
Feature/performance tracking

ganfra:
This PR introduces a BaseProfiler and first usage with SendPerformanceProfiler.
It also enables logs for TimelineEventController (only in debug).


ganfra:
What should we do we that?

bmarty:
I close this stale PR, it has to be reworked anyway

PR #2203, opened 2020-10-02
Feature/bma/pusher data

bmarty:
This PR adds a diagnostic test (for Gplay version) to ask the Push Gateway to send a push back to the device. This uses an unauthenticated API, so no auth token is send to the Push Gateway instance.

New test:

image

And the notification:

image

This PR also contains a script to directly ask FCM to send a push to the provided FCM token using the provided server API key (not publicly available).

Next step should be to validate the link between the homeserver and the Push Gateway, but for this we need a new CS API.


BillCarsonFr (commented):
Is it possible to only make the test as successful if the user tap on the notification?

bmarty:
> Is it possible to only make the test as successful if the user tap on the notification?

Done. I added a `WAITING_FOR_USER` state

bmarty:
Finally the new test to send PUSH just wait for the push to be success.
I added another test to display a notification and check that the user can click it.

BillCarsonFr (approved):

PR #2207, opened 2020-10-05
Feature/bma/open settings

### LINKED ISSUE #2171: [Discoverability] Settings are hard to find
After several support sessions with users, we can safely say that the application settings are hard to find.

- Some user do not find the left drawer (as the usual hamburger icon is replaced by the user icon in element)
- After that the settings icon is a small gear icon
- /!\ Some users even think that there is no logout option ( as its deep down left drawer > gear > General > scroll to bottom), found in review in playstore

image
image

bmarty:
Fix #2171 

New left drawer:

image

New item menu in room list to access the main settings:

image

New items menu in room to access room settings and to invite people:

image

Also fixes a bug on the Signout bottomsheet, where the sign out button was displayed with the sign out anyway button.


BillCarsonFr (approved):
LGTM, thx for fixing exit button

PR #2212, opened 2020-10-05
Add "jump to read receipt" and "mention" actions from room member detail screen

bmarty:
Using a trick to avoid messing up with ActivityResult: we store the pending action in a singleton.

Also I created https://github.com/vector-im/element-web/issues/15380

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/read/ReadService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/read/ReadService.kt
@@ -65,11 +65,11 @@ interface ReadService {
 
     /**
      * Get the eventId where the read receipt for the provided user is
-     * @param otherUserId the userId param to look for
+     * @param userId the id of the user to look for
      *
      * @return the eventId where the read receipt for the provided user is attached, or null if not found
      */
-    fun getUserReadReceipt(otherUserId: String): String?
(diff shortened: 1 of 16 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/read/ReadService.kt
```diff
@@ -63,6 +63,14 @@ interface ReadService {
      */
     fun getMyReadReceiptLive(): LiveData<Optional<String>>
 
+    /**
+     * Get the eventId where the read receipt for the provided user is
+     * @param otherUserId the userId param to look for
+     *
+     * @return the eventId where the read receipt for the provided user is attached, or null if not found
+     */
+    fun getUserReadReceipt(otherUserId: String): String?
```
onurays, line 72: Why are we using "otherUserId" term instead of "userId"?

onurays:
LGTM! Just 1 remark and 2 conflicts.

onurays (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/read/ReadService.kt
bmarty: Because in the implementation we already have a `userId`. But you are right, I will change to `userId`, it's better

PR #2215, opened 2020-10-06
If there is a pending email validation continue on this step

### LINKED ISSUE #2021: Can't confirm email due to killing by Android
**Describe the bug**

matrix.org site requires confirmed email to create an account. When I try to sign up via Element client and then go to an email client and to a browser, Element is being closed and after it offers to start over again, resetting current progress.

**To Reproduce**

1. Launch Element with clean storage.
2. Press "Get started".
3. Select [matrix.org] server.
4. Press "Sign Up".
5. Choose username and password.
6. Pass Google's captcha.
7. Input your email.
8. Go to K-9 Mail client installed on the same device.
9. Open the received confirmation link in Fennec (browser based on Firefox available in F-Droid).
10. Go back to Element (by then, Android has unloaded it).
11. It's starting from the beginning (to simulate the case, you can limit background processes).

**Expected behavior**

Please save current state of the registration progress in a persistent memory, so that a user can continue, despite the system terminates Element client.

**Smartphone:**

 - Device: Micromax Q402+
 - OS: Android 7.0
 - Main memory: 1.0 GB

**Additional context:**

 - App version and store: 1.0.5, F-Droid
 - Homeserver: matrix.org

onurays:
Fixes #2021


vector/src/main/java/im/vector/app/features/login/LoginViewModel.kt
```diff
@@ -417,6 +417,13 @@ class LoginViewModel @AssistedInject constructor(
 
     private fun handleInitWith(action: LoginAction.InitWith) {
         loginConfig = action.loginConfig
+
+        // If there is a pending email validation continue on this step
+        currentThreePid?.let {
```
bmarty, line 422: This should be in the init{} of the ViewModel

onurays: I first tried that but since init is working before the Activity is created, the Activity cannot observe events.

bmarty (approved):
LGTM, thanks!

vector/src/main/java/im/vector/app/features/login/LoginViewModel.kt
```diff
@@ -672,6 +684,7 @@ class LoginViewModel @AssistedInject constructor(
 
     private fun onSessionCreated(session: Session) {
         activeSessionHolder.setActiveSession(session)
+        authenticationService.reset()
```
bmarty, line 687: 👍

PR #2244, opened 2020-10-11
Direct share support

### LINKED ISSUE #2029: Direct share support
Please add direct share support.

https://developer.android.com/about/versions/marshmallow/android-6.0#direct-share

Screenshot_20200831-132802_Android_System

ginnyTheCat:
Fixes #2029

It should also work on Android Lollipop according to [this](https://codelabs.developers.google.com/codelabs/android-direct-share/index.html#7). I couldn't test it, however.

### Screenshot

Signed-off-by: Constantin Wartenburger <ginnythecat@lelux.net>


vector/src/main/java/im/vector/app/features/home/ShortcutCreator.kt
```diff
@@ -65,6 +66,10 @@ class ShortcutCreator @Inject constructor(
                 .setShortLabel(roomSummary.displayName)
                 .setIcon(bitmap?.toProfileImageIcon())
                 .setIntent(intent)
+
+                // Make it show up in the direct share menu
+                .setCategories(setOf(directShareCategory))
```
bmarty, line 71: This is maybe not the ideal place to add this code. `create()` is called the user create a shortcut, or for the first 4 favorite rooms.

I think other apps create direct share shortcut once something has been shared, for a quicker future sharing to the same target.

vector/src/main/java/im/vector/app/features/share/IncomingShareFragment.kt
```diff
@@ -71,13 +72,28 @@ class IncomingShareFragment @Inject constructor(
...
         val isShareManaged = when (intent?.action) {
-            Intent.ACTION_SEND -> {
+            Intent.ACTION_SEND          -> {
                 var isShareManaged = attachmentsHelper.handleShareIntent(requireContext(), intent)
                 if (!isShareManaged) {
                     isShareManaged = handleTextShare(intent)
                 }
+
+                // Direct share
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P && intent.hasExtra(Intent.EXTRA_SHORTCUT_ID)) {
```
bmarty, line 92: is it mandatory to check SDK_INT here? In old OSs, `intent.hasExtra(Intent.EXTRA_SHORTCUT_ID)` should always return `false`, no?

bmarty (commented):
Thanks a lot for this PR, I have two small remarks.

vector/src/main/java/im/vector/app/features/share/IncomingShareFragment.kt
ginnyTheCat: It is like this in the [example](https://github.com/googlecodelabs/android-direct-share/blob/f5be36dad6676e4fcb7f2a8f7a285466c7055c6a/direct-share-done/app/src/main/java/com/example/android/directshare/SendMessageActivity.kt#L113). It shouldn't be necessary, however.

vector/src/main/java/im/vector/app/features/home/ShortcutCreator.kt
bmarty: Ok, maybe I have misunderstood this. I think it is OK actuallly

ginnyTheCat: Such a ranking is still done on the android side (of course only with the shortcuts published to them).

bmarty: Yes, my remark was just about what will be the rooms exposed here

vector/src/main/java/im/vector/app/features/home/ShortcutCreator.kt
```diff
@@ -33,6 +33,7 @@ import javax.inject.Inject
 private val useAdaptiveIcon = Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
 private const val adaptiveIconSizeDp = 108
 private const val adaptiveIconOuterSidesDp = 18
+private const val directShareCategory = "im.vector.app.SHORTCUT_SHARE"
```
bmarty, line 36: It should be better to use `BuildConfig.APPLICATION_ID` as a prefix here, but will it be possible in the file `shortcuts.xml`?

ginnyTheCat: Is there a way to get the `applicationId` with the suffix in the `build.grade` or the `BuildConfig.APPLICATION_ID` without the suffix? With suffix, I mean that `.debug` at the end.

bmarty: Thinking again about it, I think the only way is to have a file `shortcuts.xml` in the release folder and one in the debug folder. In the code, we can use `BuildConfig.APPLICATION_ID`

bmarty (approved):
LGTM, thanks for the update!

ginnyTheCat:
Could you add the hacktoberfest topic to the repo as you did at element-web? That would be nice.

bmarty:
What do you mean?

ginnyTheCat:
The other matrix repos like synapse or element-web have the hacktoberfest topic on their repo. That makes PRs count to the Hacktoberfest (a competition) goal. It would be nice to have that here too.

bmarty:
I've added the topic:

image

ginnyTheCat:
Thank you

PR #2248, opened 2020-10-12
Fix / reset cross signing not well supported

BillCarsonFr:
Properly detect when cross signing keys are reseted.
-> currently on reset, we were not cleaning the known private keys (as they are now irrelevant)
-> On my cross signing keys change, we also clear now any trust of my sessions (including local)
-> When cross signing keys has changed, we detect it and add a new alert to ask to re-verify


vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -67,11 +68,39 @@ class HomeActivityViewModel @AssistedInject constructor(
     }
 
     private var checkBootstrap = false
+    private var onceTrusted = false
```
bmarty, line 71: Does it work if I kill the app when this occurs? I do not think the `true -> false` scenario will occur again in this case.

bmarty (approved):
LGTM, one inlined question

PR #2257, opened 2020-10-14
Ability to share text

ginnyTheCat:
Now able to share text (before that just images, videos, etc.).
Switched to the sharing menu used by most other apps.

### New sharing menu:

Signed-off-by: Constantin Wartenburger <ginnythecat@lelux.net>

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
@@ -106,6 +106,7 @@ import im.vector.app.core.utils.onPermissionResultVideoIpCall
 import im.vector.app.core.utils.openUrlInExternalBrowser
 import im.vector.app.core.utils.saveMedia
 import im.vector.app.core.utils.shareMedia
+import im.vector.app.core.utils.shareText
 import im.vector.app.core.utils.toast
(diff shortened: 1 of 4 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -1556,21 +1557,25 @@ class RoomDetailFragment @Inject constructor(
...
-                fileName = action.messageContent.body,
-                mimeType = action.messageContent.mimeType,
-                url = action.messageContent.getFileUrl(),
-                elementToDecrypt = action.messageContent.encryptedFileInfo?.toElementToDecrypt(),
-                callback = object : MatrixCallback<File> {
-                    override fun onSuccess(data: File) {
-                        if (isAdded) {
-                            shareMedia(requireContext(), data, getMimeTypeFromUri(requireContext(), data.toUri()))
+        if (action.messageContent is MessageTextContent) {
+            shareText(requireContext(), action.messageContent.body)
```
bmarty, line 1561: Sadly the body can be totally different than the displayed text, because we can use the formatted_body to render the Event.
Anyway changing that is out of scope of this PR

bmarty (approved):
Thanks for this new PR.
One small remark, without action on your side.

Issue #2234, opened 2020-10-08
Rows are hidden in Textinput


DoM1niC:
**Describe the bug**
Textinput Box in UI don't update/scrollable the Rows since the current Develop commits, if I reopen it sometimes there is possible to show the hidden Text (Rows), so the previous Rows would be hidden and not scrollable.

**To Reproduce**
Create a chat and write some Text more then 3-4 Rows 

**Expected behavior**

**Screenshots**
![Screenshot_20201014_051242.jpg](https://user-images.githubusercontent.com/1436958/95939248-ea9ba580-0ddb-11eb-8ebf-e6808f758a5e.jpg)

**Smartphone (please complete the following information):**
 - Device: Huawei Mate 20 Lite, tested on another Devices too! 
 - OS: Android 10

**Additional context**
 - App version : Develop 
 - Homeserver: Master 

Testet daily with blanked and cleaned Source without success, only early Versions of the Code are ok.

DoM1niC:
- I reset to previous commits so the issue is not there
- the developed without my fork give me the same issue like my picture 🖼

DoM1niC:
THX PR #2263 fixed my Issue THX @ganfra

bmarty:
Thanks for confirming!

PR #2263, opened 2020-10-16
Fix composer: update constraints layout and fix some glitches

ganfra:
This PR update constraint layout dependency to the last stable version.
It fixes some glitch on the composer animation.
And fixes this issue : #2234


DoM1niC:
THX @ganfra will test it :)

PR #2275, opened 2020-10-20
Feature/bma/create room

### LINKED ISSUE #2078: New room creation screen
Similar to https://github.com/vector-im/element-ios/issues/3628

## What we have so far

image

## Target

image

With Android native component of course.

## TODO

- Update design
- Add capability to set room topic during the room creation process. It is already possible SDK side using `CreateRoomParams`
- Add capability to set room avatar during the room creation process. For this one I think it is necessary to do it after the room has been created (to check), or to use the parameter `initial_state` from https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-createroom (not supported by the SDK yet)

bmarty:
This PR fixes #2078 : it's now possible to set avatar and topic in the room creation screen.

It include some rework and cleanup and some other small fixes.

The "Create" action is not a menu item anymore, but a Button at the bottom of the Form, which is more common on Android plateforme, and force the user to scroll to the end of the screen to check all the possible options when creating a room.

## Before

This is what we have now:

image

## Updated screen:

image
image

When an image is set, it is possible to remove it:

image

(I will fix the background color error)

## Change SDK side

The image is uploaded once the user wants to create the room and then the room avatar event is added to the initialState event parameter of the room creation request. Any error when uploading the avatar is ignored.

## Other

- `GalleryOrCameraDialogHelper` has been creating to reduce duplication of code. Sadly I cannot use it in BigImageViewerActivity because it is not a Fragment.
- Change in the script we use to release the app in a dedicated commit
- Fix crash in develop due to bad value in Greek translation. It's not crashing in production, but Greek language is not available for now


vector/src/main/java/im/vector/app/core/dialogs/GalleryOrCameraDialogHelper.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private val takePhotoActivityResultLauncher = fragment.registerStartForActivityResult { activityResult ->
+        if (activityResult.resultCode == Activity.RESULT_OK) {
+            avatarCameraUri?.let { uri ->
+                MultiPicker.get(MultiPicker.CAMERA)
+                        .getTakenPhoto(fragment.requireContext(), uri)
+                        ?.let {
+                            listener.onImageReady(it)
+                        }
```
onurays, line 55:
```
MultiPicker
                .get(MultiPicker.CAMERA)
                .getTakenPhoto(fragment.requireContext(), uri)
                ?.let { listener.onImageReady(it)
```

I would prefer ^ for better readability (seems different better than this preview) but this is subjective.

bmarty: Done

vector/src/main/java/im/vector/app/features/form/FormSubmitButtonItem.kt
```diff
@@ -0,0 +1,60 @@
...
+    @StringRes
+    var buttonTitleId: Int? = null
+
+    @EpoxyAttribute
+    var buttonClickListener: ClickListener? = null
+
+    override fun bind(holder: Holder) {
+        super.bind(holder)
+        if (buttonTitleId != null) {
+            holder.button.setText(buttonTitleId!!)
```
onurays, line 48: I want to discuss using let/run instead of if/else...

onurays (approved):
Nice cleanup, just 1 more remark which we can discuss later.

vector/src/main/java/im/vector/app/features/form/FormSubmitButtonItem.kt
bmarty: Yes, I do not like using `!!` too. Will update that later, thanks!

PR #2282, opened 2020-10-22
Feature/bma/room settings

bmarty:
Room settings update: It's now possible to update the room avatar from the room setting (and not from the BigImageViewerActivity anymore).

It's also possible to delete the room avatar.

We are now displaying a warning if the user leave the screen with unsaved changes.


vector/src/main/java/im/vector/app/features/roomprofile/RoomProfileFragment.kt
```diff
@@ -269,45 +245,9 @@ class RoomProfileFragment @Inject constructor(
 
     private fun onAvatarClicked(view: View, matrixItem: MatrixItem.RoomItem) = withState(roomProfileViewModel) {
         if (matrixItem.avatarUrl?.isNotEmpty() == true) {
-            val intent = BigImageViewerActivity.newIntent(requireContext(), matrixItem.getBestName(), matrixItem.avatarUrl!!, it.canChangeAvatar)
+            val intent = BigImageViewerActivity.newIntent(requireContext(), matrixItem.getBestName(), matrixItem.avatarUrl!!)
```
onurays, line 248: Can we use `matrixItem.avatarUrl?.takeIf { it.isNotEmpty() }?.let { ... } `please?

onurays (changes requested):
Nice to see the discard warning. Just 1 remark.

vector/src/main/java/im/vector/app/features/roomprofile/RoomProfileFragment.kt
bmarty: Done

onurays (approved):

PR #2283, opened 2020-10-22
Crypto cleaning and perf improvement

BillCarsonFr:
### Crypto Module cleaning and improvment

1. TimeLine Event Decryptor was improperly defined with session scope:
    - thus start/stop calls would stop existing one or create 2 threads.
    -  RoomSummaryUpdater was also using one but with no start stop (now directly trying to decrypt).
    -  Pass event instead of id to avoid transaction to get up to date event, moved decrypt call outside of session database transaction.

2. Break unneeded internal dependencies to CryptoService: introduced EventDecryptor (session scope)

3.  Decrypt flow: Remove write transactions from this flow (get/store inboundGroupSession), created a inboundSession store that is caching a bit and batching writes

4. Remove KeyRequests/cancel from the crypto scope, as they can jam the handler

5. Remove some un-needed decrypt requests (Like event live observers trying to decrypt all events of initial syncs ,thus jamming crypto, these decryption have no chance to succeed, and after initial sync events are decrypted on the fly before beeing passed on) [Notice we still have the  issue of aggregators not able to react to decryption of old events]

6. Cross Signing Udates were also jaming the crypt (+ was called twice in a row for same user changes) -> Now moved to a worker + it ensures it will be executed now

*Update*

7. Offload things from crypto thread:
    - Offload incoming key/secret requests in specific thread, plus batch tx
    - Offload withheld sending off crypto (in computation)
    - Offload megolm backup import out off crypto (was delaying send first message in encrypted room)

*Update 2*

8. Fix issues with Audit Trails
    - Use paged list with monarchy to display gossiping events in settings (no more freeze)
    - Outgoing key share was not stored
    - Offload and buffer saving of gossiping trails with a async write
    - Add export option to save audit trail as a text file


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CryptoSessionInfoProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.matrix.android.sdk.internal.di.SessionDatabase
+import org.matrix.android.sdk.internal.session.room.membership.RoomMemberHelper
+import org.matrix.android.sdk.internal.util.fetchCopied
+import javax.inject.Inject
+
+/**
+ * The crypto module needs some information regarding rooms that are stored
+ * in the session DB, this class encapsulate this functionality
+ */
+class CryptoSessionInfoProvider @Inject constructor(
```
bmarty, line 33: internal

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DeviceListManager.kt
```diff
@@ -377,7 +377,7 @@ internal class DeviceListManager @Inject constructor(private val cryptoStore: IM
         }
 
         // Update devices trust for these users
-        dispatchDeviceChange(downloadUsers)
+        // dispatchDeviceChange(downloadUsers)
```
bmarty, line 380: not necessary anymore?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/InboundGroupSessionStore.kt
```diff
@@ -0,0 +1,96 @@
...
+        // We want to batch this a bit for performances
+        dirtySession.add(wrapper)
+
+        timerTask?.cancel()
+        timerTask = object : TimerTask() {
+            override fun run() {
+                batchSave()
+            }
+        }
+        timer.schedule(timerTask!!, 2_000)
```
bmarty, line 84: I'm always afraid of the use case of user signout during this 2s.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendVerificationMessageTask.kt
```diff
@@ -15,28 +15,28 @@
...
 import org.matrix.android.sdk.internal.session.room.send.SendResponse
 import org.matrix.android.sdk.internal.task.Task
-import org.greenrobot.eventbus.EventBus
 import javax.inject.Inject
 
 internal interface SendVerificationMessageTask : Task<SendVerificationMessageTask.Params, String> {
     data class Params(
-            val event: Event,
-            val cryptoService: CryptoService?
+            val event: Event
```
bmarty, line 31: Happy to see that change!

bmarty (changes requested):
Thanks for the PR, nice work. I made a first review (still a Draft PR?)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CryptoSessionInfoProvider.kt
BillCarsonFr: done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DeviceListManager.kt
BillCarsonFr: It was called twice, because it is called in onKeysDownloadSucceed (called 2 lines below this one)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/InboundGroupSessionStore.kt
BillCarsonFr: yeah, i added a tryOrNull to protect.. but this part is defintly not perfect.
Maybe i could later put that as part of the CryptoStore directly? thus i could save on close

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendVerificationMessageTask.kt
BillCarsonFr: 👍

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DeviceListManager.kt
bmarty: Yes indeed. I'm wondering if we should not keep this one and remove the other call, but OK anyway.

bmarty:
LGTM. I've made some cleanup and also user can now expoft the trail in the rageshake

bmarty (approved):

PR #2285, opened 2020-10-22
Feature/bma/u crop

bmarty:
Use new way to start Activity for result with uCrop. Not done in BigImageViewerActivity since it has been removed in #2282.

Less duplicated code!


onurays (approved):
LGTM!

PR #2303, opened 2020-10-27
Add preference to send with enter

### LINKED ISSUE #1195: Add option to send messages with enter button
Like old Riot for Android, Whatsapp & Co it will be fine if also riotX gets the option to send messages with the enter button.

In case you need more info: The issue and solution of riot-android might help:
Issue: https://github.com/vector-im/riot-android/issues/1914
Solution: https://github.com/vector-im/riot-android/pull/3019

### LINKED ISSUE #1440: Can't send messages by pressing Enter on an external keyboard
Hi there! I'm quite a huge fan of all Riot and RiotX, so first, a huge thank you to all the people working on it!

I'm using an external (bluetooth) keyboard on Android. When I'm typing a message and pressing Enter, it'll interpret it as a soft-enter in the textarea (i.e. add a new line in the message textarea). This is fine. But then I try to use Ctrl+Enter (or really, any other combination of key with Enter), with the intent of sending the message, and nothing works. Would it make sense that pressing Ctrl+Enter sends the current message on RiotX for Android? (bonus points if there's a setting that makes it possible to choose among Enter or Ctrl+Enter sends the message, and which one creates the new line in the textarea.)

Cheers!

### LINKED ISSUE #1881: Hardware keyboard enter inserts a newline instead of sending message
Normally (practically in all other IM's including Element Desktop) pressing enter key sends the typed message. In Element Android hardware enter key inserts newline in the text input field. 

Tested on Gemini PDA and Cosmo Communicator.

BillCarsonFr:
- Add preference settings to send message by pressing the soft keyboard enter key (instead of new line)
 - Add support for external keyboard send (currently can only break line)

=> I imported back the old app code https://github.com/vector-im/riot-android/blob/af848a04704b8a286c39ce75b23894f63de53cfa/vector/src/main/java/im/vector/activity/VectorRoomActivity.java#L733

Fixes #1195
Fixes #1881
Fixes #1440

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,8 @@ Features ✨:
 Improvements 🙌:
  - Rework sending Event management (#154)
  - New room creation screen: set topic and avatar in the room creation form (#2078)
+ - Add option to send with enter (#1195)
+ - Use Hardware keyboard enter to send message (use shift-enter for new line) (#1881, #1440)
 
 Bugfix 🐛:
  - Messages encrypted with no way to decrypt after SDK update from 0.18 to 1.0.0 (#2252)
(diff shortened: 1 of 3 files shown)


onurays (approved):
LGTM!

PR #2311, opened 2020-10-28
Room member profile: Add action to create (or open) a DM (#2310)

### LINKED ISSUE #2310: Create (or open) a DM from the room member profile screen
As per Element Web, with the option "Direct message":

image

bmarty:
Fixes #2310 

New Action: "Direct message":

image


BillCarsonFr (commented):
LGTM
Just a remark, I think that the #getDirectRoomWith method is currently flawed, if it finds a group DM it will give you that room
-> Maybe could be fixed as part of this PR

bmarty:
Yes, I thought about that. Also, I do not really understand the role of this line: https://github.com/vector-im/element-android/blob/44114e66c2772fa79a944446d168aac0dc2ac779/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/RoomGetter.kt#L57

There is already a filter on the `otherUserId` a few line above.

BillCarsonFr (approved):
LGTM

PR #2318, opened 2020-10-29
Fix ui echo isues

BillCarsonFr:
Fix issues with ui local echo (introduced with send re-work), visible when adding a reaction that would be counted twice

We had several issues with UI echos
    - We were listening only to filtered sending events, which is annoying as relations events filtered but their updates are needed to update ui echo. Changed this and delayed the filtering at the time of event building
    - There was an issue with update of reaction (was looking at reference event instead of local id of the reaction)
    - We also had an issue in the event filtering that was not checking on clear type , clear content (for sending event post encrypt)


bmarty (approved):
LGTM, hence should be nice if @ganfra could have a quick look as well

PR #2320, opened 2020-10-29
Fix background color of reactions.

onurays:
WhatsApp Image 2020-10-29 at 16 26 53

WhatsApp Image 2020-10-29 at 16 28 03

WhatsApp Image 2020-10-29 at 16 28 59

### DIFF
--- a/vector/src/main/res/drawable/reaction_rounded_rect_shape_off.xml
+++ b/vector/src/main/res/drawable/reaction_rounded_rect_shape_off.xml
@@ -2,8 +2,6 @@
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="rectangle">
 
-    <size android:width="40dp" android:height="22dp"/>
-
     <solid android:color="?riotx_reaction_background_off" />
 
     <corners android:radius="20dp" />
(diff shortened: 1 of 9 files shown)


vector/src/main/res/drawable/reaction_rounded_rect_shape_off.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+
+    <size android:width="40dp" android:height="22dp"/>
```
bmarty, line 5: why we have a size here? And not in `reaction_rounded_rect_shape.xml`?

bmarty (commented):
LGTM, just one question

vector/src/main/res/drawable/reaction_rounded_rect_shape_off.xml
onurays: I have no idea. It is a copy of `rounded_rect_shape_off.xml`. I will remove and test it.

onurays: Fixed.

bmarty (approved):
thx!

Issue #2307, opened 2020-10-28
Search messages return strange result regarding edited message 


bmarty:
Search API return messages that has been edited and so does not contains the search term anymore.

First work for this issue is to describe it a bit more, with step to reproduce, and check how Element Web is handling that.

Issue #2200, opened 2020-10-01
[Search] Highlight text in the body of the displayed result


bmarty:
The HS returns the highlighted strings as a list, so we should use this list to highlight text at rendering level

bmarty:
Done in #2324

PR #2324, opened 2020-10-30
Feature/bma/search highlight

### LINKED ISSUE #2238: Search Result | scroll jumps after pagination
Step to reproduce:
 - Open search message screen
 - Type a query and tap search
- Scroll up to trigger a pagination -> Then start scrolling down before the pagination is finished

=> Actual: You will notice a scroll jump (the recycler view scrolls all alone after the pagination success

bmarty:
Improvements on Search screen:

- fixes #2238 by modifying the layout and removing the previous trick
- implements #2200: make the search terms in bold in the result
- take new content for edited event
- add "No more result" item (iso web)

image

#2307 is still a thing but less urgent, and related to something strange from Synapse

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@ Improvements 🙌:
 
 Bugfix 🐛:
  - Messages encrypted with no way to decrypt after SDK update from 0.18 to 1.0.0 (#2252)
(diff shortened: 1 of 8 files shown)


onurays (approved):
LGTM!

Issue #1839, opened 2020-07-29
crash on startup


richvdh:
As soon as I open riotx, it closes again.

logs fromADB:

```
07-29 10:11:10.709 23332 23332 E Zygote  : isWhitelistProcess - Process is Whitelisted
07-29 10:11:10.712 23332 23332 E Zygote  : accessInfo : 1
07-29 10:11:10.722 23332 23332 I tor.riotx.debu: Late-enabling -Xcheck:jni
07-29 10:11:11.091 23332 23332 I MultiDex: VM with version 2.1.0 has multidex support
07-29 10:11:11.091 23332 23332 I MultiDex: Installing application
07-29 10:11:11.091 23332 23332 I MultiDex: VM has multidex support, MultiDex support library is disabled.
07-29 10:11:11.136 23332 23332 I FirebaseInitProvider: FirebaseApp initialization successful
07-29 10:11:11.640 23332 23332 W tor.riotx.debu: Accessing hidden field Lsun/misc/Unsafe;->theUnsafe:Lsun/misc/Unsafe; (light greylist, reflection)
07-29 10:11:12.244 23332 23332 W tor.riotx.debu: Accessing hidden method Lcom/android/org/conscrypt/OpenSSLSocketImpl;->setUseSessionTickets(Z)V (light greylist, reflection)
07-29 10:11:12.244 23332 23332 W tor.riotx.debu: Accessing hidden method Lcom/android/org/conscrypt/OpenSSLSocketImpl;->setHostname(Ljava/lang/String;)V (light greylist, reflection)
07-29 10:11:12.245 23332 23332 W tor.riotx.debu: Accessing hidden method Lcom/android/org/conscrypt/OpenSSLSocketImpl;->getAlpnSelectedProtocol()[B (light greylist, reflection)
07-29 10:11:12.245 23332 23332 W tor.riotx.debu: Accessing hidden method Lcom/android/org/conscrypt/OpenSSLSocketImpl;->setAlpnProtocols([B)V (light greylist, reflection)
07-29 10:11:12.284 23332 23332 W tor.riotx.debu: Accessing hidden method Ldalvik/system/CloseGuard;->get()Ldalvik/system/CloseGuard; (light greylist, reflection)
07-29 10:11:12.284 23332 23332 W tor.riotx.debu: Accessing hidden method Ldalvik/system/CloseGuard;->open(Ljava/lang/String;)V (light greylist, reflection)
[...]

PR #2339, opened 2020-11-03
Realm 10: upgrade lib

ganfra:
Upgrade the dependency of realm. This should be tested on several devices and accounts before release.


bmarty (commented):
LGTM, thanks!
Tested ok on my device... I hope we will not have any pb, because once installed it cannot be reverted...

bmarty (approved):

chengww5217:
Does this PR fix the crash of realm?
I use the [matrix-android-sdk2](https://github.com/matrix-org/matrix-android-sdk2) with the version v1.0.10 build an app, and sometimes got crash with the realm exceptions.
And I met the same problem with this [issue](https://github.com/vector-im/element-android/issues/1839).
Hope the PR could fix this.

PR #2354, opened 2020-11-06
Convert `AccountService` to suspend functions

Dominaezzz:

### Pull Request Checklist

Would do a big PR but this will be easier to review.
Also each endpoint needs to be checked to see which dispatchers it should use so it's best to keep it small.

### DIFF
--- a/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
+++ b/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
@@ -344,7 +344,6 @@ class CommonTestHelper(context: Context) {
         await(latch, timeout)
     }
 
-    // Transform a method with a MatrixCallback to a synchronous method
     fun <T> runBlockingTest(timeout: Long = TestConstants.timeOutMillis, block: suspend () -> T): T {
         return runBlocking {
             withTimeout(timeout) {
--- a/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
+++ b/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
@@ -40,6 +40,7 @@ import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotNull
 import org.junit.Assert.assertTrue
@@ -341,6 +342,15 @@ class CommonTestHelper(context: Context) {
         val latch = CountDownLatch(1)
         block(latch)
         await(latch, timeout)
+    }
+
+    // Transform a method with a MatrixCallback to a synchronous method
+    fun <T> runBlockingTest(timeout: Long = TestConstants.timeOutMillis, block: suspend () -> T): T {
+        return runBlocking {
+            withTimeout(timeout) {
+                block()
+            }
+        }
     }
 
     // Transform a method with a MatrixCallback to a synchronous method
--- a/vector/src/main/java/im/vector/app/features/settings/account/deactivation/DeactivateAccountViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/settings/account/deactivation/DeactivateAccountViewModel.kt
@@ -25,7 +25,6 @@ import com.squareup.inject.assisted.AssistedInject
 import im.vector.app.core.extensions.exhaustive
 import im.vector.app.core.platform.VectorViewModel
 import im.vector.app.core.platform.VectorViewModelAction
-import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.launch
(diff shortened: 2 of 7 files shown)


matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
```diff
@@ -343,6 +344,15 @@ class CommonTestHelper(context: Context) {
         await(latch, timeout)
     }
 
+    // Transform a method with a MatrixCallback to a synchronous method
```
bmarty, line 347: No MatrixCallback anymore here!

vector/src/main/java/im/vector/app/features/settings/account/deactivation/DeactivateAccountViewModel.kt
```diff
@@ -67,19 +71,22 @@ class DeactivateAccountViewModel @AssistedInject constructor(@Assisted private v
...
-        session.deactivateAccount(action.password, action.eraseAllData, object : MatrixCallback<Unit> {
-            override fun onSuccess(data: Unit) {
-                _viewEvents.post(DeactivateAccountViewEvents.Done)
-            }
+        viewModelScope.launch {
+            val event = try {
+                session.deactivateAccount(action.password, action.eraseAllData)
+                DeactivateAccountViewEvents.Done
+            } catch (failure: Exception) {
+                if (failure is CancellationException) throw failure
```
bmarty, line 79: Why `CancellationException` is treated apart, and why throwing? Will it make the app crash?

bmarty (commented):
Thanks for the PR, one small remark and a question.
This look super promising, and we will finally get ride of MatrixCallback!

bmarty:
Please also run to cleanup the code
```sh
curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.34.2/ktlint && chmod a+x ktlint
./ktlint --android --experimental -v
```

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
Dominaezzz: Ha, missed a spot.

vector/src/main/java/im/vector/app/features/settings/account/deactivation/DeactivateAccountViewModel.kt
Dominaezzz: `CancellationException` in general won't make the app crash.
When `launch` gets a `CancellationException`, it just gracefully stops executing.
When `async` gets a `CancellationException`, it just gracefully stops executing and other coroutines calling `await()` will get the cancellation exception.

The only case when it may crash the app is if it is **directly** thrown inside `runBlocking`, since it has to run in non-coroutine code.

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
bmarty: thx

vector/src/main/java/im/vector/app/features/settings/account/deactivation/DeactivateAccountViewModel.kt
bmarty: Ok, thanks for the answer!

ganfra: Isn't it ok to let the CancellationException go to OtherFailure then? it won't change anything in the execution flow, but we'll loose the error in the UI if not posting to viewEvents

Dominaezzz: It's technically not an error, the task has just been cancelled and this will only happen if the view model has been cleared, so if the UI has been destroyed.
But yeah it doesn't make much of a difference I can remove the special casing.

ganfra: Yes I wasn't thinking about this specific case, but more generally. We can get a cancellation if some suspend got cancelled by any inner stuff. By the way, we still have to be able to cancel all session coroutine (clearing cache/logout)

Dominaezzz: That complicates things..... Is that the only time when you cancel all sessions coroutines?
In the case of `viewModelScope` the coroutine would already have been cancelled before user logs out or clears the cache, no?

ganfra: I guess for most of the cases it will be ok yeah

bmarty (approved):
Thanks for the update.
I will fix the conflict and merge the PR (EDIT: there are no conflicts, this is on another PR).
We have to think about the other applications which uses the SDK, and which are maybe not ready to use coroutines. Maybe for them we will create a project which converts coroutines to MatrixCallback. But for Element and its forks, we think it's a great improvement of the SDK API :)

PR #2367, opened 2020-11-09
Feature/bma/sanity test

bmarty:
This PR adds a sanity test to cover nearly all the screen of the app.

There are still screen to add.

Doing that, I add to fix a few bugs, TDD!

Here is the test running:

UI_test


BillCarsonFr (approved):
Wow, what a test!

PR #2377, opened 2020-11-11
Feature/bma/consent

### LINKED ISSUE #2375: Ask for explicit user consent to send their contact details to the identity server
We need to explicitly tell the user that if he granted permission to access their contact book, their emails and phone numbers may be sent to the identity server, hashed or not, depending on the capability of the identity server.

If we do not do that, we get exposed to the risk that Google removes the app from the PlayStore without any previous warning, as they use to do it in the past.

See the requirement from Google: https://support.google.com/googleplay/android-developer/answer/9888076?hl=en section "Prominent Disclosure & Consent Requirement"

So, we need to ask for the user consent to send their private data to the identity server they configured if any, and we have to do that independently of the contact permission grant request, i.e. even if the contact permission is already granted, we should ask (and store the answer) for the permission to send the data to the IS. The question can be asked on the opened screen when the user click on "Search in my contact", if and only if there is a IS configured on the account. The consent can also be given, and eventually revoked, in the identity server setting screen. If the identity server is removed from the account, the answer to the consent will also be deleted.

Wording proposal for the dialog:

> Send emails and phone numbers
>
> In order to discover existing contacts you know, do you accept to send your contact data (phone numbers and/or emails) to the configured Identity Server (%{identity_server_url})? For more security, the sent data will be hashed before being send.
>
> Yes / No

If the identity server does not provide the `sha256` algorithm, the wording would be:

> In order to discover existing contacts you know, do you accept to send your contact data (phone numbers and/or emails) to the configured Identity Server (%{identity_server_url})? Note that the data will be sent in plain text, i.e. without being hashed.

This change will also fulfill what is recommended here: https://matrix.org/docs/spec/identity_service/latest#security-considerations

bmarty:
Fixes #2375 

## New section in Discovery settings

"Send emails and phone numbers"

image

image

image

## When browsing contact book (so system permission to read contact book already granted, but not consent)

image

image

image

## Implementation

I have to migrate the Identity Server database to store the user consent. User consent could have been stored at the application level, but I wanted to be sure that the consent is reset if the Identity Server is changed.

Client applications using the Matrix SDK can still lie and declare to the Matrix SDK that the user has given their consent, but the application will be exposed to be removed by Google from the PlayStore.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/identity/IdentityService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/identity/IdentityService.kt
@@ -92,8 +92,23 @@ interface IdentityService {
 
     /**
      * Search MatrixId of users providing email and phone numbers
+     * Note the the user consent has to be set to true, or it will throw a UserConsentNotProvided failure
+     * Application has to explicitly ask for the user consent.
+     * Please see https://support.google.com/googleplay/android-developer/answer/9888076?hl=en for more details.
      */
     fun lookUp(threePids: List<ThreePid>, callback: MatrixCallback<List<FoundThreePid>>): Cancelable
+
+    /**
+     * Return the current user consent
+     */
+    fun getUserConsent(): Boolean
+
+    /**
+     * Set the user consent. Application may have explicitly ask for the user consent to send their private data
+     * (email and phone numbers) to the identity server.
+     * Please see https://support.google.com/googleplay/android-developer/answer/9888076?hl=en for more details.
+     */
+    fun setUserConsent(newValue: Boolean)
 
     /**
      * Get the status of the current user's threePid
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/identity/IdentityService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/identity/IdentityService.kt
@@ -93,20 +93,25 @@ interface IdentityService {
     /**
      * Search MatrixId of users providing email and phone numbers
      * Note the the user consent has to be set to true, or it will throw a UserConsentNotProvided failure
-     * Application has to explicitly ask for the user consent.
+     * Application has to explicitly ask for the user consent, and the answer can be stored using [setUserConsent]
      * Please see https://support.google.com/googleplay/android-developer/answer/9888076?hl=en for more details.
      */
     fun lookUp(threePids: List<ThreePid>, callback: MatrixCallback<List<FoundThreePid>>): Cancelable
 
     /**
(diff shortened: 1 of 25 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/identity/IdentityService.kt
```diff
@@ -92,9 +92,24 @@ interface IdentityService {
...
+     * Note the the user consent has to be set to true, or it will throw a UserConsentNotProvided failure
+     * Application has to explicitly ask for the user consent.
+     * Please see https://support.google.com/googleplay/android-developer/answer/9888076?hl=en for more details.
      */
     fun lookUp(threePids: List<ThreePid>, callback: MatrixCallback<List<FoundThreePid>>): Cancelable
 
+    /**
+     * Return the current user consent
+     */
+    fun getUserConsent(): Boolean
```
onurays, line 104: I think Boolean function names should start with a question. What do you think about `shouldRequestUserConsent`?

vector/src/main/java/im/vector/app/features/contactsbook/ContactsBookFragment.kt
```diff
@@ -98,6 +115,7 @@ class ContactsBookFragment @Inject constructor(
     }
 
     override fun invalidate() = withState(contactsBookViewModel) { state ->
+        phoneBookSearchForMatrixContacts.isVisible = state.filteredMappedContacts.isNotEmpty() && state.identityServerUrl != null && !state.userConsent
```
onurays, line 118: Why it is invisible if `!state.userConsent`?

onurays (approved):
LGTM! Just two questions and 1 conflict.

vector/src/main/java/im/vector/app/features/contactsbook/ContactsBookFragment.kt
bmarty: It is **visible** if `!state.userConsent` :)
The button is displayed only if the user have not given their consent yet.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/identity/IdentityService.kt
bmarty: I prefer to keep the symmetry with `setUserConsent()`.
Also, if we call it `shouldRequestUserConsent()` it will have to return the negation of the previous returned value.

I update the Javadoc to be clearer about what those 2 methods does.

deniseal:
Can we please ensure we are linking to the most up to date version of the IS privacy policy: https://vector.im/identity-server-privacy-notice-1

PR #2385, opened 2020-11-13
F-Droid version: ensure timeout of sync request can be more than 60 s…

### LINKED ISSUE #2169: Polling the homeserver, the app always times out after 60 seconds
I am using Element installed from F-Droid. Version 1.0.7 brought along #2080. Setting "Sync request timeout" does make the app set the "&timeout" parameter in a background-polling request for "/_matrix/client/r0/sync" as expected. :+1:

"Sync request timeout" apparently can be set to any integer value. (I like that. Given the option (and support by the homeserver for very-long-living requests), I'd probably set it to "infinite", so I'd have a permanent TCP connection to the homeserver.) 

However, no matter the value of "Sync request timeout", the app gives up waiting for an answer after 60 seconds. It then closes the TCP connection, which is a very expensive operation, because the connection (and probably the SSL session, too) can't be reused for further requests.

Would it be possible to increase the app's internal timeout to something like "Sync request timeout + 10 seconds"?

bmarty:
Fixes #2169 by modifying the HTTP request timeout.

Tested OK with a timeout of 90 seconds

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Improvements 🙌:
  - Handle events of type "m.room.server_acl" (#890)
 
 Bugfix 🐛:
+ - F-Droid version: ensure timeout of sync request can be more than 60 seconds (#2169)
  - Fix issue when restoring draft after sharing (#2287)
  - Fix issue when updating the avatar of a room (new avatar vanishing)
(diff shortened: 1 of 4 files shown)


BillCarsonFr (approved):
LGTM

PR #2397, opened 2020-11-16
Registration: annoying error message scares every new user when they add an email (#2391)

### LINKED ISSUE #2391: Annoying error message scares every new user
**Describe the bug**
When registering (tested on grin.hu HS, but probably it happens on others too), Element shows an error message dialog between the reCAPTCHA, and the e-mail confirmation step (while waiting for e-mail).
> Error:
M_UNAUTHORIZED

**It is shown for _every_ new user, so it's very annoying.** It makes guiding a non-techie user through registration even harder. I am saying it after guiding 5 non-techie through the registration.
It **can't be dismissed**. After clicking OK, it **appears again** after few seconds (no matter how many times you dismiss it).
The most annoying thing is that the error message is **nonsense**. Nothing is wrong, and the registration succeeds after confirming the e-mail in **every case**. There is no reason to show it.

It doesn't happen on Element Web.

**To Reproduce**
1. Start registration on grin.hu HS (probably others too), and fill the forms
2. Complete the "I'm not a robot" challenge
3. Wait for the registration e-mail to arrive (**error is shown here**)
4. Confirm the link in the mail.
5. Registration finished successfully, and logged in to new account.

**Expected behavior**
Not seeing an error message
Especially when nothing causes it, and registration succeeds.

**Smartphone:**
 - Device: Happens with every manufacturer.
 - OS: Various versions.

**Additional context**
 - App version and store: 1.0.9 Play (in every case)
 - Homeserver: grin.hu

bmarty:
Fixes #2391 
Not sure from where this regression comes from.

Actually those are the other resumed Fragment (ex: `LoginCaptchaFragment`) which show the error dialog by mistake, because they are still listening to the ViewEvents. The (sort of expected) 401 error is correctly ignored by `LoginWaitForEmailFragment`

@ganfra do you have an idea

### DIFF
--- a/vector/src/main/java/im/vector/app/features/login/AbstractLoginFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/login/AbstractLoginFragment.kt
@@ -69,6 +69,11 @@ abstract class AbstractLoginFragment : VectorBaseFragment(), OnBackPressed {
     }
 
     override fun showFailure(throwable: Throwable) {
+        // Only the resumed Fragment can eventually show the error, to avoid multiple dialog display
+        if (!isResumed) {
+            return
+        }
+
         when (throwable) {
             is Failure.Cancelled                      ->
                 /* Ignore this error, user has cancelled the action */
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Bugfix 🐛:
  - Fix issue when restoring draft after sharing (#2287)
  - Fix issue when updating the avatar of a room (new avatar vanishing)
  - Discard change dialog displayed by mistake when avatar has been updated
+ - Registration: annoying error message scares every new user when they add an email (#2391)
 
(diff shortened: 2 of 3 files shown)


vector/src/main/java/im/vector/app/features/login/AbstractLoginFragment.kt
```diff
@@ -69,6 +69,11 @@ abstract class AbstractLoginFragment : VectorBaseFragment(), OnBackPressed {
     }
 
     override fun showFailure(throwable: Throwable) {
+        // Only the resumed Fragment can eventually show the error, to avoid multiple dialog display
+        if (!isResumed) {
```
ganfra, line 73: Yes I guess it's ok to do that. It's the issue when you listen to the same things on several fragments...

ganfra (approved):

PR #2407, opened 2020-11-16
Move "Enable Encryption" from room setting screen to room profile screen

### LINKED ISSUE #2394: Enable Encryption from room properties screen
element_room_settings

Finding way to enable encryption in an already existing room is difficult.
Especially when the room properties screen has a Security section but the button isn't there.

Move the **Enable Encryption** button out from **Room Settings** into the **Security** section (it should have a shield icon too, but I was too lazy to mock it up):
kép

bmarty:
Fixes #2394

## Encryption not enabled, permission OK

image

## Encryption not enabled, permission KO

image

## Encryption enabled

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@ Improvements 🙌:
  - Open an existing DM instead of creating a new one (#2319)
  - Ask for explicit user consent to send their contact details to the identity server (#2375)
  - Handle events of type "m.room.server_acl" (#890)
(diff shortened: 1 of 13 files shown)


notramo:
Shouldn't the button have an arrow at the right side, like the ones below?

bmarty:
> Shouldn't the button have an arrow at the right side, like the ones below?

As it does not open another screen, I prefer not to put the arrow here. Also, I added `…` at the end of the text, because there is a confirmation dialog.

PR #2414, opened 2020-11-18
Convert PushRuleService to suspend functions

Dominaezzz:

### Pull Request Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/notification/DefaultPushRuleService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/notification/DefaultPushRuleService.kt
@@ -16,7 +16,6 @@
 package org.matrix.android.sdk.internal.session.notification
 
 import com.zhuinden.monarchy.Monarchy
-import org.matrix.android.sdk.api.MatrixCallback
 import org.matrix.android.sdk.api.pushrules.PushRuleService
 import org.matrix.android.sdk.api.pushrules.RuleKind
 import org.matrix.android.sdk.api.pushrules.RuleSetKey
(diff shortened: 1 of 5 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/notification/DefaultPushRuleService.kt
```diff
@@ -103,37 +101,21 @@ internal class DefaultPushRuleService @Inject constructor(
...
     }
 
-    override fun removePushRule(kind: RuleKind, pushRule: PushRule, callback: MatrixCallback<Unit>): Cancelable {
-        return removePushRuleTask
-                .configureWith(RemovePushRuleTask.Params(kind, pushRule)) {
-                    this.callback = callback
-                }
-                .executeBy(taskExecutor)
+    override suspend fun removePushRule(kind: RuleKind, pushRule: PushRule) {
+        return removePushRuleTask.execute(RemovePushRuleTask.Params(kind, pushRule))
```
bmarty, line 118: In all those 4 methods, the `return` statement could be removed.

vector/src/main/java/im/vector/app/features/settings/VectorSettingsAdvancedNotificationPreferenceFragment.kt
```diff
@@ -50,29 +51,25 @@ class VectorSettingsAdvancedNotificationPreferenceFragment @Inject constructor()
...
+                                            preference.ruleAndKind?.pushRule ?: ruleAndKind.pushRule,
+                                            newRule)
+                                }
+                                if (!isAdded) {
+                                    return@launch
+                                }
+                                result.onSuccess {
+                                    preference.setPushRule(ruleAndKind.copy(pushRule = newRule))
+                                }
+                                hideLoadingView()
```
bmarty, line 66: This is a detail, but now, I would rather call `hideLoadingView()` just after the check if Fragment is added.

bmarty (commented):
LGTM, 2 small remarks though

vector/src/main/java/im/vector/app/features/settings/VectorSettingsAdvancedNotificationPreferenceFragment.kt
Dominaezzz: Technically the behaviour would be different (out of scope) but I'm not bothered if you're not bothered.

bmarty: Yes. I do not think it will introduce a bug if we reorder. Also we could use `.fold()` instead of `onSuccess` and `onFailure` which is a bit more verbose (but equivalent)

bmarty (approved):
Thanks for the update

PR #2429, opened 2020-11-20
Better e2e log reporting

BillCarsonFr:
Globally raise crypto log levels so they are in rageshakes (a lot are verbose only)
    Better log for failed to decrypt ToDevice (print sender info)
    Check that key share and unwedge flows are correctly logged
    Check that key re-share flows are correctly logged


bmarty (approved):
LGTM

PR #2439, opened 2020-11-24
Add room alias management

### LINKED ISSUE #1579: Room alias missing features
#1464 starts the work on Room settings. What is missing:

- removing alias
- listing aliases
- give user a chance to set canonical alias from the list

Also we may improve the room settings view state: add a data class to hold current room state, and update it when the state is changed.
Also setting the room history visibility (without saving) and reverting the change let the save button active (when we revert the change

### LINKED ISSUE #2428: Update of room alias management
- [x] Do not render `m.room.aliases` event anymore in the timeline
- [x] Update rendering of `m.room.canonical_alias` event, since it now has an `alt_aliases` field
- [x] Add server API to add and delete local alias
- [x] Do not use aliases from m.room.alias as the room name fallback
- [x] UI: Add section in the room setting, to open a dedicated screen to manage the room aliases

### Element Web

image

### MSC

https://github.com/matrix-org/matrix-doc/pull/2432

bmarty:
And room directory setting. We are now iso Element Web on this point (with a better UX to publish addresses (=aliases))

Fixes #2428 
Fixes #1579 

Following MSC https://github.com/matrix-org/matrix-doc/pull/2432

This PR also contain some cleanup on the existing code.

### New item "Room addresses" which replace the previous "Canonical alias" Edit text:

image

New screen, user is a regular user

image

### New bottom sheet

image

### When user is admin

image

### Extended

image

### Bottom sheet

image

Only address added by the user can be deleted, else the homeserver will refuse the deletion.

### Published addresses

image

### With canonical alias

image

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/alias/AddRoomAliasTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/alias/AddRoomAliasTask.kt
@@ -20,13 +20,17 @@ import org.greenrobot.eventbus.EventBus
 import org.matrix.android.sdk.internal.di.UserId
 import org.matrix.android.sdk.internal.network.executeRequest
 import org.matrix.android.sdk.internal.session.directory.DirectoryAPI
-import org.matrix.android.sdk.internal.session.room.alias.RoomAliasAvailabilityChecker.Companion.toFullAlias
+import org.matrix.android.sdk.internal.session.room.alias.RoomAliasAvailabilityChecker.Companion.toFullLocalAlias
 import org.matrix.android.sdk.internal.task.Task
 import javax.inject.Inject
 
 internal interface AddRoomAliasTask : Task<AddRoomAliasTask.Params, Unit> {
     data class Params(
             val roomId: String,
+            /**
+             * the local part of the alias.
+             * Ex: for the alias "#my_alias:example.org", the local part is "my_alias"
+             */
             val aliasLocalPart: String
     )
 }
@@ -43,7 +47,7 @@ internal class DefaultAddRoomAliasTask @Inject constructor(
 
         executeRequest<Unit>(eventBus) {
             apiCall = directoryAPI.addRoomAlias(
-                    roomAlias = params.aliasLocalPart.toFullAlias(userId),
+                    roomAlias = params.aliasLocalPart.toFullLocalAlias(userId),
                     body = AddRoomAliasBody(
                             roomId = params.roomId
                     )
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/alias/AddRoomAliasTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/alias/AddRoomAliasTask.kt
@@ -17,27 +17,33 @@
 package org.matrix.android.sdk.internal.session.room.alias
 
 import org.greenrobot.eventbus.EventBus
+import org.matrix.android.sdk.internal.di.UserId
 import org.matrix.android.sdk.internal.network.executeRequest
 import org.matrix.android.sdk.internal.session.directory.DirectoryAPI
+import org.matrix.android.sdk.internal.session.room.alias.RoomAliasAvailabilityChecker.Companion.toFullAlias
 import org.matrix.android.sdk.internal.task.Task
 import javax.inject.Inject
 
 internal interface AddRoomAliasTask : Task<AddRoomAliasTask.Params, Unit> {
     data class Params(
             val roomId: String,
-            val roomAlias: String
+            val aliasLocalPart: String
     )
 }
 
 internal class DefaultAddRoomAliasTask @Inject constructor(
+        @UserId private val userId: String,
         private val directoryAPI: DirectoryAPI,
+        private val aliasAvailabilityChecker: RoomAliasAvailabilityChecker,
         private val eventBus: EventBus
 ) : AddRoomAliasTask {
 
     override suspend fun execute(params: AddRoomAliasTask.Params) {
+        aliasAvailabilityChecker.check(params.aliasLocalPart)
+
         executeRequest<Unit>(eventBus) {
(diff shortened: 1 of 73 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/alias/AddRoomAliasTask.kt
```diff
@@ -16,28 +16,34 @@
...
+import org.matrix.android.sdk.internal.session.room.alias.RoomAliasAvailabilityChecker.Companion.toFullAlias
 import org.matrix.android.sdk.internal.task.Task
-import org.greenrobot.eventbus.EventBus
 import javax.inject.Inject
 
 internal interface AddRoomAliasTask : Task<AddRoomAliasTask.Params, Unit> {
     data class Params(
             val roomId: String,
-            val roomAlias: String
+            val aliasLocalPart: String
```
onurays, line 30: Do we need to add a comment on top of this? Or should we expect developers to understand what is the local part of an alias?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/alias/RoomAliasAvailabilityChecker.kt
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+                ?.let {
+                    // Alias already exists: error case
+                    throw RoomAliasError.AliasNotAvailable
+                }
+    }
+
+    companion object {
+        internal fun String.toFullAlias(userId: String) = "#" + this + ":" + userId.substringAfter(":")
```
onurays, line 59: I didn't read the spec but I feel like we should append `roomId.substringAfter(":")` instead of userId, no?

bmarty: No, we need the homeserver of the user here. The roomId can have another homeserver url part.
But maybe the method could be renamed to `toFullLocalAlias` to avoid confusion.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/alias/AddRoomAliasTask.kt
bmarty: I tried to use `aliasLocalPart` and `alias` when appropriate to distinguish the case, since there are both String. But yeah I will add a comment here.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/format/NoticeEventFormatter.kt
```diff
@@ -465,21 +465,76 @@ class NoticeEventFormatter @Inject constructor(
...
-                    }
+        val prevContent: RoomCanonicalAliasContent? = event.resolvedPrevContent().toModel()
+        val canonicalAlias = eventContent?.canonicalAlias?.takeIf { it.isNotEmpty() }
+        val prevCanonicalAlias = prevContent?.canonicalAlias?.takeIf { it.isNotEmpty() }
+        val altAliases = eventContent?.alternativeAliases.orEmpty()
+        val prevAltAliases = prevContent?.alternativeAliases.orEmpty()
+        val added = altAliases - prevAltAliases
+        val removed = prevAltAliases - altAliases
+
+        return if (added.isEmpty() && removed.isEmpty() && canonicalAlias == prevCanonicalAlias) {
```
onurays, line 476: I converted the top-level `if/else` branches to the `when` expression and it seems more readable to me, especially for the sake of comments (after auto converting to "when", we may move comments one line above).

vector/src/main/java/im/vector/app/features/roomprofile/settings/RoomSettingsController.kt
```diff
@@ -46,7 +46,7 @@ class RoomSettingsController @Inject constructor(
         fun onNameChanged(name: String)
         fun onTopicChanged(topic: String)
         fun onHistoryVisibilityClicked()
-        fun onAliasChanged(alias: String)
+        fun onOpenAlias()
```
onurays, line 49: I found `openAlias` naming confusing. Is it a term that I don't know?

onurays (approved):
Amazingly clean work! I just have some questions/remarks.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/format/NoticeEventFormatter.kt
bmarty: Ok done, please review the commit :)

vector/src/main/java/im/vector/app/features/roomprofile/settings/RoomSettingsController.kt
bmarty: I renamed it to `onRoomAliasesClicked`

onurays:
Some remarks:
 - It is crashing if I use the "Done" button of the keyboard while adding Local Address:
`Exception: java.lang.IllegalStateException: focus search returned a view that wasn't able to take focus!`
- There are lots of M_UNKNOWN internal errors on both Android and Web while publishing alias for the first time. I found error descriptions of Web more clear (we just show "internal error" in error dialog).

onurays (changes requested):
Some remarks:

- It is crashing if I use the "Done" button of the keyboard while adding Local Address:
`Exception: java.lang.IllegalStateException: focus search returned a view that wasn't able to take focus!`
- There are lots of M_UNKNOWN internal errors on both Android and Web while publishing alias for the first time. I found error descriptions of Web more clear (we just show "internal error" in error dialog).

bmarty:
> It is crashing if I use the "Done" button of the keyboard while adding Local Address:
Exception: java.lang.IllegalStateException: focus search returned a view that wasn't able to take focus!

Fixed

For the second remark, I do not reproduce it (at the moment). We may improve this later

onurays (approved):

PR #2444, opened 2020-11-25
Fix issues with matrix.to deep linking

### LINKED ISSUE #2349: Properly support matrix.to links on android app
Matrix.to links do not properly work for user links

-> It opens the RoomMember Profile, but there is no 'chat' button, just ignore
-> Seems that it only opens if the app was not already opened (does nothing..)

+ As per https://github.com/matrix-org/matrix.to/issues/165 we should probably support custom url scheme matrixto://

BillCarsonFr:
matrix.to deep linking had some issues:
 - Clicking a matrix.to link works only the first time
 - Could end up with two instances of the app 

Home activity implements navigation interceptor to display the matrix to bottom sheet

Fixes #2349

image

image

### DIFF
--- a/vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheet.kt
+++ b/vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheet.kt
@@ -41,8 +41,7 @@ class MatrixToBottomSheet : VectorBaseBottomSheetDialogFragment() {
 
     @Parcelize
     data class MatrixToArgs(
-            val matrixToLink: String?,
-            val userId: String?
+            val matrixToLink: String
     ) : Parcelable
 
     @Inject lateinit var avatarRenderer: AvatarRenderer
@@ -136,20 +135,7 @@ class MatrixToBottomSheet : VectorBaseBottomSheetDialogFragment() {
             return MatrixToBottomSheet().apply {
                 arguments = Bundle().apply {
                     putParcelable(MvRx.KEY_ARG, MatrixToBottomSheet.MatrixToArgs(
-                            matrixToLink = matrixToLink,
-                            userId = null
-                    ))
-                }
-                interactionListener = listener
-            }
-        }
-
-        fun withUserId(userId: String, listener: InteractionListener?): MatrixToBottomSheet {
-            return MatrixToBottomSheet().apply {
-                arguments = Bundle().apply {
-                    putParcelable(MvRx.KEY_ARG, MatrixToBottomSheet.MatrixToArgs(
-                            matrixToLink = null,
-                            userId = userId
+                            matrixToLink = matrixToLink
                     ))
                 }
                 interactionListener = listener
--- a/vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheet.kt
+++ b/vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheet.kt
@@ -17,23 +17,38 @@
 package im.vector.app.features.matrixto
 
 import android.os.Bundle
+import android.os.Parcelable
 import android.view.View
+import androidx.core.view.isInvisible
+import androidx.core.view.isVisible
+import com.airbnb.mvrx.Fail
+import com.airbnb.mvrx.Loading
+import com.airbnb.mvrx.MvRx
+import com.airbnb.mvrx.Success
+import com.airbnb.mvrx.Uninitialized
+import com.airbnb.mvrx.fragmentViewModel
+import com.airbnb.mvrx.withState
 import im.vector.app.R
 import im.vector.app.core.di.ScreenComponent
 import im.vector.app.core.extensions.setTextOrHide
 import im.vector.app.core.platform.VectorBaseBottomSheetDialogFragment
 import im.vector.app.features.home.AvatarRenderer
+import kotlinx.android.parcel.Parcelize
 import kotlinx.android.synthetic.main.bottom_sheet_matrix_to_card.*
-import org.matrix.android.sdk.api.util.MatrixItem
 import javax.inject.Inject
 
-class MatrixToBottomSheet(private val matrixItem: MatrixItem) : VectorBaseBottomSheetDialogFragment() {
+class MatrixToBottomSheet : VectorBaseBottomSheetDialogFragment() {
+
+    @Parcelize
+    data class MatrixToArgs(
+            val matrixToLink: String?,
+            val userId: String?
+    ) : Parcelable
 
     @Inject lateinit var avatarRenderer: AvatarRenderer
 
-    interface InteractionListener {
-        fun didTapStartMessage(matrixItem: MatrixItem)
-    }
+    @Inject
+    lateinit var matrixToBottomSheetViewModelFactory: MatrixToBottomSheetViewModel.Factory
 
     override fun injectWith(injector: ScreenComponent) {
         injector.inject(this)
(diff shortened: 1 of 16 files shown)


vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheet.kt
```diff
@@ -43,21 +58,100 @@ class MatrixToBottomSheet(private val matrixItem: MatrixItem) : VectorBaseBottom
...
+        }
+
+        fun withUserId(userId: String, listener: InteractionListener?): MatrixToBottomSheet {
+            return MatrixToBottomSheet().apply {
+                arguments = Bundle().apply {
+                    putParcelable(MvRx.KEY_ARG, MatrixToBottomSheet.MatrixToArgs(
+                            matrixToLink = null,
+                            userId = userId
+                    ))
+                }
```
bmarty, line 154: Maybe we can have only one method and determine later if the String is a userId or a matrixTo link? Or keep only the withLink method, and convert the userId to a matrixTo link first, so this BottomSheet will always take a matrixTo link as parameter, which make sense regarding its name (MatrixToBottomSheet)

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
```diff
@@ -0,0 +1,192 @@
...
+                    is PermalinkData.UserLink -> {
+                        val user = resolveUser(linkedId.userId)
+                        setState {
+                            copy(
+                                    matrixItem = Success(user.toMatrixItem()),
+                                    startChattingState = Success(Unit)
+                            )
+                        }
+                    }
+                    is PermalinkData.RoomLink -> TODO()
```
bmarty, line 95: The `TODO()` will make the app crash...

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
```diff
@@ -0,0 +1,192 @@
...
+import org.matrix.android.sdk.api.session.profile.ProfileService
+import org.matrix.android.sdk.api.session.room.model.create.CreateRoomParams
+import org.matrix.android.sdk.api.session.user.model.User
+import org.matrix.android.sdk.api.util.JsonDict
+import org.matrix.android.sdk.api.util.toMatrixItem
+import org.matrix.android.sdk.internal.util.awaitCallback
+
+class MatrixToBottomSheetViewModel @AssistedInject constructor(
+        @Assisted initialState: MatrixToBottomSheetState,
+        @Assisted val args: MatrixToBottomSheet.MatrixToArgs,
```
bmarty, line 51: See the remark from @ganfra on the other PR.

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
```diff
@@ -0,0 +1,192 @@
...
+        return (session.getUser(userId)
+                ?: tryOrNull {
+                    awaitCallback<JsonDict> {
+                        session.getProfile(userId, it)
+                    }
+                }?.let {
+                    User(userId, it[ProfileService.DISPLAY_NAME_KEY] as? String, it[ProfileService.AVATAR_URL_KEY] as? String)
+                }
+                // Create raw Uxid in case the user is not searchable
+                ?: User(userId, null, null))
```
bmarty, line 134: please remove the extra parenthesis

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    private suspend fun resolveUser(userId: String): User {
+        return (session.getUser(userId)
+                ?: tryOrNull {
+                    awaitCallback<JsonDict> {
+                        session.getProfile(userId, it)
+                    }
+                }?.let {
+                    User(userId, it[ProfileService.DISPLAY_NAME_KEY] as? String, it[ProfileService.AVATAR_URL_KEY] as? String)
```
bmarty, line 131: this is ugly code, we could provide something better SDK side to convert a result of getProfile() to a User

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
```diff
@@ -0,0 +1,192 @@
...
+            viewModelScope.launch(Dispatchers.IO) {
+                val adminE2EByDefault = rawService.getElementWellknown(session.myUserId)
+                        ?.isE2EByDefault()
+                        ?: true
+
+                val roomParams = CreateRoomParams()
+                        .apply {
+                            invitedUserIds.add(mxId)
+                            setDirectMessage()
+                            enableEncryptionIfInvitedUsersSupportIt = adminE2EByDefault
```
bmarty, line 172: this make me think that the method `createDirectRoom()` is not correct since it ignore the wellknown configuration.

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
```diff
@@ -0,0 +1,192 @@
...
+                        }
+
+                val roomId =
+                        try {
+                            awaitCallback<String> { session.createRoom(roomParams, it) }.also {
```
bmarty, line 177: no need to use `also` here, if you move the setState() call after the `try/catch` block

vector/src/main/java/im/vector/app/features/permalink/PermalinkHandlerActivity.kt
```diff
@@ -45,23 +43,28 @@ class PermalinkHandlerActivity : VectorBaseActivity() {
...
+        // We forward intent to HomeActivity (singleTask) to avoid the dueling app problem
+        // https://stackoverflow.com/questions/25884954/deep-linking-and-multiple-app-instances
+        intent.setClass(this, HomeActivity::class.java)
+        intent.flags = Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_SINGLE_TOP
+        startActivity(intent)
+
+        finish()
+    }
+
+    override fun onNewIntent(intent: Intent?) {
```
bmarty, line 65: I do not think it is necessary to add `onNewIntent()` here.

vector/src/main/java/im/vector/app/features/usercode/UserCodeSharedViewModel.kt
```diff
@@ -145,22 +148,36 @@ class UserCodeSharedViewModel @AssistedInject constructor(
...
+                    val user = session.getUser(linkedId.userId)
+                            ?: tryOrNull {
+                                awaitCallback<JsonDict> {
+                                    session.getProfile(linkedId.userId, it)
+                                }
+                            }?.let {
+                                User(linkedId.userId, it[ProfileService.DISPLAY_NAME_KEY] as? String, it[ProfileService.AVATAR_URL_KEY] as? String)
+                            }
+                            // Create raw Uxid in case the user is not searchable
+                            ?: User(linkedId.userId, null, null)
```
bmarty, line 165: Duplicated code :/

bmarty (changes requested):
Thanks for the PR, a few remarks

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
BillCarsonFr: yes but not called yet, anyhow will log silently instead

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheet.kt
BillCarsonFr: That's what i thought and how it was at the beginning, but side effect is that callers would have to resolve user first to get the matrix item, for that callers fragment/activity needed to have waiting view show/hide... leading to code duplication
Maskes more sense that matrix.to is self contained and do network calls

vector/src/main/java/im/vector/app/features/usercode/UserCodeSharedViewModel.kt
BillCarsonFr: added resolveUser in UserService

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
BillCarsonFr: done

BillCarsonFr: done

vector/src/main/java/im/vector/app/features/matrixto/MatrixToBottomSheetViewModel.kt
BillCarsonFr: moved

bmarty (approved):
Thanks for the update!

PR #2451, opened 2020-11-26
Fix / update profile when no rooms

BillCarsonFr:
Update profile has no effect if user is in zero rooms (as nothing comes back from sync?)
-> propose to update user store after update call


bmarty (approved):
Nice, thanks

PR #2465, opened 2020-11-27
Fix / double bottomsheet effect

BillCarsonFr:
When verifying with passphrase, after entering correct passpharse you go back to the bottom sheet, and it's displaying again the options to verify with pass / skip ...
- > actually some secret extraction is going on in background and you have to wait a few seconds for the conclusion fragment to appear

=> Fixed by introducing a new waiting fragment during this process

Also the keybackup import was done in viewmodel scope, moved it to global scope for better reliability

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ Improvements 🙌:
  -
 
 Bugfix 🐛:
- -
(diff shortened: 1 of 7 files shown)


bmarty (approved):
LGTM, thanks

PR #2490, opened 2020-12-04
Url preview

### LINKED ISSUE #481: Url preview
Display URL preview in the timeline, as other Riot clients does (except - sadly - Riot iOS)

Do not perform URL previewing in e2e rooms, as it leaks the URL to the HS

**EDIT**: Update of description:

We need design on how to display URL previews in the timeline.
We have several states to consider: loading/error/loaded. It should be good to avoid blink in the timeline, so all the states should take the same space.
Several preview can be displayed in a row if the message contains several URLs.
Also it is possible to close the preview. On Riot-Android, it had the effect of storing the url value, to avoid display the url preview anymore for that URL. On Element Web it seems that this is not the current behavior, it just closes the preview.

The current data we get for a preview:
- The url itself
- An image URI (optional)
- A title
- A short description

As suggested by @dhrunia, we could consider displaying URL preview in e2e room, but the decision has to be taken globally for the Element product.

Ref: https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-media-r0-preview-url

Element URL display (FTR):

image

Edit:
Design is ready: https://www.figma.com/file/tnv1s5KUVK38Byt54gs7GM/Timeline-URL-previews?node-id=0%3A1 , proposal 2 is mature (we think :) but feedback welcome via Figma comments

bmarty:
Fixes #481

### Implementation

Data useful for Element are cached in DB (Session DB, migration implemented). The SDK offer the API to retrieve all the data from the server (without cache).

### Behavior

This PR add display in the timeline of the first URL of a text message (or emote). The URL has to contain the protocol (iso Element Web). We limit to one preview because it takes too much space on a mobile screen, but we may change this later (FTR WhatsApp also limits to 1 preview).

Closing the preview is stored for the scope of the screen. When re-entering the room, the preview will be displayed again.

Loading, Error, Unknown state, etc. of the Preview are just hidden for the moment.

Design is in #481 

### Screenshots

image

image

### Known bugs

- [x] I will fix some bug in the layout
- [x] Fix preview vanished after screen rotation


tulir:
Any chance of getting some way to disable previews, either per-room (Element web supports a `org.matrix.preview_urls` room state event with a `disable` boolean) or per-message (MSC2376 or MSC2385)? I don't want to disable url previews globally as soon as they're released just because there's no other option :(

bmarty:
> `org.matrix.preview_urls`

@tulir Where is this state event specified?

tulir:
It's `org.matrix.` rather than `m.` so it's probably not specified other than maybe some element web docs. Element web supports that event type in different config levels (room state, room account data, account data) and it's just the `disable` boolean in the content, e.g. `"content": {"disable": true}`

Implementing one of the MSCs might be a better idea in terms of having a spec

BillCarsonFr (approved):
LGTM!

PR #2512, opened 2020-12-09
Improve management of files

bmarty:
## This PR contains several changes

- Update of Realm library
- Log requests and request failure in release mode, without detail, to help debugging rageshakes.
- Small improvement on rendering: add space between text and image when waiting for message to be decrypted:
image

- Improve the way the SDK managed the encrypted file: now the encrypted version is stored in cache. Decrypted version can also be stored in cache but will be deleted each time the application starts.
- Use the correct name for file. So when we share the file, the exported file is correctly name. We use a hash if the mxc url as a parent folder to avoid name collision.
- Prevent Glide from storing images on disk for encrypted files.

So here is the files hierarchy of the cache:

### During usage of app

Screenshot 2020-12-09 at 12 17 46

- "d4856..." is a hash of the session Id (for future multi account support)
- "F" is a subfolder to sotre downloaded files
- "D" is a subfolder to store decrypted file. This folder is deleted each time the app is started
- "3f468..." is a hash of an mxc URL of a clear fille
- "d17a3..." and "d2f64..." are hashes of 2 mxc URL of encrypted files.
- "encrypted.bin" is the filename used for encrypted file (hard coded in the SDK).

### After app restarted

Screenshot 2020-12-09 at 12 28 54

We can see that the "D" folder has been removed. Even without network coverage, the app will be able to decrypt again the file to display it in the timeline or share it.

### When a file is shared to Gmail for instance

image

We can see that the file name is correct.


BillCarsonFr (commented):
Great changes, static review ok!
Do you think it's worth it doing a cache migration?

bmarty:
We cannot handle the cache migration since the FileService cannot know which files were originally encrypted and which were not, since the FileService used to store only clear files using the mxcUrl as filename.
So the migration is handled by deleting all the previously downloaded files, stored in the folder "MF"

PR #2514, opened 2020-12-09
Emoji Keyboard

### LINKED ISSUE #2520: Emoji Keyboard
Screen Shot 2020-12-10 at 12 57 39

onurays:
Fixes #2520 

emoji

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
@@ -53,7 +53,6 @@ import androidx.lifecycle.Observer
 import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
-import butterknife.BindView
 import com.airbnb.epoxy.EpoxyModel
 import com.airbnb.epoxy.OnModelBuildFinishedListener
 import com.airbnb.epoxy.addGlidePreloader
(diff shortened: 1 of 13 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -478,6 +482,23 @@ class RoomDetailFragment @Inject constructor(
...
+    private fun setupEmojiPopup() {
+        EmojiPopup
+                .Builder
+                .fromRootView(rootConstraintLayout)
+                .setKeyboardAnimationStyle(R.style.emoji_fade_animation_style)
+                .setOnEmojiPopupShownListener { composerLayout.composerEmojiButton.setImageResource(R.drawable.ic_keyboard) }
+                .setOnEmojiPopupDismissListener { composerLayout.composerEmojiButton.setImageResource(R.drawable.ic_insert_emoji) }
+                .build(composerLayout.composerEditText)
+                .also {
+                    emojiPopup = it
```
bmarty, line 494: Why not just affecting the value like this: `val emojiPopup = EmojiPopup.Builder...`?
It seems that `emojiPopup` does not have to be a member of the class.

bmarty (commented):
One first remark, I will do some more test

bmarty:
Also can you add en entry in [this file](https://github.com/vector-im/element-android/blob/bc339191eaf3b6d03eeeba079f025464501d11ab/vector/src/main/assets/open_source_licenses.html#L393) please?

bmarty (approved):
Thanks for the update!

erAck:
WTF does this add an entire emoji library and buttons on its own?

github-tomster:
@'code creating ones'
unfortunately there seems to be an huge info-gap: 
almost **every** android device has
a font named NotoColorEmoji.ttf 🤦‍♂️ 

you are welcome and i am happy helping you out with this 
not so obvious information!

so what is the intension by **not** using the
system NotoColorEmoji.ttf ❓  🤷‍♂️ 

is this about to take away control from the user ❓

github-tomster:
@ freetime-programmers
does a merge makes a bad code a good code? 🤦‍♂️ 
💉

Issue #2194, opened 2020-10-01
"No known servers" when joining rooms via federation


ajayyy:
When trying to join a room with an address, such as #gitter:matrix.org, it fails with the error "No known servers".

Joining rooms from the discovery list on other instances also fails.

![image](https://user-images.githubusercontent.com/12688112/94759076-f3689200-036c-11eb-93cc-6c525a3259ba.png)

**Smartphone (please complete the following information):**
 - Device: OnePlus 7t
 - OS: Android 10

**Additional context**
 - App version and store 1.0.8 - Google Play
 - Homeserver: self-hosted

Related: https://github.com/vector-im/element-android/issues/808

bmarty:
Fixed by #2526

PR #2526, opened 2020-12-11
element:// support + basic peeking + fix join via server

### LINKED ISSUE #2516: No known servers error is given when joining rooms on new Gitter bridge
**Describe the bug**
Bridged rooms on the `gitter.im` server cannot be joined from Element Android.

An error is given:

> No known servers
> [Retry]

**To Reproduce**
1. Click on the join room button.
2. Type the room alias (`#<owner>_<repo>:gitter.im`) (because gitter.im doesn't show up as a Matrix network)
3. Click on the entry in the list
4. Click "Join" under "Room cannot be previewed" message.
5. See the error message above

OR

1. Click on a link to a room alias for a Gitter bridge room from another room.
2. Click "Join" under "Room cannot be previewed" message.
3. See the error message above

I believe that I also got the same error when using a matrix.to link, but don't remember at the moment.

**Expected behavior**
The room can be joined from Element Android as they are able to be joined on Element Web.

 - Device: Moto g5 play
 - OS: Android 9.0[e.g. Android 6.0]

**Additional context**
 - App version and store: 1.0.11 - Google Play
 - Homeserver: matrix.org

I am able to re-join the bridged scala/scala room (`#scala_scala:gitter.im`) after joining and leaving on Element Web.

BillCarsonFr:
This PR adds 
 - support for element://user/ and element://room/ custom scheme that the matrix.to is providing now
 - Basic peeking for unknown rooms (resolve name, avatar, topic...)
      - If alias given, resolve the alias (and get via servers)
      - Then try to see if room is public, if it's the case try to find it in public directory 
      - If not public or not found, try to get state events (in case guest allowed) and get name/avatar/topic from there
- The RoomPreviewActivity is more self contained and now able to peek (loading status) and update display

Fixes #2516 (due to app not getting/using via server)

### Room not public and without guest access
image

### Room not pubic and with guest access
image

### Public room found in directory

image


bmarty (approved):
LGTM, does it also fixes https://github.com/vector-im/element-android/issues/2194 ?

PR #2543, opened 2020-12-16
 Make sure to load all members in the room before sending the event

### LINKED ISSUE #2518: Wait for all room members to be known before sending a message to a e2e room
Due to lazy-loading of room members, all the room members are not known after an initial sync.
The other members are loaded when the user load the timeline of the room.
If the user send a message in this room before all the members are known, the message will be encrypted only for known users, and not for every one in the room. Other users will not be able to decrypt the message.

The SDK has to wait for all the members to be known before sending a message to a e2e room

onurays:
Fixes #2518


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/RealmSessionStoreMigration.kt
```diff
@@ -105,4 +107,18 @@ class RealmSessionStoreMigration @Inject constructor() : RealmMigration {
                 .addField(PreviewUrlCacheEntityFields.MXC_URL, String::class.java)
                 .addField(PreviewUrlCacheEntityFields.LAST_UPDATED_TIMESTAMP, Long::class.java)
     }
+
+    private fun migrateTo7(realm: DynamicRealm) {
+        Timber.d("Step 6 -> 7")
+        realm.schema.get("RoomEntity")
+                ?.addField("membersLoadStatusStr", String::class.java)
```
bmarty: Changed to const

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/RoomMembersLoadStatusType.java
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.internal.database.model;
+
+public enum RoomMembersLoadStatusType {
+    NONE,
+    LOADING,
+    LOADED
+}
```
bmarty: Converted to Kotlin internal class

bmarty (approved):
I made some improvements (@onurays please review my commits), the test `everyone_should_decrypt_message_in_a_crowded_room()` is still passing

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/LoadRoomMembersTask.kt
```diff
@@ -84,14 +113,16 @@ internal class DefaultLoadRoomMembersTask @Inject constructor(
                 }
                 roomMemberEventHandler.handle(realm, roomId, roomMemberEvent)
             }
-            roomEntity.areAllMembersLoaded = true
+            roomEntity.membersLoadStatus = RoomMembersLoadStatusType.LOADED
```
ganfra, line 116: In case of failure, it should be back to NONE?

ganfra (changes requested):
One remark, otherwise LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/LoadRoomMembersTask.kt
bmarty: Yes, true, good catch, thanks, I'll handle that

PR #2568, opened 2020-12-21
Let the SDK compute the URLs

bmarty:
When releasing the SDK 1.0.13, I realized that the auth URLs should be computed by the SDK, and not by the app.

This PR moves those computation SDK side.


BillCarsonFr (approved):
LGTM

Issue #2486, opened 2020-12-03
UTD for events before invitation if member state events are hidden


onurays:
To reproduce:
- User A creates an empty room
- User A sends some messages
- User A invites User B
- User B joins the room
- User B switches off the "Show room member state events" settings from Settings>Preferences
- User B navigates to the room timeline and see UTD for the previous events which User A sent

PR #2581, opened 2020-12-26
Fix member state filter

SpiritCroc:
Fixes the "Show room member state events" setting.

What we want with that setting:
- When enabled, include all member state events
- When disabled, exclude all member state events

What it did:
- When enabled, exclude all member state events except from the
  logged-in user
- When disabled, include all member state events

### Pull Request Checklist

~~- [ ] UI change has been tested on both light and dark themes~~
~~- [ ] Pull request includes screenshots or videos if containing UI changes~~

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Improvements 🙌:
 
 Bugfix 🐛:
  - Url previews sometimes attached to wrong message (#2561)
+ - Fix the "Show room member state events" setting (#2581)
 
 Translations 🗣:
  -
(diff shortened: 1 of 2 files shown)


bmarty:
Hello,
The current code you have changed has been introduced to ensure that the invite event and the join event of the current user are always displayed in the timeline, so the application can filter out the messages sent before the user get invited to an encrypted room (to fix #2486). This look like a work-around and I admit this is not ideal, but this is why we coded it like that. We considered the filtering other users' membership event was enough for large rooms. User may also wish to see the change they made on their own account in the timeline.
We should probably update the wording of the following setting though, to mention that it only affects other users membership changes:

image

bmarty:
So after reading your description again, it seems that you actually fixed an inversion.

What we want:
- When enabled, include all member state events
- When disabled, exclude all member state events except from the logged-in user

SpiritCroc:
Ah, I see. In that case it should be enough to just negate `userPreferencesProvider.shouldShowRoomMemberStateEvents()`, I guess.

bmarty:
Let ask @onurays who worked on this code.

bmarty:
Handled by #2604

PR #2615, opened 2021-01-04
Fix crash in AttachmentViewerActivity (after ViewBindings)

bmarty:
And improve code a bit


onurays (approved):
It seems a good idea to use getter instead of lateinit, thanks.

PR #2616, opened 2021-01-04
Migrate to Android 11, API 30.

### LINKED ISSUE #2600: Support Android 11 (API 30)
More and more users report some issues using Element on Android 11.
We should compile with SDK API 30 set target SDK to 30 and try the app on at least an Emulator running Android 11.

Ref: https://developer.android.com/about/versions/11

### Checklist

- [x] One-time permissions

> Starting in Android 11, whenever your app requests a permission related to location, microphone, or camera, the user-facing permissions dialog contains an option called Only this time. If the user selects this option in the dialog, your app is granted a temporary one-time permission.
https://developer.android.com/training/permissions/requesting#one-time

- [x] Starting in Android 11, only pre-installed system camera apps can respond to the following intent actions:

> - android.media.action.VIDEO_CAPTURE
> - android.media.action.IMAGE_CAPTURE
> - android.media.action.IMAGE_CAPTURE_SECURE

- [x] Scoped storage enforcement

- [ ] Foreground services
> Apps that run on Android 11 or higher and access location, the camera, or the microphone in a foreground service

- [ ] Share content URIs
> If your app shares a content URI with another app, the intent must grant URI access permissions by setting at least one of the following intent flags: FLAG_GRANT_READ_URI_PERMISSION and FLAG_GRANT_WRITE_URI_PERMISSION.

onurays:
Fixes #2600

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,10 +15,10 @@ Translations 🗣:
  -
 
 SDK API changes ⚠️:
- -
+ - Increase targetSdkVersion to 30 (#2600)
 
 Build 🧱:
- -
+ - Compile with Android SDK 30 (Android 11)
 
 Test:
  -
--- a/attachment-viewer/build.gradle
+++ b/attachment-viewer/build.gradle
@@ -32,11 +32,11 @@ buildscript {
 }
 
 android {
-    compileSdkVersion 29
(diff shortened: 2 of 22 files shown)


bmarty (approved):
LGTM, thanks!

PR #2621, opened 2021-01-05
Persist outbound group session

BillCarsonFr:
This PR Improves the sending time of the first message in an e2e room after app launch.

Currently everytime the app is restarted, a new outbound session key is created when sending a new message in a e2e room, thus forcing a new key share. 
This key has to be shared with all participant devices in the room, making the first message sending time slow (from a couple of seconds to several depending on number of participant).
With this PR the latest outbound session is stored, allowing instant encrypting and sending in the room (if there was no membership change or if the key is not too old - 1 week old-)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -421,6 +428,34 @@ internal class MXOlmDevice @Inject constructor(
         return null
     }
 
+    fun storeOutboundGroupSessionForRoom(roomId: String, sessionId: String) {
+        outboundGroupSessionStore[sessionId]?.let {
```
bmarty, line 432: You should probably update the comment above `outboundGroupSessionStore` declaration:
> The outbound group session.
They are not stored in 'store' to avoid to remember to which devices we sent the session key.
Plus, in cryptography, it is good to refresh sessions from time to time.

So this assertion is not true anymore?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -421,6 +428,34 @@ internal class MXOlmDevice @Inject constructor(
...
+            store.storeCurrentOutboundGroupSessionForRoom(roomId, it)
+        }
+    }
+
+    fun restoreOutboundGroupSessionForRoom(roomId: String):  MXOutboundSessionInfo? {
+        val restoredOutboundGroupSession = store.getCurrentOutboundGroupSessionForRoom(roomId)
+        if (restoredOutboundGroupSession != null) {
+            val sessionId = restoredOutboundGroupSession.outboundGroupSession.sessionIdentifier()
+            if (!outboundGroupSessionStore.containsKey(sessionId)) {
```
bmarty, line 441: I do not understand why you are checking on `outboundGroupSessionStore` (which is now actually a cache of what we have in store, no? Rename to `outboundGroupSessionCache`?).
Can you add some comment please?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXOutboundSessionInfo.kt
```diff
@@ -23,9 +23,8 @@ import timber.log.Timber
 internal class MXOutboundSessionInfo(
         // The id of the session
         val sessionId: String,
-        val sharedWithHelper: SharedWithHelper) {
-    // When the session was created
```
bmarty, line 27: Why removing this comment? This is not true anymore?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/RealmCryptoStore.kt
```diff
@@ -756,6 +759,44 @@ internal class RealmCryptoStore @Inject constructor(
...
+
+    override fun storeCurrentOutboundGroupSessionForRoom(roomId: String, outboundGroupSession: OlmOutboundGroupSession?) {
+        // we can do this async, as it's just for restoring on next launch
+        // the olmdevice is caching the active instance
+        // this is called for each sent message (so not high frequency), thus we can use basic realm async without
+        // risk of reaching max async operation limit?
+        doRealmTransactionAsync(realmConfiguration) { realm ->
+            realm.where<CryptoRoomEntity>()
+                    .equalTo(CryptoRoomEntityFields.ROOM_ID, roomId)
+                    .findFirst()?.let { entity ->
```
bmarty, line 785: You can use `CryptoRoomEntity.getById()`

bmarty (changes requested):
Small remarks and questions

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/RealmCryptoStoreMigration.kt
```diff
@@ -483,4 +485,16 @@ internal class RealmCryptoStoreMigration @Inject constructor(private val crossSi
...
+    // Version 12L added outbound group session persistence
+    private fun migrateTo12(realm: DynamicRealm) {
+        Timber.d("Step 11 -> 12")
+        val outboundEntitySchema = realm.schema.create("OutboundGroupSessionInfoEntity")
+                .addField(OutboundGroupSessionInfoEntityFields.SERIALIZED_OUTBOUND_SESSION_DATA, String::class.java)
+                .addField(OutboundGroupSessionInfoEntityFields.CREATION_TIME, Long::class.java)
+                .setNullable(OutboundGroupSessionInfoEntityFields.CREATION_TIME, true)
+
+        realm.schema.get("CryptoRoomEntity")
+                ?.addRealmObjectField(CryptoRoomEntityFields.OUTBOUND_SESSION_INFO.`$`, outboundEntitySchema)
```
bmarty, line 498: The code does not compile because of a warning about nullability of `outboundEntitySchema`.
You should do this ^ on an also block maybe

bmarty: Actually just removing the usage of `?` for `outboundEntitySchema` fixes the pb

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -421,6 +428,34 @@ internal class MXOlmDevice @Inject constructor(
...
+                )
+            } else {
+                restoredOutboundGroupSession.outboundGroupSession.releaseSession()
+            }
+        }
+        return null
+    }
+
+    fun discardOutboundGroupSessionForRoom(roomId: String) {
+        outboundGroupSessionStore.remove(roomId)?.releaseSession()
```
gileluard, line 456: Shouldn't be `outboundGroupSessionStore.remove(sessionId)?.releaseSession()` here as session IDs are the keys?

bmarty: Very good catch! I agree with you

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -421,6 +428,34 @@ internal class MXOlmDevice @Inject constructor(
...
+            val sessionId = restoredOutboundGroupSession.outboundGroupSession.sessionIdentifier()
+            if (!outboundGroupSessionStore.containsKey(sessionId)) {
+                outboundGroupSessionStore[sessionId] = restoredOutboundGroupSession.outboundGroupSession
+                return MXOutboundSessionInfo(
+                        sessionId = sessionId,
+                        sharedWithHelper = SharedWithHelper(roomId, sessionId, store),
+                        restoredOutboundGroupSession.creationTime
+                )
+            } else {
+                restoredOutboundGroupSession.outboundGroupSession.releaseSession()
```
bmarty, line 449: There is something to change here.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
BillCarsonFr: updated

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
BillCarsonFr: updated the code

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXOutboundSessionInfo.kt
BillCarsonFr: reverted

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/RealmCryptoStore.kt
BillCarsonFr: Done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/RealmCryptoStoreMigration.kt
BillCarsonFr: done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
BillCarsonFr: refactored

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
BillCarsonFr: Updated

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -429,33 +428,34 @@ internal class MXOlmDevice @Inject constructor(
...
-                return MXOutboundSessionInfo(
-                        sessionId = sessionId,
-                        sharedWithHelper = SharedWithHelper(roomId, sessionId, store),
-                        restoredOutboundGroupSession.creationTime
-                )
-            } else {
-                restoredOutboundGroupSession.outboundGroupSession.releaseSession()
-            }
+            // cache it
+            outboundGroupSessionCache[sessionId] = restoredOutboundGroupSession.outboundGroupSession
```
bmarty, line 441: Is it OK if `outboundGroupSessionCache[sessionId]` already contains something?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -429,33 +428,34 @@ internal class MXOlmDevice @Inject constructor(
...
+                    restoredOutboundGroupSession.creationTime
+            )
         }
         return null
     }
 
     fun discardOutboundGroupSessionForRoom(roomId: String) {
-        outboundGroupSessionStore.remove(roomId)?.releaseSession()
+        store.getCurrentOutboundGroupSessionForRoom(roomId)?.outboundGroupSession?.sessionIdentifier()?.let { sessionId ->
+            outboundGroupSessionCache.remove(sessionId)?.releaseSession()
```
bmarty, line 454: Do you have to also release what have been returned by `store.getCurrentOutboundGroupSessionForRoom(roomId)`?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
BillCarsonFr: It should not happen (as only called after launch), but it won't break (but the previous session native counter part might not be released.)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -65,10 +65,15 @@ internal class MXOlmDevice @Inject constructor(
     // The OLM lib utility instance.
     private var olmUtility: OlmUtility? = null
 
+    private data class GroupSessionCacheItem(
+            val groupId: String,
```
bmarty, line 69: `roomId`?

PR #2625, opened 2021-01-06
Typo in the set-up instructions for the test homeserver

yquemener:
I think I removed a typo. You probably don't want a new virtualenv every time you start a homeserver?


bmarty:
You are probably right :)

PR #2627, opened 2021-01-06
Edit power level

### LINKED ISSUE #2471: Edit room permissions
AKA power level

## Element Web

image
image

### Expanded

image

## Riot-Android (legacy app)

Cannot find it in the UI, it seems it was not supported...

bmarty:
Fixes #2471

It was already possible to edit the power level event to update the power level of users, but this PR add the possibility to edit the configuration of the room itself. I am reusing the same dialog to edit the level though, I just added a dynamic title.

### New item "Room permissions" in the room settings

image

### When permissions are editable

image

Expanded:

image

Updating something

image

### When permissions are not editable

image

(also possible to show advanced part)

### DIFF
--- a/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
+++ b/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
@@ -247,6 +247,7 @@ class UiAllScreensSanityTest {
 
         // Room settings
         clickListItem(R.id.matrixProfileRecyclerView, 3)
+        navigateToRoomParameters()
         pressBack()
 
         // Notifications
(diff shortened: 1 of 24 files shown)


vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
```diff
@@ -285,6 +286,31 @@ class UiAllScreensSanityTest {
...
+        clickOn(R.string.show_advanced)
+        clickOn(R.string.hide_advanced)
+        pressBack()
+
+        // Room history readability
+        clickListItem(R.id.roomSettingsRecyclerView, 8)
+        pressBack()
+
+        // Room access
+        clickListItem(R.id.roomSettingsRecyclerView, 10)
```
onurays, line 310: Oh, I am surprised with 4, 6, 8, 10... Don't we have a better solution?

onurays (approved):
LGTM! Just one question if we have a chance to improve it.

vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
bmarty: Yes, that's really ugly... Using R.string.x to click on item based on the text does not work well, the list does not scroll automatically. And as we have separator items in the list, it increases 2 by 2...

Issue #2066, opened 2020-09-05
Stickers in sticker picker aren't cached


Bubu:
**Describe the bug**
The stickers (well, I guess images) displayed in the sticker picker aren't cached. Which leads to element android not shwoing any stickers on bad/missing internet connection. And I guess it's redownloading about 10MB of sticker data whenever I want to sent a sticker.

This is using https://github.com/maunium/stickerpicker as a stickerpicker

**To Reproduce**
1. Have https://github.com/maunium/stickerpicker as a stickerpicker
2. Open the stickerpicker, wait for all previews to be downloaded
3. close element android
4. open stickerpicker again and wait for everything to download again

**Expected behavior**
I expected that the stickerpicker widget itself and the sticker preview data is cached

**Smartphone (please complete the following information):**
 - Device: Nexus 5X
 - OS: Android 8.1

**Additional context**
 - App version 1.0.5-F-Droid
 - Homeserver: bubu1.eu

Bubu:
Mmmh, maybe there is some caching. Stickers load faster *sometimes*. But they definitely don't load at all when you are offline.

dkanada:
I can confirm there is zero caching. I found the code that disables all caching for widgets and removed it for my personal builds to make stickers viable. Hopefully I'll have a PR to fix that along with a few more sticker issues soon.

kojid0:
> I can confirm there is zero caching. I found the code that disables all caching for widgets and removed it for my personal builds to make stickers viable. Hopefully I'll have a PR to fix that along with a few more sticker issues soon.

Thank you!

bmarty:
Done in #2648 (Element Android 1.2.1)

PR #2648, opened 2021-01-09
store cache for WidgetWebView between loads

dkanada:
The sticker widget is extremely inefficient due to these lines, reloading every asset on the page any time the sticker widget is opened. I made a new sticker widget and made this change on my personal builds to avoid ten second load times every time I wanted to send a sticker. This should only cause issues for widgets hosted on servers that don't properly invalidate caches as far as I can tell, but the WebView cache can also be manually reset with the `Reset Cache` option in the app settings.


dkanada:
Are there any specific issues with this pull request or is it just a low priority? I'll just close it if there are concerns with other widgets or something.

bmarty:
Thanks for the PR. To be honest, I'm not sure about the impact with other widget, or even with the sticker widget.
@ganfra WDYT?

Bubu:
This would fix #2066 if merged.

bmarty (approved):
Let's merge the PR, I'll add a file for towncrier on develop. Thanks @dkanada !

bmarty:
Done in  8a38f1ca04af7897d354236520802a347ebba2c6

PR #2659, opened 2021-01-12
Fix issue with delay set to 0

bmarty:


ganfra (approved):
LGTM, thanks for this

PR #2661, opened 2021-01-12
Fix request too large Uri error when joining a room

bmarty:
This PR fix the pb of using too many servers when joining a room.
By default we now use what was provided in the permalink (via parameter).

Also the SDK enforces that there is no more than 3 servers in the join request

You can reproduce the current problem with this permalink: https://matrix.to/#/#matrix:matrix.org?via=matrix.org&via=t2bot.io&via=gnome.org


BillCarsonFr (approved):
LGTM

PR #2663, opened 2021-01-13
Fix order of sending/undelivered messages

### LINKED ISSUE #729: Order of sent messages are not always respected
let's queue messages in order (particularly when resending)

Also check when sending multiple images. Reorder can happen, depending on the time to upload images, and it should not.

https://riot.im/develop/#/room/#matrix-internal:matrix.org/$1575047748339386PvDse:matrix.org

onurays:
Fixes #729

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryEventsHelper.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryEventsHelper.kt
@@ -33,11 +33,11 @@ internal object RoomSummaryEventsHelper {
             filterEdits = true
     )
 
-    fun getLatestPreviewableEvent(realm: Realm, roomId: String): TimelineEventEntity? {
+    fun getLatestPreviewableEvent(realm: Realm, roomId: String, includeSending: Boolean = true): TimelineEventEntity? {
         return TimelineEventEntity.latestEvent(
                 realm = realm,
                 roomId = roomId,
-                includesSending = true,
+                includesSending = includeSending,
                 filters = previewFilters
         )
     }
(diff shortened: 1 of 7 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryEventsHelper.kt
```diff
@@ -33,11 +33,11 @@ internal object RoomSummaryEventsHelper {
             filterEdits = true
     )
 
-    fun getLatestPreviewableEvent(realm: Realm, roomId: String): TimelineEventEntity? {
+    fun getLatestPreviewableEvent(realm: Realm, roomId: String, includeSending: Boolean = true): TimelineEventEntity? {
```
ganfra, line 36: Please don't change this method. If you don't need sending then this is not the latestPreviewable...

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
```diff
@@ -408,7 +408,7 @@ internal class DefaultTimeline(
     }
 
     private fun createSnapshot(): List<TimelineEvent> {
-        return buildSendingEvents() + builtEvents.toList()
+        return (buildSendingEvents() + builtEvents.toList()).sortedByDescending { it.root.originServerTs }
```
ganfra, line 411: We shouldn't rely on ts to sort events

ganfra (changes requested):

ganfra:
I don't understand exactly which case you try to fix here

ShadowJonathan:
Copy-paste from #729:

> When i was poking around in the internals, it seems to be that element has a global send queue, maybe for this initialisation it's better if there's an ad-hoc queue per room? To make sure events sent in other rooms dont "block" on a large file sent earlier in one room.

bmarty:
Stale PR, I close it, let's try to fix the original issue in another way

PR #2674, opened 2021-01-14
Init sync optimization

### LINKED ISSUE #983: Stuck on initial sync
Using latest Android build from play store. After signing in with a matrix.org account I am stuck on an endless "Initial Sync: Importing account" 
SmartSelect_20200211-145358_RiotX

bmarty:
Fixes #983 

New strategy to handle initial sync: First store the request result into a file, to avoid doing it again in case of crash during treatment. Then perform a partial parsing for big account to avoid too much RAM usage.

Performance change is not significant for small account, it take a bit more time for big account but it consume much less RAM, so there is a better chance to finally finish the sync.

This PR also improve `DefaultInitialSyncProgressService` and fix a few bug on it


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncTask.kt
```diff
@@ -81,20 +91,289 @@ internal class DefaultSyncTask @Inject constructor(
...
+
+            if (syncResponse.isSuccessful) {
+                Timber.v("INIT_SYNC request successful, download and save to file")
+                syncResponse.body()?.byteStream()?.use { inputStream ->
+                    workingFile.outputStream().use { outputStream ->
+                        inputStream.copyTo(outputStream)
+                    }
+                    Timber.v("INIT_SYNC safeInitialSync() end copy")
+
+                    handleSyncFile(workingFile)
```
bmarty, line 133: move one line below

BillCarsonFr (approved):

PR #2680, opened 2021-01-15
Dagger: update to 2.31 and use new AssistedInject

ganfra:


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/alias/DefaultAliasService.kt
```diff
@@ -26,9 +27,9 @@ internal class DefaultAliasService @AssistedInject constructor(
         private val addRoomAliasTask: AddRoomAliasTask
 ) : AliasService {
 
-    @AssistedInject.Factory
+    @AssistedFactory
     interface Factory {
-        fun create(roomId: String): AliasService
+        fun create(roomId: String): DefaultAliasService
```
bmarty, line 32: We are now returning the implementing type?

ganfra: Yes no choice, but it doesn't have any impact

PR #2694, opened 2021-01-18
Render markdown in room list

### LINKED ISSUE #452: Render markdown in chat previews
When previewing the previous message from the direct messages tab, markdown is not rendered, resulting in previews such as:

```
Hello **there**
```

onurays:
Fixes #452 

markdown_in_room_list


TR-SLimey:
I haven't checked so I don't know if that's the case, but if markdown links are also rendered, they should probably not be clickable in the room list, because clicking on the chat can make you accidentally click on a link.

bmarty (commented):
some remarks.
Also what about MSGTYPE_EMOTE ? This is maybe another bug, but you could take the opportunity to also handle emote properly here.

bmarty:
Will iterate on it on my own branch

PR #2701, opened 2021-01-20
Fix clear cache issue: sometimes, after a clear cache, there is still a token, so the init sync service is not started.

bmarty:
Migratiing to supend methods fixes the problem, I'm not sure why...

### DIFF
--- a/vector/src/main/java/im/vector/app/features/MainActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/MainActivity.kt
@@ -175,12 +175,10 @@ class MainActivity : VectorBaseActivity<FragmentLoadingBinding>(), UnlockedActiv
                 emptyViewModel.viewModelScope.launch {
                     try {
                         session.clearCache()
-                        Timber.e("CACHE success")
                         doLocalCleanup(clearPreferences = false)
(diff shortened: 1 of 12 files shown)


vector/src/main/java/im/vector/app/features/MainActivity.kt
```diff
@@ -147,38 +150,41 @@ class MainActivity : VectorBaseActivity<FragmentLoadingBinding>(), UnlockedActiv
         }
         when {
             args.isAccountDeactivated -> {
-                // Just do the local cleanup
-                Timber.w("Account deactivated, start app")
-                sessionHolder.clearActiveSession()
-                doLocalCleanup(clearPreferences = true)
-                startNextActivityAndFinish()
+                emptyViewModel.viewModelScope.launch {
```
ganfra, line 153: Why not dispatching to a real viewmodel?

ganfra (approved):
One remark otherwise LGTM

vector/src/main/java/im/vector/app/features/MainActivity.kt
bmarty: This is another rework... But yes, ideally it will have to be done

Issue #2708, opened 2021-01-22
"Discovery" screen doesn't always show shared emails


AndrewFerr:
**Describe the bug**
When viewing Settings->General->Discovery after not having viewed it for a while, shared emails will appear as unshared. Exiting the page and immediately re-entering it fixes the problem.

This happens is because of mishandling of expired peppers for the identity server. When a lookup request to an identity server responds with `M_INVALID_PEPPER`, Element Android correctly asks the server for a new pepper, and retries the lookup. However, the retry uses the original, invalid pepper instead of the newly-retrieved one.

**To Reproduce**
1. With any client, add some email addresses to your Matrix account, connect to an identity server, and mark some of those email addresses as shared
2. Wait a few hours
3. In Element Android, view Settings->General->Discovery
4. Notice that "Discoverable email addresses" shows none of your email addresses as shared
5. Exit and re-enter the Discovery page
6. Notice that your shared email addresses now properly appear as shared

**Expected behavior**
Shared emails (and any other 3PIDs) should appear as shared without having to exit and re-enter the Discovery page.

**Additional context**
 - 1.0.13 - F-Droid and 1.0.15 from develop branch
 - Homeserver: self-hosted homeserver
 - Identity server: self-hosted ma1sd

bmarty:
You open an issue, then you fix it in a PR, wonderful! Thanks a lot!

I'm glad I set up a `canRetry: Boolean` parameter to avoid infinite loop due to this bug!

bmarty:
Fixed in #2709 so in Element Android 1.0.15

PR #2709, opened 2021-01-22
Identity: Recompute hashes after M_INVALID_PEPPER

AndrewFerr:
When a new pepper is retrieved after an M_INVALID_PEPPER response,
recompute hashes with that pepper, and send those new hashes in the next
lookup attempt instead of reusing the original hashes.

### Pull Request Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Improvements 🙌:
 Bugfix 🐛:
  - Fix clear cache issue: sometimes, after a clear cache, there is still a token, so the init sync service is not started.
  - Sidebar too large in horizontal orientation or tablets (#475)
+ - When receiving a new pepper from identity server, use it on the next hash lookup (#2708)
 
 Translations 🗣:
  -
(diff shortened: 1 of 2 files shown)


AndrewFerr:
Fixes #2708

bmarty (approved):
Thanks a lot for this fix!
I'm wondering if there is an other issue in file `IdentityBulkLookupTask.kt` at line 110 (on your branch). We should use `newHashDetailResponse` instead of `hashDetailResponse`. The comment on line 112 (`// Also, what we have in cache is maybe outdated, the identity server maybe now support sha256`) could then be removed. WDYT?

bmarty:
> I'm wondering if there is an other issue in file `IdentityBulkLookupTask.kt` at line 110 (on your branch). We should use `newHashDetailResponse` instead of `hashDetailResponse`. The comment on line 112 (`// Also, what we have in cache is maybe outdated, the identity server maybe now support sha256`) could then be removed. WDYT?

I fixed that on #2714

AndrewFerr:
> > I'm wondering if there is an other issue in file `IdentityBulkLookupTask.kt` at line 110 (on your branch). We should use `newHashDetailResponse` instead of `hashDetailResponse`. The comment on line 112 (`// Also, what we have in cache is maybe outdated, the identity server maybe now support sha256`) could then be removed. WDYT?
> 
> I fixed that on #2714

Thanks for the follow-up fixes! I didn't know the history behind those comments; it was like that when I got there, and the only issue I could find was the one that I fixed. Glad that other fixes were possible too :)

PR #2727, opened 2021-01-27
Feature/fga/voip fix audio

ganfra:
This PR is an attempt to fix issues with audio devices. It also fix a bunch of other issues while testing with multiple android devices.


BillCarsonFr (approved):
lgtm

PR #2777, opened 2021-02-04
Fix duplicate thumbs reactions

### LINKED ISSUE #2776: Duplicate thumbs | Mobile reactions for 👍 and 👎 are not the same as web
image

see https://github.com/matrix-org/matrix-appservice-slack/issues/488

> Looks like Element appends 239 184 143 VARIATION SELECTOR-16 after the thumbs up and down.
https://www.utf8-chartable.de/unicode-utf8-table.pl

BillCarsonFr:
Fixes #2776 

Mobile:
image

Web:
image


bmarty:
This is OK, but I think the file `emoji_picker_datasource.json` has also to be updated.
If you add a "thumb up" reaction from the emoji picker (using `EmojiReactionPickerActivity`), the bug will still be there (not tested though).

Also there is also maybe some other bug due to this 2 sources of truth.

PR #2779, opened 2021-02-04
Join room other federation via alias error

### LINKED ISSUE #2778: Join room by alias other federation error
Step to reproduce:

From a matrix.org account try to join this room #cloud-devroom_testing4:fosdem.org

=> android displays a "This room is not accessible at this time", and doesn't give you an option to try and join?

BillCarsonFr:
Fixes #2778
Peek room task was not catching enough failures (of child tasks call)

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/peeking/PeekRoomTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/peeking/PeekRoomTask.kt
@@ -17,6 +17,7 @@
 package org.matrix.android.sdk.internal.session.room.peeking
 
 import org.matrix.android.sdk.api.MatrixPatterns
+import org.matrix.android.sdk.api.extensions.tryOrNull
 import org.matrix.android.sdk.api.session.events.model.EventType
 import org.matrix.android.sdk.api.session.events.model.toModel
 import org.matrix.android.sdk.api.session.room.model.RoomAvatarContent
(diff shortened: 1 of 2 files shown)


ganfra (approved):
LGTM, thanks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/peeking/PeekRoomTask.kt
```diff
@@ -65,23 +66,29 @@ internal class DefaultPeekRoomTask @Inject constructor(
         }
 
         // Is it a public room?
-        val publicRepoResult = when (getRoomDirectoryVisibilityTask.execute(GetRoomDirectoryVisibilityTask.Params(roomId))) {
-            RoomDirectoryVisibility.PRIVATE -> {
-                // We cannot resolve this room :/
-                null
-            }
-            RoomDirectoryVisibility.PUBLIC  -> {
+        val visibilityRes = tryOrNull("## PEEK: failed to get visibility") {
```
bmarty, line 69: I'm not a big fan of using `tryOrNull {}` around tasks which make network request. If there is no network, there is no need to go further.

ganfra: We don't have good mechanism to be sure we can reach the server, so how do you want to handle this?

bmarty: No network: special Exception. Getting 404 is something else for instance

PR #2781, opened 2021-02-05
Fix a bunch of bugs and feature parity for Widget

bmarty:
Especially joining a Jitsi conf created by ElementWeb does not work since the Jitsi URL does not contain the `confId` parameter, it's only in the fragment.

Also properly encode (and decode) parameters in the fragment.

Support for `$theme` template

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/widgets/helper/WidgetFactory.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/widgets/helper/WidgetFactory.kt
@@ -70,7 +70,7 @@ internal class WidgetFactory @Inject constructor(private val userDataSource: Use
         var computedUrl = url ?: return null
         val myUser = userDataSource.getUser(userId)
 
-        val keyValue = data.mapKeys { "\$$it" }.toMutableMap()
+        val keyValue = data.mapKeys { "\$${it.key}" }.toMutableMap()
 
         keyValue[WIDGET_PATTERN_MATRIX_USER_ID] = userId
         keyValue[WIDGET_PATTERN_MATRIX_DISPLAY_NAME] = myUser?.getBestName() ?: userId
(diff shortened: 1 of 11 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/widgets/helper/WidgetFactory.kt
```diff
@@ -65,23 +65,31 @@ internal class WidgetFactory @Inject constructor(private val userDataSource: Use
...
+
+        for ((key, value) in keyValue) {
+            computedUrl = computedUrl.replace(key, URLEncoder.encode(value.toString(), "utf-8"))
         }
         return computedUrl
     }
+
+    companion object {
+        // Value to be replaced in URLS
+        const val WIDGET_PATTERN_MATRIX_USER_ID = "\$matrix_user_id"
```
ganfra, line 89: private?

vector/src/main/java/im/vector/app/features/call/conference/VectorJitsiActivity.kt
```diff
@@ -154,13 +155,19 @@ class VectorJitsiActivity : VectorBaseActivity<ActivityJitsiBinding>(), JitsiMee
     }
 
     override fun onConferenceTerminated(p0: MutableMap<String, Any>?) {
-        finish()
+        Timber.v("JitsiMeetViewListener.onConferenceTerminated()")
+        // Do not finish if there is an error
+        if (p0?.get("error") == null) {
```
ganfra, line 160: The jisti view already show an error?  otherwise, we should maybe show one

ganfra (approved):
Good work, thanks for that, just see small comments

vector/src/main/java/im/vector/app/features/call/conference/VectorJitsiActivity.kt
bmarty: Yes, it shows an error... But we cannot be a hundred percent sure about that for all cases... So I do not know. This could be improved in the future, there are other useful information than we could display: someone's joined the conf, someone left it, etc.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/widgets/helper/WidgetFactory.kt
bmarty: Yes, this is in the `internal` package, so it should be moved to `api` package, because I wanted to use those `const` in `RoomDetailViewModel`. We should maybe expose an API to build such URL and not let the app do it. IDK

PR #2786, opened 2021-02-05
Feature/bma/rtl

### LINKED ISSUE #2757: RTL: some arrows should be rotated in RTL
Switch to a RTL language, ex Arabic

Observe than some arrow are not on the correct direction.

Use `android:rotationY="@integer/rtl_mirror_flip"` to fix the pb

Seen in the main drawer, and in the settings. Once a drawable is identified for the fix to apply, please check that the fix is done everywhere this drawable is used.

bmarty:
Fix #2757

## Add  by QR code icon

(ignore new translation)

Before:

image

After:

image

## Right arrow

Before

image

image

image

After

image

image

image

And many other places

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Improvements 🙌:
 Bugfix 🐛:
  - Duplicate thumbs | Mobile reactions for 👍 and 👎 are not the same as web (#2776)
  - Join room by alias other federation error (#2778)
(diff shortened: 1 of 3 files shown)


ganfra (approved):
LGTM
Did you check all screens?

bmarty:
Yes, there are maybe other issues, but the pb with arrow was quite annoying.

PR #2790, opened 2021-02-07
Create a WidgetItemFactory and use it for better rendering of Jitsi widget change (video conference)

bmarty:
Parity with Element Web

## Before

image

## After

image


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/WidgetItemFactory.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    fun create(event: TimelineEvent,
+               highlight: Boolean,
+               callback: TimelineEventController.Callback?): VectorEpoxyModel<*>? {
+        val widgetContent: WidgetContent = event.root.getClearContent().toModel() ?: return null
+        val previousWidgetContent: WidgetContent? = event.root.resolvedPrevContent().toModel()
+
+        return when (WidgetType.fromString(widgetContent.type ?: previousWidgetContent?.type ?: "")) {
+            WidgetType.Jitsi       -> createJitsiItem(event, callback, widgetContent, previousWidgetContent)
+            WidgetType.TradingView,
```
ganfra, line 60: use else instead?

ganfra (approved):
One small remark

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/WidgetItemFactory.kt
bmarty: Yes, it was to show all the cases we could improve... I will change to `else`

bmarty: Done

PR #2793, opened 2021-02-07
Url preview fixes

### LINKED ISSUE #2705: Allow viewing images from url previews in attachement viewer
When seeing an url preview which includes an image (potentially it's just a link to an image) then it would be helpful to be able to view the fullscreen version of that image in element instead of having to click through and open it in a browser.

### LINKED ISSUE #2756: URL preview on reply fallback
- send a message with a URL
- reply to this message with no URL in the reply
- a URL preview is displayed under the reply, because of the URL in the original message

Replied message fallback content should be stripped before looking for URL for a preview.

### LINKED ISSUE #2766: HTLM unescaping
Send a message with this URL: ` https://github.com/matrix-org/synapse/issues/9304`

Observe that HTML chars are not un-escaped from what the server has returned:

image

image

bmarty:
Fixes #2756 
Fixes #2705 
Fixes #2766


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/media/UrlsExtractor.kt
```diff
@@ -35,7 +36,14 @@ internal class UrlsExtractor @Inject constructor() {
                             || it.msgType == MessageType.MSGTYPE_NOTICE
                             || it.msgType == MessageType.MSGTYPE_EMOTE
                 }
-                ?.body
+                ?.let { messageContent ->
+                    if (messageContent.relatesTo?.inReplyTo?.eventId != null) {
```
BillCarsonFr, line 40: Not sure if relevant but usually i use getRelationContent() extension, it handles encrypted relation

BillCarsonFr (approved):
LGTM, just a small remark

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/media/UrlsExtractor.kt
bmarty: OK, thanks, I will check

PR #2807, opened 2021-02-10
fix post initial sync crash on 404

BillCarsonFr:
Dendrite | fix crash after initial sync (regression from SSO UIA work)


bmarty (commented):
I trust you on the fact that it will fix the bug, but I can't figure out why it is crashing.

PR #2808, opened 2021-02-10
Room settings ux

bmarty:
A few changes and improvements in the UX for room settings

- Move "Room addresses" and "Room permissions" entry point from the screen where user can update basic states of the room to the advanced section of the room profile. This is mainly because it's better for non-tech user, and also the room setting screen is a form where all changes are committed at once, so, for the UX it was not ideal to have entry point for other screens here.
- Add some click listener to open specific screens from the settings.


onurays (approved):
LGTM!

PR #2408, opened 2020-11-16
Fetch my user profile when my room member changed.

onurays:
Fixes #1715

PR #2816, opened 2021-02-11
Feature/bma/user name

### LINKED ISSUE #1715: myroomnick changes the Global name
Hello,

I changed the nick using /myroomnick in a room, and Element-Android changed my global nick. Element on the mobile is showing the roomnick while Element-Web is correctly showing the previous global nick. 

Element version 1.0.0

Thanks

bmarty:
Fixes #1715

Replace #2408 

The strategy here is to store all the used display name of a user for all rooms.
So we maintain a counter for each display name and then we select the display name which is the most used
as room member name for the user display name.
We make the assumption that the user will not update display name per room (using /myroomnick) for a big number of rooms.

The same algorithm is used for avatarUrl.

It requires an init sync to properly update the database.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/UserEntityFactory.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/UserEntityFactory.kt
@@ -16,16 +16,81 @@
 
 package org.matrix.android.sdk.internal.session.user
 
+import io.realm.Realm
+import io.realm.kotlin.createObject
 import org.matrix.android.sdk.api.session.room.model.RoomMemberContent
+import org.matrix.android.sdk.internal.database.model.StringCountedEntity
 import org.matrix.android.sdk.internal.database.model.UserEntity
+import org.matrix.android.sdk.internal.database.query.where
 
 internal object UserEntityFactory {
 
-    fun create(userId: String, roomMember: RoomMemberContent): UserEntity {
-        return UserEntity(
-                userId = userId,
-                displayName = roomMember.displayName ?: "",
-                avatarUrl = roomMember.avatarUrl ?: ""
-        )
+    /**
+     * The strategy here is to store all the used display name of a user for all rooms.
+     * So we maintain a counter for each display name and then we select the display name which is the most used
+     * as room member name for the user display name.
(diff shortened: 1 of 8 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/UserEntityFactory.kt
```diff
@@ -16,16 +16,81 @@
...
+                       previousRoomMember: RoomMemberContent?) {
+        val entity = UserEntity.where(realm, userId).findFirst() ?: realm.createObject(userId)
+
+        // Display Name
+        val newDisplayName = roomMember.displayName ?: ""
+        val previousDisplayName = previousRoomMember?.displayName
+
+        if (newDisplayName != previousDisplayName) {
+            if (previousDisplayName != null) {
+                entity.displayNameInRoom.indexOfFirst { it.value == previousDisplayName }
```
ganfra, line 48: I would have made a realm query here. And it's a bit hard to read

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/UserEntityFactory.kt
```diff
@@ -16,16 +16,81 @@
...
+        }
+
+        // Avatar
+        val newAvatar = roomMember.avatarUrl ?: ""
+        val previousAvatar = previousRoomMember?.avatarUrl
+
+        if (newAvatar != previousAvatar) {
+            if (previousAvatar != null) {
+                entity.avatarUrlInRoom.indexOfFirst { it.value == previousAvatar }
+                        .takeIf { it != -1 }
```
ganfra, line 76: Same ^

ganfra (commented):
One remark
Also, it's just the sdk part, I don't see the mapping of UserEntity to User and his usage with a "preferredAvatarUrl" and "preferredDisplayName"?

bmarty:
As discussed IRL, @ganfra will propose a third solution to handle this pb:
- Aggregate things to do during the sync response treatment
- Perform at the end of the sync treatment the "things to do"

For this particular case, we will keep a list of userId with room member ship change, and at the end of the sync update the UserEntity database by performing request on the room member ship for each rooms for those users.

But this strategy could be extended to other (existing?) treatments, to avoid iterating on the sync response several times.

PR #2825, opened 2021-02-15
Color resources cleanup

bmarty:
Start cleaning color resources.

Also removed some unused resources.


bmarty:
Will not be merged, but will be splitted in several PRs (first one: #2828)

PR #2831, opened 2021-02-16
Fetch homeserver type and version and display in a new wetting screen and add info in rageshakes

bmarty:
### Rageshake

image

### Setting screen

image

### Dark

image

### DIFF
--- a/vector/src/main/java/im/vector/app/features/settings/homeserver/HomeserverSettingsController.kt
+++ b/vector/src/main/java/im/vector/app/features/settings/homeserver/HomeserverSettingsController.kt
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2021 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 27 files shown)


vector/src/main/java/im/vector/app/features/settings/homeserver/HomeserverSettingsController.kt
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        val limit = data.homeServerCapabilities.maxUploadFileSize
+
+        settingsInfoItem {
+            id("uploadValue")
+            if (limit == HomeServerCapabilities.MAX_UPLOAD_FILE_SIZE_UNKNOWN) {
+                helperTextResId(R.string.settings_server_upload_size_unknown)
+            } else {
+                helperText(stringProvider.getString(R.string.settings_server_upload_size_content, "${limit / 1048576L} MB"))
```
onurays, line 116: It seems like a magic number at first glance. Can we use something like `1024.0.pow(2).toLong()` to avoid this? :)

onurays (approved):
LGTM! Just one simple remark.

PR #2844, opened 2021-02-18
Fix account deactivation crash

bmarty:
The cleanup of the session was happening after each request to deactivate the account, so a bit too much...

This PR also contains some cleanup and documentation, to improve code understandability.


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/account/DeactivateAccountTask.kt
```diff
@@ -30,40 +29,49 @@ import javax.inject.Inject
...
+                            retryBlock = { authUpdate ->
+                                execute(params.copy(userAuthParam = authUpdate))
+                            }
+                    )
             ) {
                 Timber.d("## UIA: propagate failure")
-                throw  throwable
+                throw throwable
+            } else {
+                false
```
bmarty, line 65: The fix is here

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/account/DeactivateAccountTest.kt
```diff
@@ -46,12 +46,13 @@ class DeactivateAccountTest : InstrumentedTest {
 
     @Test
     fun deactivateAccountTest() {
-        val session = commonTestHelper.createAccount(TestConstants.USER_ALICE, SessionTestParams(withInitialSync = false))
+        val session = commonTestHelper.createAccount(TestConstants.USER_ALICE, SessionTestParams(withInitialSync = true))
```
bmarty, line 49: This is just to fix the test

BillCarsonFr (approved):
LGTM, thx for the log good catch

PR #2848, opened 2021-02-19
Add subscribeToChanges/unsubscribeToChanges to VectorPreferences

neilalexander:
This is useful for the P2P demo as we need to subscribe to configuration changes from a background service. The `Service` can implement `SharedPreferences.OnSharedPreferenceChangeListener` and override `onSharedPreferenceChanged` to receive notifications.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/settings/VectorPreferences.kt
+++ b/vector/src/main/java/im/vector/app/features/settings/VectorPreferences.kt
@@ -239,6 +239,20 @@ class VectorPreferences @Inject constructor(private val context: Context) {
     }
 
     private val defaultPrefs = DefaultSharedPreferences.getInstance(context)
+    
+    /**
+     * Allow subscribing and unsubscribing to configuration changes. This is
+     * particularly useful when you need to be notified of a configuration change
+     * in a background service, e.g. for the P2P demos.
+     */
+    
+    public fun subscribeToChanges(listener: SharedPreferences.OnSharedPreferenceChangeListener) {
+        defaultPrefs.registerOnSharedPreferenceChangeListener(listener)
+    }
+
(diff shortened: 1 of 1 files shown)


vector/src/main/java/im/vector/app/features/settings/VectorPreferences.kt
```diff
@@ -239,6 +239,20 @@ class VectorPreferences @Inject constructor(private val context: Context) {
...
 
     private val defaultPrefs = DefaultSharedPreferences.getInstance(context)
+    
+    /**
+     * Allow subscribing and unsubscribing to configuration changes. This is
+     * particularly useful when you need to be notified of a configuration change
+     * in a background service, e.g. for the P2P demos.
+     */
+    
+    public fun subscribeToChanges(listener: SharedPreferences.OnSharedPreferenceChangeListener) {
```
bmarty, line 249: `import android.content.SharedPreferences` is missing

Also `public` modifier can be removed, this is the default value in Kotlin

vector/src/main/java/im/vector/app/features/settings/VectorPreferences.kt
```diff
@@ -239,6 +239,20 @@ class VectorPreferences @Inject constructor(private val context: Context) {
...
+        defaultPrefs.registerOnSharedPreferenceChangeListener(listener)
+    }
+
+    public fun unsubscribeToChanges(listener: SharedPreferences.OnSharedPreferenceChangeListener) {
```
bmarty, line 253: same remark about `public`

bmarty (changes requested):
The code does not compile :)

bmarty (approved):
Thanks for the update. I will fix the issue with ktlint on develop

PR #2908, opened 2021-02-23
PIP support for Jitsi call

### LINKED ISSUE #2418: Can't minimise jitsi call in Android 10?
Apparently in Android 10 nothing is shown in the upper left corner to minimise a Jitsi call

bmarty:
Fixes #2418

Also handle switch to another conference.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/call/conference/VectorJitsiActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/conference/VectorJitsiActivity.kt
@@ -80,6 +80,12 @@ class VectorJitsiActivity : VectorBaseActivity<ActivityJitsiBinding>(), JitsiMee
             renderState(it)
         }
 
+        jitsiViewModel.observeViewEvents {
+            when(it) {
+                is JitsiCallViewEvents.StartConference -> configureJitsiView(it)
+            }
+        }
+
         registerForBroadcastMessages()
     }
 
@@ -96,7 +102,6 @@ class VectorJitsiActivity : VectorBaseActivity<ActivityJitsiBinding>(), JitsiMee
             is Success -> {
                 views.jitsiProgressLayout.isVisible = false
                 jitsiMeetView?.isVisible = true
-                configureJitsiView(viewState)
             }
             else       -> {
                 jitsiMeetView?.isVisible = false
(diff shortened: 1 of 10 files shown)


vector/src/main/java/im/vector/app/features/call/conference/VectorJitsiActivity.kt
```diff
@@ -147,13 +181,23 @@ class VectorJitsiActivity : VectorBaseActivity<ActivityJitsiBinding>(), JitsiMee
...
     }
 
-//    override fun onUserLeaveHint() {
-//        super.onUserLeaveHint()
-//        jitsiMeetView?.enterPictureInPicture()
-//    }
+    override fun onUserLeaveHint() {
+        super.onUserLeaveHint()
+        if (packageManager.hasSystemFeature(PackageManager.FEATURE_PICTURE_IN_PICTURE)) {
+            jitsiMeetView?.enterPictureInPicture()
```
ganfra, line 187: I think compat is already handled by Jitsi

vector/src/main/java/im/vector/app/features/call/conference/VectorJitsiActivity.kt
```diff
@@ -147,13 +181,23 @@ class VectorJitsiActivity : VectorBaseActivity<ActivityJitsiBinding>(), JitsiMee
         super.onDestroy()
     }
 
-//    override fun onUserLeaveHint() {
-//        super.onUserLeaveHint()
-//        jitsiMeetView?.enterPictureInPicture()
-//    }
+    override fun onUserLeaveHint() {
```
ganfra, line 184: Also, I think this method is only called when pressing home button, what about pressing back?

ganfra (commented):
2 small remarks

vector/src/main/java/im/vector/app/features/call/conference/VectorJitsiActivity.kt
bmarty: Yes, but better safe than sorry :)

vector/src/main/java/im/vector/app/features/call/conference/VectorJitsiActivity.kt
bmarty: Yes, if you press back, you leave the conference, and I think it is OK, no?

ganfra: I don't know, generally I want to leave only when explicitly pressing hang-up

ganfra (approved):
LGTM, thanks

PR #2963, opened 2021-03-04
Improve init sync performance

bmarty:
This PR reduce the delay to see the room list and be able to open a room and send a message to it.
On my account (about 300 rooms) I can see the room list after 55 seconds. With this PR, it is reduced to 35 seconds, including 12 seconds to let the server compute the response and the client to download the data.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/sync/FilterService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/sync/FilterService.kt
@@ -22,9 +22,9 @@ interface FilterService {
         NoFilter,
 
         /**
-         * Filter for Riot, will include only known event type
+         * Filter for Element, will include only known event type
          */
-        RiotFilter
+        ElementFilter
     }
 
     /**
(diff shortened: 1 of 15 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/sync/FilterService.kt
```diff
@@ -22,9 +22,9 @@ interface FilterService {
         NoFilter,
 
         /**
-         * Filter for Riot, will include only known event type
+         * Filter for Element, will include only known event type
          */
-        RiotFilter
+        ElementFilter
```
ganfra, line 27: should we not remove this from SDK?  and all stuff related to Element filter ofc

ganfra (approved):
One remark, it can be done later, but I guess it would be better, as it's a bit weird now

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/sync/FilterService.kt
bmarty: Yes, probably.

bmarty:
Thanks @ganfra for the review

Issue #2752, opened 2021-02-02
Wrong language code zh_Hans and zh_Hant in fastlane


linsui:
**Describe the bug**
Wrong language code zh_Hans and zh_Hant in fastlane

**To Reproduce**
1. Go to https://github.com/vector-im/element-android/tree/develop/fastlane/metadata/android

**Expected behavior**
Many other apps use `zh-CN` and `zh-TW`/`zh-HK` and it seems that `zh-Hans` and `zh-Hant` are [suppoeted officially](https://github.com/fastlane/fastlane/blob/633860a5aae0fa9c0b936797560efb24181b1bf7/fastlane_core/lib/fastlane_core/languages.rb#L14). But `zh_Hans` and `zh_Hant` are used in this repo. I thought that this is the reason the Chinese translation of the fastlane is not shown in F-Droid client.

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: [e.g. Samsung S6]
 - OS: [e.g. Android 6.0]

**Additional context**
 - App version and store [e.g. 1.0.0 - F-Droid]
 - Homeserver: [e.g. matrix.org]

eighthave:
[Fastlane](https://github.com/fastlane/fastlane/blob/633860a5aae0fa9c0b936797560efb24181b1bf7/fastlane_core/lib/fastlane_core/languages.rb#L14) uses [BCP47](https://www.w3.org/International/articles/bcp47/) locale tags, which use a `-` as the separator rather than `_`.    There are a number of locales that use the `_` here.  Please rename those dirs to use `-`.

@linsui FYI https://gitlab.com/fdroid/issuebot/-/issues/33#note_522152297

ara4n:
So the fix for this is just to rename `zh-Hans` and `zh-Hant` as `zh_Hans` and `zh_Hant` throughout?

linsui:
Not only zh-Hans and zh-Hant but also nb-NO and pt-BR. Thanks!

linsui:
I guess this is just forgotten.

PR #2964, opened 2021-03-05
rename zh-Hans and zh-Hant - TO BE MERGED DURING WEBLATE SYNC

ara4n:
as per https://github.com/vector-im/element-android/issues/2752

Want to get review from zh-experts before merging this though.


eighthave:
Looks good!

eighthave:
@linsui is who alerted us to this issue, they're the F-Droid `zh-Hans-CN` translator.  You can see clearly the tags that Google Play requires here: https://github.com/fastlane/fastlane/blob/633860a5aae0fa9c0b936797560efb24181b1bf7/fastlane_core/lib/fastlane_core/languages.rb#L14

eighthave:
You can see clearly that Google Play uses the dash `-` rather than `_`, but oddly, fastlane and Google Play seem to differ on the `zh` variants: https://support.google.com/googleplay/android-developer/table/4419860?hl=en

bmarty:
Those files has been created by Weblate, so not sure if merging that will not mess up the next Weblate sync.
We should probably wait for https://github.com/WeblateOrg/weblate/issues/5522 to be fixed and released...

eighthave:
I've already done some of these kinds of moves with projects in Weblate, it 
works fine.  Weblate is based on git, so it seems to inherit git's smooth 
handling of file renames.

eighthave:
Also, the current names don't work properly with the `fastlane` tool nor 
F-Droid, so those trnaslations are already brken.

bmarty:
Yes, it's broken, but what I want to avoid is having conflict when pulling from Weblate. I will merge this PR during the Weblate sync i.e. when Weblate is locked.

Issue #2449, opened 2020-11-26
[SDK] Migrate Services API from MatrixCallback mode to coroutines


bmarty:
This work has already been started by @Dominaezzz , and this issue is here to track the progress and the remaining works.

Note: every new API should now use `suspend` functions instread of `MatrixCallback`

## List of services

- [ ] AccountDataService
- [ ] AccountService
- [x] AliasService (new service #2439)
- [ ] AuthenticationService
- [ ] CacheService
- [ ] CallSignalingService
- [ ] CrossSigningService
- [ ] CryptoService
- [x] DraftService (#2405)
- [ ] FileService
- [ ] FilterService
- [x] Group (note: to rename) (#2413)
- [x] GroupService (nothing to do)
- [ ] HomeServerCapabilitiesService
- [ ] IdentityService
- [ ] InitialSyncProgressService
- [ ] IntegrationManagerService
- [ ] KeysBackupService
- [ ] MembershipService
- [ ] PermalinkService
- [ ] ProfileService
- [x] PushRuleService (#2414)
- [ ] PushersService
- [x] RawService (#2406)
- [ ] ReadService
- [ ] RelationService
- [x] ReportingService (#2402)
- [ ] RoomCallService
- [x] RoomCryptoService (#2376)
- [ ] RoomDirectoryService
- [x] RoomPushRuleService (#2382)
- [ ] RoomService
- [x] SearchService (#2415)
- [ ] SecureStorageService
- [ ] SendService
- [ ] SharedSecretStorageService
- [ ] SignOutService
- [ ] StateService
- [ ] SyncService
- [x] TagsService (#2401)
- [ ] TermsService
- [ ] TimelineService
- [ ] TypingService
- [ ] UploadsService
- [ ] UserService
- [ ] VerificationService
- [ ] WidgetService

Dominaezzz:
Currently working on these.
- `ReadService`
- `UserService`
- `CallSignalingService`
- `AuthenticationService`

bmarty:
CallSignalingService (1 method to migrate) is in progress in #2830

Dominaezzz:
Looking at `IdentityService`.

Dominaezzz:
Looking at `RelationService`.

Dominaezzz:
You can tick `StateService` now btw.

Dominaezzz:
Should also make an issue for `LiveData` -> `(State)Flow`.
I'm done making PRs for now.

[...]

PR #2975, opened 2021-03-08
Basic filtering for "nsfw" keyword

bmarty:
New settings, in Settings/Preferences screen, off by default:

(Edit: wording has been updated, see new screenshot below)
image

Filtering room directory

image

Also migrate RoomDirectoryService to coroutines (#2449)

### DIFF
--- a/vector/src/main/java/im/vector/app/features/roomdirectory/RoomDirectoryViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/roomdirectory/RoomDirectoryViewModel.kt
@@ -203,7 +203,7 @@ class RoomDirectoryViewModel @AssistedInject constructor(
             val newPublicRooms = data.chunk.orEmpty()
                     .filter {
                         showAllRooms
-                                || "${it.name} ${it.topic}".toLowerCase(Locale.ROOT)
+                                || "${it.name.orEmpty()} ${it.topic.orEmpty()} ${it.canonicalAlias.orEmpty()}".toLowerCase(Locale.ROOT)
                                 .let { str ->
                                     explicitContentTerms.all { term -> term !in str }
                                 }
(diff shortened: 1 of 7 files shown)


SimonBrandner:
May I have two questions here? 
+ I _think_ "Filter NSFW rooms" might be easier to find. 
+ How exactly does the filtering work? The Blender room, for example, has _No NSFW Media_ in its name. Would it be hidden then?

bmarty:
>     * I _think_ _Filter NSFW rooms_ might be easier to find.

"NSFW" is not understandable enough for non power users.

>     * How exactly does the filtering work? The Blender room, for example, has _No NSFW Media_ in its name. Would it be hidden then?

Yes, I have seen the problem with the Blender room, it will be actually filtered out. Filtering like that is far from ideal. This is a first quick implementation, we will do better in the future (filtering server side, new state event describing the room content, etc.)

SimonBrandner:
Thanks for the quick response!

> "NSFW" is not understandable enough for non power users.

Maybe _"Filter explicit content"_ or something like that. I would just expect the _explicit_ word to be in the "header".

But I am probably just nitpicking at this point... :joy:  Sorry for annoying

bmarty:
I have created https://github.com/matrix-org/matrix-doc/issues/3054 to try to do better on this subject

bmarty:
Update wording

image

bmarty:
FTR I have asked the admin of the room about Blender to update the topic of the room to avoid being filtered by mistake.

vector/src/main/java/im/vector/app/features/roomdirectory/RoomDirectoryViewModel.kt
```diff
@@ -155,47 +166,60 @@ class RoomDirectoryViewModel @AssistedInject constructor(@Assisted initialState:
...
+
+            currentJob = null
+
+            since = data.nextBatch
+
+            // Filter
+            val newPublicRooms = data.chunk.orEmpty()
+                    .filter {
+                        showAllRooms
+                                || "${it.name.orEmpty()} ${it.topic.orEmpty()} ${it.canonicalAlias.orEmpty()}".toLowerCase(Locale.ROOT)
```
onurays, line 206: I couldn't suggest a better way for now, will think later.

PR #2976, opened 2021-03-09
Snappier local echo

bmarty:
There was an issue on the local echo, especially the update of the sent state. Message can stay in gray, whereas they are already sent to the server.

This is more visible if the database is blocked in a long transaction.

This PR also contain a rework (in a dedicated commit), to split DefaultTimeline into several smaller files.


BillCarsonFr (approved):
LGTM, good catch

Issue #2991, opened 2021-03-12
Remove shields from the room list


bmarty:
Remove shields from the room list.

It should allow to greatly improve the performance, especially after login to a big account with lots of e2e rooms, because computing the room shield require to load all the room members and their device, download their keys, etc.

Shields will stay visible when in the room detail screen (timeline), and on all its sub-screens.

EW does not display shield on the room list as well.

manuroe:
Kamino cloned this issue to [vector-im/riot-ios](https://github.com/vector-im/element-ios/issues/4115)

PR #2999, opened 2021-03-15
Shields optimization

bmarty:
- Hide the shield from the room list (part of #2991)
- Stop loading room members of e2e just after an init sync.
- Also fix "Encrypted message" displayed on the room list
- Rework UpdateTrustWorker (should not introduce change)
- limit the size of logged data.

In the future we may want to compute the room shields for room on demand. For now we keep computing it but in a lazy way, since we do not load all the members at once.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/crosssigning/UpdateTrustWorker.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/crosssigning/UpdateTrustWorker.kt
@@ -36,12 +36,14 @@ import org.matrix.android.sdk.internal.crypto.store.db.model.UserEntityFields
 import org.matrix.android.sdk.internal.database.model.RoomMemberSummaryEntity
 import org.matrix.android.sdk.internal.database.model.RoomMemberSummaryEntityFields
 import org.matrix.android.sdk.internal.database.model.RoomSummaryEntity
+import org.matrix.android.sdk.internal.database.model.RoomSummaryEntityFields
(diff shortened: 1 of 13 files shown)


BillCarsonFr (approved):
LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/crosssigning/UpdateTrustWorker.kt
```diff
@@ -201,66 +203,79 @@ internal class UpdateTrustWorker(context: Context,
...
-                        try {
-                            val updatedTrust = computeRoomShield(allActiveRoomMembers, roomSummary)
-                            if (roomSummary.roomEncryptionTrustLevel != updatedTrust) {
-                                Timber.d("## CrossSigning - Shield change detected for $roomId -> $updatedTrust")
-                                roomSummary.roomEncryptionTrustLevel = updatedTrust
-                            }
-                        } catch (failure: Throwable) {
-                            Timber.e(failure)
-                        }
+        sRealm.executeTransaction { sessionRealm ->
```
ganfra, line 206: One question, why are we not using awaitTransaction as we are in a worker?

ganfra (approved):
One question, otherwise LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/crosssigning/UpdateTrustWorker.kt
bmarty: I did not change that, but done

PR #3003, opened 2021-03-16
Split key request

### LINKED ISSUE #2925: [SDK] Key querying is not chunked and servers might not support an arbitrary amount of users in the request.
**Describe the bug**
During my testing using Complement with this PR https://github.com/matrix-org/complement/pull/75, I hit a bug when element would need to send out a `/keys/query` request. Complement creates an server with N E2EE capable users that are joined into a single room.

When I increased the number of users to 1000, synapse failed to return a response for the initial `/keys/query` request which contains all 1000 users.

**To Reproduce**
1. Freshly log in to an account that shares an E2EE room with 1000 users.
2. Try to send a message in the room.
3. Observe the 500 `/keys/query` response in the network tab.

This completely wedges the E2EE support and the message can't ever be sent.

I'm not sure if this is a specific limitation of the Synapse install Complement was running but it seems a good idea to chunk `/keys/query` requests anyways. It seems that chunking the requests out at 200 or 500 users (or anything between that) would work.

Note that this is the same problem as https://github.com/matrix-org/matrix-js-sdk/issues/1619

bmarty:
Fixes #2925

Limit the number of user to max 250 per request.

If there is 251 users, the request will be cut into 2 requests, one with 125 users and one with 126 users

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/MathUtils.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/MathUtils.kt
@@ -33,7 +33,8 @@ internal fun getBetsChunkSize(listSize: Int, limit: Int): BestChunkSize {
         )
     } else {
         val numberOfChunks = ceil(listSize / limit.toDouble()).toInt()
-        val chunkSize = listSize / numberOfChunks
+        // Round on next Int
+        val chunkSize = ceil(listSize / numberOfChunks.toDouble()).toInt()
 
         BestChunkSize(
                 numberOfChunks = numberOfChunks,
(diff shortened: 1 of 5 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/MathUtils.kt
```diff
@@ -0,0 +1,44 @@
...
+import kotlin.math.ceil
+
+internal data class BestChunkSize(
+        val numberOfChunks: Int,
+        val chunkSize: Int
+) {
+    fun shouldSplit() = numberOfChunks > 1
+}
+
+internal fun getBetsChunkSize(listSize: Int, limit: Int): BestChunkSize {
```
manuroe, line 28:
```suggestion
internal fun getBestChunkSize(listSize: Int, limit: Int): BestChunkSize {
```

giomfo (approved):
LGTM

PR #3004, opened 2021-03-16
Quick Work around to remove stuck messages

BillCarsonFr:
Quick work around to remove stuck messages at the bottom of timeline.
Currently only choice is to clear cache and do a full initial sync.
This PR is just adding an option to cancel sent but not synced message, only effect is that it will delete the local echo.

image

This does not fix the root issues, See:
https://github.com/vector-im/element-android/issues/516
https://github.com/matrix-org/synapse/issues/9424
https://github.com/matrix-org/synapse/issues/9624

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
@@ -251,7 +251,7 @@ class MessageActionsViewModel @AssistedInject constructor(@Assisted
                     addActionsForSyncedState(timelineEvent, actionPermissions, messageContent, msgType)
                 }
                 timelineEvent.root.sendState == SendState.SENT   -> {
-                    addActionsForSentNotSyncedState(timelineEvent, actionPermissions, messageContent, msgType)
+                    addActionsForSentNotSyncedState(timelineEvent)
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
```diff
@@ -287,7 +290,24 @@ class MessageActionsViewModel @AssistedInject constructor(@Assisted
...
-            add(EventSharedAction.Cancel(timelineEvent.eventId))
+            add(EventSharedAction.Cancel(timelineEvent.eventId, false))
+        }
+    }
+
+    private fun ArrayList<EventSharedAction>.addActionsForSentNotSyncedState(timelineEvent: TimelineEvent,
+                                                                             actionPermissions: ActionPermissions,
+                                                                             messageContent: MessageContent?,
+                                                                             msgType: String?) {
+        if (timelineEvent.root.sendState == SendState.SENT) {
```
bmarty, line 301: This test is already done before calling this method, so we could remove it

bmarty (approved):
LGTM, just one small remark. Also, any idea of the root cause?

bmarty:
> LGTM, just one small remark. Also, any idea of the root cause?

Never mind, I did not see the link to Synapse issues. Thanks

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
BillCarsonFr: Root cause is complex and hard to catch on synapse side, two main scenario could cause that:
- It could be the downstream bug (https://github.com/matrix-org/synapse/issues/9624)
- Or a gappy sync scenario (two much events, and the synced event is in the gap, and in this case clients should have some mechanism to backfill if a sent event is not synced after following sync...)

bmarty:
(Please fix the warning in the code)

PR #3006, opened 2021-03-16
Regression: Video will not play upon tap

### LINKED ISSUE #2928: Regression: Video will not play upon tap, but only after swipe
Tap on a video will spawn the video player (controls visible at bottom edge), but the video doesn't play and the screen remains black (even after tapping on the play button). The video *will* start playing if (from the video player screen) you swipe through the room's media.

BillCarsonFr:
Fixes #2928 (Regression: Video will not play upon tap, but only after swipe)


ganfra (approved):
LGTM, maybe add a TODO to fix that properly?

bmarty (approved):
LGTM. I've checked and I see no rageshake with the bug this line should fix. It was just changed as a safety measure.

PR #3018, opened 2021-03-17
Fix issue with Realm in UpdateTrustWorker

bmarty:
And some cleanup


ganfra (approved):
LGTM, tested ok, except some decryption not working, but I don't think it's related

PR #3025, opened 2021-03-18
Send several NO_OLM in one request

BillCarsonFr:
NO_OLM withheld code were sent one by one during keyshare process.
In big room this can be a common scenario (people running out of OTK), this change simply send all the NO_OLM in a single toDevice event.

With a test account in Megolm Test room, this tear down the number of needed requests to 15 instead of 835

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmEncryption.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmEncryption.kt
@@ -303,7 +303,7 @@ internal class MXMegolmEncryption(
         }
 
         if (noOlmToNotify.isNotEmpty()) {
-            //XXX offload?, as they won't read the message anyhow?
+            // XXX offload?, as they won't read the message anyhow?
             notifyKeyWithHeld(
                     noOlmToNotify,
                     session.sessionId,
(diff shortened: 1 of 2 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmEncryption.kt
```diff
@@ -304,6 +301,16 @@ internal class MXMegolmEncryption(
         } else {
             Timber.i("## CRYPTO | shareUserDevicesKey() : no need to sharekey")
         }
+
+        if (noOlmToNotify.isNotEmpty()) {
+            //XXX offload?, as they won't read the message anyhow?
+            notifyKeyWithHeld(
```
bmarty, line 307: This is now sent after the code with `if (haveTargets)`, I think it is OK, but could you confirm please?

bmarty (approved):
Thanks, LGTM.
2 minor remarks.
Also you should be interested to limit the size of the logs by using `logLimit()` in some of the logs in this class.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmEncryption.kt
BillCarsonFr: yes that's what i wanted
(i even wonder if we could offload to other thread to speed up sharing to the one that will be able to read the message)

bmarty: ok, thanks

PR #3026, opened 2021-03-18
Display the room shield in all room setting screens

bmarty:
Some screen have the room shield (ex: room uploads) and some do not have it. This PR make it uniform.


ganfra (approved):
LGTM

PR #3034, opened 2021-03-19
Picture preview when replying. 

### LINKED ISSUE #2916: Picture preview when replying
It's very unclear which picture you're replying to when you reply to a picture in a chat because it is only a link, with no timestamp and no preview (two things most other messaging clients use). Having a miniaturized picture would help a lot to clarify this without a bunch of extra clicking, or at least, having the entire message be able to be tapped to see the message in question rather than just a small link on the left.

bmarty:
Also add the image preview in the message detail bottomsheet

Fixes #2916

#### Reply

image

#### Bottomsheet

image

### DIFF
--- a/vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetMessagePreviewItem.kt
+++ b/vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetMessagePreviewItem.kt
@@ -19,6 +19,7 @@ package im.vector.app.core.epoxy.bottomsheet
 import android.text.method.MovementMethod
 import android.widget.ImageView
 import android.widget.TextView
+import androidx.core.view.isVisible
 import com.airbnb.epoxy.EpoxyAttribute
 import com.airbnb.epoxy.EpoxyModelClass
 import im.vector.app.R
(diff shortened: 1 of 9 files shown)


vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetMessagePreviewItem.kt
```diff
@@ -59,6 +67,10 @@ abstract class BottomSheetMessagePreviewItem : VectorEpoxyModel<BottomSheetMessa
         holder.avatar.setOnClickListener { userClicked?.invoke() }
         holder.sender.setOnClickListener { userClicked?.invoke() }
         holder.sender.setTextOrHide(matrixItem.displayName)
+        data?.let {
+            imageContentRenderer?.render(it, ImageContentRenderer.Mode.THUMBNAIL, holder.imagePreview)
```
BillCarsonFr, line 71: clear on unbind?

BillCarsonFr (approved):
just one remark

vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetMessagePreviewItem.kt
bmarty: Done

kojid0:
Quick question: When you reply to a photo (swipe the photo, then enter some text and hit send), it will still say "In reply to [Name] sent an image.", right? So no picture preview here? This seems weird to me. If this is intended, okay. Otherwise I will open a new issue.

PR #3049, opened 2021-03-24
Be robust to Exception.

bmarty:
And display details about it in the bottom sheet

When there is an exception to build on item in the timeline, it's correctly handled, but it was not handled in the message detail bottom sheet.

Exception was due to bad translation in French, see below usage of `%2$s` instead of `%1$s`:

```xml
    <plurals name="notice_room_canonical_alias_alternative_added_by_you">
        <item quantity="one">Vous avez ajouté %2$s comme adresse alternative pour ce salon.</item>
        <item quantity="other">Vous avez ajouté %2$s comme adresses alternatives pour ce salon.</item>
    </plurals>
```

Sadly Lint do not detect that.

The translation is already fixed in Weblate, so it will be fixed for the next release 1.1.4.


ganfra (approved):
LGTM, thanks

PR #3055, opened 2021-03-25
Get event

bmarty:
Try to get event faster after receiving a Push by retrieving the event directly. A background sync is run as usual.

Also give this event to the CallEventProcessor for faster incoming call treatment.

TODO:

- [x] Test all cases
- [x] Handle push rules
- [x] Ignore message edition
- [x] Do this also when app is in foreground? -> NO
- [x] Do not do this if not on WiFi network because of the network data overhead (the event will be downloaded 2 times)

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/events/DefaultEventService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/events/DefaultEventService.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2021 The Matrix.org Foundation C.I.C.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.internal.session.events
+
+import org.matrix.android.sdk.api.session.events.EventService
+import org.matrix.android.sdk.api.session.events.model.Event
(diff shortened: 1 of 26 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/GetEventTask.kt
```diff
@@ -16,28 +16,49 @@
...
             apiCall = roomAPI.getEvent(params.roomId, params.eventId)
         }
+
+        // Try to decrypt the Event
+        if (event.isEncrypted()) {
+            tryOrNull(message = "Unable to decrypt the event") {
+                eventDecryptor.decryptEvent(event, "")
+            }
+                    ?.let { result ->
+                        event.mxDecryptionResult = OlmDecryptionResult(
```
ganfra, line 53: Maybe we should do that directly from the eventDecryptor?

bmarty: This is out of scope of this PR

ganfra (commented):
I think we should offer the fastlane option to be used only on wifi in settings (enabled by default)
Otherwise LGTM, but didn't tested

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/events/DefaultEventService.kt
```diff
@@ -0,0 +1,48 @@
...
+import timber.log.Timber
+import javax.inject.Inject
+
+internal class DefaultEventService @Inject constructor(
+        private val getEventTask: GetEventTask,
+        private val callEventProcessor: CallEventProcessor,
+        private val wifiDetector: WifiDetector
+) : EventService {
+
+    override suspend fun getEvent(roomId: String, eventId: String, onlyOnWifi: Boolean): Event? {
```
ganfra, line 33: Also I'm not sure the onlyWifi should be exposed on the sdk...

bmarty: i wanted to give the ability to easily tell the SDK to save network data usage. But yes, I will move this check to the app level.

PR #3056, opened 2021-03-25
Add version details on the login screen, in debug or developer mode

bmarty:
image

(Build number is only relevant if built by Buildkite)


ganfra (approved):
LGTM

PR #3060, opened 2021-03-26
remove enum from rest model

BillCarsonFr:
Avoid using enum in rest model

````
To model failed : com.squareup.moshi.JsonDataException: Expected one of [world_readable, shared, invited, joined] but was anyone at path $.history_visibility

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/RoomHistoryVisibility.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/RoomHistoryVisibility.kt
@@ -16,35 +16,31 @@
 
 package org.matrix.android.sdk.api.session.room.model
 
-import com.squareup.moshi.Json
-import com.squareup.moshi.JsonClass
-
 /**
  * Ref: https://matrix.org/docs/spec/client_server/latest#room-history-visibility
  */
-@JsonClass(generateAdapter = false)
 enum class RoomHistoryVisibility {
     /**
      * All events while this is the m.room.history_visibility value may be shared by any
      * participating homeserver with anyone, regardless of whether they have ever joined the room.
      */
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/RoomHistoryVisibilityContent.kt [hunk 1]
```diff
@@ -21,5 +21,16 @@ import com.squareup.moshi.JsonClass
 
 @JsonClass(generateAdapter = true)
 data class RoomHistoryVisibilityContent(
-        @Json(name = "history_visibility") val historyVisibility: RoomHistoryVisibility? = null
-)
+        @Json(name = "history_visibility") val historyVisibility: String? = null
+) {
+
+    fun safeHistoryVisibility(): RoomHistoryVisibility? {
```
bmarty, line 27: Rename to `historyVisibility` (and make it a `val`?)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/RoomHistoryVisibility.kt
```diff
@@ -16,35 +16,31 @@
...
  * Ref: https://matrix.org/docs/spec/client_server/latest#room-history-visibility
  */
-@JsonClass(generateAdapter = false)
 enum class RoomHistoryVisibility {
     /**
      * All events while this is the m.room.history_visibility value may be shared by any
      * participating homeserver with anyone, regardless of whether they have ever joined the room.
      */
-    @Json(name = "world_readable") WORLD_READABLE,
+    WORLD_READABLE,
```
bmarty, line 27: The Json value is needed for `updateHistoryReadability()` (serialization)

bmarty (commented):
This breaks things. I will propose something else.

PR #3092, opened 2021-03-31
Room List performance PR (use Live PagedList via Monarchy)

BillCarsonFr:
## Room List performance PR

This PR introduces the usage of PagedList for the RoomList fragment.
This will use less RAM, independent of the number of rooms, and it will release the pressure on the computation thread, and stop listening to the summary when they are not on screen

### How it used to be
Currently all the rooms are queried and mapped to models, and the filtering/sorting is done in memory and observed by view models
In order to maintain performances, the sorting/filtering was done on computation thread, and all was loaded early (AppStateHandler) and always kept up to date for ever even if room list fragment/activity was not active (to be up to date faster when coming back from timeline)

### How it is now

Now the query returns a PagedList, that is backed by a realmResults, but only the needed summaries are actually mapped and loaded in memory (around 20 currently)
So we keep the benefits of realm zero copies as long as possible

This allow us to only load and listen to DB when the room list is on screen; i.e when opening the timeline we don't have anymore to listen to summary changes and perform sortering/filtering in memory

The paged list loading is based on position in recycler view, when an item is visible on screen (onBound), the realm result is asked to load more around the given position, and this will actually load the object in memory (copy from realm)

Now all filtering/sorting is done by Realm, thus the entities have been modified: some fields have been denormalized (last activity), isFav/isLowPrio/isServerNotice, and all fields used in query have been Indexed

The RoomListViewModel now provides a list of sections (header + pagedList), and the fragment uses then a ConcatAdapter to join all sections and headers.

The summaryEntity has been also modified to prevent unecessary DB notifications when setting a field


bmarty (approved):
LGTM, I've push some cleanup

bmarty (changes requested):
There is a few things to fix (regression) before we can merge it:
- [ ] long click hint should be displayed only if some rooms are visible
- [x] I had several times the same section. Maybe due to my cleanup, will test
- [ ] Cold start rendering is not as good as before, there is no loading anymore, but the empty wiev is displayed

vector/src/main/java/im/vector/app/features/home/room/list/RoomListFragment.kt
```diff
@@ -95,11 +95,12 @@ class RoomListFragment @Inject constructor(
     )
 
     private val adapterInfosList = mutableListOf<SectionAdapterInfo>()
-    private val concatAdapter = ConcatAdapter()
+    private var concatAdapter : ConcatAdapter? = null
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        setupCreateRoomButton()
+        views.stateView.contentView = views.roomListView
```
bmarty, line 102: We do not call `setupCreateRoomButton()` anymore?

vector/src/main/java/im/vector/app/features/home/room/list/RoomListFragment.kt
```diff
@@ -95,11 +95,12 @@ class RoomListFragment @Inject constructor(
     )
 
     private val adapterInfosList = mutableListOf<SectionAdapterInfo>()
-    private val concatAdapter = ConcatAdapter()
+    private var concatAdapter : ConcatAdapter? = null
```
bmarty, line 98: I think I've moved this to the constructor... Strange

vector/src/main/java/im/vector/app/features/home/room/list/RoomListFragment.kt
BillCarsonFr: wow, bad revert. good catch

bmarty (approved):
I will merge it, we could do some improvement later (restore animation, item placeholder, etc.)

PR #3097, opened 2021-03-31
Migrate Retrofit interface to coroutine calls

bmarty:
Nicer code, and type can be inferred. It should also have a better handling of cancellation.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/UploadSignaturesTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/UploadSignaturesTask.kt
@@ -17,7 +17,6 @@ package org.matrix.android.sdk.internal.crypto.tasks
 
 import org.matrix.android.sdk.api.failure.Failure
 import org.matrix.android.sdk.internal.crypto.api.CryptoApi
-import org.matrix.android.sdk.internal.crypto.model.rest.SignatureUploadResponse
 import org.matrix.android.sdk.internal.network.GlobalErrorReceiver
 import org.matrix.android.sdk.internal.network.executeRequest
 import org.matrix.android.sdk.internal.task.Task
@@ -36,11 +35,14 @@ internal class DefaultUploadSignaturesTask @Inject constructor(
 
     override suspend fun execute(params: UploadSignaturesTask.Params) {
         try {
-            val response = executeRequest<SignatureUploadResponse>(globalErrorReceiver) {
-                this.isRetryable = true
-                this.maxRetryCount = 10
-                this.apiCall = cryptoApi.uploadSignatures(params.signatures)
-            }
+            val response = executeRequest(
+                    globalErrorReceiver,
+                    { cryptoApi.uploadSignatures(params.signatures) },
+                    {
+                        isRetryable = true
+                        maxRetryCount = 10
+                    }
+            )
             if (response.failures?.isNotEmpty() == true) {
                 throw Throwable(response.failures.toString())
             }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/UploadSignaturesTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/UploadSignaturesTask.kt
@@ -37,12 +37,11 @@ internal class DefaultUploadSignaturesTask @Inject constructor(
         try {
             val response = executeRequest(
                     globalErrorReceiver,
-                    { cryptoApi.uploadSignatures(params.signatures) },
-                    {
-                        isRetryable = true
-                        maxRetryCount = 10
-                    }
-            )
+                    isRetryable = true,
+                    maxRetryCount = 10
+            ) {
+                cryptoApi.uploadSignatures(params.signatures)
+            }
             if (response.failures?.isNotEmpty() == true) {
                 throw Throwable(response.failures.toString())
             }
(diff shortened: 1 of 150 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/terms/DefaultTermsService.kt
```diff
@@ -63,12 +63,12 @@ internal class DefaultTermsService @Inject constructor(
                                       baseUrl: String,
                                       agreedUrls: List<String>,
                                       token: String?) {
-        withContext(coroutineDispatchers.main) {
+        withContext(coroutineDispatchers.io) {
```
Dominaezzz, line 66: I think this `withContext` can be removed.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/terms/DefaultTermsService.kt
```diff
@@ -50,10 +50,10 @@ internal class DefaultTermsService @Inject constructor(
 ) : TermsService {
     override suspend fun getTerms(serviceType: TermsService.ServiceType,
                                  baseUrl: String): GetTermsResponse {
-        return withContext(coroutineDispatchers.main) {
+        return withContext(coroutineDispatchers.io) {
```
Dominaezzz, line 53: Same here.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
```diff
@@ -21,36 +21,51 @@ import kotlinx.coroutines.delay
...
+internal suspend inline fun <DATA : Any> executeRequest(globalErrorReceiver: GlobalErrorReceiver?,
+                                                        noinline requestBlock: suspend () -> DATA,
+                                                        initBlock: Request<DATA>.() -> Unit): DATA {
+    return Request(globalErrorReceiver, requestBlock)
+            .apply(initBlock)
+            .execute()
+}
 
-internal class Request<DATA : Any>(private val globalErrorReceiver: GlobalErrorReceiver?) {
+internal class Request<DATA : Any>(
```
Dominaezzz, line 42: Don't think there's much need for this class anymore. Might as well merge it into the `executeRequest` function.
The private `var`s would become local variables and the public `var`s would become function parameters.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/UploadSignaturesTask.kt
```diff
@@ -36,11 +35,14 @@ internal class DefaultUploadSignaturesTask @Inject constructor(
...
-            val response = executeRequest<SignatureUploadResponse>(globalErrorReceiver) {
-                this.isRetryable = true
-                this.maxRetryCount = 10
-                this.apiCall = cryptoApi.uploadSignatures(params.signatures)
-            }
+            val response = executeRequest(
+                    globalErrorReceiver,
+                    { cryptoApi.uploadSignatures(params.signatures) },
+                    {
+                        isRetryable = true
```
Dominaezzz, line 42: This would definitely look nicer as a parameter instead of DSL.

Dominaezzz (commented):
LGTM :slightly_smiling_face: .
Just a nit about the `Request` class.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
bmarty: Yes, I agree, it's a bit dirty like that

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/UploadSignaturesTask.kt
bmarty: Very true

bmarty:
Thanks for the review @Dominaezzz I'll see what I can do

bmarty:
The problem is that the `Request.execute()` is recursive, so I cannot remove the class usage.

bmarty:
@Dominaezzz I've pushed some commits to take into account your valuable remarks :)

bmarty:
One [useful log](https://github.com/vector-im/element-android/pull/3097/files#diff-b5d7c43fa1f7920615059be875782e020e64e3ff7a273d7ddb1ec0164fe3fb31R59) is lost with this change, and I do not see how can I get it back...

ganfra (approved):
LGTM, thanks!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
```diff
@@ -19,38 +19,45 @@ package org.matrix.android.sdk.internal.network
...
+                is HttpException              -> throwable.toFailure(globalErrorReceiver)
+                else                          -> throwable
             }
-        } catch (exception: Throwable) {
-            // Log some details about the request which has failed
-            Timber.e("Exception when executing request ${apiCall.request().method} ${apiCall.request().url.toString().substringBefore("?")}")
+
+            // Log some details about the request which has failed. This is less useful than before...
+            // Timber.e("Exception when executing request ${apiCall.request().method} ${apiCall.request().url.toString().substringBefore("?")}")
+            Timber.e("Exception when executing request")
```
ganfra, line 60: When it's an HttpException we can grab the response and request I think?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/terms/DefaultTermsService.kt
bmarty: Done, thanks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
bmarty: Done, thanks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
bmarty: Sadly, no

PR #3098, opened 2021-03-31
Fix avatar rendering for DMs, after initial sync (#2693)

### LINKED ISSUE #2693: Room Avatar is error
**Describe the bug**
When i login in the element , i found some rooms has error avatar.
These room avatars are supposed to be empty, but instead return URLs.

When I click in the return avatar will be correct.

bmarty:
Also better handling of previous DMs management

Fixes #2693 (I think)

Tested OK when playing with account data to switch room between DM to room with devtools on EW


BillCarsonFr (approved):
LGTM

PR #3124, opened 2021-04-06
Fix mandatory parameter in API (#3065)

### LINKED ISSUE #3065: Bug in /read_markers
https://github.com/vector-im/element-android/blob/c80564e1b14067287e463374ab71b9ad8c6cd0d4/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/read/SetReadMarkersTask.kt#L45

The m.fully_read field is handled as optional, but the spec says it is required: https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-rooms-roomid-read-markers

bmarty:
Fixes #3065.

Use another request when "m.fully_read" does not have to be updated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@ Bugfix 🐛:
  - Handle encrypted reactions (#2509)
  - Disable URL preview for some domains (#2995)
  - Fix avatar rendering for DMs, after initial sync (#2693)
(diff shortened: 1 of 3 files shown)


ganfra (approved):
Thanks, LGTM, didn't aware of this one

PR #3128, opened 2021-04-07
Various fixes before release 1.1.4

bmarty:
Please review commit by commit

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
@@ -1675,10 +1675,12 @@ class RoomDetailFragment @Inject constructor(
             shareText(requireContext(), action.messageContent.body)
         } else if (action.messageContent is MessageWithAttachmentContent) {
             lifecycleScope.launch {
-                val data = session.fileService().downloadFile(messageContent = action.messageContent)
-                if (isAdded) {
-                    shareMedia(requireContext(), data, getMimeTypeFromUri(requireContext(), data.toUri()))
-                }
+                val result = runCatching { session.fileService().downloadFile(messageContent = action.messageContent) }
+                if (!isAdded) return@launch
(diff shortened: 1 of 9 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -1675,10 +1675,12 @@ class RoomDetailFragment @Inject constructor(
             shareText(requireContext(), action.messageContent.body)
         } else if (action.messageContent is MessageWithAttachmentContent) {
             lifecycleScope.launch {
```
ganfra, line 1677: why not using viewLifecycleScope? isAdded check would not be needed then

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -1701,16 +1703,22 @@ class RoomDetailFragment @Inject constructor(
...
-                saveMedia(
-                        context = requireContext(),
-                        file = data,
-                        title = action.messageContent.body,
-                        mediaMimeType = action.messageContent.mimeType ?: getMimeTypeFromUri(requireContext(), data.toUri()),
-                        notificationUtils = notificationUtils
-                )
-            }
+            val result = runCatching { session.fileService().downloadFile(messageContent = action.messageContent) }
+            if (!isAdded) return@launch
```
ganfra, line 1707: Same remark ^

ganfra (approved):
2 small remarks otherwise LGTM

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
bmarty: yeah, we will change that later, I'm not even sure if this isAdded statement is still necessary with `lifecycleScope`

PR #3140, opened 2021-04-08
Fix EW-EA compatibility for verification

BillCarsonFr:
The recent EW [rework](https://github.com/matrix-org/matrix-react-sdk/pull/5727) on verification flow are not compatible with mobile implementation.

Two main issues:
- When login a new web session and trying to verify with EA:
      When you type on 'Use anothe login' on web, nothing appears on mobile; And if from EA you type 'verify the new login' you will have a infinite loading

- When login in a new EA and trying to verify with EW:
    EA is waiting for an incoming verification request, and this will never happen with EW recent re-work

Quick Fix

For 1: Actually the EA alert for the request from EW is there but under the 'verify this new login' green popup. to fix it EA will dismiss the green alert so that the user can see the verification request alert

For 2: EA now always send a verification request (instead of waiting for an incoming one)


bmarty (approved):
LGTM, but please run `ktlint`

PR #3142, opened 2021-04-08
Generic hanlding of rate limit eror by the SDK

### LINKED ISSUE #1300: Handle 429 M_LIMIT_EXCEEDED error with a global policy
JS implementation: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/scheduler.js#L138-L175

It's already half implemented for some requests, "half" because we do not take care of the `retry_after_ms` that the homeserver is asking us to wait, except when we display the error to the user.

### LINKED ISSUE #2240: Errors during message redaction are not handled
**Describe the bug**
When redacting many messages, if you hit the server rate limit, the event will be shown as redacted, when in fact redaction request is rejected with 429, and the message is not actually deleted.

**To Reproduce**
1. Redact lots of messages so that Synapse rate limit is hit.

This will appear like this in Synapse log:
```
2020-10-09 18:23:35,748 - synapse.http.server - 81 - INFO - PUT-4047355 - <XForwardedForRequest at 0x7efc824bea90 method='PUT' uri='/_matrix/client/r0/rooms/<...>/redact/<...>/$local.<...>' clientproto='HTTP/1.0' site=8008> SynapseError: 429 - Too Many Requests
```

 - Device: N/A
 - OS: N/A

**Additional context**
 - App version and store: 1.0.8 (Google Play)
 - Homeserver: torlan.ru

Similar to #462, maybe a dup of #1300

bmarty:
Fixes #1300
I think it also fixes #2240

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
@@ -19,6 +19,7 @@ package org.matrix.android.sdk.internal.network
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.delay
 import org.matrix.android.sdk.api.failure.Failure
+import org.matrix.android.sdk.api.failure.MatrixError
 import org.matrix.android.sdk.api.failure.getRetryDelay
 import org.matrix.android.sdk.api.failure.shouldBeRetried
 import org.matrix.android.sdk.internal.network.ssl.CertUtil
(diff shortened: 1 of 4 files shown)


BillCarsonFr (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
```diff
@@ -28,22 +29,21 @@ import java.io.IOException
 
 /**
  * Execute a request from the requestBlock and handle some of the Exception it could generate
+ * Ref: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/scheduler.js#L138-L175
```
Florian14, line 32: I think it should now point to this url @bmarty: https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/scheduler.ts#L54-L92

PR #3147, opened 2021-04-08
Update import_emojis.py to retain keyword order

oogm:
Follow up to PR #2998 to make sure that the emoji import script does not shuffle keywords between runs.


oogm:
Ah, did not know a change to CHANGES.md is mandatory.

bmarty (approved):
Wonderful, thanks a lot!

Issue #1793, opened 2020-07-23
Unable to connect to LAN IP addresses


Limezero:
I have a local Synapse server running at http://192.168.0.109:8008 on my local network. Every other client and device I have can successfully connect to it, including the old Riot.im Android client (0.9.12) and the desktop version of Element, but RiotX/Element cannot. If I try to enter a 192.168.0.0/24 address as a custom server, the client will immediately throw an error message saying "No network. Please check your internet connection." whether or not there's a server running on that socket.

![image](https://i.imgur.com/ckyPGnR.png)

A few other strange observations:

- The server only has an HTTP port enabled, no HTTPS - but the Element client CAN connect to an insecure HTTP server if it's running on the phone itself, under http://127.0.0.1:8008
- If I explicitly bind the server to the LAN address of the phone (i.e., put `['::1', '127.0.0.1', '192.168.0.103']` in the config file), it will still only be able to connect using the localhost address, not the LAN one
- If I enter https://192.168.0.109 (HTTPS, without the port), the client DOES find the server and even displays a fingerprint, asking me if I want to trust it - if I click trust, I get an error saying "This is not a valid Matrix server address"
- Trying to enter https://192.168.0.109:8008 (or any other IP/port) causes the client to hang for about 10 seconds while it's waiting for a timeout

Limezero:
As of version 1.0.7, this is still broken.

Rio:
I'm having the same problem, networks like 10.x.x.x and 192.168.x.x fail to connect with the message show above. Other apps like firefox and curl using tmux have no problem connecting.

Zombie-Feynman:
Confirming this bug for version 1.0.11 (F-Droid). FWIW, it also occurs if you try to connect via a domain name that points to a LAN address.

Zombie-Feynman:
The canonical Element-Web at app.element.io gave me a similar error, which I eventually traced back to being caused by mixed content: I was trying to access an http server from an https web app. I added a reverse proxy to my homeserver (using `easy-rsa` to create self-signed certificates), which allows the Android app to work. (Somehow it still wasn't enough to let app.element.io work; Developer Tools shows a GET request for `https://<my_server>/_matrix/client/versions` in red, but doesn't actually tell me what the problem is.)

bmarty:
Are you using an emulator to run Element Android? If it is the case, URLs like http://192.168.0.109:8008 can not work because it is not the 192.168 is not the lan of the emulator itself.
[...]

PR #3166, opened 2021-04-13
Related to #1793, allow cleartext to LAN domains

tobymurray:
This functionality exists in the desktop client, so hoping to mirror that. 

This addresses a number of the use cases touched on in #1793. `localdomain` is a conventional domain that is an equivalent of the `localhost` host. Enabling clear text to `*.localdomain` means it's easier to develop the Android application, as a Matrix server can be deployed locally without much fuss anywhere on the developer's LAN. This can reduce the reliance on a DNS or SSL certificates when neither are really relevant to the interaction of the client/server. In particular, managing SSL certs without a public domain is a pain in the butt.

At the same time, this does not significantly diminish the security of Element Android, as `*.localdomain` is not a TLD so any "real" deployment still needs SSL.

### Pull Request Checklist

Signed-off-by: Toby Murray <toby.murray@protonmail.com>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,6 +38,7 @@ git branch -u origin/main main
 # And optionally
 git remote prune origin
 ```
+ - Allow cleartext (non-SSL) connections to Matrix servers on LAN hosts (#3166)
 
 Changes in Element 1.1.6 (2021-04-16)
 ===================================================
--- a/vector/src/main/res/xml/network_security_config.xml
+++ b/vector/src/main/res/xml/network_security_config.xml
@@ -13,6 +13,16 @@
         <domain includeSubdomains="true">10.0.2.2</domain>
         <!-- Onion services -->
         <domain includeSubdomains="true">onion</domain>
+
+        <!-- Domains that are used for LANs -->
+        <!-- These are IANA recognized special use domain names, see https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml -->
+        <domain includeSubdomains="true">home.arpa</domain>
+        <domain includeSubdomains="true">local</domain> <!-- Note this has been reserved for use with mDNS -->
+        <domain includeSubdomains="true">test</domain>
+        <!-- These are observed in the wild either by convention or RFCs that have not been accepted, and are not currently TLDs -->
+        <domain includeSubdomains="true">home</domain>
+        <domain includeSubdomains="true">lan</domain>
+        <domain includeSubdomains="true">localdomain</domain>
     </domain-config>
 
     <debug-overrides>


TR-SLimey:
I feel like this isn't a great solution, since the local domain isn't always strictly `*.localdomain`. I've seen `.local`, `.home` and `.lan`, and it seems there are [even more domains reserved for this purpose](https://tools.ietf.org/id/draft-chapin-rfc2606bis-00.html). Besides, I don't think this applies to direct IP connections, which are probably used more than domain names on a LAN.

Ideally, the entire [private address space](https://en.m.wikipedia.org/wiki/Reserved_IP_addresses) could be included here, as well as anything that resolves to an IP in it. However, that doesn't seem to be easily supported by the current mechanism.

tobymurray:
> I feel like this isn't a great solution, since the local domain isn't always strictly .localdomain.

Agreed

> I've seen .local, .home and .lan, and it seems there are even more domains reserved for this purpose.

That link is a draft, so far as I can tell the most up-to-date "accepted" list is [here](https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml), and notably does not include `localdomain` but as you've noticed in practice people use things that are not official. I'm not familiar enough with the area to know the history, but it seems like `localdomain` in particular came from the "ground up" and is a common practice that is not enshrined as a reserved TLD. As you linked, it has been proposed.

> Ideally, the entire private address space could be included here, as well as anything that resolves to an IP in it. However, that doesn't seem to be easily supported by the current mechanism.

Also agreed, this is really the desired outcome though Android does not appear to support anything like it, which brought me to the current changeset.

> I'm not sure this is a good approach as it feels almost like hard-coding certain network configs into the app.

Yeah, my impression is this is the least bad strategy I could think of. The status quo is that Element Android can't be used in a bunch of places where Element desktop is already used, so closing that gap is more desirable than this strategy is distasteful. I'm super open to anything that reaches the same end goal (parity of Element Android and desktop when it comes to access to non-SSL endpoints). I'm not aware of a way to do this dynamically, so the functionality to express this behavior seems extremely limited.

> I would at least include all the domains reserved for local use

I can certainly add the list from [here](https://www.iana.org/assignments/special-use-domain-names/special-use-domain-names.xhtml). You mentioned `.home` and `.lan` - would you want to see these added? I think starting with a limited list and opening it after getting feedback is likely to be a less contentious change. 

> and then remove 10.0.2.2 in favour of using domains since that looks kinda ugly too
Are you suggesting a change to the status quo to for developers to assign a hostname that is resolvable to the emulator? I'm not sure how that currently pans out, so I want to make sure I'm not regressing for that use case/have documented any changes.

TR-SLimey:
>Please let me know if there's a "better" list

Sadly, nothing I know of. That draft I found with only a quick search, so I can't offer anything beyond that. There's also a somewhat related SE answer [here](https://serverfault.com/a/1006003) which lists some sources, but those seem limited in usefulness.

Personally, I would at least go with:
- .localdomain
- .local
- .home
- .lan
- .home.arpa (as per SE post)

None of those are global TLDs so we should be good (for reference: https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains).

>I'm not aware of a way to do this dynamically

Same here, though I kinda assumed there must be a better way with how ugly this is. However, having done a little research, it doesn't seem so :/

>Are you suggesting a change to the status quo to for developers to assign a hostname that is resolvable to the emulator?

Actually, scratch that part. I forgot that's for use with the emulator for which setting up a resolvable hostname might be a pain. I guess that'll have to stick around. The idea was that since we're covering private addresses with those domains, there's no need for it, but actually, it's probably better to leave it be.

dkasak:
I don't see a problem with adding these exceptions.

Given the intended purpose is testing, what is notably missing from the current list is [`.test`](https://tools.ietf.org/html/rfc6761#section-6.2).

Also, the commit message is now no longer accurate since it adds exemptions for more than just `.localdomain`. This should be corrected.

Some additional thoughts:

- Even though I'm fine with adding a cleartext exception for them since they are not currently TLDs, I would personally refrain from using `.home` and `.lan`.

  These are short and nice-sounding, and since they haven't been officially reserved, there is a risk of them being sold off. IIRC, the reason `.home.arpa` was added is because agreement couldn't be reached for reserving `.home`. `.localdomain` is a bit better in this regard, IMO.

- It should also be noted that `.local` is reserved for use with mDNS.

tobymurray:
> what is notably missing from the current list is .test.

Is it valid to create a resolvable domain with '.test'? I can certainly add it.

> the commit message is now no longer accurate

I'll update, and resolve the CHANGES.md conflict

> I would personally refrain from using .home and .lan.

I'm happy to exclude those, I don't want to churn - @TR-SLimey, does it sound sensible to you to avoid those?

> It should also be noted that .local is reserved for use with mDNS.

Do you mean noted in the context of this PR? I don't think that changes the value proposition (maybe it does?). I can add a comment inline in the file if you'd like that to be called out explicitly.

0x1a8510f2:
(I've changed GitHub accounts, but it is in fact me, @TR-SLimey)

>I don't want to churn - @TR-SLimey, does it sound sensible to you to avoid those?

Personally, I thing they should remain. I've seen `.home` being used by a major UK ISP (with no way to change it in the router control panel) and .lan seems common in custom setups (including my own though that can be changed in my case relatively easily). If the main argument is that they could eventually become TLDs (which I doubt would happen with .lan anyway) then that can always be changed quite easily in the future, but meanwhile the alternative is excluding some users completely in favour of something which may or may not happen.

dkasak:
> Is it valid to create a resolvable domain with '.test'? I can certainly add it.

They can be made resolvable in a private network, but not globally. DNS servers should generate automatic negative responses to them, except if configured to have them resolved by a private authoritative server. They can't be registered. See https://tools.ietf.org/html/rfc6761#section-6.2 for more information.

And yes, since we're already adding the other exceptions, I'd prefer if you added them for completeness.

This looks good otherwise.

My "additional thoughts" above were something I wanted to mention for completeness' sake, but not meant to be actionable points. Sorry for being less than clear. I agree with @0x1a8510f2's reasoning in that cleartext exceptions for `.home` and `.lan` don't hurt and can be easily changed if need be.

tobymurray:
I've rebased, updated the commit message and added `test.

dkasak (approved):

PR #3178, opened 2021-04-14
Expose directUserId in RoomSummary

yostyle:
Expose directUserId in RoomSummary to identify the user if the room is direct.


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/mapper/RoomSummaryMapper.kt
```diff
@@ -43,6 +43,7 @@ internal class RoomSummaryMapper @Inject constructor(private val timelineEventMa
                 topic = roomSummaryEntity.topic ?: "",
                 avatarUrl = roomSummaryEntity.avatarUrl ?: "",
                 isDirect = roomSummaryEntity.isDirect,
+                directUserId = roomSummaryEntity.directUserId ?: "",
```
bmarty, line 46: no need for `?: ""` since `RoomSummary.directUserId` can be null

bmarty (approved):
One small remark, else LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/mapper/RoomSummaryMapper.kt
yostyle: I fixed it

Issue #2944, opened 2021-03-02
Wrong EMOJi displayed


github-tomster:
**Describe the bug**
just updated from version 1.0.11 to 1.1.0 🤦‍♂️ 
i type text and push the keyboard-app emoji button.
emojies are shown according to system installed notocoloremoji.ttf
i push an emoji that gets inserted into the inputfield an there it is
displayed as an other styled, quite ugly, emoji not from my notocoloremoji.ttf system font! 💢 😠 
then i send the message and the correct emoji appears in the timeline.

**To Reproduce**
1. Go to 'message text input field'
2. Click on 'emoji button'
3. see wrong emojis displayed

**Expected behavior**
a clear and consisted respect to the user device notocoloremoji.ttf!

**Smartphone (please complete the following information):**
 - Device: [athene]
 - OS: [7.1]
 
unfortunately i found this:
https://github.com/vector-im/element-android/pull/2514

so i am more than scared about the upcomming reactions from the code-making-ones.
am i taken away the control of my system emoji style for my own good ❓  🤷‍♂️

github-tomster:
@code-creating-ones
it seems like an unsolvable issue. 🤷🏻‍♂️ 
but if that is correct, what issues could be solved by this project at all❓

PR #3182, opened 2021-04-15
Message Composer Design Update 

onurays:
composer_empty
composer_with_text


gaelledel (changes requested):
- [ ] The max height of the composer frame should be increased so it covers more UI real estate See an example here https://www.figma.com/file/uaWc62Ux2DkZC4OGtAGcNc/Voice-Messages?node-id=838%3A225
- [ ] The keyboard icon inside the frame is too big. Could we reduce? The colour should also be changed so it reflects Figma's specs
Screenshot_20210416-154955_Element dbg
Example in Figma here.

- [ ] When I am typing something and have just added an emoji, the keyboard should return automatically. Currently, the user has to explicitly tap on the keyboard button to see the keyboard again. This also means that is I happen to have added an emoji in the text frame without having pressed back on the keyboard button before sending, Once it is sent, the Keyboard icon remains in the text frame. See below
Screenshot_20210416-155003_Element dbg

TR-SLimey:
@gaelledel I'm no designer, so feel free to disregard my comments, but I wanted to suggest:

- Keeping the max composer height a bit smaller than that in Figma. Maybe 1/2 as opposed to 3/4. This is because, when replying to a lengthy message, with a long message of your own, it can be difficult to go back to and reference the message you're replying to. The timeline is above the composer, so it's also more difficult to scroll the timeline than the composer which giving the timeline a bit more space should improve.

- Not automatically switching back from emoji keyboard after adding an emoji - sometimes one might wish to enter multiple emojis. However, the keyboard should still switch back upon sending.

WDYT?

gaelledel (commented):
Amazing thank you for the changes. Just two minor things still and then all good to go

- [x] On dark mode the Reveal + buttons and X close should have other colours referenced here https://www.figma.com/file/X4XTH9iS2KGJ2wFKDqkyed/Compound?node-id=1237%3A16664. The animation OnPress is missing but you might actually haven't done that part of the work yet though.

- [ ] When typing and sending you actually see the composer frame expanding and there's a weird flash on the emoji icon. I wonder if we can tweak this?

github-tomster:
as one clearly can see on @gaelledel's screenshots
it is still a [mismatch](https://github.com/vector-im/element-android/issues/2944) between the emojis!
will this 'update' fix this mess?

bmarty (approved):
LGTM, I updated to colors

github-tomster:
@https://github.com/vector-im/element-android/commit/5d69bc934a3f792d1daa86cc1fef966edb42e776
was the inconsistent and wrong emoji display fixed?

PR #3190, opened 2021-04-16
Fix a race condition

bmarty:
Push can be received before the Gateway API returns, which lead to infinite loading in the notification troubleshoot screen.


BillCarsonFr (approved):
LGTM

PR #3197, opened 2021-04-19
WIP - introduce sessionScope

bmarty:
@Dominaezzz what do you think about the new sessionScope?

### Pull Request Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
@@ -19,10 +19,6 @@ package org.matrix.android.sdk.internal.session
 import androidx.annotation.MainThread
 import dagger.Lazy
 import io.realm.RealmConfiguration
-import kotlinx.coroutines.CancellationException
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.launch
 import okhttp3.OkHttpClient
 import org.matrix.android.sdk.api.auth.data.SessionParams
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
@@ -84,6 +84,7 @@ import kotlin.coroutines.CoroutineContext
 @SessionScope
 internal class DefaultSession @Inject constructor(
         override val sessionParams: SessionParams,
+        private val sessionCoroutineScopeHolder: SessionCoroutineScopeHolder,
         private val workManagerProvider: WorkManagerProvider,
         private val globalErrorHandler: GlobalErrorHandler,
         @SessionId
@@ -161,13 +162,11 @@ internal class DefaultSession @Inject constructor(
     override val isOpenable: Boolean
         get() = sessionParamsStore.get(sessionId)?.isTokenValid ?: false
 
-    private var sessionScope: CoroutineScope? = null
-
     @MainThread
     override fun open() {
         assert(!isOpen)
         isOpen = true
-        sessionScope = CoroutineScope(SupervisorJob())
+        sessionCoroutineScopeHolder.start()
         cryptoService.get().ensureDevice()
         uiHandler.post {
             lifecycleObservers.forEach { it.onSessionStarted() }
@@ -208,8 +207,7 @@ internal class DefaultSession @Inject constructor(
 
     override fun close() {
         assert(isOpen)
-        sessionScope?.coroutineContext?.cancelChildren(CancellationException("Closing session"))
-        sessionScope = null
+        sessionCoroutineScopeHolder.stop()
         stopSync()
         // timelineEventDecryptor.destroy()
         uiHandler.post {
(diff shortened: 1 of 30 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
```diff
@@ -153,10 +158,13 @@ internal class DefaultSession @Inject constructor(
     override val isOpenable: Boolean
         get() = sessionParamsStore.get(sessionId)?.isTokenValid ?: false
 
+    private var sessionScope: CoroutineScope? = null
+
     @MainThread
     override fun open() {
         assert(!isOpen)
         isOpen = true
+        sessionScope = CoroutineScope(SupervisorJob())
```
ganfra, line 167: Use a holder shared with Dagger SessionScope, so we can use it internally?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
```diff
@@ -297,4 +307,10 @@ internal class DefaultSession @Inject constructor(
     override fun logDbUsageInfo() {
         RealmDebugTools(realmConfiguration).logInfo("Session")
     }
+
+    override fun launch(block: suspend () -> Unit) {
```
ganfra, line 311: Naming is a bit light... Maybe also you should at least add context param like the regular launch method.

ganfra, line 311: Otherwise we could create a sessionScope in the app, and not rely at all on the SDK for this. I'm not sure what is best

bmarty: yes, IDK

bmarty: > Naming is a bit light... 

I like this naming, can replace `viewModelScope.launch { ... }` by `session.launch { ... }`, this is nice
 
> Maybe also you should at least add context param like the regular launch method.

Done

Dominaezzz: This should probably return a `Job`.
A potential issue here is, are you going to whack-a-mole every single coroutine creating function? `launch`, `async`, `produce`, `launchIn`, etc.

Maybe this should be a `withContext` kinda thing that throws `CancellationException` when the session is not active?

Dominaezzz: Seeing as all the uses of `launch` are in the app, I guess it can be avoided?

ganfra: I'm not really confortable with exposing coroutine scope from the sdk. There are 2 problematics we have to tackle here:

1. Makes sure some actions are not cancelled when leaving the screen (so when viewModelScope is closed)
2. Makes sure every suspend methods at the session/room level are cancelled in the SDK when closing session without needing to cancel them in the app 

For the 1 point, I think we can just have some app scope injected (will be a bit like GlobalScope but will be more testable)
For the 2 point, I'm not sure how we should handle this? Any idea @Dominaezzz? using withContext with an inner scope?  Example : 

```
 override suspend fun createRoom(createRoomParams: CreateRoomParams): String {
        return withContext(scopeHolder.scope.coroutineContext){
            createRoomTask.execute(createRoomParams)
        }
    }
```
We have to keep in mind we might have some SDK inner launches too on this scope ^

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
bmarty: Done

bmarty:
@Dominaezzz @ganfra I've push 2 more commits.
Thanks for the discussion. We have to find a solution quickly, as this PR contains some crash fix for the next release.
I can also merge this PR and iterate later on the arch, but I would like to avoid that :).

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
Dominaezzz: Point 2 is the problematic case. Trying to make the SDK cancel stuff feels a bit artificial, this is something I think apps should take care of.
This is a pretty big design issue though, so it's definitely out of scope of what this PR is trying to fix.
Once everything has been migrated to suspend functions and flows I think it'll be easier to solve point 2. Any solution that is created now would just be something that will have to be fixed later.

Dominaezzz: https://github.com/Kotlin/kotlinx.coroutines/issues/1001 is also worth looking at for the withContext route.

ganfra:
So I would avoid exposing the scope on the sdk (even if its indirect through the launch method).
We still have to replace the task executor scope by the session one internally.
Otherwise I would go for a scope handled by the app. Maybe we can create an extension method on the session, like viewModelScope does?

ganfra:
I made one more commit where coroutine session scope is handled by the app. I kept the one ScopeHolder in the SDK because we should use it to replace TaskManager too. Let me know what do you think.

Dominaezzz:
I think it's a decent starting point, not sure how I feel about the locks (i.e `synchonized`) but it can revisited later.

bmarty:
@ganfra : `SessionCoroutineScopeHolder` is not used anymore, right?

bmarty:
Oki

PR #3208, opened 2021-04-20
Fix color issues when the system theme is changed (#2738)

### LINKED ISSUE #2738: Invisible or almost-invisible text with system theme
**Describe the bug**
I use the system theme and an app to automatically switch it (Auto Dark Theme). After a system theme switch, the text in Element is invisible, but toggling the Element theme between system and dark/light and back fixes it. I think Element doesn't notice that the text foreground color has changed.

I also noticed this with the bottom notification in the final step of the cross-signing flow. There's a "Finish >" step at the end and if you go back (cancel), the new session will remain inconsistent.

**Screenshots**

White on white text:
image

Gray on gray text:
image

 - Device: OnePlus 6T
 - OS: Android 10

**Additional context**
 - App version and store: 1.0.14 - Play Store
 - Homeserver: N/A

bmarty:
Tested OK on the room list and in the timeline

Fixes #2738

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Improvements 🙌:
 
 Bugfix 🐛:
  - Message states cosmetic changes (#3007)
+ - Fix color issues when the system theme is changed (#2738)
 
 Translations 🗣:
  -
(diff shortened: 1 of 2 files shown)


BillCarsonFr (approved):
LGTM

PR #3209, opened 2021-04-20
Initial support to use the Rust SDK for end to end encryption

poljar:
This is the initial version of EA utilizing the rust-sdk for the crypto duties. Most of the things in the `DefaultCryptoService` now utilize the Rust `OlmMachine` bound over to Kotlin using [uniffi-rs](https://github.com/mozilla/uniffi-rs/).

Initial performance measurements using an Xiaomi Mi A3 seem to be quite good, e.g. an initial key share before a message can be sent is done in ~12s for 2000 members each with a single device. Verification, cross signing, and key backups have not been touched yet and thus don't work anymore, everything else seems to work, minus some details that need polishing.

Build instructions for the bindings can be found in the [rust](https://github.com/poljar/element-android/tree/rust/rust-sdk#readme) subfolder of the repo.

Signed-off-by: Damir Jelić <poljar@termina.org.uk>

### Pull Request Checklist


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -457,63 +389,28 @@ internal class DefaultCryptoService @Inject constructor(
      */
     override fun getDeviceInfo(userId: String, deviceId: String?): CryptoDeviceInfo? {
         return if (userId.isNotEmpty() && !deviceId.isNullOrEmpty()) {
-            cryptoStore.getUserDevice(userId, deviceId)
+            runBlocking {
```
BillCarsonFr, line 392: we should avoid using runBlocking, exceptions could crash app. We used it mainly in unit test.
Can getDevice throw?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -457,63 +389,28 @@ internal class DefaultCryptoService @Inject constructor(
...
-                        isUpdated = true
-                    }
-                }
-
-                if (isUpdated) {
-                    cryptoStore.storeUserDevices(userId, storedDeviceIDs)
-                }
-            }
+        return runBlocking {
+            this@DefaultCryptoService.olmMachine?.getLiveDevices(userIds) ?: LiveDevice(userIds, deviceObserver)
```
BillCarsonFr, line 412: What's the default value here?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -524,21 +421,19 @@ internal class DefaultCryptoService @Inject constructor(
      * @param deviceId   the unique identifier for the device.
      */
     override fun setDeviceVerification(trustLevel: DeviceTrustLevel, userId: String, deviceId: String) {
-        setDeviceVerificationAction.handle(trustLevel, userId, deviceId)
+        // TODO
```
BillCarsonFr, line 424: rust SDK is not yet storing trust of devices?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -549,21 +444,12 @@ internal class DefaultCryptoService @Inject constructor(
             return false
         }
 
-        val encryptingClass = MXCryptoAlgorithms.hasEncryptorClassForAlgorithm(algorithm)
-
-        if (!encryptingClass) {
+        if (algorithm != MXCRYPTO_ALGORITHM_MEGOLM) {
```
BillCarsonFr, line 447: OLM encrypted room not supported anymore? (not sure if current SDK handles it correctly though, web does)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -574,12 +460,7 @@ internal class DefaultCryptoService @Inject constructor(
             Timber.v("Enabling encryption in $roomId for the first time; invalidating device lists for all users therein")
 
             val userIds = ArrayList(membersId)
-
-            deviceListManager.startTrackingDeviceList(userIds)
-
-            if (!inhibitDeviceQuery) {
-                deviceListManager.refreshOutdatedDeviceLists()
-            }
+            olmMachine!!.updateTrackedUsers(userIds)
```
BillCarsonFr, line 463: Why unwrap here and not anywhere else? Maybe olmMachine could be defined as lateinit var?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: That's correct, is that even used anywhere anymore? The spec does mention this for example:

> In order to enable end-to-end encryption in a room, clients can send an m.room.encryption state event specifying m.megolm.v1.aes-sha2 as its algorithm property.

And nowhere mentions `m.olm.*` inside rooms, though I guess it doesn't really forbid it either.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -636,44 +517,28 @@ internal class DefaultCryptoService @Inject constructor(
...
-            if (safeAlgorithm != null) {
+            val algorithm = getEncryptionAlgorithm(roomId)
+
+            if (algorithm != null) {
+                val userIds = getRoomUserIds(roomId)
                 val t0 = System.currentTimeMillis()
                 Timber.v("## CRYPTO | encryptEventContent() starts")
                 runCatching {
-                    val content = safeAlgorithm.encryptEventContent(eventContent, eventType, userIds)
+                    preshareRoomKey(roomId, userIds)
```
BillCarsonFr, line 527: can't this be done via algorithm? and the sharing done only for megolm?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -856,8 +580,16 @@ internal class DefaultCryptoService @Inject constructor(
             return
         }
         cryptoCoroutineScope.launch(coroutineDispatchers.crypto) {
-            val userIds = getRoomUserIds(roomId)
-            setEncryptionInRoom(roomId, event.content?.get("algorithm")?.toString(), true, userIds)
+            val params = LoadRoomMembersTask.Params(roomId)
+            try {
```
BillCarsonFr, line 584: This would defeat lazy loading of room members no?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -901,27 +636,182 @@ internal class DefaultCryptoService @Inject constructor(
...
+        toDevice: ToDeviceSyncResponse?,
+        deviceChanges: DeviceListResponse?,
+        keyCounts: DeviceOneTimeKeysCountSyncResponse?) {
+            // Decrypt and handle our to-device events
+            val toDeviceEvents = this.olmMachine!!.receiveSyncChanges(toDevice, deviceChanges, keyCounts)
+
+            // Notify the our listeners about room keys so decryption is retried.
+            if (toDeviceEvents.events != null) {
+                toDeviceEvents.events.forEach { event ->
+                    if (event.type == "m.room_key") {
```
BillCarsonFr, line 666: EventType.ROOM_KEY

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: If we don't correctly update our tracked users we or the other side might have undecryptable messages thus it's an error to not initialize the `olmMachine` before we hit this part of the code. 

Fetching devices out of the store on the other hand doesn't have this problem and callers handle the case where there isn't a device well enough.

There is a bit of a dissonance here since the previous store could be accessed before you initialized the `CryptoService` while the `OlmMachine` requires a valid `deviceId` to be initialized and the store is abstracted behind the `OlmMachine`. That's why I suggested to move the initialization logic somewhere else than what we currently do, in the sync loop.

It would probably make sense for the `CryptoService` to require the `deviceId` and for it to be initialized as soon as we get a valid device ID from the login and later when we load the session info, wherever that might be.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -901,27 +636,182 @@ internal class DefaultCryptoService @Inject constructor(
...
+                        val content = event.getClearContent().toModel<RoomKeyContent>() ?: return@forEach
 
-        rest = rest.copy(
-                signatures = objectSigner.signObject(canonicalJson)
+                        val roomId = content.sessionId ?: return@forEach
+                        val sessionId = content.sessionId
+
+                        notifyRoomKeyReceival(roomId, sessionId)
+                    } else if (event.type == "m.forwarded_room_key") {
```
BillCarsonFr, line 673: use EventTyp.FORWARDED_ROOM_KEY

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: It does, but as verification isn't yet implemented, there wasn't a point in forwarding the info from the rust side, I also need to expose, on the rust side that is, if it's locally verified or cross signing verified, currently there's just a boolean.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: How else would you convert a suspendable function into a blocking one? Do note that the actual reads are happening on a rust thread-pool and mostly cached. 

It can throw if there is a error with the store, though that was true for the old cryptostore as well and I'm not sure if that's propagated up the stack.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: An empty list of devices.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: This was already the case before I got here: https://github.com/vector-im/element-android/commit/af338b06078d790d7cf22a56deda510f285f3645#diff-10f33a2aedee42d5ed7a9b1db212f3f0992a6852592e700d04e66ca34ad2beaaR764

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: What do you mean by that? Do you mean by `algorithm`, like it was done before where we fetch an opaque interface? If so, I would argue that the additional complexity of such an abstraction for now isn't warranted, unless we really want to support `m.olm.*` inside of rooms.

If there's a `megolm.v2` the interface won't change and the difference will be handled inside the rust code, if there's a completely different algorithm (`mls` for example) we'll likely need to change the logic around depending on the algorithm.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: Oh also, no idea about the `lateinit`, I don't really know how that's working.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
bmarty: I guess the proper solution is to make `fun getDeviceInfo(...)` suspendable as well. Same for the other methods. I can help to make this change, please let me know @poljar

poljar: Well sure, but that changes the public API and my intention was to keep changes to the public API minimal for the first pass so we don't introduce bugs because of API changes.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
bmarty: `lateinit` is used to tell that a var will not be null when necessary, but will be not initialized in the constructor, but later. So in this case, if `olmMachine` is declared as lateinit, there will be no need to use `!!`. If used without being initialized, it will still crash of course.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
bmarty: The abstraction was a very legacy thing, imported from the previous SDK. I agree that this is maybe not necessary for now.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
poljar: Yeah, I think I'm using `lateinit`  nowadays somewhere in the verification service.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OlmMachine.kt
```diff
@@ -0,0 +1,536 @@
...
+     * @return The list of Devices or an empty list if there aren't any.
+     */
+    @Throws(CryptoStoreErrorException::class)
+    suspend fun getUserDevices(userId: String): List<CryptoDeviceInfo> {
+        val ownDevice = ownDevice()
+        val devices = inner.getUserDevices(userId).map { toCryptoDeviceInfo(it) }.toMutableList()
+
+        // EA doesn't differentiate much between our own and other devices of
+        // while the rust-sdk does, append our own device here.
+        devices.add(ownDevice)
```
bmarty, line 464: This is strange, you add the current device, event if the `userId` in parameter is not the current user?

poljar: Right, that's wrong. Nice catch.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -1008,31 +856,10 @@ internal class DefaultCryptoService @Inject constructor(
      * @param warn true to warn when some unknown devices are detected.
      */
     override fun setWarnOnUnknownDevices(warn: Boolean) {
+        // TODO this doesn't seem to be used anymore?
```
bmarty, line 859: This has been imported from the legacy SDK, still waiting to be plugged, but maybe we can just remove it for now.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -685,7 +550,10 @@ internal class DefaultCryptoService @Inject constructor(
      */
     @Throws(MXCryptoError::class)
     override fun decryptEvent(event: Event, timeline: String): MXEventDecryptionResult {
-        return internalDecryptEvent(event, timeline)
+
+        return runBlocking {
+            olmMachine!!.decryptRoomEvent(event)
```
bmarty, line 555: the `timeline: String` parameter is not used here?

poljar: Well it's not really clear what this is supposed to be used for, I don't think I saw any docs for this, the rust-sdk doesn't really know what a timeline is, all the info necessary to decrypt a megolm encrypted event is part of the event.

bmarty: The javadoc says `It is used to prevent replay attack.`, I have no other info on it

poljar: That's handled inside the rust-sdk, all the info to combat replay attacks is contained in the event as well.

poljar:
This has now been updated to include verification support, latest develop has been merged in as well.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
bmarty: ok, so we will do the cleanup later

poljar:
Updated to include cross signing support and merged the latest develop in.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -907,27 +813,168 @@ internal class DefaultCryptoService @Inject constructor(
...
+                    else                         -> {
+                        this.verificationService?.onEvent(event)
+                    }
+                }
+            }
+        }
+    }
+
+    private suspend fun preshareRoomKey(roomId: String, roomMembers: List<String>) {
+        keyClaimLock.withLock {
```
BillCarsonFr, line 869: Isn't this locking a bit too much? Like it will keep lock while doing server request?

poljar: Why is that too much? Is it a problem to keep the lock while sending the request out? 

The point is to not generate and send out the same request twice. The state machine will generate the same request multiple times unless you tell it that the given request was sent out.

poljar:
Updated to include back up support and merged the latest develop in.

The backup logic stayed largely the same, it just uses Rust objects to decrypt room keys and generating the request to upload the room keys is generated on the Rust side.

Not completely happy with that API but I don't have a better idea considering that the Rust side doesn't do any requests.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -907,27 +649,197 @@ internal class DefaultCryptoService @Inject constructor(
...
+
+    suspend fun receiveSyncChanges(
+            toDevice: ToDeviceSyncResponse?,
+            deviceChanges: DeviceListResponse?,
+            keyCounts: DeviceOneTimeKeysCountSyncResponse?) {
+        // Decrypt and handle our to-device events
+        val toDeviceEvents = this.olmMachine.receiveSyncChanges(toDevice, deviceChanges, keyCounts)
+
+        // Notify the our listeners about room keys so decryption is retried.
+        if (toDeviceEvents.events != null) {
```
BillCarsonFr, line 677: Would it be possible to have an API from olmMachine that would fire some events like `ROOM_KEY_IMPORTED`and `SECRET_RECEIVED`.

It woud be particularly usefull for secrets, as the rust-sdk could then give also the secret name. Because actually we only have the secret value, so no idea what to do with it (here for example the ssk/usk/msk seems to be sent to the keybackup service)

poljar: Firing events, while not impossible, would:
1. Require us to pass in callbacks over FFI
2. Mess with the order of execution, i.e. the callback would probably fire before we have everything handled on the Rust side
3. Introduce dissonance into the simple push/pull design

The `receiveSyncChanges` method, of course, can return arbitrary data, so we can include extra info you need in the return value.

What's exactly needed for room keys? I agree that the secret send thing sucks and may need additional data.

BillCarsonFr: > What's exactly needed for room keys?

It's used by the opened timeline. When a new megolm session is received it will trigger a refresh (will retry to decrypt UISIs) to display properly the event on screen.

poljar: > It's used by the opened timeline. When a new megolm session is received it will trigger a refresh (will retry to decrypt UISIs) to display properly the event on screen.

I know that, but isn't that already happening? The return value of `receiveSyncChanges` contains the decrypted to-device events. `m.room_key` events will be there as well, just that the `session_key` will be an empty string.

Am I missing something? Isn't the notifying happening right here: https://github.com/poljar/element-android/blob/rust-verification/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt#L758.

BillCarsonFr: Yes it's working fine for new session in this case. But I am not sure it will work when i am getting a key from backup in background. Would have to also call notifyRoomKeyReceived after a keybackup import

poljar: Ah OK. 

I think it makes more sense to modify the return value of `OlmMachine.importKeys` and `OlmMachine.importDecryptedKeys` to tell us which room keys were imported than to add callbacks that get called when the Rust side receives a room key.

Would that work?

BillCarsonFr: Yes would allow me to notify from cryptoService. I would have to refactor a bit cause now only the DefaultCryptoService is aware of the listeners, the KeybackupService is not and can't reference back the cryptoservice. Could introduced a new class of session scope to handle that and be injected in both

poljar: Importing now returns a map of keys that were imported, see this commit: https://github.com/vector-im/element-android/pull/3209/commits/d138306b0826fb81ae5922c9c595012760ba6bfd.

You just need to traverse the map after the import and call the notifiers.

bmarty:
It was a temporary PR, so I delete it. It will be back later, in a mergeable form.

PR #3229, opened 2021-04-26
Fix intent resolving behavior for Android 11

warwicklol14:
Fixes : #3067 

Signed-off-by: Varun Paul kjvpaul@tutanota.com

### Pull Request Checklist

- [X] Changes has been tested on an Android device or Android emulator with API 21
- [ ] UI change has been tested on both light and dark themes
- [X] Pull request is based on the develop branch
- [X] Pull request updates [CHANGES.md](https://github.com/vector-im/element-android/blob/develop/CHANGES.md)
- [ ] Pull request includes screenshots or videos if containing UI changes
- [X] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)

PR #3238, opened 2021-04-28
Android 11 fixes an other fixes for attachement

### LINKED ISSUE #2381: Install APKs from chats
Sometimes I want to send an APK file to a friend so he can install it. If I click on the file to open it, it just does nothing.
In the system log I can see the message: `5470-5470/com.android.packageinstaller E/InstallStart: Requesting uid 10106 needs to declare permission android.permission.REQUEST_INSTALL_PACKAGES`

Add the permission `android.permission.REQUEST_INSTALL_PACKAGES` to manifest. If you want, I could create a pull request for that. Tested this already, it would work. No more work is needed.

Don't know any alternatives as the permission is required since Android 7.1.

**Additional context**
The screenshot shows the package installer asking the user to grant the permission. (after adding the permission to manifest):

Screenshot_1605205738

### LINKED ISSUE #3067: Android 11: "Sorry, no external application has been found to complete this action" when trying to open a PDF
#### Describe the bug
When tapping a PDF file someone sent me (or I sent), it doesn't open, but `error_no_external_application_found` ("Sorry, no external application has been found to complete this action.") shows up instead. The only workaround right now is to long-press the file, save locally, and then open it from the file download notification or a file manager.

#### To Reproduce
1. Receive a PDF (maybe sending one yourself works as well, I tested it with one I sent via the desktop app)
2. Click on the file
3. See error

#### Expected behavior
The PDF file opening in any PDF viewer on the phone

 - Device: Pixel 5 (happens on Pixel 3a XL as well)
 - OS: Android 11

#### Additional context
 - App version and store: 1.1.3 via Play store
 - Homeserver: private (ran by me, so if it might be homeserver-related (which I doubt) let me know)

### LINKED ISSUE #3186: Cannot open encrypted file after restarting the app
#### Describe the bug
Whenever the app is started all the cached decrypted files are deleted. Due to this we cannot open files in an encrypted room.

#### To Reproduce
1. Upload a file in an encrypted room.
2. Close the app and remove it from the recent apps.
3. Open the app again and click on the uploaded file.
4. Expected UX is that the file opens but nothing happens.

#### Expected behavior
My expected behavior is that file opens in the desired app when you click on it.

 - Device: Samsung M31
 - OS: Android 11

#### Additional context
 - App version and store: 1.1.3 - F-Droid
 - Homeserver: matrix.org

I tried to find the cause of the bug and saw that to open the file we needed to get a temporaryShareableURI
https://github.com/vector-im/element-android/blob/9565e8975e31f8b873369e22e29d51824f8b20ca/vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt#L1154

But when I tried to debug the app, this URI always was null so the event would never get posted causing nothing to happen when the file is clicked.

The reason the URI was always null is that
https://github.com/vector-im/element-android/blob/9565e8975e31f8b873369e22e29d51824f8b20ca/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultFileService.kt#L279

The reason targetFile doesn't exist is that the following file is assigned to it which doesn't exist when you restart the app.
https://github.com/vector-im/element-android/blob/9565e8975e31f8b873369e22e29d51824f8b20ca/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultFileService.kt#L244

The cached decrypted file is deleted whenever the app is started
https://github.com/vector-im/element-android/blob/9565e8975e31f8b873369e22e29d51824f8b20ca/vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt#L84

The getClearFile() method just returns the decryptedFile but it doesn't exist
https://github.com/vector-im/element-android/blob/9565e8975e31f8b873369e22e29d51824f8b20ca/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultFileService.kt#L231

So the desired behavior is that `getTemporarySharableURI()` should instead of returning null, decrypt the file and send it.

#### Screenshots
image

This is the way element-desktop handles it.

bmarty:
Fixes #3067 (and replace #3229): open file using external app from Element. I've done a lot of test using an Android 11 emulator, and I haven't found other issue using Intent (save key, etc.)

Also fixes #3186 and fixes #2381

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@ Features ✨:
  -
 
 Improvements 🙌:
- -
+ - Add ability to install APK from directly from Element (#2381)
 
 Bugfix 🐛:
  - Message states cosmetic changes (#3007)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@ Bugfix 🐛:
  - Fix exception in rxSingle (#3180)
  - Do not invite the current user when creating a room (#3123)
  - Fix color issues when the system theme is changed (#2738)
+ - Fix issues on Android 11 (#3067)
 
 Translations 🗣:
  -
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Bugfix 🐛:
  - Do not invite the current user when creating a room (#3123)
  - Fix color issues when the system theme is changed (#2738)
  - Fix issues on Android 11 (#3067)
+ - Fix issue when opening encrypted files (#3186)
 
 Translations 🗣:
  -
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/file/FileService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/file/FileService.kt
@@ -29,14 +29,19 @@ import java.io.File
  */
 interface FileService {
 
-    enum class FileState {
-        IN_CACHE,
-        DOWNLOADING,
-        UNKNOWN
+    sealed class FileState {
+        /**
+         * The original file is in cache, but the decrypted files can be deleted for security reason.
+         * To decrypt the file again, call [downloadFile], the encrypted file will not be downloaded again
+         * @param decryptedFileInCache true if the decrypted file is available. Always true for clear files.
+         */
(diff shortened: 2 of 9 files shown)


BillCarsonFr (approved):
LGTM

PR #2411, opened 2020-11-17
Added option to record and send videos.

c-andy-proj:
Signed-off-by: Andrea Conalbi <c.andy@aleeas.com>

### Pull Request Checklist

- [X] Changes has been tested on an Android device or Android emulator with API 21
- [ ] UI change has been tested on both light and dark themes
- [X] Pull request is based on the develop branch
- [X] Pull request updates [CHANGES.md](https://github.com/vector-im/element-android/blob/develop/CHANGES.md)
- [ ] Pull request includes screenshots or videos if containing UI changes
- [X] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)

I basically adapted the code of the "camera attachment" for video recording adding a button in the attachments list.
Sorry if I made mistakes, it's my first PR ever.

PR #3264, opened 2021-04-30
Compress video and improve file too big error detection

### LINKED ISSUE #3177: [Bug] Mixed media sending will stuck
#### Describe the bug
Mixed media contents will not send and give back a error with shows continuously resending

#### To Reproduce
1. Select some mixed Videos and Images and try to send

 - Device: Huawei Mate 20 lite
 - OS: Android 10

#### Additional context
 - App version: develop
 - Homeserver: master

bmarty:
Add support to compress video before sending.

Improve file too big detection. It's now handle by the SDK, since the file can be compressed, and so become acceptable for the server. The message bottom sheet now display the "too big file" error to the user.

image

Also the SDK checks the size after the encryption of the file (can be bigger when encrypted).

Thumbnails are now sent after the main content, to avoid sending them and then get an error when sending the main file.

Add support for preview of video files (no play preview though, just an indicator in the strip to show it's a video), to let the user selects the option to send the original uncompressed video.

image

Remaining issue (to be fixed in another PR?): if file is too large for the server when retrying, it forces to compress the file and so it has more chance to work. But it is not the original intent from the user.

This PR also:
- fixes a regression on the fail to send indicator color.
- avoid "soft_logout" to be serialized by making it optional

EDIT:

Now when we do multiple selection, it handle correctly the several Event type (fixes #3177)

We can now take a video (this PR replaces #2411) Design input will be updated later, for now we are using a dialog when clicking on "Camera":

image

There also a new setting "Take photo or video" to update the stored value:

image

image

EDIT 2:

Added some information about attachment in the bottom sheet: when applicable user will see: duration, dimensions, size and mime type (for files only).

Example for a video:

image


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/content/VideoCompressor.kt
```diff
@@ -0,0 +1,83 @@
...
+                        }
+
+                        override fun onFailure(failureMessage: String) {
+                            Timber.d("Compressing: failure: $failureMessage")
+                            job.completeExceptionally(Exception(failureMessage))
+                        }
+
+                        override fun onCancelled() {
+                            Timber.d("Compressing: cancel")
+                            job.cancel()
```
bmarty, line 67: @Dominaezzz there is probably room for improvement here, on https://github.com/AbedElazizShe/LightCompressor API, but also on my workaround :)

Dominaezzz: I'm slightly confused by the use of both `Job` and `Dispatchers.IO`, I think one of them has to go.
In reality this API should've exposed a flow but oh well, it's not too hard to wrap.

DoM1niC:
Hey @bmarty  :) 
I hope this will fixed this issue #3177 too 👍🏻 nice work btw!

DoM1niC:
> Hey @bmarty :)
> I hope this will fixed this issue #3177 too 👍🏻 nice work btw!

is fixed with this PR THX!!!!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/content/VideoCompressor.kt
bmarty: I've added the support to track the progress, but I'm a bit confused too about how I can improve that. The Dispatchers.IO is here because we are creating a file, and the Job is here to make the call "synchronous". What do you suggest?

Dominaezzz: Looking at https://natario1.github.io/Transcoder/docs/events , (unfortunately) transcoding is already done on a background thread, so you don't need to use `Dispatchers.IO`. You could also just limit it to just the `createDestinationFile()`.

(I can't wait until `ProgressListener` is replaced with `Flow`)

bmarty: Thanks for the feedback, I will change that. By the way the link you provide also shows me that there is a `successCode` in the `onTranscodeCompleted` fun. I did not noticed that. When the value is `SUCCESS_NOT_NEEDED` it appears that the target file is not written, so it explain a bug I observed today with a 0-lenght file. I will fix that too.

BillCarsonFr (approved):
LGTM, you should update open_source_licence.html with transcoder lib

bmarty:
@BillCarsonFr the file `open_source_licence.html` is on the application side. Another PR could add licence for eveything which is used by the MatrixSDK, IDK...

PR #3058, opened 2021-03-25
ci: add initial github actions pipelines

abn:
### Pull Request Checklist

- [ ] Changes has been tested on an Android device or Android emulator with API 21
- [ ] UI change has been tested on both light and dark themes
- [x] Pull request is based on the develop branch
- [ ] Pull request updates [CHANGES.md](https://github.com/vector-im/element-android/blob/develop/CHANGES.md)
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)

PR #3294, opened 2021-05-06
GitHub action

bmarty:
Follow up what have been done on #3058


bmarty:
Let's merge it like that for now.

PR #3305, opened 2021-05-10
Upgrade Realm to version 10.4.0

bmarty:
Migration with isEmbedded tested OK

Ref: https://github.com/realm/realm-java/issues/7402


ganfra (approved):
LGTM, thanks!

PR #3307, opened 2021-05-10
Allow inviting users even if profile information cannot be retrieved

### LINKED ISSUE #3306: Fix user invitation in case of restricted profile api
If the server is configured as `limit_profile_requests_to_users_who_share_rooms: true` then profile API will not return user information, instead, the response will be `403`. Invitation of valid user ids should be allowed.

onurays:
Fixes #3306.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/profile/ProfileService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/profile/ProfileService.kt
@@ -66,7 +66,7 @@ interface ProfileService {
     /**
      * Get the combined profile information for this user.
      * This may return keys which are not limited to displayname or avatar_url.
-     * If server is configured as limit_profile_requests_to_users_who_share_rooms: true then response will be HTTP 403.
+     * If server is configured as limit_profile_requests_to_users_who_share_rooms: true then response can be HTTP 403.
      * @param userId the userId param to look for
      *
      */
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/profile/ProfileService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/profile/ProfileService.kt
@@ -66,6 +66,7 @@ interface ProfileService {
     /**
      * Get the combined profile information for this user.
      * This may return keys which are not limited to displayname or avatar_url.
+     * If server is configured as limit_profile_requests_to_users_who_share_rooms: true then response will be HTTP 403.
      * @param userId the userId param to look for
      *
      */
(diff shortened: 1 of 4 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/profile/ProfileService.kt
```diff
@@ -66,6 +66,7 @@ interface ProfileService {
     /**
      * Get the combined profile information for this user.
      * This may return keys which are not limited to displayname or avatar_url.
+     * If server is configured as limit_profile_requests_to_users_who_share_rooms: true then response will be HTTP 403.
```
bmarty, line 69:
```suggestion
     * If server is configured as limit_profile_requests_to_users_who_share_rooms: true then response can be HTTP 403.
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/profile/ProfileAPI.kt
```diff
@@ -33,6 +33,7 @@ internal interface ProfileAPI {
      * Get the combined profile information for this user.
      * This API may be used to fetch the user's own profile information or other users; either locally or on remote homeservers.
      * This API may return keys which are not limited to displayname or avatar_url.
+     * If server is configured as limit_profile_requests_to_users_who_share_rooms: true then response will be HTTP 403.
```
bmarty, line 36:
```suggestion
     * If server is configured as limit_profile_requests_to_users_who_share_rooms: true then response can be HTTP 403.
```

bmarty (approved):
Just some remark about the doc, else LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/profile/ProfileAPI.kt
onurays: Done.

vector/src/main/java/im/vector/app/features/userdirectory/UserListViewModel.kt
```diff
@@ -137,7 +136,15 @@ class UserListViewModel @AssistedInject constructor(@Assisted initialState: User
                                                 avatarUrl = json[ProfileService.AVATAR_URL_KEY] as? String
                                         ).toOptional()
                                     }
-                                    .onErrorReturn { Optional.empty() }
+                                    .onErrorReturn {
+                                        // Profile API can be restricted and doesn't have to return result.
+                                        // In this case allow inviting valid user ids.
+                                        User(
+                                                userId = search,
+                                                displayName = search,
```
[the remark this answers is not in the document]
onurays: Done (web is not doing so though).

PR #3308, opened 2021-05-10
Fix read marker visibility/position when filtering some events

ganfra:
When some events were filtered the read marker position processing was breaking thing


bmarty (approved):
I'm blindly trust you on this one, I do not know what was the issue, and I hardly understand the changes

PR #3322, opened 2021-05-12
Add option to send beta feedback

BillCarsonFr:
Add option to send feedback for the space beta

-> Added new checkbox to give consent to be contacted 
image

image

image


vector/src/main/java/im/vector/app/features/rageshake/BugReportActivity.kt
```diff
@@ -60,30 +63,48 @@ class BugReportActivity : VectorBaseActivity<ActivityBugReportBinding>() {
...
+            tryOrNull { ReportType.valueOf(it) }
+        } ?: ReportType.BUG_REPORT
 
         // Default screen is for bug report, so modify it for suggestion
-        if (forSuggestion) {
-            supportActionBar?.setTitle(R.string.send_suggestion)
+        when (reportType) {
+            ReportType.BUG_REPORT -> {
+                supportActionBar?.setTitle(R.string.title_activity_bug_report)
+                views.bugReportButtonContactMe.isVisible = false
```
bmarty, line 74: Why not keeping this new option also for bug report?

vector/src/main/res/values/strings.xml
```diff
@@ -2167,6 +2167,14 @@
     <string name="send_suggestion_sent">Thanks, the suggestion has been successfully sent</string>
     <string name="send_suggestion_failed">The suggestion failed to be sent (%s)</string>
 
+    <string name="send_feedback_space_title">Spaces feedback</string>
+    <string name="feedback">Feedback</string>
+    <string name="send_feedback_space_info">You’re using a beta version of spaces.  Your feedback will help inform the next versions.  Your platform and username will be noted to help us use your feedback as much as we can. To leave the beta, visit your settings.</string>
```
bmarty, line 2172: There are extra spaces char: `spaces.  Your feedback` and `sions.  Your`. It happens when copy pasting from Figma.

vector/src/main/res/values/strings.xml
```diff
@@ -2167,6 +2167,14 @@
...
+    <string name="you_may_contact_me">You may contact me if you have any follow up questions</string>
+    <string name="space_feedback_sent">Thanks, your feedback has been successfully sent</string>
+    <string name="space_feedback_failed">The feedback failed to be sent (%s)</string>
```
bmarty, line 2175: Maybe use more generic ids: `sending_feedback_sent` and `sending_feedback_failed`

vector/src/main/res/values/strings.xml
bmarty, line 2172: Can we really "leave the beta"? I do not think so.

bmarty (changes requested):
Thanks. Some remarks

vector/src/main/java/im/vector/app/features/rageshake/BugReportActivity.kt
BillCarsonFr: ok

vector/src/main/res/values/strings.xml
BillCarsonFr: done

vector/src/main/res/values/strings.xml
BillCarsonFr: done

vector/src/main/res/values/strings.xml
```diff
@@ -2164,15 +2164,13 @@
     <string name="send_suggestion">Make a suggestion</string>
     <string name="send_suggestion_content">Please write your suggestion below.</string>
     <string name="send_suggestion_report_placeholder">Describe your suggestion here</string>
-    <string name="send_suggestion_sent">Thanks, the suggestion has been successfully sent</string>
-    <string name="send_suggestion_failed">The suggestion failed to be sent (%s)</string>
```
bmarty, line 2168: lint won't like that, you have to also delete all the translations. Maybe just keep it for now and we will do the cleanup later

niquewoodhouse:
Could we please alter the "spaces beta box" on the menu?

It's taking up a lot of space, especially with adding the feedback link.

Now it's not a dismissable box, we could drop some of the UI to be like the right instead of the left:

image

Removing some of the vertical height, making Home visible sooner. [Figma](https://www.figma.com/file/baCEQP0BvtuitBJUAtapoc/Spaces-Beta?node-id=3994%3A3644)

vector/src/main/res/values/strings.xml
BillCarsonFr: done

BillCarsonFr:
Updated :

image

bmarty (approved):

PR #3323, opened 2021-05-12
WIP: Add low bandwidth support

kegsay:
*Foreword: first time contributing to this codebase, and I know very little Kotlin so my code may look a bit weird!*

This PR adds support for [MSC3079: Low Bandwidth CS API](https://github.com/matrix-org/matrix-doc/pull/3079).

The core impl is [in Go](https://github.com/matrix-org/lb-android) and is compiled to an `.aar` which is then pulled in via jitpack. A `LowBandwidthInterceptor` is then added to both the authenticated and unauthenticated clients to make HTTP requests call through to the Go code. Manually tested.

This pull request isn't ready to merge yet because:
 - This should be gated behind a labs option so we don't always add the interceptor unless we explicitly opt-in.
 - I'm not sure if there's a better way to pull in the `.aar`. Currently the `lb-android` repo just checks in the file which is yuck as it's 25MB(!)
 - When building for debug, `insecureSkipVerify` should be true, else it should be false in production (so we do certificate checks).

I would appreciate it if someone from the EA team would help me get this PR over the line to a standard which is acceptable to merge, as you guys will know how to integrate this better than me!

### Pull Request Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
@@ -21,6 +21,7 @@ import androidx.lifecycle.ProcessLifecycleOwner
 import androidx.work.Configuration
 import androidx.work.WorkManager
 import com.zhuinden.monarchy.Monarchy
+import mobile.Mobile
 import org.matrix.android.sdk.BuildConfig
 import org.matrix.android.sdk.api.auth.AuthenticationService
 import org.matrix.android.sdk.api.auth.HomeServerHistoryService
@@ -55,6 +56,11 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
     @Inject internal lateinit var apiInterceptor: ApiInterceptor
 
     init {
+        // First time setup for development
+        val cp = Mobile.params()
+        cp.insecureSkipVerify = true
+        Mobile.setParams(cp)
+        
         Monarchy.init(context)
         DaggerMatrixComponent.factory().create(context, matrixConfiguration).inject(this)
         if (context.applicationContext !is Configuration.Provider) {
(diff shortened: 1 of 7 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionModule.kt
```diff
@@ -218,10 +219,12 @@ internal abstract class SessionModule {
         fun providesOkHttpClient(@UnauthenticatedWithCertificate okHttpClient: OkHttpClient,
                                  @Authenticated accessTokenProvider: AccessTokenProvider,
                                  @SessionId sessionId: String,
+                                 lowBandwidthInterceptor: LowBandwidthInterceptor,
                                  @MockHttpInterceptor testInterceptor: TestInterceptor?): OkHttpClient {
             return okHttpClient
                     .newBuilder()
                     .addAccessTokenInterceptor(accessTokenProvider)
+                    .addInterceptor(lowBandwidthInterceptor)
```
bmarty, line 227: Why do you add this interceptor to the Authenticated okHttpClient and not instead to the unauthenticated httpClient?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
```diff
@@ -21,6 +21,7 @@ import androidx.lifecycle.ProcessLifecycleOwner
 import androidx.work.Configuration
 import androidx.work.WorkManager
 import com.zhuinden.monarchy.Monarchy
+import mobile.Mobile
```
bmarty, line 24: package name is a bit meaningless, is it possible to change it on your library?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionModule.kt
Kegsay: because confusingly the authenticated client is not derived from the unauthenticated client. So if you just add it to the unauthenticated client you lose the interceptor after logging in.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
Kegsay: I'll see what I can do. I probably can.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionModule.kt
bmarty: This is strange. Authenticated client should be derived from Unauthenticated client. Let me double check

kegsay:
Closing this; this branch will be used as an example for people who want to test low bandwidth locally. Code pushed to it will do things like disable TLS checks for use with self-signed certificates making this unsuitable to be merged.

witchent:
Will this low bandwidth mode come back in any form?

PR #3324, opened 2021-05-12
Add documentation on LoginWizard and RegistrationWizard (#3303)

### LINKED ISSUE #3303: Document RegistrationWizard
Related to #3217

bmarty:
Fixes #3303

@tladesignz what do you think? Is it more understandable to you with this documentation? I will appreciate your feedback :)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,7 +23,7 @@ Test:
  -
 
 Other changes:
- -
+ - Add documentation on LoginWizard and RegistrationWizard (#3303)
 
 Changes in Element 1.1.7 (2021-05-12)
 ===================================================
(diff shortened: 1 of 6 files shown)


tladesignz:
@bmarty: Wait... what? You should call `#createAccount` as a first step, always? I didn't know that and always expected it to be the other way around! Man...

This is definitely an improvement, but I suggest  some additions to the `RegistrationWizard` class comment along the line of this, to make it crystal clear:

> Steps to register successfully:
> 1. Call `#getRegistrationFlow`.
> 2. Call `#createAccount`.
> 3. Fulfill all mandatory stages.
> 4. (anything I still don't grasp, of course!)
> 
> (see https://github.com/vector-im/element-android/blob/main/docs/signup.md for a detailed explanation)


Thank you very much!

bmarty:
Thanks for the feedback @tladesignz . I've added what you're suggesting.

I recommend you to read to better understand how it works https://matrix.org/docs/spec/client_server/latest#user-interactive-authentication-api :)

tladesignz:
> Thanks for the feedback @tladesignz . I've added what you're suggesting.

Great, thanks!

> I recommend you to read to better understand how it works https://matrix.org/docs/spec/client_server/latest#user-interactive-authentication-api :)

You may be surprised, but I *did* read this the first time back in 2019 when I first started to implement signin and signup for our iOS client. And I got back to it whenever I had a specific question. *However*, I never ever read, that `createAccount` needs to come before the rest of the stages. Would be very happy, if you could point that out for me in there.

PR #3372, opened 2021-05-19
Fix | Explore room list: the RoomId is displayed instead of name (#3371)

### LINKED ISSUE #3371: Spaces | Explore room list: the RoomId is displayed instead of name
In #community:matrix.org 

image

BillCarsonFr:
Fixes #3371


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/MatrixItem.kt
```diff
@@ -159,4 +159,4 @@ fun RoomMemberSummary.toMatrixItem() = MatrixItem.UserItem(userId, displayName,
 
 fun SenderInfo.toMatrixItem() = MatrixItem.UserItem(userId, disambiguatedDisplayName, avatarUrl)
 
-fun SpaceChildInfo.toMatrixItem() = MatrixItem.RoomItem(childRoomId, name, avatarUrl)
+fun SpaceChildInfo.toMatrixItem() = MatrixItem.RoomItem(childRoomId, name ?: canonicalAlias ?: aliases?.firstOrNull(), avatarUrl)
```
bmarty, line 162: I remember that we should never use `aliases` as a room name fallback. Only canonical alias is permitted. But `getPrimaryAlias()` is still a thing...

bmarty (approved):
LGTM, one remark about usage of aliases

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/MatrixItem.kt
BillCarsonFr: Ok updated

PR #3373, opened 2021-05-19
Implement a workaround to render <del> and <u> in the timeline (#1817)

### LINKED ISSUE #1817: <del> doesn't work in sent messages
In Element Android 1.0.0, sending a message such as `<del>test</del>` (which is the only way to send strike-through text as far as I'm aware) causes the HTML markup to be shown in plain text instead of striking through the message ~~like this~~. Other clients can send such messages and they'll show up in Element as crossed-out just fine, they just can't be sent from this app.

bmarty:
Workaround to fix #1817

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageTextItem.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageTextItem.kt
@@ -41,6 +41,9 @@ abstract class MessageTextItem : AbsMessageItem<MessageTextItem.Holder>() {
     var message: CharSequence? = null
 
     @EpoxyAttribute
+    var canUseTextFuture: Boolean = true
+
+    @EpoxyAttribute
     var useBigFont: Boolean = false
 
     @EpoxyAttribute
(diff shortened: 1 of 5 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageTextItem.kt
```diff
@@ -80,17 +83,26 @@ abstract class MessageTextItem : AbsMessageItem<MessageTextItem.Holder>() {
                 it.bind(holder.messageView)
             }
         }
-        val textFuture = PrecomputedTextCompat.getTextFuture(
-                message ?: "",
-                TextViewCompat.getTextMetricsParams(holder.messageView),
-                null)
+        val textFuture = if (canUseTextFuture) {
```
ganfra, line 86: Why do we have 2 if(canUseTextFuture) ?

ganfra (approved):
One question otherwise LGTM (if it doesn't break computation?)

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageTextItem.kt
bmarty: One to compute it, one to apply it. I did not want to change method call order. I could test nullability of `textFuture` for the second one if you prefer, but it's the same

PR #3443, opened 2021-06-01
stable ids for MSC 2858

### LINKED ISSUE #3442: Switch to stable endpoint/fields for MSC2858
remember the brand ids change

BillCarsonFr:
Fixes #3442

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/data/LoginFlowResponse.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/data/LoginFlowResponse.kt
@@ -42,7 +42,7 @@ internal data class LoginFlow(
          * the client can show a button for each of the supported providers
          * See MSC #2858
          */
-        @Json(name = "org.matrix.msc2858.identity_providers")
+        @Json(name = "identity_providers")
         val ssoIdentityProvider: List<SsoIdentityProvider>? = null
 
 )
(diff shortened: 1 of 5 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/data/LoginFlowResponse.kt
```diff
@@ -42,7 +42,7 @@ internal data class LoginFlow(
          * the client can show a button for each of the supported providers
          * See MSC #2858
          */
-        @Json(name = "org.matrix.msc2858.identity_providers")
+        @Json(name = "identity_providers")
```
bmarty, line 45: Shouldn't we add a field for `identity_providers`, and use it first, or take `org.matrix.msc2858.identity_providers` if present?

bmarty (approved):
One remark, else LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/data/LoginFlowResponse.kt
BillCarsonFr: synapse is already using stable id since synapse 1.30, so we can just drop

bmarty: ok, fine

PR #3445, opened 2021-06-01
Migration to coroutines

bmarty:
Part of #2449: migrate DefaultTypingService, KeysImporter and KeysExporter to coroutines

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/typing/DefaultTypingService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/typing/DefaultTypingService.kt
@@ -43,7 +43,7 @@ internal class DefaultTypingService @AssistedInject constructor(
         fun create(roomId: String): DefaultTypingService
     }
 
-    private var job = Job()
+    private val coroutineScope = CoroutineScope(Job())
     private var currentTask: Job? = null
 
     // What the homeserver knows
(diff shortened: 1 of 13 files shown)


bmarty:
CC @Dominaezzz if you want to have a look :)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/typing/DefaultTypingService.kt
```diff
@@ -44,35 +43,33 @@ internal class DefaultTypingService @AssistedInject constructor(
         fun create(roomId: String): DefaultTypingService
     }
 
-    private var currentTask: Cancelable? = null
-    private var currentAutoStopTask: Cancelable? = null
+    private var job = Job()
```
Dominaezzz, line 46: I think this should be a `CoroutineScope` instead of `Job`.

ganfra (approved):
Not sure about the typing methods, otherwise LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/typing/DefaultTypingService.kt
bmarty: So you would replace `private var job = Job()` by `private var coroutineScope = CoroutineScope(Job())` ?

Dominaezzz: `private var coroutineScope = CoroutineScope()`

`Job` is added by default.

bmarty: The constructor of `CoroutineScope` requires a context

Dominaezzz: Hmm, maybe I'm using a different version.
Well, it's the same either way.

Issue #3451, opened 2021-06-02
Support Call asserted identity


ganfra:
Adds support for receiving MSC3086 Asserted Identity events & emitting event from the call object.
Should be disabled by default and enabled at compile time if needed.

PR #3452, opened 2021-06-02
Feature/fga/voip asserted identity

ganfra:
Handles https://github.com/vector-im/element-android/issues/3451

Adds support for receiving MSC3086 Asserted Identity events & emitting event from the call object.
Behavior disabled by default in gradle.


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/call/CallAssertedIdentityContent.kt
```diff
@@ -0,0 +1,57 @@
...
+         */
+        @Json(name = "call_id") override val callId: String,
+        /**
+         * Required. ID to let user identify remote echo of their own events
+         */
+        @Json(name = "party_id") override val partyId: String? = null,
+        /**
+         * Required. The version of the VoIP specification this messages adheres to.
+         */
+        @Json(name = "version") override val version: String?,
```
bmarty, line 38: Required but `?` and `callId` is Required but not `?`...

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/CallEventProcessor.kt
```diff
@@ -37,7 +37,9 @@ internal class CallEventProcessor @Inject constructor(private val callSignalingH
             EventType.CALL_CANDIDATES,
             EventType.CALL_INVITE,
             EventType.CALL_HANGUP,
-            EventType.ENCRYPTED
+            EventType.ENCRYPTED,
+            EventType.CALL_ASSERTED_IDENTITY,
+            EventType.CALL_ASSERTED_IDENTITY_PREFIX
```
bmarty, line 42: What is the point to use both with and without PREFIX?
I think the version without Prefix should be added later, once the MSC is merged

vector/src/main/java/im/vector/app/features/call/VectorCallViewModel.kt
```diff
@@ -202,6 +208,22 @@ class VectorCallViewModel @AssistedInject constructor(
...
+    private fun WebRtcCall.extractCallInfo(): VectorCallViewState.CallInfo {
+        val assertedIdentity = this.remoteAssertedIdentity
+        val matrixItem = if (assertedIdentity != null) {
+            val userId = if (MatrixPatterns.isUserId(assertedIdentity.id)) {
+                assertedIdentity.id!!
+            } else {
+                // Need an id starting with @
+                "@${assertedIdentity.displayName}"
+            }
+            MatrixItem.UserItem(userId, assertedIdentity.displayName, assertedIdentity.avatarUrl)
```
bmarty, line 220: Create a `fun AssertedIdentity.toMatrixItem()` in the SDK, as per every other supported types?

vector/src/main/res/layout/activity_call.xml
```diff
@@ -37,23 +37,26 @@
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
-    <FrameLayout
+    <androidx.cardview.widget.CardView
```
bmarty, line 40: please use MaterialCardView. Also you should merge or rebase to develop

bmarty (commented):
LGTM, static review
Can you merge develop or rebase your branch please?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/call/CallAssertedIdentityContent.kt
ganfra: Yes it's because it changed from double to string, we need retrocompat

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/CallEventProcessor.kt
ganfra: Copied Web on that point, so not sure.

vector/src/main/java/im/vector/app/features/call/VectorCallViewModel.kt
ganfra: I can, but it was a bit weird, I just need that for AvatarRendering...

bmarty: That's one of the main point of MatrixItem, to be given to the avatar renderer :)

PR #3456, opened 2021-06-03
Feature/fga/dial pad tab

### LINKED ISSUE #2593: One-to-one rooms (direct chats) listed under rooms
Element android shows some one-to-one chat rooms under rooms instead of direct chats for me. This happens mainly with rooms created by the [mautrx-telegram](https://github.com/tulir/mautrix-telegram) bridge. But since element desktop sorts them correctly I assume it's a problem with element android.

I double checked, the one-to-one rooms only have two members each including me. I did not found a pattern. Around half of the one-to-one rooms are listed under rooms, the other half under direct chats.

**Smartphone**
 - Device: Samsung Galaxy S10e
 - OS: Android 10

**Additional context**
 - App version and store 1.0.13 Play Store
 - Homeserver: own, private homeserver with synapse

ganfra:
This PR moves the start call with dial pad from room detail to a tab in home.
It's still dynamic and only shows if the server supports dialpad lookup (PSTN protocol). 
Also fix the case where dialpad lookup returns a virtual user
Also makes some changes in the styles of the dial pad.
Should also fix DM endings in room (BMA fixes #2593)


vector/src/main/java/im/vector/app/features/home/HomeDetailFragment.kt
```diff
@@ -356,14 +365,46 @@ class HomeDetailFragment @Inject constructor(
                         detach(it)
                     }
             if (fragmentToShow == null) {
-                val params = RoomListParams(displayMode)
-                add(R.id.roomListContainer, RoomListFragment::class.java, params.toMvRxBundle(), fragmentTag)
+                if (tab is HomeTab.RoomList) {
```
bmarty, line 368: A `when` would be better here

bmarty (approved):
LGTM, please fix the conflict

PR #3515, opened 2021-06-16
Feature/bma/fix crash theme

bmarty:
- Fix a crash with alert for verification
- Update some colors
- Do some cleanup

No need to update the changelog, this are fixes following previous PR, merged after the latest release


ganfra (approved):
Thanks for the fixes, LGTM

PR #3519, opened 2021-06-17
Fix crash when setting up key backup

BillCarsonFr:
Got a crash when trying to setup keybackup on logout

image

````
    Thread: main, Exception: android.view.InflateException: Binary XML file line #97 in im.vector.app.debug:layout/fragment_keys_backup_setup_step1: Binary XML file line #97: You must supply a layout_width attribute.
    Caused by: java.lang.UnsupportedOperationException: Binary XML file line #97: You must supply a layout_width attribute.
        at android.content.res.TypedArray.getLayoutDimension(TypedArray.java:829)
        at android.view.ViewGroup$LayoutParams.setBaseAttributes(ViewGroup.java:7932)
        at android.view.ViewGroup$MarginLayoutParams.<init>(ViewGroup.java:8130)
        at androidx.constraintlayout.widget.ConstraintLayout$LayoutParams.<init>(ConstraintLayout.java:2691)

### DIFF
--- a/vector/src/main/res/layout/fragment_keys_backup_setup_step1.xml
+++ b/vector/src/main/res/layout/fragment_keys_backup_setup_step1.xml
@@ -84,6 +84,8 @@
     <Button
         android:id="@+id/keys_backup_setup_step1_manual_export_button"
         style="@style/Widget.Vector.Button.Text"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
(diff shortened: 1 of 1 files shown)


bmarty:
Thanks, I will check if there is other case like that. It's maybe due to the style refacto (the width and height was in the style), but I'm not sure. As you say it's maybe older...

bmarty (approved):

PR #3520, opened 2021-06-17
Feature/fga/timeline virtual room

ganfra:
It's an attempt to allow merging of timelines, so we can show call history of virtual rooms in corresponding native room.
There are lots of limitations and performance are not great.


bmarty (approved):
LGTM

bmarty:
Nice work. Maybe write down the limitation somewhere?

PR #3529, opened 2021-06-18
Apply Google Services Gradle plugin more robustly

### LINKED ISSUE #3527: Google Services Gradle plugin is not always applied to build
#### Describe the bug

https://github.com/vector-im/element-android/blob/5325c761f4bce9edaf77239f84a3553a0cb135c3/vector/build.gradle#L510-L512

The Google Services Gradle plugin is applied if and only if the Gradle task name being run from CLI contains the substring `"Gplay"`. This is not always the case, even when building a `Gplay` flavor build. For example, `./gradlew` can be run with task `asGD`, which is an unambiguous abbreviation of `assembleGplayDebug`, so Gradle runs that task. However, the logical branch that applies the Google Services Gradle plugin would not be triggered, because `"asGD"` does not contain the `"Gplay"` substring! This can lead to serious issues in production if a release would accidentally be built with a task that doesn't contain the `"Gplay"` substring.

#### To Reproduce
1. Checkout branch `develop`
2. Build using `./gradlew asGD`
3. Install the built APK, run the app.
4. Note the logcat: 

```
            FirebaseApp  W  Default FirebaseApp failed to initialize because no default options were found. This usually means that com.google.gms:google-services was not applied to your gradle project.
   FirebaseInitProvider  I  FirebaseApp initialization unsuccessful
```

This proves that the Google Services Gradle plugin wasn't installed successfully, with consequences like FCM not working like expected.

#### Expected behavior

The build script applies the plugin regardless of the task name specified through the Gradle CLI.

#### Screenshots

N/A

N/A

#### Additional context

tag `v1.1.9` (and branch develop between that version and the next future version)

---

I'm opening a PR that fixes this issue.

erikhuizinga:
Fixes #3527

Signed-off-by: Erik Huizinga <huizinga.erik@gmail.com>

### DIFF
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -247,6 +247,9 @@ android {
     productFlavors {
         gplay {
             apply plugin: 'com.google.gms.google-services'
+            afterEvaluate {
+                tasks.matching { it.name.contains("GoogleServices") && !it.name.contains("Gplay") }*.enabled = false
+            }
 
             dimension "store"
             isDefault = true
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -246,6 +246,8 @@ android {
 
     productFlavors {
         gplay {
+            apply plugin: 'com.google.gms.google-services'
+
             dimension "store"
             isDefault = true
             versionName "${versionMajor}.${versionMinor}.${versionPatch}${getGplayVersionSuffix()}"
@@ -505,8 +507,4 @@ dependencies {
     androidTestImplementation('com.schibsted.spain:barista:3.9.0') {
         exclude group: 'org.jetbrains.kotlin'
     }
-}
-
-if (getGradle().getStartParameter().getTaskRequests().toString().contains("Gplay")) {
-    apply plugin: 'com.google.gms.google-services'
 }


erikhuizinga:
I think it still fails, but I know another solution!

bmarty:
Yeah, it's not a easy issue to fix. I had some hope with your solution :/

erikhuizinga:
I've just pushed something that works. However, it depends on the Google Services plugin to provide a task containing the flavor name (likely always the case) and substring `"GoogleServices"`.

Now, when running an Fdroid build variant:

```
> Task :vector:processFdroidDebugGoogleServices SKIPPED
```

This does mean that the plugin is applied, but the task it adds never runs and no google-services.json file is every read and used. Does that mean that the plugin doesn't do anything? I don't know. One way this can maybe be verified is by building 2 FdroidDebug APKs, one from the `HEAD` of this branch and one from the merge base with `develop`, and compare the APKs. Ideally, they should be identical.

Other than that I'm giving up after trying a few things. 😹

---

One question though:
Can you please explain why the current code on develop:

https://github.com/vector-im/element-android/blob/5325c761f4bce9edaf77239f84a3553a0cb135c3/vector/build.gradle#L510-L512

does work?

What I don't understand: if I e.g. run `./gradlew assembleGplayDebug assembleFdroidDebug`, how come that the plugin is applied only to the `GplayDebug` variant, and not to the `FdroidDebug` variant? To me it seems like the current code just checks if the CLI task name contains `"Gplay"` and then applies the plugin; wouldn't it also apply the plugin to the Fdroid task? Or does this apply it after resolving the task graph for the Fdroid flavor (which comes earlier alphabetically)?

---

Other approaches I've tried:

- Adding a task that does `doLast { /* Apply plugin */ }`: happens after configuration == too late
- For all application variants, filter by flavor name containing `"Gplay"` and apply the plugin: it also applies the plugin to the Fdroid flavor.
- Awaiting the `gradle.taskGraph.whenReady { graph -> /* Apply plugin if task graph contains tasks with name "Gplay" */ }`: fails because the task graph is ready after evaluating plugin, so it cannot apply a plugin that adds yet another task...

erikhuizinga:
Note that I've put the added lines in the `gplay {}` flavor closure, but they could also simply be put near the top of the build script where other plugins are applied. Let me know what you think is easier to understand. Putting it near the top is more common practice, but putting it inside the `gplay {}` closure indicates (only semantically) that it is related.

erikhuizinga:
Some ways to check that this works:

- Build it: `./gradlew asF asG` build without (Fdroid) and with (Gplay) the Google Services Gradle plugin.
- Check the dependencies, e.g.:
   - `./gradlew vector:dependencyInsight --configuration fdroidDebugCompileClasspath --dependency google | grep -i services` is empty. This check could even be used by a GitHub workflow to verify that this works, assuming that the Google Services dependencies still contain `google` and `services` in the future.
   - `./gradlew vector:dependencyInsight --configuration gplaydDebugCompileClasspath --dependency google | grep -i services` shows a bunch of google services dependencies.

bmarty:
> What I don't understand: if I e.g. run ./gradlew assembleGplayDebug assembleFdroidDebug, how come that the plugin is applied only to the GplayDebug variant, and not to the FdroidDebug variant?

One first remark:

```sh
./gradlew assembleGplayDebug assembleFdroidDebug
```

actually does not work (and this is a known limitation of the current solution. The plugin is applied to the fdroid build and we get:

```sh
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':vector:processFdroidDebugGoogleServices'.
> File google-services.json is missing. The Google Services Plugin cannot function without it. 
   Searched Location: 
  /Users/bmarty/workspaces/element-android/vector/src/fdroid/debug/google-services.json
  /Users/bmarty/workspaces/element-android/vector/src/debug/fdroid/google-services.json
  /Users/bmarty/workspaces/element-android/vector/src/fdroid/google-services.json
  /Users/bmarty/workspaces/element-android/vector/src/debug/google-services.json
  /Users/bmarty/workspaces/element-android/vector/src/fdroidDebug/google-services.json
  /Users/bmarty/workspaces/element-android/vector/google-services.json
```

erikhuizinga:
I see, thanks. Then my current solution does work better than before.

Like I said: it relies on the assumption that it's safe / good enough to build Fdroid builds with the plugin applied and the task disabled. I would assume that the build would fail if for some reason the plugin becomes enabled again, because of the missing google-services.json. However, if in the future that configuration file changes or moves elsewhere, we don't know if the plugin remains disabled.

I'll leave it up to you to weigh the benefits and risks of this assumption.

bmarty:
Thanks for the follow up.
I'm not sure what we should do now :).

@Bubu have you any though on this?

Bubu:
If this works for your use-cases than this should be fine for f-droid as well. The plugin is removed from the gradle file before building anyway as a safeguard. I'll likely need to adapt the regex then for the next version then :).

erikhuizinga:
Where's that regex? That's easy to fix; there are other ways to do that too!

Bubu:
> Where's that regex? That's easy to fix; there are other ways to do that too!

https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/im.vector.app.yml#L434-435

erikhuizinga:
@Bubu thanks!

https://gitlab.com/fdroid/fdroiddata/-/merge_requests/9287

erikhuizinga:
See discussion in MR on GitLab (in my previous comment): if this PR is ok, I think that it can be merged.

erikhuizinga:
This is a friendly reminder to review this PR and decide whether or not this approach is acceptable.

bmarty:
Yes, I think it is acceptable. If @Bubu is OK for F-Droid, I'm OK for Gplay.

I will need a sign-off, as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off please.

bmarty:
(FCM tested OK using gplay version of this PR)

bmarty (approved):

erikhuizinga:
> I will need a sign-off, as per [https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md?rgh-link-date=2021-06-30T15%3A39%3A04Z#sign-off) please.

Are you asking me this? I had read those contributing guidelines, but I thought my local Git config already signed all commits. It turns out it does not! 😅

Is it enough if I edit the first PR message?

bmarty:
> Is it enough if I edit the first PR message?

Yes, thanks!

PR #3540, opened 2021-06-21
Send button does not show up half of the time (#3535)

### LINKED ISSUE #3535: Send button does not show up half of the time
#### Describe the bug
In the current pre-release with the disappearing "send' button that only shows up when the message field isn't empty, it oftentimes does not show up even when a message has been written. Then, the input field needs to be completely deleted and content re-added for it to appear.

#### To Reproduce
1. Type a message.
2. If unlucky, button stays hidden.

#### Expected behavior
Let me send a message.

 - Device: Galaxy Note 10+
 - OS: Android 11

#### Additional context
 - App version and store: 1.1.10 [40101102] (G-b2627)
 - Homeserver: irrelevant

bmarty:
Fix a regression in 1.1.10 (Fixes #3535)

### DIFF
--- a/changelog.d/3535.feature
+++ b/changelog.d/3535.feature
@@ -1,0 +1,1 @@
+Send button does not show up half of the time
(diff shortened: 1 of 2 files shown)


ganfra (approved):
LGTM I guess

PR #3549, opened 2021-06-23
Fix styles issue

bmarty:
Especially for dialogs.

Please note that `withColoredButton()` has been removed, please use theme variation instead.


ganfra (approved):
LGTM!

Issue #3542, opened 2021-06-21
Redaction of a user's first message hides their avatar, spoofing previous speaker


emorrp1:
#### Describe the bug
If you moderate just the first in a sequence of messages, it also removes the displayed avatar. This makes it appear to android users as if the previous person is still talking.

#### To Reproduce
Steps to reproduce the behavior (Haven't tested redacting own message):
1. Have someone send two sequential messages
2. Long press just the first, Remove...
3. Give a reason and confirm with REMOVE
4. Observe the absence of any indication the speaker has changed

![Screenshot_20210621-155645_Element](https://user-images.githubusercontent.com/71895/122785500-58716880-d2ab-11eb-8961-311ba55f7f8c.png)

#### Expected behavior
Does not happen on element-web.

![Screenshot_2021-06-21_15-58-00](https://user-images.githubusercontent.com/71895/122785563-645d2a80-d2ab-11eb-8048-0917999f0192.png)

#### Smartphone (please complete the following information):
 - Device: Wileyfox Swift 2
 - OS: LineageOS 17.1 (~10)

#### Additional context
 - App version and store: 1.1.9 - F-Droid
 - Homeserver: freedombox.emorrp1.name

PR #3553, opened 2021-06-24
Fix some issues with timeline cache invalidation and visibility.

ganfra:
This PR fixes https://github.com/vector-im/element-android/issues/3542 and also avoid adding sender info when not required.


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/TimelineEventController.kt
```diff
@@ -209,6 +218,7 @@ class TimelineEventController @Inject constructor(private val dateFormatter: Vec
     )
 
     init {
+        isDebugLoggingEnabled = true
```
bmarty, line 221: Maybe remove this?

bmarty (approved):
I did it then

Issue #3564, opened 2021-06-24
Call notification is stuck on invite because of fastlane mode.


ganfra:
Fetched event by Fastlane mode doesnt have agelocalts set so the invite is not filtered like others, resulting in call invite being sent even if the call is ended.
Also it can happens if sync is faster than fastlane.

PR #3565, opened 2021-06-24
Fix call invite processed after call is ended because of fastlane mode.

ganfra:
Fixes https://github.com/vector-im/element-android/issues/3564


bmarty (approved):
LGTM even if I do not like using `System.currentTimeMillis()` which can be manually altered by the user...

PR #3585, opened 2021-06-30
Improve wellknown usage

### LINKED ISSUE #2843: Sends login requests to https://matrix.org/ rather than honouring .well-known
By default, clients should check https://matrix.org/.well-known/matrix/client before logon to see that they should send login/registration requests to https://matrix-client.matrix.org.

It seems that currently element-android defaults to https://matrix.org.

### LINKED ISSUE #3572: Improve get wellknown task and store the homeserver url entered by the user during the session creation phase.
[Well-known](https://matrix.org/docs/spec/client_server/r0.6.1#server-discovery) feature has been introduced to let any Matrix client to discover where is located the client-server Matrix API, either when the user enters a domain or if the user enter a matrixId (userId) from which the domain can be extracted.

For instance, for the domain `matrix.org`, the well-known URL is https://matrix.org/.well-known/matrix/client

And the result is:
```json
{
    "m.homeserver": {
        "base_url": "https://matrix-client.matrix.org"
    },
    "m.identity_server": {
        "base_url": "https://vector.im"
    }
}
```

So the base url for the client-server API has to be `https://matrix-client.matrix.org`.

Element Android handles this correctly (#2843 exists though), but in fact the well-known file has been hacked on some servers to also store extra data, for instance some client configuration settings.

Moreover, when the user logins, the login response can contain another client-server URL, see https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-login , the parameter `Discovery Information`, which is also supported by Element Android. This URL will override the previous store homeserver url, and the url entered by the client is lost (can be changed twice!).

We can see that in the current general setting screen in Element Android:

image

This is the same in the homeserver setting screen:

image

It's also worth noting that https://matrix-client.matrix.org/.well-known/matrix/client does not return anything (it's actually a non sense).

### What we have to do

- `GetWellknownTask.Params` should not take a `matrixId` anymore, but rather take a `domain` name. It will allow to perform a wellknown request after the login phase, to any desire domain. Also there will be no need anymore to create fake userId from a domain to do that.

- We should store in `HomeServerConnectionConfig` object the homeserver url which has been entered by the user, to be able to later perform the same Wellknown request, even if the URL has been overridden by wellknown, or by the login response. On some circumstances, this is not OK to retrieve the domain from the userId. New field will be:
```kotlin
val userHomeServer: Uri
```

Also this URL could be used in the settings screen. A new section could be added to see the actual client-server API URL (the one which is displayed now).

### Other

There are some old sessions which are still using matrix.org as a base URL for CS-API. This can be an opportunity to migrate the session database (SessionParamsEntity):
- deserialize `HomeServerConnectionConfig`
- force the usage of `https://matrix-client.matrix.org` if and only if the server url is matrix.org. For other servers, we can assume that the URL are the same. So `userHomeServer` will be `https://matrix.org` and `homeServerUri` will be overridden to `https://matrix-client.matrix.org`
- serialize the object and write it to DB.

bmarty:
This PR fixes several issues:

Fixes #2843 : we now first perform a .well-known request, even if the entered URL is a valid Matrix API

Fixes #3572: we now store the url entered at session creation (Auth database migration required to add the new field), for further reference. Settings screen is updated:

image

and

image

API URL is only displayed in developer mode (see the diff with https://user-images.githubusercontent.com/3940906/123458079-e79cba00-d5e4-11eb-825e-4d481662b418.png).

Some change in the SDK API, wellknown now takes a domain instead of a matrixId.

### Tested OK

- sign in on matrix.org
- sign up on matrix.org
- SSO on matrix.org (using a GitHub account, see the screenshots)
- sign in on mozilla.modular.im (other SSO)
- sign in using a matrixId
- use element web url (develop.element.io) to retrieve config file

### Limitation

If the user enter the matrix API url, for instance "matrix-client.matrix.org", there is no way to deduce the domain matrix.org. So every further wellknow request will fail.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/DefaultAuthenticationService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/DefaultAuthenticationService.kt
@@ -19,6 +19,7 @@ package org.matrix.android.sdk.internal.auth
 import android.net.Uri
 import dagger.Lazy
 import okhttp3.OkHttpClient
+import org.matrix.android.sdk.api.MatrixPatterns
 import org.matrix.android.sdk.api.auth.AuthenticationService
 import org.matrix.android.sdk.api.auth.data.Credentials
 import org.matrix.android.sdk.api.auth.data.HomeServerConnectionConfig
@@ -28,6 +29,7 @@ import org.matrix.android.sdk.api.auth.login.LoginWizard
 import org.matrix.android.sdk.api.auth.registration.RegistrationWizard
 import org.matrix.android.sdk.api.auth.wellknown.WellknownResult
 import org.matrix.android.sdk.api.failure.Failure
+import org.matrix.android.sdk.api.failure.MatrixIdFailure
 import org.matrix.android.sdk.api.session.Session
 import org.matrix.android.sdk.api.util.appendParamToUrl
 import org.matrix.android.sdk.internal.SessionManager
@@ -275,9 +277,7 @@ internal class DefaultAuthenticationService @Inject constructor(
         val domain = homeServerConnectionConfig.homeServerUri.host
                 ?: throw Failure.OtherServerError("", HttpsURLConnection.HTTP_NOT_FOUND /* 404 */)
 
-        // Create a fake userId, for the getWellknown task
-        val fakeUserId = "@alice:$domain"
-        val wellknownResult = getWellknownTask.execute(GetWellknownTask.Params(fakeUserId, homeServerConnectionConfig))
+        val wellknownResult = getWellknownTask.execute(GetWellknownTask.Params(domain, homeServerConnectionConfig))
 
         return when (wellknownResult) {
             is WellknownResult.Prompt -> {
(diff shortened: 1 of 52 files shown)


BillCarsonFr (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/DefaultAuthenticationService.kt
```diff
@@ -165,30 +167,41 @@ internal class DefaultAuthenticationService @Inject constructor(
...
-                )
+                        .fold(
+                                {
+                                    it
+                                },
+                                {
+                                    if (it is Failure.OtherServerError
+                                            && it.httpCode == HttpsURLConnection.HTTP_NOT_FOUND /* 404 */) {
+                                        // It's maybe a Riot url?
+                                        getRiotDomainLoginFlowInternal(homeServerConnectionConfig)
```
ganfra, line 195: Remove Riot reference?

vector/src/main/java/im/vector/app/features/raw/wellknown/ElementWellKnownExt.kt
```diff
@@ -16,11 +16,14 @@
...
 
+import org.matrix.android.sdk.api.auth.data.SessionParams
 import org.matrix.android.sdk.api.extensions.tryOrNull
 import org.matrix.android.sdk.api.raw.RawService
 
-suspend fun RawService.getElementWellknown(userId: String): ElementWellKnown? {
-    return tryOrNull { getWellknown(userId) }
+suspend fun RawService.getElementWellknown(sessionParams: SessionParams): ElementWellKnown? {
+    // By default we use the domain of the userId to retrieve the .well-known data
+    val domain = sessionParams.userId.substringAfter(":")
```
ganfra, line 25: Should we have an extension method for this substringAfter(":") as we use it at several places

ganfra (approved):
LGTM, just 2 remarks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/DefaultAuthenticationService.kt
bmarty: Done

vector/src/main/java/im/vector/app/features/raw/wellknown/ElementWellKnownExt.kt
bmarty: Done: added `fun String.getDomain(): String` in MatrixPattern

Issue #3577, opened 2021-06-28
Crash after video call


bmarty:
Reported several times in by the PlayStore and in rageshakes too.

`MediaStreamTrack has been disposed.`

```
java.lang.IllegalStateException: MediaStreamTrack has been disposed.
  at org.webrtc.MediaStreamTrack.checkMediaStreamTrackExists (MediaStreamTrack.java:2)
  at org.webrtc.MediaStreamTrack.getNativeMediaStreamTrack (MediaStreamTrack.java:1)
  at org.webrtc.VideoTrack.addSink (VideoTrack.java:4)
  at im.vector.app.features.call.webrtc.WebRtcCall.attachViewRenderersInternal (WebRtcCall.kt:9)
  at im.vector.app.features.call.webrtc.WebRtcCall.access$attachViewRenderersInternal (WebRtcCall.kt:1)
  at im.vector.app.features.call.webrtc.WebRtcCall$attachViewRenderers$1.invokeSuspend (WebRtcCall.kt:10)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:3)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:18)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:923)
```

https://github.com/matrix-org/element-android-rageshakes/issues?q=is%3Aissue+is%3Aopen+checkMediaStreamTrackExists

PR #3587, opened 2021-06-30
Fix crash after video call

ganfra:
Fix https://github.com/vector-im/element-android/issues/3577


bmarty (approved):
LGTM, no crash anymore. Thanks!

PR #3624, opened 2021-07-02
Use different copy for self verification

dbkr:
The copy all refers to 'their' device for when you're verifying
someone else. Update it to refer to your own devices in the case
where you're verifying one of your own devices.

### Pull Request Checklist

- [ ] UI change has been tested on both light and dark themes (Hopefully not necessary for a string change...)

Screenshot_2021-07-05-19-48-43-131_im vector app debug

Equivalent for iOS: https://github.com/vector-im/element-ios/pull/4525

### DIFF
--- a/vector/src/main/java/im/vector/app/features/crypto/verification/choose/VerificationChooseMethodController.kt
+++ b/vector/src/main/java/im/vector/app/features/crypto/verification/choose/VerificationChooseMethodController.kt
@@ -45,9 +45,19 @@ class VerificationChooseMethodController @Inject constructor(
         val host = this
 
         if (state.otherCanScanQrCode || state.otherCanShowQrCode) {
+            var scanCodeInstructions: String
+            var scanOtherCodeTitle: String
+            if (state.isMe) {
+                scanCodeInstructions = host.stringProvider.getString(R.string.verification_scan_self_notice)
+                scanOtherCodeTitle = host.stringProvider.getString(R.string.verification_scan_with_this_device)
+            } else {
+                scanCodeInstructions = host.stringProvider.getString(R.string.verification_scan_notice)
+                scanOtherCodeTitle = host.stringProvider.getString(R.string.verification_scan_their_code)
+            }
+
             bottomSheetVerificationNoticeItem {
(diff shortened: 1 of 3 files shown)


bmarty (approved):
LGTM, thanks.
One remark: the "Can't scan" option is also a bit weird when self-verifying :/. We should maybe hide the subtitle in this case?

dbkr:
oh, good point

dbkr:
updated

vector/src/main/java/im/vector/app/features/crypto/verification/choose/VerificationChooseMethodController.kt
```diff
@@ -45,9 +45,22 @@ class VerificationChooseMethodController @Inject constructor(
         val host = this
 
         if (state.otherCanScanQrCode || state.otherCanShowQrCode) {
+            var scanCodeInstructions: String
+            var scanOtherCodeTitle: String
+            var compareEmojiSubtitle: String
+            if (state.isMe) {
+                scanCodeInstructions = host.stringProvider.getString(R.string.verification_scan_self_notice)
```
bmarty, line 52: Sorry, I did not noticed it before, but there is no need to use `host.` here. It only useful if you want to use the host member from the Epoxy item DSL, for instance inside `bottomSheetVerificationNoticeItem {}` block.
It's not a big deal at all, it's OK to merge the PR like that, I can do the clean up later.

bmarty (approved):
Thanks for the update, LGTM

vector/src/main/java/im/vector/app/features/crypto/verification/choose/VerificationChooseMethodController.kt
bmarty: (Fixed in 07d6eaacf95acfec165dce94e567ae89db8e3f14)

dbkr: ah - I see, I did wonder what that was all about, but figured it was best to be consistent anyhow

Issue #1673, opened 2020-07-12
Notification keeps resending


Wandelstock:
If a notification isnt honored, riot.imx keeps resending it. So i get again a notificationssound and it resends it to my smartwatch (4 times)

0.91.4

Rares95:
I have this, but on the phone. Very weird (Android, latest Element version)

orbitarte:
Same here last version of element

Wandelstock:
Still the annoying bug. My smartwatch and tasker is going Bezirk

Ji-eF:
I also have this bug on my device : Nexus 6P with Lineageos 15 (Androis 8) and Element from F-Droid

Ji-eF:
It's quite hard to explain, so I hope I'll be clear enough :)
Upon a bit of investigation, it seems that the culprit is Element's "Silent notifications" category : 

- When getting a notification from a room with noisy notification, Element triggers "Noisy notification" category for Element's (in Android Settings > Apps Notifications) which seems OK.
- But then, the notification keeps popping up every few 10-seconds with the "Silent notification" category (when "Silent" is configured to "Vibrate" and/or "Sound" and/or "Blink LED".

The same behaviour appears while being on a video/phone call and you play around with notification area to make the voice call screen appear and hide ; the silent notification make the phone vibrate and sound, when configured to do so.

ukanuk:
it makes me very frustrated when Element notifications wake me up in the middle of the night due to this issue.

Notifications will start retriggering every minute because a friend in another time zone sent a message, and if it happens to be the 1% of the time when my Amazfit Bip hasn't detected me going to sleep and automatically turning on Do Not Disturb, then the watch starts vibrating and lighting up every minute and wakes me up. Every other app sends notifications only once.

Even if it only happened once a year, it makes me very frustrated when Element notifications wake me up in the middle of the night.

ukanuk:
I have this issue with the Google Play version of Element, while the other users say they have it with the F-Droid version. It seems to affect both versions.

Vulkanodox:
How is this shit not fixed yet? Just started to use element but this is unusable. No wonder why everybody wants to use whatsapp.

ouchadam:
able to reproduce this using the latest from develop 2b38c607b23b9ad087ac849f9b29e1da31360023 with a Xiaomi Mi Band 5

- Have a unread notification on the phone
- Join matrix-hq (even with no notifications)
[...]

PR #3657, opened 2021-07-10
Do not notify again for old events

SpiritCroc:
Resending the notification here can trigger other system components or
apps that listen to new notifications, such as connected smart watches
or automation tools.

Fixes https://github.com/vector-im/element-android/issues/1673

### Pull Request Checklist

- [ ] Changes has been tested on an Android device or Android emulator with API 21 (Note: I haven't tested this thoroughly on pure Element, but this has been included in SchildiChat and Beeper for around a month and didn't seem to cause issues so far)

### DIFF
--- a/vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
+++ b/vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
@@ -457,7 +457,7 @@ class NotificationDrawerManager @Inject constructor(private val context: Context
 
             if (eventList.isEmpty() || eventList.all { it.isRedacted }) {
                 notificationUtils.cancelNotificationMessage(null, SUMMARY_NOTIFICATION_ID)
-            } else {
+            } else if (hasNewEvent) {
                 // FIXME roomIdToEventMap.size is not correct, this is the number of rooms
                 val nbEvents = roomIdToEventMap.size + simpleEvents.size
                 val sumTitle = stringProvider.getQuantityString(R.plurals.notification_compat_summary_title, nbEvents, nbEvents)
--- a/changelog.d/1673.bugfix
+++ b/changelog.d/1673.bugfix
@@ -1,0 +1,1 @@
+Avoid resending notifications that are already shown


bmarty:
@BillCarsonFr WDYT about this change? I though we were building the same notifications each time we want to refresh the navigation drawer, but it's maybe not true

bmarty:
We are a bit scared that this change brings some regressions...

vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
```diff
@@ -457,7 +457,7 @@ class NotificationDrawerManager @Inject constructor(private val context: Context
 
             if (eventList.isEmpty() || eventList.all { it.isRedacted }) {
                 notificationUtils.cancelNotificationMessage(null, SUMMARY_NOTIFICATION_ID)
-            } else {
+            } else if (hasNewEvent) {
```
ouchadam, line 460: can confirm this fixes the issue, great catch! 

it might be worth renaming the variable to `hasGroupChanged` as what this logic is doing is updating the summary group notification when one of the group children has also been notified 

I'm also happy to make the change in a separate PR as I'll be in the area attempting to address the double sound issue

ouchadam (approved):

ouchadam:
@bmarty from my testing there were no regressions and it fixes #1673, however there is the possibility that the summary count could go out of sync if events are removed as that won't trigger set `hasNewEvent = true`

although... the count is already incorrect~

```
// FIXME roomIdToEventMap.size is not correct, this is the number of rooms
```
* screenshots from latest beta

| | |
| --- | --- |
|2021-10-01T17:05:28,454972740+01:00|2021-10-01T17:04:46,431305246+01:00

vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
bmarty: thanks a lot Adam, I think we can merge the PR now!

PR #3667, opened 2021-07-12
Permission rework

bmarty:
Especially better handling of the "denied forever" permission user story.

Add a new debug screen to play with permission request from the application.

so now when requesting permissions:
- already granted: sync call (no change on this part)
- not yet granted:
  - user give permission: async success
  - user denies: no more thing to do, the action is not performed
  - user denies and check `do not ask again`: no more thing to do, the action is not performed
- not yet granted, user has previously checked `do not ask again`:
  - a dialog or a snackbar is displayed to ask the user to open the system settings to grant the permission

https://user-images.githubusercontent.com/3940906/125365236-2caa4580-e374-11eb-9cf6-3d84c2f54885.mp4

### DIFF
--- a/changelog.d/3667.feature
+++ b/changelog.d/3667.feature
@@ -1,0 +1,1 @@
+Better management of permission requests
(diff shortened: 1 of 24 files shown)


BillCarsonFr (commented):
LGTM

BillCarsonFr (approved):

Issue #3672, opened 2021-07-14
Changing Account Settings > Notifications > Advanced Notifications on android causes discrepancies with web


langleyd:
#### Describe the bug
Changing between `None`, `Silent` and `Noisey` on android causes discrepancies with web client.
Causing the selection for the notification setting in question to appear in a unselected state on web and the push rule setting is then duplicated under a new section that appears on web as "Advanced notification settings"
The only settings affected appear to be:
- `Messages containing @room`
- `Messages in group chats`
- `Encrypted messages in group chats`
- `Call Invitations`
- `Messages in one-to-one chats`

#### To Reproduce
1. Go to `Account Settings > Notifications > Advanced Notifications` on Android
2. Cycle between `None`, `Silent` and `Noisy` for each setting (may have to do a number of times)
3. Go to `Settings > Notification Settings` on web/desktop client
4. For the settings listed above the selection will sometimes disappear and the setting duplicated as described above and shown in the screenshot

#### Expected behavior
Push Rule Setting is not duplicated and radio buttons remain in a selected state

#### Screenshots
![image](https://user-images.githubusercontent.com/6135282/125617943-c91774c4-3479-4faa-b1f2-05ec2df532ba.png)

#### Additional context
- I have seen some discrepancies in the requests that are send when cycling the settings on android with the web for both the enable endpoint and the actions end point. Closing this gap I think should fix the behaviour seen on web.

Issue #3646, opened 2021-07-07
Reorganise Advanced Notifications in to Global Notifications, Keywords and Mentions, Other


langleyd:
This issue is to re-organise `Account Settings > Notifications > Advanced Notifications` in to `Global Notifications`, `Keyword and Mentions` and `Other`

The only change in functionality is that rather than 3 options for `off`, `silent` ,`noisey` there will just be a checkbox for on/off states with the on meaning `noisey.

This shows the new grouping under the `Account Settings > Notifications`:
![image](https://user-images.githubusercontent.com/6135282/124799975-0720d480-df4d-11eb-86a9-e247362b5ccd.png)

Global Notifications
![image](https://user-images.githubusercontent.com/6135282/124801041-47cd1d80-df4e-11eb-9422-d495946d308d.png)

Mentions and Keywords
![image](https://user-images.githubusercontent.com/6135282/124801075-50bdef00-df4e-11eb-9947-87fb98bf09ea.png)

Other
![image](https://user-images.githubusercontent.com/6135282/124801118-603d3800-df4e-11eb-8364-cc8c011b8361.png)

AnisTaluqdar:
What about if someone replies to me? It would be great if there have an option to notify reply text.

matthijskooijman:
> The only change in functionality is that rather than 3 options for off, silent ,noisey there will just be a checkbox for on/off states with the on meaning `noisey.

What's the rationale for this change?

I was very much using these options, so I can get noisy notifications for one-on-one rooms and mentions, and silent notifications for other rooms (that typically have too much activity to be noisy, but I do want to get a notification on the screen for their activity). This configuration was also the default, not sure if that has also changed? I recall reading an issue where this was discussed as the preferred default configuration to not overload people with noisy notifications by default. I can't quite find it again, but it does suggest that I'm not the only one that wants to use silent notifications.

It seems you can still configure silent notifications through element web, but I'm slightly worried that if this setting is removed in the Android version intentionally, it will also get dropped from the web version.

What's also a bit weird now, is that the Element Android now allow configuring "silent" and "noisy" (and "call") notification profiles. It does make sense to have these, since even if Element Android does not allow configuring silent notifications, they can be configured (through push rules) from other clients, so Element Android must be able to handle silent notifications I guess. But without a way to configure a notification as "silent" or "noisy", I expect this will be confusing to users (which will not know which of these two will be used when they put a checkbox for some notification category).

[...]

PR #3673, opened 2021-07-14
Reorganise Account Notification Settings

langleyd:
Implements feature https://github.com/vector-im/element-android/issues/3646

### screenshots
Account Settings > Notifications
image
Default Notifications
image
Mentions and Keywords
image
Other
image

### notes
- Kept the existing settings architecture (`VectorSettingsBaseFragment`, `VectorPreference` `VectorPreferenceCategory`, etc). Changing the architecture to `ViewModels` etc I think would be out of the scope of this PR.
- Introduced `VectorSwitchPreference` to replace `PushRulePreference` and moved all business logic to the fragment level.
- Removing an option (Silent) allowed us to simplify the business logic somewhat but kept implementation functionally equivalent to previous.
- Found some bugs/discrepancies with web(not regressions, they occur in develop), that I will fix as a part of https://github.com/vector-im/element-android/issues/3672.
- Introduced build time resVal to hide V2 until we are ready to introduce this new UI.
- Ran contributing steps/checks
- Update `Global Notifications` to `Default Notifications` on design's advice


gaelledel (approved):

bmarty (commented):
Wow, IDK why `vector/src/main/java/im/vector/app/core/preference/PushRulePreference.kt` was executable :/

vector/src/main/java/im/vector/app/core/preference/VectorCheckboxPreference.kt
```diff
@@ -0,0 +1,45 @@
...
+
+package im.vector.app.core.preference
+
+import android.content.Context
+import android.util.AttributeSet
+import android.widget.TextView
+import androidx.preference.CheckBoxPreference
+import androidx.preference.PreferenceViewHolder
+
+open class VectorCheckboxPreference : CheckBoxPreference {
```
bmarty, line 25: This class does not have to be `open`. It does not have to be sealed as well, but let's keep the default Kotlin behavior (so seal) unless we have to extend the class.

vector/src/main/java/im/vector/app/features/settings/VectorSettingsGlobalNotificationPreferenceFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+
+class VectorSettingsGlobalNotificationPreferenceFragment
+    : VectorSettingsPushRuleNotificationPreferenceFragment() {
+
+    override var titleRes: Int = R.string.settings_notification_global
+
+    override val preferenceXmlRes = R.xml.vector_settings_notification_global
+
+    override val prefKeyToPushRuleId: Map<String, String>
+        get() = mapOf(
```
bmarty, line 30: Could be simplified with `override val prefKeyToPushRuleId = mapOf(`

vector/src/main/java/im/vector/app/features/settings/VectorSettingsOtherNotificationPreferenceFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    override var titleRes: Int = R.string.settings_notification_other
+
+    override val preferenceXmlRes = R.xml.vector_settings_notification_other
+
+    override val prefKeyToPushRuleId: Map<String, String>
+        get() = mapOf(
+                "SETTINGS_PUSH_RULE_INVITED_TO_ROOM_PREFERENCE_KEY" to RuleIds.RULE_ID_INVITE_ME,
+                "SETTINGS_PUSH_RULE_CALL_INVITATIONS_PREFERENCE_KEY" to RuleIds.RULE_ID_CALL,
+                "SETTINGS_PUSH_RULE_MESSAGES_SENT_BY_BOT_PREFERENCE_KEY" to RuleIds.RULE_ID_SUPPRESS_BOTS_NOTIFICATIONS,
```
bmarty, line 33: I'm scared about the `suppress` wording here. So it mean that the check box has the opposite behavior? I know it's in the spec... EDIT: the UI seems to uncheck the box by default, so I guess it's working correctly.

vector/src/main/java/im/vector/app/features/settings/VectorSettingsPushRuleNotificationPreferenceFragment.kt
```diff
@@ -0,0 +1,141 @@
...
+import org.matrix.android.sdk.api.pushrules.rest.PushRule
+import org.matrix.android.sdk.api.pushrules.rest.PushRuleAndKind
+
+abstract class VectorSettingsPushRuleNotificationPreferenceFragment
+    : VectorSettingsBaseFragment() {
+
+    /**
+     * @return the bing rule status boolean
+     */
+    fun ruleStatusIndexFor(ruleAndKind: PushRuleAndKind): Boolean {
```
bmarty, line 36: Please make all the `fun` private where it's possible, default visibility is public in Kotlin

vector/src/main/res/values/strings.xml
```diff
@@ -2946,10 +2965,6 @@
     <string name="error_failed_to_import_keys">Failed to import keys</string>
 
     <string name="settings_notification_configuration">Notifications configuration</string>
-    <string name="settings_messages_at_room">Messages containing @room</string>
-    <string name="settings_messages_in_e2e_one_to_one">Encrypted messages in one-to-one chats</string>
-    <string name="settings_messages_in_e2e_group_chat">Encrypted messages in group chats</string>
-    <string name="settings_when_rooms_are_upgraded">When rooms are upgraded</string>
```
bmarty, line 2952: Lint will complain that we have translation but not original string. So either remove all the existing translations or keep the original string and we do the cleanup in the strings from time to time.
I would go for the first option.

vector/src/main/res/xml/vector_settings_notification_global.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.preference.PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
+    <im.vector.app.core.preference.VectorPreferenceCategory
+        android:key="SETTINGS_OTHER"
```
bmarty, line 4: This is never used, but this is not the correct key (copy/paste error).

bmarty (approved):
LGMT, some small remarks, but nothing terrific. Thanks!

vector/src/main/res/values/strings.xml
bmarty: My bad, actually those strings has just been moved above. Please ignore my comment

vector/src/main/java/im/vector/app/features/settings/notifications/NotificationIndex.kt
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+
+package im.vector.app.features.settings.notifications
+
+import org.matrix.android.sdk.api.pushrules.rest.PushRule
+import org.matrix.android.sdk.api.pushrules.toJson
+
+enum class NotificationIndex(val index: Int) {
+    OFF(0),
```
ganfra, line 23: Enum already have indexes by default starting from 0

ganfra (approved):
Just one small remark otherwise LGTM

vector/src/main/java/im/vector/app/features/settings/notifications/NotificationIndex.kt
langleyd: Plus I didn't actually end up using the index value 🥴. Removed, thanks.

PR #3699, opened 2021-07-19
Complement for #3656

bmarty:
See #3656 

CC @SpiritCroc

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultFileService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultFileService.kt
@@ -131,8 +131,14 @@ internal class DefaultFileService @Inject constructor(
                     Timber.v("Response size ${response.body?.contentLength()} - Stream available: ${!source.exhausted()}")
 
                     // Write the file to cache (encrypted version if the file is encrypted)
-                    writeToFile(source.inputStream(), cachedFiles.file)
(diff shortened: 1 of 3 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultFileService.kt
```diff
@@ -96,6 +97,9 @@ internal class DefaultFileService @Inject constructor(
             }
         }
 
+        var atomicFileCreator1: AtomicFileCreator? = null
```
BillCarsonFr, line 100: Can it be renamed to something like encyptedCachedAtomicFile

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultFileService.kt
```diff
@@ -96,6 +97,9 @@ internal class DefaultFileService @Inject constructor(
...
+        var atomicFileCreator2: AtomicFileCreator? = null
```
BillCarsonFr, line 101: Same renamed to clearCachedAtomicFime

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultFileService.kt
bmarty: Renamed to `atomicFileDownload`, it can be a clear file

BillCarsonFr (approved):

Issue #3589, opened 2021-06-30
Element Android reuses txnId for to-device events in the verification process


timokoesters:
There are two types of transaction ids in the to-device verification process:
The txnId, which is a path parameter for the /sendToDevice request. As far as I understand, this should never be reused unless the previous try failed (https://spec.matrix.org/unstable/client-server-api/#put_matrixclientr0sendtodeviceeventtypetxnid).
Then there is a transaction_id field that is shared by the other messages of the verification session (https://spec.matrix.org/unstable/client-server-api/#key-verification-framework).

Element Android uses the same id for both txnId and transaction_id, so on Conduit most sendToDevice requests are dropped because the txnId was already used before. This is probably the reason for https://github.com/vector-im/element-android/issues/3183

How to fix:
Only use the session's transaction_id for the transaction_id field of verification message contents. Generate a new id for the txnId of /sendToDevice requests.

bmarty:
That's very good catch. We will have a look

bmarty:
Should be fixed on develop and in Element Android 1.1.14.

@timokoesters should be nice if you can confirm that! Thanks

PR #3707, opened 2021-07-20
crypto: Don't use the transaction ID of the verification for the request

poljar:
Verification flows have something called a transaction id. This is a client-set custom ID that identifies the flow and is established by the first message that gets sent out. This transaction ID needs to be kept the
same and be part of all events that are sent during the verification flow.

To-device requests have something called a transaction id. This is a client-set custom ID that identifies a given request. It is used to ensure idempotency of requests, i.e. retrying to send a request won't result in two events being sent as long as the transaction id is kept the same.

This patch removes usage of the first type of transaction ID for the second use-case.

Before:
before
before2

After:
after

To be honest, I have no idea how that ever worked.

This closes: #3589.

Signed-off-by: Damir Jelić poljar@termina.org.uk

### Pull Request Checklist

### DIFF
--- a/changelog.d/3589.bugfix
+++ b/changelog.d/3589.bugfix
@@ -1,0 +1,1 @@
+Don't use the transaction ID of the verification for the request
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/VerificationTransportToDevice.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/VerificationTransportToDevice.kt
@@ -68,7 +68,7 @@ internal class VerificationTransportToDevice(
             contentMap.setObject(otherUserId, it, keyReq)
         }
         sendToDeviceTask
-                .configureWith(SendToDeviceTask.Params(MessageType.MSGTYPE_VERIFICATION_REQUEST, contentMap, localId)) {
+                .configureWith(SendToDeviceTask.Params(MessageType.MSGTYPE_VERIFICATION_REQUEST, contentMap)) {
                     this.callback = object : MatrixCallback<Unit> {
                         override fun onSuccess(data: Unit) {
                             Timber.v("## verification [$tx.transactionId] send toDevice request success")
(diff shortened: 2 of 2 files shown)


bmarty:
Thanks for the PR, LGTM!

> To be honest, I have no idea how that ever worked.

There is maybe a bug to report on https://github.com/matrix-org/synapse/ ?

bmarty:
@poljar what about those other location?

- https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CancelGossipRequestWorker.kt#L95
- https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipRequestWorker.kt#L125
- https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipWorker.kt#L130

poljar:
> @poljar what about those other locations?
> 
>     * https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CancelGossipRequestWorker.kt#L95
> 
>     * https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipRequestWorker.kt#L125
> 
>     * https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipWorker.kt#L130

The short answer is: I don't know.

The long answer: All of those seem to generate a unique ID using the [LocalEcho.createLocalEchoId()](https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CancelGossipRequestWorker.kt#L69) method but the author of the code or someone else who knows how the gossiping in EA works might be a better candidate to answer this.

bmarty:
Thanks for the fast answer.
@BillCarsonFr WDYT? On my side I think we should remove the param `SendToDeviceTask.Params.transactionId` and let the `DefaultSendToDeviceTask` create a transactionId for each request using `UUID.randomUUID().toString()` (rather than `Random.nextInt(Integer.MAX_VALUE).toString()`)

poljar:
> Thanks for the fast answer.
> @BillCarsonFr WDYT? On my side I think we should remove the param `SendToDeviceTask.Params.transactionId` and let the `DefaultSendToDeviceTask` create a transactionId for each request using `UUID.randomUUID().toString()` (rather than `Random.nextInt(Integer.MAX_VALUE).toString()`)

This is largely unimportant, an UUID gives you 128 bits of randomness vs how many bits `Integer` in Kotlin land has, both are large enough for this to work correctly. That being said, the rust-sdk does use UUIDs for this.

bmarty (approved):
Thanks @poljar!

bmarty:
@poljar I let you delete the branch if you want to ^

bmarty:
> To be honest, I have no idea how that ever worked.

According to @richvdh :

> synapse includes the event type when de-duplicating, so you can use the same txn id for different messages as long as they have different event types

PR #3711, opened 2021-07-21
 Ensure that txnId is the same if the request is retried

bmarty:
Complement to #3707 

Ensure that txnId is the same if the request is retried.

Also improve the value of generated requestId. We now use UUIDs

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/VerificationService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/VerificationService.kt
@@ -52,7 +52,7 @@ interface VerificationService {
                              transactionId: String?): String?
 
     /**
-     * Request a key verification from another user using event in a room.
+     * Request key verification with another user via room events (instead of the to-device API)
      */
     fun requestKeyVerificationInDMs(methods: List<VerificationMethod>,
                                     otherUserId: String,
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/VerificationService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/VerificationService.kt
@@ -52,7 +52,7 @@ interface VerificationService {
                              transactionId: String?): String?
 
     /**
-     * Request a key verification from another user using toDevice events.
+     * Request a key verification from another user using event in a room.
      */
     fun requestKeyVerificationInDMs(methods: List<VerificationMethod>,
                                     otherUserId: String,
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CancelGossipRequestWorker.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CancelGossipRequestWorker.kt
@@ -23,12 +23,12 @@ import org.matrix.android.sdk.api.auth.data.Credentials
 import org.matrix.android.sdk.api.failure.shouldBeRetried
 import org.matrix.android.sdk.api.session.events.model.Event
 import org.matrix.android.sdk.api.session.events.model.EventType
-import org.matrix.android.sdk.api.session.events.model.LocalEcho
 import org.matrix.android.sdk.api.session.events.model.toContent
 import org.matrix.android.sdk.internal.crypto.model.MXUsersDevicesMap
 import org.matrix.android.sdk.internal.crypto.model.rest.ShareRequestCancellation
 import org.matrix.android.sdk.internal.crypto.store.IMXCryptoStore
 import org.matrix.android.sdk.internal.crypto.tasks.SendToDeviceTask
+import org.matrix.android.sdk.internal.crypto.tasks.createUniqueTxnId
 import org.matrix.android.sdk.internal.session.SessionComponent
 import org.matrix.android.sdk.internal.worker.SessionSafeCoroutineWorker
 import org.matrix.android.sdk.internal.worker.SessionWorkerParams
@@ -43,6 +43,9 @@ internal class CancelGossipRequestWorker(context: Context,
             override val sessionId: String,
             val requestId: String,
             val recipients: Map<String, List<String>>,
+            // The txnId for the sendToDevice request. Nullable for compatibility reason, but MUST always be provided
+            // to use the same value if this worker is retried.
+            val txnId: String? = null,
             override val lastFailureMessage: String? = null
     ) : SessionWorkerParams {
         companion object {
@@ -51,6 +54,7 @@ internal class CancelGossipRequestWorker(context: Context,
                         sessionId = sessionId,
                         requestId = request.requestId,
                         recipients = request.recipients,
+                        txnId = createUniqueTxnId(),
                         lastFailureMessage = null
                 )
             }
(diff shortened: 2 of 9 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CancelGossipRequestWorker.kt
```diff
@@ -43,6 +43,9 @@ internal class CancelGossipRequestWorker(context: Context,
             override val sessionId: String,
             val requestId: String,
             val recipients: Map<String, List<String>>,
+            // The txnId for the sendToDevice request. Nullable for compatibility reason, but MUST always be provided
```
poljar, line 46:
```suggestion
            // The txnId for the sendToDevice request. Nullable for compatibility reasons, but MUST always be provided
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipRequestWorker.kt
```diff
@@ -46,6 +46,9 @@ internal class SendGossipRequestWorker(context: Context,
             override val sessionId: String,
             val keyShareRequest: OutgoingRoomKeyRequest? = null,
             val secretShareRequest: OutgoingSecretRequest? = null,
+            // The txnId for the sendToDevice request. Nullable for compatibility reason, but MUST always be provided
```
poljar, line 49:
```suggestion
            // The txnId for the sendToDevice request. Nullable for compatibility reasons, but MUST always be provided
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipRequestWorker.kt
```diff
@@ -58,7 +61,8 @@ internal class SendGossipRequestWorker(context: Context,
     }
 
     override suspend fun doSafeWork(params: Params): Result {
-        val localId = LocalEcho.createLocalEchoId()
+        // (temporary code)
```
poljar, line 64: Temporary how/why?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipWorker.kt
```diff
@@ -48,6 +48,9 @@ internal class SendGossipWorker(context: Context,
             val requestUserId: String?,
             val requestDeviceId: String?,
             val requestId: String?,
+            // The txnId for the sendToDevice request. Nullable for compatibility reason, but MUST always be provided
```
poljar, line 51:
```suggestion
            // The txnId for the sendToDevice request. Nullable for compatibility reasons, but MUST always be provided
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipWorker.kt
```diff
@@ -62,7 +65,8 @@ internal class SendGossipWorker(context: Context,
     }
 
     override suspend fun doSafeWork(params: Params): Result {
-        val localId = LocalEcho.createLocalEchoId()
+        // (temporary code)
```
poljar, line 68: Same as above, why is this temporary?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendToDeviceTask.kt
```diff
@@ -46,16 +46,21 @@ internal class DefaultSendToDeviceTask @Inject constructor(
                 messages = params.contentMap.map
         )
 
+        // Create a unique txnId first, to use the same value if the request is retried
```
poljar, line 49: This comment is a bit hard to parse, also isn't it the other way around? We first try an existing transaction ID and if there isn't one we create a new random one?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/util/RequestIdHelper.kt
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.internal.crypto.util
+
+import java.util.UUID
+
+internal object RequestIdHelper {
+    fun createUniqueRequestId() = "req_" + UUID.randomUUID().toString()
```
poljar, line 22: What's the point of prepending `req_` here? 

As far as I know everyone who uses a UUID for transaction IDs does so without modification. This will only allow servers to know which client is doing the request, avoiding the information leak is probably more important than whatever reason we have to do this.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/VerificationService.kt
```diff
@@ -52,7 +52,7 @@ interface VerificationService {
                              transactionId: String?): String?
 
     /**
-     * Request a key verification from another user using toDevice events.
+     * Request a key verification from another user using event in a room.
```
poljar, line 55:
```suggestion
     * Request key verification with another user via room events (instead of the to-device API)
```

poljar (changes requested):
I'm not quite as fluent in the EA codebase as I would like to give proper reviews of the functional change (i.e. does this code indeed make sure that the transaction id is reused when requests are retried), but the comments and the way the UUID is modified could do with some refinement.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendToDeviceTask.kt
bmarty: I agree, I will update it

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipRequestWorker.kt
bmarty: params.txnId should be provided in all cases now. But `Params` can be deserialized by the WorkManager from data serialized in a previous version of the application, so without the `txnId`. That's why it has to stay nullable for a while... Probably until we delete all this code to switch to the RustSdk.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/util/RequestIdHelper.kt
bmarty: OK, removed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/SendGossipRequestWorker.kt
poljar: The comment is in this state isn't that helpful if that's going to hang around until the rust-sdk comes around. I would either expand it a bit to explain the state of affairs or just remove it to not confuse people that might be reading this code.

bmarty: You are right, I've updated the comment (at the 3 places).

poljar (approved):
Looks good from my point of view.

PR #3715, opened 2021-07-21
Add initialState support to CreateRoomParams

### LINKED ISSUE #3713: Support initial states in CreateRoomParams
As in previous version of Element, we should be able to send a list of initial events after the room creation in CreateRoomParams.

There is a remaining TODO about that need in this class `// TODO Give a way to include other initial states`.

bmarty:
Fixes #3713

Usage, for instance to set the room name (which can be done normally more simplier):

```kotlin
CreateRoomParams().apply {
    initialStates.add(CreateRoomStateEvent(
            EventType.STATE_ROOM_NAME,
            content = RoomNameContent(
                    name = state.roomName
            ).toContent()
    ))
}
```

Tested OK on my side, with temporary code. There is no dedicated test for room creation yet...

### DIFF
--- a/changelog.d/3713.removal
+++ b/changelog.d/3713.removal
@@ -1,0 +1,1 @@
+Add initialState support to CreateRoomParams (#3713)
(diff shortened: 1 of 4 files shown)


Florian14 (approved):
LGTM

PR #3720, opened 2021-07-22
Fix a crash which can happen when user signs out.

bmarty:
The crypto DB has been deleted, and the key download request is cancelled, but in the catch block we tried to write to the deleted DB.

Tested OK on my device.


onurays (approved):
For a static review LGTM!

BillCarsonFr (approved):

PR #3726, opened 2021-07-22
Add the resent request to the db

bmarty:
It was not the case. So I'm not sure how it could work...


bmarty:
Note: it does not work, `getOrAddOutgoingRoomKeyRequest` return the cancelled request... Will update

PR #3743, opened 2021-07-28
Update Account Data with user matrix id for invited user by email

### LINKED ISSUE #3643: DM created to invite someone by email is stuck on the email address in the AccountData
The current implementation of getDirectRoom ignores the actual directUserId.
Indeed it returns a direct by checking only the members (otherUserId) without considering the directUserId
This implementation allowed to handle the case of a direct chat created by inviting the other by email (while the account was not created)
Indeed in that case the right DM is returned by getDirectRoom whereas the directUserId is the invited email
We suggest here to update in this case the directUserId by replacing it with the provided matrix ID
in order to update the direct dictionary of the account data with the actual data

To sum up:
Force an update of the account data if the provided id is a matrix Id and the returned summary doesn't have this matrix id as directUserId value (we may limit this update to the case where the current directUserId is an email address. But we don't see the need to consider this limitation. Force the update in any cases sound the right thing to do)

#### To Reproduce
Given: existing UserA, non existing UserB
1. UserA: in Element, invite a contact from the contacts book to a direct message
2. UserA: an « Empty room » is created with the invited user email
3. UserB: click on the link received by email and join element
4. UserB: configure the identity server and turn on discovery by email
5. UserA: room displayName is updated with userB id
6. Bug: the room directUserId has not been updated with the userB id
7. Second Bug: in room details, two invites are now displayed, the first one with the userB email and the second one with the userB id.

#### Expected behavior
- directUserId should be changed from the user email to the user id as soon as it has been catched
- in room details, email invite should not be displayed anymore as soon as there is an invite with the user id

#### Screenshot
In the screenshot, the second user has joined the room but the email invite is still pending
Screenshot_1625563943

Florian14:
Fixes #3643 

Update the Account Data when an user has been invited to a DM by email (which is not bound to an existing account).

The DM is initially populated with the invited user email and stuck with this email when the user has joined Element by creating an account. Therefore, the `RoomSummary.directUserId` is also stuck on the email and does not match the otherUserId which has been updated with the user matrix id.

To handle the user registration, we catch the `m.room.member` event which contains the matrix id as part of the third_party_invite content. Then, we check the account data dictionary and update the entries matching the roomId attached to the event, and where the user id does not match the matrix id pattern.
Update the account data will then cause the update of the directUserId in the RoomSummary.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
@@ -17,6 +17,7 @@
 package org.matrix.android.sdk.internal.session.sync
 
 import org.matrix.android.sdk.api.MatrixPatterns
+import org.matrix.android.sdk.internal.session.sync.model.accountdata.toMutable
 import org.matrix.android.sdk.internal.session.user.accountdata.DirectChatsHelper
 import org.matrix.android.sdk.internal.session.user.accountdata.UpdateUserAccountDataTask
 import javax.inject.Inject
@@ -38,7 +39,7 @@ internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
     }
 
     private suspend fun updateDirectUserIds(directUserIdsToUpdate: Map<String, String>) {
-        val directChats = directChatsHelper.getLocalDirectMessages()
+        val directChats = directChatsHelper.getLocalDirectMessages().toMutable()
         var hasUpdate = false
         directUserIdsToUpdate.forEach { (roomId, candidateUserId) ->
             // consider room is a DM if referenced in the DM dictionary
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
@@ -16,18 +16,53 @@
 
 package org.matrix.android.sdk.internal.session.sync
 
+import org.matrix.android.sdk.api.MatrixPatterns
+import org.matrix.android.sdk.internal.session.user.accountdata.DirectChatsHelper
+import org.matrix.android.sdk.internal.session.user.accountdata.UpdateUserAccountDataTask
 import javax.inject.Inject
 
 internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
-        private val ephemeralTemporaryStore: RoomSyncEphemeralTemporaryStore
+        private val directChatsHelper: DirectChatsHelper,
+        private val ephemeralTemporaryStore: RoomSyncEphemeralTemporaryStore,
+        private val updateUserAccountDataTask: UpdateUserAccountDataTask
 ) {
-    fun handle(synResHaResponsePostTreatmentAggregator: SyncResponsePostTreatmentAggregator) {
+    suspend fun handle(synResHaResponsePostTreatmentAggregator: SyncResponsePostTreatmentAggregator) {
         cleanupEphemeralFiles(synResHaResponsePostTreatmentAggregator.ephemeralFilesToDelete)
+        updateDirectUserIds(synResHaResponsePostTreatmentAggregator.directChatsToCheck)
     }
 
     private fun cleanupEphemeralFiles(ephemeralFilesToDelete: List<String>) {
         ephemeralFilesToDelete.forEach {
             ephemeralTemporaryStore.delete(it)
+        }
+    }
+
+    private suspend fun updateDirectUserIds(directUserIdsToUpdate: Map<String, String>) {
+        val directChats = directChatsHelper.getLocalDirectMessages()
+        var hasUpdate = false
+        directUserIdsToUpdate.forEach { (roomId, candidateUserId) ->
+            // consider room is a DM if referenced in the DM dictionary
+            val currentDirectUserId = directChats.firstNotNullOfOrNull { (userId, roomIds) -> userId.takeIf { roomId in roomIds } }
+            // update directUserId with the given candidateUserId if it mismatches the current one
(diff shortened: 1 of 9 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -16,18 +16,53 @@
...
+                        .getOrPut(candidateUserId) { arrayListOf() }
+                        .apply {
+                            if (!contains(roomId)) {
+                                hasUpdate = true
+                                add(roomId)
+                            }
+                        }
+
+                // remove roomId from currentDirectUserId entry
+                hasUpdate = hasUpdate or(directChats[currentDirectUserId]?.remove(roomId) == true)
```
bmarty, line 59: is it necessary to assign `hasUpdate` here?
Also I think that if `hasUpdate` is true, the part on the right of `or` will not be evaluated.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/model/accountdata/DirectMessagesContent.kt
```diff
@@ -19,4 +19,4 @@ package org.matrix.android.sdk.internal.session.sync.model.accountdata
 /**
  * Keys are userIds, values are list of roomIds
  */
-internal typealias DirectMessagesContent = Map<String, List<String>>
+internal typealias DirectMessagesContent = MutableMap<String, MutableList<String>>
```
bmarty, line 22: The change from immutable to mutable is not ideal here.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -16,18 +16,53 @@
...
+                // remove currentDirectUserId entry if there is no attached room anymore
+                hasUpdate = hasUpdate or(directChats.takeIf { it[currentDirectUserId].isNullOrEmpty() }?.remove(currentDirectUserId) != null)
```
bmarty, line 61: Same remark

bmarty (commented):
Some small remarks/question. Can you add a file for towncrier please?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
Florian14: Yes, the `remove` method will modify the direct chat dictionary if the given item has been found and removed, so I want to edit this flag in that case.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
Florian14: Same answer, If I remove a key from the directChats dictionary, I also want to edit this flag in order to update the accountData with the updated dictionary

Florian14:
> Some small remarks/question. Can you add a file for towncrier please?

Done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
bmarty: Yes, but I was wondering if `hasUpdate` was not already equal to true in this case. It can maybe add some cleanup on some existing account data.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/model/accountdata/DirectMessagesContent.kt
bmarty: can you try this ugly code?

```kotlin
internal typealias DirectMessagesContent = Map<String, List<String>>

internal fun DirectMessagesContent.mutate(): MutableMap<String, MutableList<String>> {
    return map { it.key to it.value.toMutableList() }
            .toMap()
            .toMutableMap()
}
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
bmarty: Ok, I think I understand now.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
bmarty: `!= null` could be replaced by `== true`, no?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
Florian14: I'm not sure to understand, I use the `or` function which will always call the second part of the operation (whether hasUpdate is true or not). The flag should be updated for each modification in the direct chats dictionary. In our case, the cleanup will be performed only for members related to the given `directUserIdsToUpdate` dictionary.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponsePostTreatmentAggregatorHandler.kt
Florian14: Nop, in that case, remove returns the removed object, if any or null otherwise

bmarty (approved):
LGTM, thanks for the update

PR #3765, opened 2021-07-30
Spaces | Default to 'Home' in settings

### LINKED ISSUE #3754: M11.12 Spaces | Default to 'Home'
* Default to ‘Home’ out of the box
* Allow users to opt in to ‘All’ in Settings
* [Prev design](https://www.figma.com/file/baCEQP0BvtuitBJUAtapoc/Spaces-Beta?node-id=5449%3A184167)

BillCarsonFr:
Fixes #3754

Moved and renamed the lab preference to a regular preference

image


vector/src/main/res/values/strings.xml
```diff
@@ -3417,7 +3417,10 @@
     <string name="space_manage_rooms_and_spaces">Manage rooms and spaces</string>
 
 
-    <string name="labs_space_show_orphan_in_home">Experimental Space - Only show orphans in Home</string>
```
bmarty, line 3420: Lint is not happy, you have to remove all the existing translations for this string.

vector/src/main/res/xml/vector_settings_preferences.xml
```diff
@@ -30,6 +30,16 @@
 
     </im.vector.app.core.preference.VectorPreferenceCategory>
 
+    <im.vector.app.core.preference.VectorPreferenceCategory android:title="@string/spaces_header">
```
bmarty, line 33: Maybe create a new string for the setting section instead of reusing `spaces_header`?

bmarty (changes requested):
LGTM, 2 small remarks.
Cannot be merged because of lint

vector/src/main/res/values/strings.xml
BillCarsonFr: done

vector/src/main/res/xml/vector_settings_preferences.xml
BillCarsonFr: done

PR #3772, opened 2021-08-01
Improve accessibility of voice messages

pvagner:
Signed-off-by: Peter Vágner <pvdeejay@gmail.com>

### Pull Request Checklist


vector/src/main/res/layout/view_voice_message_recorder.xml
```diff
@@ -231,10 +233,11 @@
         android:layout_height="wrap_content"
         android:layout_marginBottom="84dp"
         android:visibility="gone"
+        android:accessibilityLiveRegion="polite"
```
bmarty, line 236: This could become a bit annoying to speak out the text each time the view become visible, since the "toast" is displayed each time a voice recording starts, and also in some other circumstances. Maybe we should improve that in the code.

bmarty (commented):
LGTM, thanks. Just one remark, what is your point of view?

vector/src/main/res/layout/view_voice_message_recorder.xml
pvagner: I haven't strong objection regarding this. My thinking is that this way toast gets spoken whenever it is shown and hint on how to operate this feature will be easier to discover for screen reader users. Other idea is to implement custon AccessibilityNodeInfo class for the mic button, adding the double tap accessibility action with appropriate description. The latter is more difficult to implement so I went with this initially. If you like I can revert this part and we can decide  later.

pvagner:
Since I have submitted this one @onurays  has been also improved most of the things indepently of this PR. He has added content descriptions to the play button and fixed the vibration feetback.
Now what's missing in the develop branch is content description for the waveform and dynamically hiding it from accessibility services when in the playing state. I know it's not a proper button but for visually disabled people it's nice to add a description there and it can be used to stop the recording by double tapping, so it's natural to make it like this.
Other than that I have changed description for the mic button so it says record voice message rather than start voice message, it sounds nicer to me but I have no strong objection about the wording, so I can accept your advice on that too.

Huge thanks for reviewing and looking at this.

vector/src/main/res/layout/view_voice_message_recorder.xml
bmarty: Ok, I see your point, thanks for the fast feedback. Let's keep it like that for now.

bmarty (approved):

bmarty (approved):
Thanks for the update

Issue #3243, opened 2021-04-28
No sound notifications (noisy) when using Wi-Fi


bahur142:
#### Describe the bug
Since version 1.1.6 there is a strange bug, that I can reproduce in 4 different devices. There are no sound notifications (noisy notifications) when you are using Wi-Fi. But when you are using mobile data sound notifications are fine. 
**This is valid only for devices with Android 8 and above. Devices with Android 5,6,7 are not affected.**

#### To Reproduce

1. Install the latest version no matter how - Play Store, Fdroid, Buldkite
2. Switch on Wi-Fi data and ask someone to message you in one-to-one chat with noisy notifications.
3. You receive a message notification (led, vibration, pop-up), **but no sound**.
4. Now switch on mobile data and ask again someone to message you in one-to-one chat.
5. You receive a message notification, but this time with sound. As it should be in one-to-one.

Expected behavior:
Sounds should work no matter WiFi or mobile data when noisy notifications are enabled.

 - Devices: all devices using Android 8 and above.

 - App version and store: all versions since 1.1.6
 - Homeserver: Synapse 1.31

hifi:
I was just about to post this as well. Also reproduced on a Samsung XCover Pro and Huawei P30 Pro where both are on a recent enough Android. 

Though sometimes in addition to vibration I get partial sound like it's cut off. Most of the time nothing. I didn't notice using Wi-Fi was the common denominator so good catch! 

I'm on the current Play Store version of 1.1.6.

pilpo:
Yep, same here. On one-to-one, I have no sound notification. But in room with multiple person, it's ok.

funkesoftware:
I can confirm that. I've got the same behavior on my Motorola Moto G6 and my wifes HTC.
I hope this gets fixed soon.

LinAGKar:
Still there in 1.1.7 [40101072] (G-b2298)

matrixes:
I'm the author of #3537 and I have now been able to reproduce _this_ issue's... issue. I can get my notification audibly playing while on mobile data but not when using WiFi. It took me to go to the android web app and put all notification settings to Noisy to get to this point. I had done this exact same setting change in the Android app multiple times but that didn't affect anything. It was only when I just recently made the Noisy settings change on the web that audio started playing on my Android devices (I tested multiple).

It's _super weird_ that WiFi on/off has any affect on this..... And also, I'm more than a little bit confused about the web vs Android when it comes to the application of Notification settings....

[...]

PR #3786, opened 2021-08-04
Fix missing sounds on room notifications (messages, invitations, etc).

langleyd:
### Old behaviour
On receiving of push notifications we do a background sync to fill in missing information in the push content.
Previously we would update the notification with a call to notify but with the sound disabled in an attempt to update the notification and avoid a second notify sound.
Depending on on the time it would take to do the background sync you would sometimes see different behaviours:
- The first sound would play, the sync would complete after it had finished and the the push would be updated silently.
- The first sound would start to play,  the sync would complete during it playing and the, second call to notify would cut off the sound playing.
- The sync would complete before the sound played, the second call to notify would in effect silence the sound.

### New behaviour
I am removing the modification of the notification `noisy` property and as per [this guide](https://developer.android.com/training/notify-user/build-notification#Updating) using the `setOnlyAlertOnce` property on all group notifications other than the summary notification. So that any sounds on notifications are not silenced(or double fired) by updates. Child notifications continue to be rolled up into a single summary notification [here](https://github.com/vector-im/element-android/blob/1d9baaf0e34f1efd6f05d3feed72f8cbd0bc6e10/vector/src/main/java/im/vector/app/features/notifications/NotificationUtils.kt#L567).

The behaviour also appears different across android versions. As reported in the issue it appears to affect versions 8+ but in those below it does not seem to cut off the notification on update.

I tested this on 7.0 and 8+ and is working across both. 
I'll look for somebody else to test this on API < 24 as I don't have an old device and it's difficult to test push on the emulator for versions that old.
I'll also wait for this to be reviewed by @bmarty when he is available as I think he has most context on this code. 

Fixes https://github.com/vector-im/element-android/issues/3243

### DIFF
--- a/changelog.d/3243.bugfix
+++ b/changelog.d/3243.bugfix
@@ -1,0 +1,1 @@
+Notifications - Fix missing sound on notifications.
--- a/vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
+++ b/vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
@@ -100,13 +100,14 @@ class NotificationDrawerManager @Inject constructor(private val context: Context
             if (existing != null) {
                 if (existing.isPushGatewayEvent) {
                     // Use the event coming from the event stream as it may contains more info than
-                    // the fcm one (like type/content/clear text)
+                    // the fcm one (like type/content/clear text) (e.g when an encrypted message from
+                    // FCM should be update with clear text after a sync)
                     // In this case the message has already been notified, and might have done some noise
                     // So we want the notification to be updated even if it has already been displayed
-                    // But it should make no noise (e.g when an encrypted message from FCM should be
-                    // update with clear text after a sync)
+                    // Use setOnlyAlertOnce to ensure update notification does not interfere with sound
+                    // from first notify invocation as outlined in:
+                    // https://developer.android.com/training/notify-user/build-notification#Updating
                     notifiableEvent.hasBeenDisplayed = false
-                    notifiableEvent.noisy = false
                     eventList.remove(existing)
                     eventList.add(notifiableEvent)
                 } else {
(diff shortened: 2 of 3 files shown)


DoM1niC:
Ringing on incoming Call only do a simple Notification Sound? I think in v1 there was ringing ????

LinAGKar:
> Ringing on incoming Call only do a simple Notification Sound? I think in v1 there was ringing ????

Something I've noticed is that if you have "Sound" set to something in the ringing notification category (which I had, although I don't remember changing it), it will just play that sound, but if you set it to none, it will play the phones normal ringtone.

langleyd:
> Ringing on incoming Call only do a simple Notification Sound? I think in v1 there was ringing ????

@DoM1niC @LinAGKar This PR is just related to room notifications like message notifications not having a sound. If you are seeing other notification issues (like calling) mind creating a new Issue with steps to reproduce?

LinAGKar:
For something actually related to this pull request, I'm wondering if setOnlyAlertOnce() also affects smart watch notifications (in which this PR should also solve #1673/#2270), or if it just affects the alerts on-device.

bahur142:
Hi.
I can confirm that this PR works on devices with Android 8.1 and above. The sound is back again on Wi-Fi !
On Android devices below 8.1, the notification sound plays twice for a single message. But it is better than noting.

bmarty (approved):
LGTM, thanks!

bmarty:
@bahur142 thanks for your test. 

> There is one side effect after two-three days of testing. Half of the time, the phone did not wake-up nor shows missed call when someone calls me. I have to open the app to show the missed call.
Before to apply the PR, there was no such issue. Unfortunately I have to revert it.

Let's merge the PR and see if the regression you are describing can be reproduced by others.

LinAGKar:
Doesn't seem to work perfectly. Now with 1.2.2 I sometimes get double sounds, on Android 11 (which is an improvement over missed notifications, but still).

ouchadam:
the double notification issues sounds like this old 8.0 problem https://stackoverflow.com/questions/46653731/android-8-0-double-notification-sound, I've not been able to reproduce it myself 

it might be as simple as adding `.setGroupAlertBehavior(GROUP_ALERT_SUMMARY)` to our grouped notifications like the post suggests :crossed_fingers:

binarydad:
I know this is an old request, but given the current state of the code, would anyone know of a (hopefully) quick way to only have the alert sound happen once? I'm trying to build an alternate version that simply plays the sound for each DM, as my kids and I use it, and they constantly miss notifications from messages. 

My original post is here: https://github.com/vector-im/element-android/issues/5161

I am not an Android developer, so I know very little about it (so far), but am merely trying to change the behavior of the audible alerts from just the first alert (setOnlyAlertOnce) to an alert for EVERY DM. 

I'm able to get sounds for every DM currently, but as mentioned in this thread, the sounds are doubled for older devices. Would anyone have a solution? Many thanks.

PR #3793, opened 2021-08-04
Fix attachment menu animation order

pr0gr8mm3r:
### Pull Request Checklist

Tested on device with Android 11, shouldn't matter (assuming it worked before 😉)

# What it does
This PR changes the order in which the items in the attachement menu appear. Currently, the ones on the right are faster than the ones on the left. Because the '+'-Button to trigger the menu is on the left, this doesn't make sense.

This behavior is more noticable if animations are disabled, because the menu doesn't have the circular mask animation. That's why the videos are recorded with animations disabled.

# Current animation
https://user-images.githubusercontent.com/37022952/128323836-4de37c81-6fd0-44d8-bd67-09713c635942.mp4

# New animation
https://user-images.githubusercontent.com/37022952/129963137-51ae4628-c16f-44e9-9ff9-2be4550c058a.mp4

### DIFF
--- a/changelog.d/3793.bugfix
+++ b/changelog.d/3793.bugfix
@@ -1,0 +1,1 @@
+Fix order in which the items of the attachement menu appear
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks for the PR, you are right, the animation are better like this.

Previously the plus button was on the right and has been recently moved to the left.

As a small remark users with RTL languages does not have the bug at the moment, but after this PR will be merged they will see the bug. But this problem is not introduced by your PR.

bmarty:
My bad in RTL, the icons are also mirrored, so it will be fine.

PR #3804, opened 2021-08-06
Fix error when sending encrypted message if someone in the room logs out

onurays:


manuroe (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/RealmCryptoStore.kt
```diff
@@ -286,10 +286,13 @@ internal class RealmCryptoStore @Inject constructor(
                 val userEntity = UserEntity.getOrCreate(realm, userId)
                 // First delete the removed devices
                 val deviceIds = devices.keys
-                userEntity.devices.iterator().forEach { deviceInfoEntity ->
-                    if (deviceInfoEntity.deviceId !in deviceIds) {
-                        Timber.d("Remove device ${deviceInfoEntity.deviceId} of user $userId")
-                        deviceInfoEntity.deleteOnCascade()
+                userEntity.devices.toTypedArray().iterator().let {
```
bmarty, line 289: I trust you on the fact that it fixes the issue, but I'm wondering why I used an iterator at the first place and now if it should also work using `userEntity.devices.forEach {` or eventually `userEntity.devices.toList().forEach {`.

Anyway let's keep it like that for now, thanks for the fix!

PR #3816, opened 2021-08-10
Voice Message - UI Improvements

### LINKED ISSUE #3798: Voice Message - UI Improvements
- [x] Locking background is flickering. Smooth the transition between cancel and lock states.
- [x] Dynamically set content description of play/pause button in bot timeline and locked state.
- [x] If the mic icon is invisible, remove the content description.
- [x] Vibrate when the recording is canceled.
- [x] If onPause is called, cannot draw waveform while recording.

onurays:
Fixes #3798


vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageRecorderView.kt
```diff
@@ -87,6 +88,15 @@ class VoiceMessageRecorderView @JvmOverloads constructor(
         initListeners()
     }
 
+    override fun onVisibilityChanged(changedView: View, visibility: Int) {
+        super.onVisibilityChanged(changedView, visibility)
+        if (changedView == this && visibility == VISIBLE) {
```
ganfra, line 93: Why doing this in this method? shouldn't be handled where you look for recordindState?

ganfra (changes requested):
I still notice some issues:
- When typing text and sending it, there some kind of blink between the Send button and Mic button.
- Slide actions (hold and cancel) should probably be triggered only when you release your finger, and not as soon as you pass the threshold value? 
- When you cancel the recording I have vibrate which is called twice in a row

vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageRecorderView.kt
onurays: Because if the composer isn't blank we are hiding (making invisible) the mic icon and we need to clear the content description. Visibility is changing but recordingState is the same (NONE).

ganfra (approved):
Thanks for the changes!

PR #3834, opened 2021-08-14
Add mxid to autocomplete suggestion if multiple users have same displayname

Tigermouthbear:
This PR adds the mxid of a user if they have the same displayname as another user in the autocomplete suggestions.

### Screenshot

Signed-off-by: Tigermouthbear tigermouthbear@tigr.dev

### Pull Request Checklist


vector/src/main/java/im/vector/app/features/autocomplete/member/AutocompleteMemberController.kt
```diff
@@ -35,12 +37,34 @@ class AutocompleteMemberController @Inject constructor() : TypedEpoxyController<
...
+            data.forEach { user ->
+                autocompleteMatrixItem {
+                    id(user.userId)
+                    matrixItem(user.toMatrixItem())
+                    avatarRenderer(host.avatarRenderer)
+                    clickListener { host.listener?.onItemClick(user) }
+                }
+            }
+        } else {
+            val dataSeparated = data.separateDuplicateDisplayNames()
```
bmarty, line 52: I think this treatment should be done in `AutocompleteMemberPresenter`.

vector/src/main/java/im/vector/app/features/autocomplete/member/AutocompleteMemberController.kt
```diff
@@ -35,12 +37,34 @@ class AutocompleteMemberController @Inject constructor() : TypedEpoxyController<
...
+            dataSeparated.single.forEach { user ->
+                autocompleteMatrixItem {
+                    id(user.userId)
+                    matrixItem(user.toMatrixItem())
+                    avatarRenderer(host.avatarRenderer)
+                    clickListener { host.listener?.onItemClick(user) }
+                }
+            }
+            dataSeparated.duplicate.forEach { user ->
```
bmarty, line 61: The order will not be preserved. The users with duplicate display name will always be displayed at the end of the list.

bmarty (changes requested):
Thanks for the PR.

I have not tested it but maybe just updating the `AutocompleteMemberPresenter` class with this new method is enough:

```kotlin
private fun Sequence<RoomMemberSummary>.disambiguate(): Sequence<RoomMemberSummary> {
    val displayNames = groupBy { it.displayName }.mapValues { it.value.size }

    return map { roomMemberSummary ->
        if (displayNames[roomMemberSummary.displayName] ?: 0 > 1) {
            roomMemberSummary.copy(
                    displayName = roomMemberSummary.displayName + " " + roomMemberSummary.userId
            )
        } else {
            roomMemberSummary
        }
    }
}
```

And then call `disambiguate()` just after `sortedBy` here: https://github.com/vector-im/element-android/blob/main/vector/src/main/java/im/vector/app/features/autocomplete/member/AutocompleteMemberPresenter.kt#L73

Tigermouthbear:
Thanks for the feedback! I implemented what you described, but I changed it a bit to make it disambiguate ignoring case of the usernames.

Tigermouthbear:
Sorry, I accidentally closed this

vector/src/main/java/im/vector/app/features/autocomplete/member/AutocompleteMemberPresenter.kt
```diff
@@ -71,6 +71,23 @@ class AutocompleteMemberPresenter @AssistedInject constructor(context: Context,
...
 }
+
+private fun Sequence<RoomMemberSummary>.disambiguate(): Sequence<RoomMemberSummary> {
+    val displayNames = hashMapOf<String, Int>().also { map ->
+        for (item in this) {
+            item.displayName?.lowercase()?.also { displayName ->
+                map[displayName] = map.getOrPut(displayName, { 0 }) + 1
+            }
+        }
+    }
```
bmarty, line 86: Why not simply write:
`val displayNames = groupBy { it.displayName?.lowercase() }.mapValues { it.value.size }`

Also, I do not think we disambigate display name regarding the case, for instance in the timeline, for sender name. But why not, I think it's worth it here.

bmarty (commented):
Thanks for the update.
One last remark.

Also it could be nice to squash all your commits to reduce the git history noise. I can do it when merging the PR, but feel free to do it and force push your branch.

vector/src/main/java/im/vector/app/features/autocomplete/member/AutocompleteMemberPresenter.kt
Tigermouthbear: I didn't use groupBy and mapValues because they each loop over the sequence, and because groupBy creates a List for every displayName. I didn't want there to be a big performance impact for disambiguation (especially since it loops over thousands of people in big rooms), therefore I wrote it differently so it would loop only once without creating a list for each displayName.

bmarty: I see, thanks for the feedback

bmarty (approved):

Issue #3841, opened 2021-08-16
Authenticated Jitsi not working in release


ganfra:
Some JWT code  is being removed by Proguard, so have to add rules.

PR #3842, opened 2021-08-16
Fix Authenticated Jitsi not working in release #3841

ganfra:
Fix https://github.com/vector-im/element-android/issues/3841


onurays (approved):
LGTM.

Issue #3159, opened 2021-04-12
No notifications when element is running in background (Huawei P30)


MarkWieczorek:
### Description

When using Element on a Huawei P30 mobile phone, I recieve all notifications for new messages and video calls when the app is running and open in the foreground. However, when the app is running in the background (i.e., when I am on the launcher, or using a different app), I never receive notifications for new messages/calls.

I note that everything works fine on another mobile I use (Galaxy S9). However, a colleague of mine that has a Huawei P30 has the exact same problem as I do with mine.

I have gone through all the settings on the P30, and there is nothing that I can see that would cause this problem. As this issue is repeatable with the two different P30 phones, there appears to be some kind of bug in how Element notifications is interacting with the P30.

DoM1niC:
> ### Description
> When using Element on a Huawei P30 mobile phone, I recieve all notifications for new messages and video calls when the app is running and open in the foreground. However, when the app is running in the background (i.e., when I am on the launcher, or using a different app), I never receive notifications for new messages/calls.
> 
> I note that everything works fine on another mobile I use (Galaxy S9). However, a colleague of mine that has a Huawei P30 has the exact same problem as I do with mine.
> 
> I have gone through all the settings on the P30, and there is nothing that I can see that would cause this problem. As this issue is repeatable with the two different P30 phones, there appears to be some kind of bug in how Element notifications is interacting with the P30.

What Version are you using? GooglePlay or FDroid

MarkWieczorek:
version 1.1.3 from F-droid.

bahur142:
This is due to Huawei's EMUI and their custom Android. The problem is not Element, but Huawei.
All of the Huawei devices experiencing so many problems with push notifications, even with the most popular messaging apps. All of my friends with Huawei has problems.

DoM1niC:
Huawei works fine with P30 Pro and Mate 20 lite on Android 10 on FDroid set the batterie options to manual and all will works, auto settings will close the App instantly !

MarkWieczorek:
> This is due to Huawei's EMUI and their custom Android. The problem is not Element, but Huawei.
> All of the Huawei devices experiencing so many problems with push notifications, even with the most popular messaging apps. All of my friends with Huawei has problems.

Perhaps, but notifications work fine with, at least, Telegram and Signal.

> Huawei works fine with P30 Pro and Mate 20 lite on Android 10 with FDroid set the batterie options to manual and all will works, auto settings will close the App instantly !

[...]

Issue #3263, opened 2021-04-30
Background process stops & Notifications are not working on LineageOS (no google servicies)


peterflock:
#### Describe the bug
Notification are not showing.

#### To Reproduce
1. Setup notifications to "Optimize for real-time", delay sync default 60
2. Battery optimization in Android setting is "Not Optimized" for Element
3. Troubleshooting notifications: all green 8/8
4. When I open the app I see sticky notification "Listening for events"
5. After I scroll the "top menu with notification" over app or I close the app the sticky notification diesappers

#### Expected behavior
I believe the sticky notification should be there all the time to keep background process running. Signal, Telegram and all other apps are doing it this way.

#### Screenshots

**Notification settins:**
![image](https://user-images.githubusercontent.com/78184669/116711857-c774c400-a9c2-11eb-8fa1-659a1dae24f3.png)

**Sticky notification UP** (will disappear shortly, after I close app / keep scrolled top-menu for some time)
![image](https://user-images.githubusercontent.com/78184669/116712183-1fabc600-a9c3-11eb-825a-0c2e97bd0791.png)

#### Smartphone
 - Device: OnePlus 7Pro
 - OS: LineageOS 18 (Android 11)

#### Additional context
 - App version: 1.1.6 (F-Droid)
 - Homeserver: matrix.org

lanmarc77:
#2143 
I created a patched fdroid only version in that thread. Check if that is more stable.

Issue #2827, opened 2021-02-16
Can't receipt of nitification


imotosaiko:
**Describe the bug**
When Element is running in the background, it can't send me a notification when someone is sending me a message.When I reenter Element, these notifications, which were supposed to appear hours ago, appear at the top of the screen.
Also, I noticed that every time I went to Element, it looked like I was reconnecting and refreshing the interface.I don't know if this is the cause of the above problems.

**Expected behavior**
I want to be notified as soon as a friend sends me a message, which I think is extremely important for a chat app.

**Screenshots**
I have no.

**Smartphone (please complete the following information):**
 - Device: Redmi K30i
 - OS: MIUI 12.1.1 [Android11]

**Additional context**
 - my app version : 1.0.17

lanmarc77:
#2143 
I created a patched fdroid only version in that thread. Check if that is more stable.

Issue #2669, opened 2021-01-14
Notifications in F-Droid version doesn't work


keleset:
Element is getting killed by Android system while being in the background. This completely disables notifications until next manual start. This happens on every Android device I've tested (Android 8.0 - Android 10).
Disabling every Android setting for lower power usage doesn't help.
**The reason for that - Element doesn't display permanent notification.**
I believe even Android dev manuals explicitly state that every permanent background app should have permanent notification now.
RiotX worked perfectly fine while having this permanent notification.
I have no Idea why this regression took place while transitioning from RiotX codebase to Element, but I guess this should be fix ASAP, because as far as I understand (correct me if I'm wrong), this is quite a simple thing to fix and right now it makes F-Droid version of Element pretty much unusable.

kszzsk:
[https://github.com/vector-im/element-android/issues/1685#issuecomment-745460067](url) :(

sydbarrett74:
@kszzsk That's a dead link. :(

kszzsk:
@sydbarrett74 corrected

mathiser:
Is the play store version different to the droid version on this matter? I have had the same trouble...

lanmarc77:
#2143 
I created a patched fdroid only version in that thread. Check if that is more stable.

PR #3844, opened 2021-08-16
(rough) patch for background sync in fdroid real time mode

lanmarc77:
This patches (not fixes) #3263,  #3159, #2827, #2669, #2143
I suggest a more thorough rewrite of the syncing process but this at least a working version until this has been done.

As a crashed or killed service is restarted individually I also needed to change the default in BackgroundDetectionObserver to true so that if only the service is started it defaults to background sync values.
Depending on the state of the alarms of the device and the state of VectorSyncService it might be, that the alarm based sync is triggered and the GuardService one resulting in "double" sync based on the selected sync frame. I found this acceptable for this kind of "rough" patch. You might think different though.

This patch might need a bit more work e.g. on the frontend to allow enabling or disabling the experimental GuardService.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/BackgroundDetectionObserver.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/BackgroundDetectionObserver.kt
@@ -29,7 +29,7 @@ import javax.inject.Inject
 @MatrixScope
 internal class BackgroundDetectionObserver @Inject constructor() : LifecycleObserver {
 
-    var isInBackground: Boolean = false
+    var isInBackground: Boolean = true
         private set
 
     private
--- a/vector/src/fdroid/AndroidManifest.xml
+++ b/vector/src/fdroid/AndroidManifest.xml
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     package="im.vector.app">
 
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <!--
     Required for long polling account synchronisation in background.
     If not present ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS intent action won't work
@@ -23,6 +25,11 @@
             android:name=".fdroid.receiver.AlarmSyncBroadcastReceiver"
             android:enabled="true"
             android:exported="false" />
+
+        <service
+            android:name=".fdroid.service.GuardService"
+            android:exported="false"
+            tools:ignore="Instantiatable" />
 
     </application>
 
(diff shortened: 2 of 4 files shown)


opusforlife2:
Thank you so much for this PR!

I don't really get the programming jargon in the description. Does this create a permanent notification that keeps Element running in the background, like Telegram-FOSS does?

lanmarc77:
Hi,
I am sorry, it is indeed pretty short.
I based this patch on my observation on my old Moto G3 with 1GB of Ram. It seems (after some time) VectorSyncService crashes with an ANR on my phone. It can then not create the next alarm and the sync cycle stops. I assume, even though the service might get restarted by android, it has somewhere lost it's sessionId which was originally transfered via the alarm. I did not investigate the cause but I am pretty certain it is because of the old hardware where services and activities are killed by android quite often.
I am new to kotlin and I wanted to create a (not meant to be forever) patch that does not change too much in the exisiting code.
So I created a GuardService that is a sticky Android service. It uses the same symbol as the VectorSyncService and stays visible in the taskbar even if the main Element app goes into background.
Inside the GuardService a thread is created that calls/starts the VectorSyncService class based on the same timeframe as the user has setup. So VectorSyncService is not only triggered by an alarm anymore but also called from the thread in GuardService.
As I did not change the alarm based triggers it might/will happen that VectorSyncService syncs more often (double as much) than the user has setup. Also I have implemented this only for the real time mode sync.
After a few weeks running with this patch I can confirm for my phone that it does work. Unfortunately no one from the original thread reported back if it also worked for them.

opusforlife2:
Hey man, sorry if I sound rude, but you sort of didn't answer my question and provided more of programming jargon instead. 😅

I'm just looking for a confirmation or denial of whether this will create a permanent notification that can be used to keep Element running in the background. Or does your code do something else?

lanmarc77:
:-)
human language is the worst: ambiguous. I actually read you did not understood MY programming jargon, not the programming jargon in general.

So yes it creates a permanent notification. And even if the phone needs to make RAM room for other applications by killing really everything that Element is made of, it will automatically recreate this notification (because it restarts the underlying service) and makes Element sync again. I do not know about the Telegram-FOSS app but it is like Signal which also always shows up.

opusforlife2:
Awesome! Thanks for explaining! I hope this is merged soon.

robsmith11:
Does someone need to assigned to review this patch?
Without working notifications, the F-Droid build is useless for most purposes, so seems like this should be a high priority issue..

lrq3000:
Just to confirm that as a user of Huawei phones, this PR fixes the issues #3159 and #2143 for me.

ouchadam:
@lanmarc77 You're right, the stickyness or redeliverly of the intents won't help us, our issue isn't the service not acting on the intents but the alarm intents themselves never being dispatched to our receiver (which would trigger the service).

I would probably argue the current `SyncService + AlarmManager` combo alone is no longer fit for purpose with newer versions of android and we need to be running as a foreground service for reliability.

My thinking is we merge this and then further integrate it as a dedicated `ForegroundTimer` implementation of the `SyncService`, this way we can avoid multiple services/redundant syncs and allow users to configure or select from different implementations. Huawei unfortunately tends to be the most hostile to background logic which will likely have the largest impact of on battery, which is ironically what their background polices attempt to reduce!

Peeking a little bit a further into the future we also have [unified push](https://github.com/vector-im/element-android/pull/3448) which may also reduce the need for multiple apps to have their own dedicated polling services

@lanmarc77 we'll also need you to confirm sign off https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off before we can merge

_tl;dr_ let's merge! This will likely temporarily increase battery usage for all fdroid users but that's arguably better than no or flaky background notifications. I will follow up making this a `SyncService` implementation with updated background settings options to change implementation.

@bmarty what do you think?

lanmarc77:
Signed-off-by: Marcel Langner <langner.marcel@myiq.de>

dadosch:
Would it be possible (technically) to set the permanent notification's priority/importance to MIN? https://developer.android.com/reference/android/app/NotificationManager#IMPORTANCE_MIN

That way the notification would not show up in the notification bar, only in the fold. In Android 7, otherwise the notification is shown permanently in the notification bar which is really annoying :)

PR #3847, opened 2021-08-16
Fix Dial pad lost entry when config changes #3845

### LINKED ISSUE #3845: Home: Dial pad lost entry when config changes

ganfra:
Fix #3845


onurays (approved):
LGTM.

Issue #3853, opened 2021-08-17
Call: handle busy reason


ganfra:
Show a dialog when user is busy.

PR #3854, opened 2021-08-17
Call ended: handle busy reason and invite timeout.

ganfra:
Fix https://github.com/vector-im/element-android/issues/3853

### DIFF
--- a/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
@@ -26,6 +26,7 @@ import android.os.Bundle
 import android.os.Parcelable
 import android.view.View
 import android.view.WindowManager
+import androidx.annotation.StringRes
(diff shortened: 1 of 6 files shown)


vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
```diff
@@ -249,14 +257,44 @@ class VectorCallActivity : VectorBaseActivity<ActivityCallBinding>(), CallContro
...
+                finish()
             }
         }
     }
 
+    private fun showEndCallDialog(@StringRes title: Int, @StringRes description: Int) {
+        MaterialAlertDialogBuilder(this)
+                .setTitle(title)
+                .setMessage(description)
+                .setNegativeButton(R.string.ok) { _, _ -> }
```
bmarty, line 291:
```suggestion
                .setNegativeButton(R.string.ok, null)
```

bmarty (approved):
LGTM, just one quick suggestion

Issue #3599, opened 2021-07-01
New call designs


ganfra:
Implement new designs: 

- [x]  In call controls
- [x] Timeline tiles
- [x] Draggable PiP 
- [ ] Jitsi notification
- [x] Jitsi close conf
- [x] Jitsi join conf

Figma link: https://www.figma.com/file/V6m2z0oAtUV1l8MdyIrAep/VoIP?node-id=4254%3A25767&viewport=241%2C48%2C0.07

ariskotsomitopoulos:
Closing this issue it seems fixed in **v1.3.8**

PR #3863, opened 2021-08-18
Feature/fga/new voip design

ganfra:
This PR handles most of the points from https://github.com/vector-im/element-android/issues/3599
It also offers some fixing.


vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -872,6 +868,23 @@ class RoomDetailFragment @Inject constructor(
                 onOptionsItemSelected(menuItem)
             }
         }
+        val joinConfItem = menu.findItem(R.id.join_conference)
+        joinConfItem.actionView.findViewById<MaterialButton>(R.id.join_conference_button).also { joinButton ->
+            joinButton.setOnClickListener { roomDetailViewModel.handle(RoomDetailAction.JoinJitsiCall) }
+            val colorFrom = ContextCompat.getColor(joinButton.context, R.color.palette_element_green)
+            // This is a special color, not defined in the palette
+            val colorTo = Color.parseColor("#0BAC7E")
```
bmarty, line 876: Add it as a resource anyway?

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -436,6 +443,26 @@ class RoomDetailFragment @Inject constructor(
...
+                roomDetailViewModel.handle(RoomDetailAction.RemoveWidget(jitsiWidgetId))
+            }
+        }
+    }
+
+    private fun leaveJitsiConference() {
+        JitsiBroadcastEmitter(vectorBaseActivity).emitConferenceEnded()
+    }
+
+    private fun onBroadcastEvent(event: BroadcastEvent) {
```
bmarty, line 462: to avoid confusion with the Matrix Events, maybe rename to jitsiEvent or sth like that...

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
bmarty, line 876: Add it as a resource anyway?

vector/src/main/res/values/strings.xml
```diff
@@ -3283,6 +3301,14 @@
         <item quantity="one">1 active call (%1$s) · 1 paused call</item>
         <item quantity="other">1 active call (%1$s) · %2$d paused calls</item>
     </plurals>
+    <plurals name="call_active_status">
+        <item quantity="one">Active call (%1$s)</item>
+        <item quantity="other">%1$d active calls</item>
```
bmarty, line 3306: I'm not sure it can work properly for every languages.
Please split into:
- 1 string when there is one active call, which takes %1$s as a parameter, as you it is now.
- 1 plural which will never be used in the code when there is only 1 active call, but will be used for other cases, without any %1$s pattern

vector/src/main/res/values/strings.xml
```diff
@@ -3283,6 +3301,14 @@
...
+    </plurals>
+
+    <string name="call_one_active">Active call (%1$s) ·</string>
+    <string name="call_multiple_active">%1$d active calls ·</string>
```
bmarty, line 3310: Please create a plurals for this one

bmarty (changes requested):
A few remark after a static review, I'm testing the code right now.

bmarty:
Limit the position of the text to the right?

image

bmarty:
When I start a WebRtc call (audio or video), the ringtone start playing immediately for a few ms, which is a bit strange.

PR #3881, opened 2021-08-23
Send an empty body for POST rooms/{roomId}/receipt/{receiptType}/{eve…

bmarty:
…ntId}

Fixes #3789

Issue #3546, opened 2021-06-22
Split Android SDK2 into several sub-modules


bmarty:
The integration of the RustSDK is not easy because the arch of the crypto code is a bit messy.

We need to split the Matrix SDK into several sub modules:

- 1 module for client server API, including the model
- 1 module for authentication
- 1 module for crypto (to ease the future Rust binding)
- 1 module for the sync, to be plugged by other modules: crypto, db, etc.
- 1 top level module with main SDK objects: session, room, etc., as entry point of the SDK

Open for discussion...

PR #3886, opened 2021-08-24
Feature/fga/extract matrix client

ganfra:
This is a first step on having several modules for the SDK. This is the network part, but it still needs to be branched. 
Maybe add more packages for the models. Also it's missing Content data classes at the moment.


bmarty:
Related to #3546 .

You have created new files, so we will loose the git history, but I suppose this is not a big deal here.

bmarty (approved):
LGTM

bmarty (commented):
- [ ] Can you report change from https://github.com/vector-im/element-android/pull/3881 please?
- [ ] Also as discussed IRL, it's probably better to avoid duplicated classes every where. It's probably best to finish this work before merging this PR to avoid confusion for other developer and fork maintainer.
- [ ] OK to keep 2 simple package `api` and `model` with no sub package per API. Maybe we can also add a package `model/content`(or just `content`?) for all the Event content models.
- [ ] Update the script for code quality, especially [here](https://github.com/vector-im/element-android/blob/main/tools/check/check_code_quality.sh#L67) and the next lines

Maybe it will be better if possible to start again this work in a new branch this PR by:
- Creating the new module
- Move the retrofit files, instead of copy past them; for a better git history and probably a better conflict handling for forks
- Move the data class
- Add the new files you've created in this PR

bmarty (changes requested):
Please see my other comment

bmarty:
I close the PR for now

PR #3891, opened 2021-08-24
Use `when` statement instead of `if`/ `else`

Florian14:
Minor code cleanup (which is more kotlin friendly) to simplify adding cases using `when` statement instead of old school Java `if` / `else`.


bmarty (approved):
Thanks!

PR #3893, opened 2021-08-24
Fix message edition is not rendered in e2e rooms after pagination

### LINKED ISSUE #3887: Message edition is not rendered in e2e rooms after pagination
In an encrypted room:
- Send a message
- Edit it
- Send about 26 more messages
- Restart the app and/or clear cache from the app setting (it will trigger an init sync)

Observe that the message edition is not taken into account.

ganfra:
Fix #3887

### DIFF
--- a/changelog.d/3887.bugfix
+++ b/changelog.d/3887.bugfix
@@ -1,0 +1,1 @@
+Message edition is not rendered in e2e rooms after pagination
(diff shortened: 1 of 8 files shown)


bmarty (approved):
LGTM.
There is still an issue if the app is not able to decrypt the edition for some reason, the original Event will still be displayed in the timeline without any indication that an edition occurred (but is not decryptable though).

bmarty:
Fix tested OK on my phone

Issue #3650, opened 2021-07-08
Read, Add and Remove Keywords for Keyword Notifications


langleyd:
The ability to read, add and remove keywords  that the user can decide to receive notifications for.

The entry point for this will be  under "Account Settings > Notifications > Mentions and Keywords"
![image](https://user-images.githubusercontent.com/6135282/124889295-daad9c80-dfce-11eb-9dd0-5397f7e8aadb.png)

Where existing keywords can be read and removed
![image](https://user-images.githubusercontent.com/6135282/124889378-f1ec8a00-dfce-11eb-9af3-68328d5b6e58.png)

And new ones added
![image](https://user-images.githubusercontent.com/6135282/124889404-f9ac2e80-dfce-11eb-8f30-ecc79b9a7b22.png)

Changing the notification settings for keywords and mentions are covered in https://github.com/vector-im/element-android/issues/3646.

PR #3894, opened 2021-08-24
Feature/dla/keyword notification settings

langleyd:
### Included in this PR
- Adds a `KeywordPreference` for rendering the edit keyword chips.
- Adds new logic for adding, removing and enabled/disabling keywords push rules in `VectorSettingsKeywordAndMentionsNotificationPreferenceFragment`.

### What it looks like
EditKeywords

resolves https://github.com/vector-im/element-android/issues/3650


library/ui-styles/src/main/res/color/keyword_foreground_selector.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item  android:alpha="0.25" android:color="?attr/colorOnSecondary" android:state_enabled="false" />
+    <item  android:color="?attr/colorOnSecondary"  />
```
bmarty, line 4: I know this is the same value, but since you have used `colorPrimary` for the background, why not using `colorOnPrimary` here (2 occurrences)?

library/ui-styles/src/main/res/values/theme_dark.xml
```diff
@@ -134,6 +134,8 @@
         <item name="vctr_social_login_button_gitlab_style">@style/Widget.Vector.Button.Outlined.SocialLogin.Gitlab.Dark</item>
 
         <item name="vctr_jump_to_unread_style">@style/Widget.Vector.JumpToUnread.Dark</item>
+        <!-- Keywords -->
+        <item name="vctr_keyword_style">@style/Widget.Vector.Keyword</item>
```
bmarty, line 138: Since the value is the same for light and dark theme, there is no need to declare `vctr_keyword_style`. You can directly use `Widget.Vector.Keyword`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/pushrules/RuleIds.kt
```diff
@@ -35,6 +35,8 @@ object RuleIds {
     // Default Content Rules
     const val RULE_ID_CONTAIN_USER_NAME = ".m.rule.contains_user_name"
 
+    const val RULE_ID_KEYWORDS = "_keywords"
```
bmarty, line 38: This is clearly not the same value format than the other RULE_ID. Is it really a rule Id?

vector/src/main/java/im/vector/app/core/preference/KeywordPreference.kt
```diff
@@ -0,0 +1,125 @@
...
+
+        keywords.forEach {
+            addChipToGroup(it, chipGroup)
+        }
+
+        chipEditText.isEnabled = keywordsEnabled
+        chipGroup.isEnabled = keywordsEnabled
+        chipGroup.children.forEach { it.isEnabled = keywordsEnabled }
+
+        chipEditText.setOnEditorActionListener { _, actionId, _ ->
```
bmarty, line 90: Would be nice to also have a button "Add" below the edit text. I'm afraid the IME_ACTION can be missed be some users.

vector/src/main/java/im/vector/app/core/preference/KeywordPreference.kt
```diff
@@ -0,0 +1,125 @@
...
+            listener?.didAddKeyword(keyword)
+            onPreferenceChangeListener?.onPreferenceChange(this, _keywords)
+            notifyChanged()
+            chipEditText.text = null
+            return@setOnEditorActionListener true
+        }
+    }
+
+    private fun addChipToGroup(keyword: String, chipGroup: ChipGroup) {
+        val chip = Chip(context, null, R.attr.vctr_keyword_style)
```
bmarty, line 108: Ah, I see, you need an attr here.

vector/src/main/java/im/vector/app/core/preference/KeywordPreference.kt
```diff
@@ -0,0 +1,125 @@
...
+        chip.text = keyword
+        chip.isClickable = true
+        chip.isCheckable = false
+        chip.isCloseIconVisible = true
```
bmarty, line 112: Can those 3 attributes be defined in the style?

vector/src/main/res/xml/vector_settings_notification_default.xml
```diff
@@ -6,22 +6,18 @@
 
         <im.vector.app.core.preference.VectorCheckboxPreference
             android:key="SETTINGS_PUSH_RULE_MESSAGES_IN_ONE_TO_ONE_PREFERENCE_KEY"
-            android:persistent="false"
```
bmarty, line 9: Are you sure you want to make the setting persistent? I think we get the value from the current push rule, so no need to store the values in a XML files.

bmarty (changes requested):
Thanks for the PR, a few remark after a static review. I will do some tests now

vector/src/main/res/xml/vector_settings_notification_default.xml
langleyd: Ah yea, I was looking at the wrong android documentation for this attribute. Will put it back to false thanks.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/pushrules/RuleIds.kt
langleyd: Sorry should have commented this. It's not really a rule Id but a placeholder just used client side. I should have commented this. Web has the same approach/naming.

vector/src/main/java/im/vector/app/core/preference/KeywordPreference.kt
langleyd: Wasn't included in the designs/don't think they felt it was necessary, but I see where you are coming from.
Maybe we can get feedback and see if it needs to be added. But will run it by design when they are available either way.

library/ui-styles/src/main/res/color/keyword_foreground_selector.xml
langleyd: Hrmm thought there was a reason for this but no seems you are right. will change, thanks.

gaelledel:
@langleyd Why do we have an "Add" button on the UI? Can we remove this and instead use the "Enter" button of the keyboard?

PR #3921, opened 2021-08-31
Test on intermediate API levels

bmarty:
21 is broken with same causes than https://github.com/square/workflow-kotlin/issues/298
30 emulator does not start...

Let's try on 28 and see what's happening


bmarty:
After the latest rebase:

image

BillCarsonFr (approved):
LGTM

PR #3950, opened 2021-09-01
FIx / bad format of restricted join rule

BillCarsonFr:
EA was using an old format of restriced join rule
````
"join_rule": restricted
"allow" : [
    {
      "space" : "!mods:example.org",
      "via": ["m.org","ex.org"]
    }
]
````
instead of

````
"join_rule": "restricted",
"allow": [
    {
        "type": "m.room_membership",
        "room_id": "!mods:example.org"
    }
]
````


bmarty (approved):
LGTM

PR #3957, opened 2021-09-02
Install ktlint plugin

### LINKED ISSUE #2820: Use a ktlint Gradle plugin to automate ktlint use instead of requiring manual downloading of the ktlint binary

The contributing guidelines could be simpler by not requiring manual downloads and runs of the ktlint binary. This can be cumbersome for new and existing contributors. Automating this removes the need for manual downloads and runs of binaries.

Specifically here: https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#ktlint

Use a Gradle plugin to run ktlint, e.g.

- https://github.com/JLLeitschuh/ktlint-gradle/ - just for ktlint, easy to configure
- https://github.com/diffplug/spotless - for ktlint and many more
- There may be other solutions: https://github.com/search?q=ktlint+gradle

This has the added benefit that it can be easily run on CI by simply running a Gradle task, which is likely what already happens for other build steps like Android lint. The two suggested Gradle plugins have the added benefit that it puts the ktlint version used under version control: various versions of ktlint have proven to behave differently, so making the correct and desired ktlint version in VCS has the benefit of knowing that every machine runs the same version.

Also, currently ktlint is in .gitignore, but it could be removed there if integrated from Gradle.

N/A

**Additional context**

N/A

bmarty:
Draft for now.

Iteration call to `./gradlew ktlintCheck` does not give the same result :/

[EDIT]

It's OK now, ktlint plugin is working as expected.

The following rule are now enabled (were disabled before):
- colon-spacing
- chain-wrapping
- import-ordering

Fixes #2820


erikhuizinga:
Try running `./gradlew ktlintCheck --continue`. If that leads to identical results, it might be an indication of ktlint running checks randomly. That might explain the non-determinism. I think that the ktlint Gradle tasks fail on the first found problem, while there might be more to find after the first.

bmarty:
There is still a problem, Android Studio does not always order imports correctly :/

erikhuizinga:
I think that the import ordering issues have been discussed extensively on the ktlint repo. Note that various settings can affect the import ordering, e.g. Android Studio settings for the Kotlin language and for the project. I can't check right now, but there might be a ktlint setting that tells ktlint that the project uses IntelliJ / AS. Ktlint then knows that the import ordering should use different rules.

erikhuizinga:
See also https://github.com/pinterest/ktlint#custom-ktlint-specific-editorconfig-properties.

bmarty:
Thanks @erikhuizinga . I've force push the branch with latest improvement.
First commit 03bdcee is the important change, next commits are done by enabling disabled rules one by one and run `./gradlew ktlintFormat`

bmarty:
So the ktlint plugin is now working as expected.

The following rule are now enabled (were disabled before):
- colon-spacing
- chain-wrapping
- import-ordering

BillCarsonFr (approved):
I'd prefer to keep && and || formating as it is currently

ouchadam (approved):
thanks for this :100: 

some of the line breaks for closing fun parameters/constructors etc are a little strange with

```kotlin
class Foo(val prop: Long) : 
  RealmObject() {
}
```

but they should be easy enough to fix later on with a find replace/regex


would have expected to see either...
```kotlin
class Foo(val prop: Long) : RealmObject() {
  ...
}
```

or 

```kotlin
class Foo (
    val prop: Long
) : RealmObject() {
  ...
}
```

bmarty:
@ouchadam yes, I agree with you,some constructors are not nicely formatted. Your latter proposal is good to me. As you said we can manually change that later, I wanted to keep manual commits as minimal as possible in this PR, since it's not possible to review it...

bmarty:
Fixes #2820

erikhuizinga:
Just under 600 files changed... 😅

Agreed: if ktlint accepts the current formatting, then this PR is acceptable. Apply the boy scout rule later when you work on lines of code with strange formatting.

For class definitions with constructor arguments I prefer oneliners, but if too long I prefer to put arguments in one column, possibly even with a trailing comma after the last argument. The same goes for long lists of super types: first one comes after `):` and the rest on their own lines.

Is that how the current PR configure ktlint, so what I just said is acceptable formatting?

bmarty:
another advantage is that dependabot is now able to upgrade the tool: #4138 :)

PR #3970, opened 2021-09-06
Fallback keys implementation

onurays:
### Problem
Olm uses a set of one-time keys when initializing a session between two devices: Alice uploads one-time keys to her homeserver, and Bob claims one of them to perform a Diffie-Hellman to generate a shared key. As implied by the name, a one-time key is only to be used once. However, if all of Alice's one-time keys are claimed, Bob will not be able to create a session with Alice.

### Proposal
A new response parameter, `device_unused_fallback_key_types` is added to /sync. If the client wants the server to have a fallback key for a given key algorithm, but that algorithm is not listed in `device_unused_fallback_key_types`, the client will upload a new key.

Here is the proposed logic to decide whether we need to generate a new fallback keys or not:
```
// The presence of device_unused_fallback_key_types indicates that the server supports fallback keys.
// If there's no unused signed_curve25519 fallback key we need a new one.
val shouldGenerateFallbackKey = if (syncResponse.deviceUnusedFallbackKeyTypes != null) {
    // Generate a fallback key only if the server does not already have an unused fallback key.
    !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)
} else {
    // Server does not support fallbackKey
    false
}
```

We need to use the existing `UploadKeysTask` by adding `fallbackKeys` parameter to the request.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
@@ -67,6 +67,7 @@ import org.matrix.android.sdk.internal.crypto.model.CryptoDeviceInfo
 import org.matrix.android.sdk.internal.crypto.model.ImportRoomKeysResult
 import org.matrix.android.sdk.internal.crypto.model.MXDeviceInfo
 import org.matrix.android.sdk.internal.crypto.model.MXEncryptEventContentResult
+import org.matrix.android.sdk.internal.crypto.model.MXKey.Companion.KEY_SIGNED_CURVE_25519_TYPE
 import org.matrix.android.sdk.internal.crypto.model.MXUsersDevicesMap
 import org.matrix.android.sdk.internal.crypto.model.event.EncryptedEventContent
 import org.matrix.android.sdk.internal.crypto.model.event.RoomKeyContent
@@ -329,7 +330,7 @@ internal class DefaultCryptoService @Inject constructor(
                 uploadDeviceKeys()
             }
 
-            oneTimeKeysUploader.maybeUploadOneTimeKeys()
+            oneTimeKeysUploader.maybeUploadOneTimeKeys(shouldGenerateFallbackKey = false)
             // this can throw if no backup
             tryOrNull {
                 keysBackupService.checkAndStartKeysBackup()
@@ -431,7 +432,17 @@ internal class DefaultCryptoService @Inject constructor(
                 if (isStarted()) {
                     // Make sure we process to-device messages before generating new one-time-keys #2782
                     deviceListManager.refreshOutdatedDeviceLists()
-                    oneTimeKeysUploader.maybeUploadOneTimeKeys()
+                    // The presence of device_unused_fallback_key_types indicates that the server supports fallback keys.
+                    // If there's no unused signed_curve25519 fallback key we need a new one.
+                    val shouldGenerateFallbackKey = if (syncResponse.deviceUnusedFallbackKeyTypes != null) {
+                        // Generate a fallback key only if the server does not already have an unused fallback key.
+                        !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)
+                    } else {
+                        // Server does not support fallbackKey
+                        false
+                    }
+
+                    oneTimeKeysUploader.maybeUploadOneTimeKeys(shouldGenerateFallbackKey)
                     incomingGossipingRequestManager.processReceivedGossipingRequests()
                 }
             }
@@ -928,7 +939,7 @@ internal class DefaultCryptoService @Inject constructor(
                 signatures = objectSigner.signObject(canonicalJson)
         )
 
-        val uploadDeviceKeysParams = UploadKeysTask.Params(rest, null)
+        val uploadDeviceKeysParams = UploadKeysTask.Params(rest, null, null)
         uploadKeysTask.execute(uploadDeviceKeysParams)
 
         cryptoStore.setDeviceKeysUploaded(true)
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
@@ -330,7 +330,7 @@ internal class DefaultCryptoService @Inject constructor(
                 uploadDeviceKeys()
             }
 
-            oneTimeKeysUploader.maybeUploadOneTimeKeys(shouldGenerateFallbackKey = false)
+            oneTimeKeysUploader.maybeUploadOneTimeKeys()
             // this can throw if no backup
             tryOrNull {
                 keysBackupService.checkAndStartKeysBackup()
@@ -434,15 +434,13 @@ internal class DefaultCryptoService @Inject constructor(
                     deviceListManager.refreshOutdatedDeviceLists()
                     // The presence of device_unused_fallback_key_types indicates that the server supports fallback keys.
                     // If there's no unused signed_curve25519 fallback key we need a new one.
-                    val shouldGenerateFallbackKey = if (syncResponse.deviceUnusedFallbackKeyTypes != null) {
-                        // Generate a fallback key only if the server does not already have an unused fallback key.
-                        !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)
-                    } else {
-                        // Server does not support fallbackKey
-                        false
+                    if (syncResponse.deviceUnusedFallbackKeyTypes != null
+                            // Generate a fallback key only if the server does not already have an unused fallback key.
+                            && !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)) {
+                        oneTimeKeysUploader.setNeedsNewFallback()
                     }
 
-                    oneTimeKeysUploader.maybeUploadOneTimeKeys(shouldGenerateFallbackKey)
+                    oneTimeKeysUploader.maybeUploadOneTimeKeys()
                     incomingGossipingRequestManager.processReceivedGossipingRequests()
                 }
             }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
@@ -434,9 +434,9 @@ internal class DefaultCryptoService @Inject constructor(
                     deviceListManager.refreshOutdatedDeviceLists()
                     // The presence of device_unused_fallback_key_types indicates that the server supports fallback keys.
                     // If there's no unused signed_curve25519 fallback key we need a new one.
-                    if (syncResponse.deviceUnusedFallbackKeyTypes != null
+                    if (syncResponse.deviceUnusedFallbackKeyTypes != null &&
                             // Generate a fallback key only if the server does not already have an unused fallback key.
-                            && !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)) {
+                            !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)) {
                         oneTimeKeysUploader.setNeedsNewFallback()
                     }
 
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
@@ -437,7 +437,7 @@ internal class DefaultCryptoService @Inject constructor(
                     if (syncResponse.deviceUnusedFallbackKeyTypes != null &&
                             // Generate a fallback key only if the server does not already have an unused fallback key.
                             !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)) {
-                        oneTimeKeysUploader.setNeedsNewFallback()
+                        oneTimeKeysUploader.needsNewFallback()
                     }
 
                     oneTimeKeysUploader.maybeUploadOneTimeKeys()
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
@@ -136,6 +136,24 @@ internal class MXOlmDevice @Inject constructor(
         return store.getOlmAccount().maxOneTimeKeys()
     }
 
+    fun getFallbackKey(): MutableMap<String, MutableMap<String, String>>? {
+        try {
+            return store.getOlmAccount().fallbackKey()
+        } catch (e: Exception) {
+            Timber.e("## getFallbackKey() : failed")
+        }
+        return null
+    }
+
+    fun generateFallbackKey() {
+        try {
+            store.getOlmAccount().generateFallbackKey()
+            store.saveOlmAccount()
+        } catch (e: Exception) {
+            Timber.e("## generateFallbackKey() : failed")
+        }
+    }
+
     /**
      * Release the instance
      */
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
@@ -150,13 +150,26 @@ internal class MXOlmDevice @Inject constructor(
         return null
     }
 
-    fun generateFallbackKey() {
+    /**
+     * Generates a new fallback key if there is not already
+     * an unpublished one.
+     * @return true if a new key was generated
+     */
+    fun generateFallbackKeyIfNeeded(): Boolean {
         try {
-            store.getOlmAccount().generateFallbackKey()
-            store.saveOlmAccount()
+            if (!hasUnpublishedFallbackKey()) {
+                store.getOlmAccount().generateFallbackKey()
+                store.saveOlmAccount()
+                return true
+            }
         } catch (e: Exception) {
             Timber.e("## generateFallbackKey() : failed")
         }
+        return false
+    }
+
+    internal fun hasUnpublishedFallbackKey(): Boolean {
+        return getFallbackKey()?.get(OlmAccount.JSON_KEY_ONE_TIME_KEY).orEmpty().isNotEmpty()
     }
 
     fun forgetFallbackKey() {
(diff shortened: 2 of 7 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -427,7 +429,10 @@ internal class DefaultCryptoService @Inject constructor(
                 if (isStarted()) {
                     // Make sure we process to-device messages before generating new one-time-keys #2782
                     deviceListManager.refreshOutdatedDeviceLists()
-                    oneTimeKeysUploader.maybeUploadOneTimeKeys()
+                    // The presence of device_unused_fallback_key_types indicates that the server supports fallback keys.
+                    // If there's no unused signed_curve25519 fallback key we need a new one.
+                    val shouldGenerateFallbackKey = !syncResponse.deviceUnusedFallbackKeyTypes?.contains(KEY_SIGNED_CURVE_25519_TYPE).orTrue()
```
bmarty, line 434: I think it is equivalent to the simpler version:

`val shouldGenerateFallbackKey = syncResponse.deviceUnusedFallbackKeyTypes?.contains(KEY_SIGNED_CURVE_25519_TYPE).orFalse()`

Remove `!` and use `orFalse`. WDYT?

> The presence of device_unused_fallback_key_types indicates that the server supports fallback keys.

Also to ensure that there is no mistake, and for clarity, you should probably cut this line in 2 tests, i.e.:

```kotlin
val shouldGenerateFallbackKey = if (syncResponse.deviceUnusedFallbackKeyTypes != null) {
    if(syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE) {
        // Server already have the fallback key
        false
    } else {
        // Generate the fallback key
        true
    }
} else {
    // Server does not support fallbackKey
    false
}
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -136,6 +136,24 @@ internal class MXOlmDevice @Inject constructor(
...
+        }
+        return null
+    }
+
+    fun generateFallbackKey() {
+        try {
+            store.getOlmAccount().generateFallbackKey()
+            store.saveOlmAccount()
+        } catch (e: Exception) {
+            Timber.e("## getFallbackKey() : failed")
```
bmarty, line 153: Update the comment (copy/paste)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
```diff
@@ -120,19 +124,22 @@ internal class OneTimeKeysUploader @Inject constructor(
      * @param keyLimit the limit
      * @return the number of uploaded keys
      */
-    private suspend fun uploadOTK(keyCount: Int, keyLimit: Int): Int {
+    private suspend fun uploadOTK(keyCount: Int, keyLimit: Int, shouldGenerateFallbackKey: Boolean): Int {
```
bmarty, line 127: This is a bit strange to send fall back key using the recursive method which send one time key. I do not see any mistake, but it's a bit confusing.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
```diff
@@ -159,9 +166,20 @@ internal class OneTimeKeysUploader @Inject constructor(
...
+            val canonicalJson = JsonCanonicalizer.getCanonicalJson(Map::class.java, k)
+            k["signatures"] = objectSigner.signObject(canonicalJson)
+
+            fallbackJson["signed_curve25519:$key_id"] = k
+        }
+
         // For now, we set the device id explicitly, as we may not be using the
         // same one as used in login.
-        val uploadParams = UploadKeysTask.Params(null, oneTimeJson)
+        val uploadParams = UploadKeysTask.Params(null, oneTimeJson, fallbackJson)
```
bmarty, line 182: fallbackJson should be null instead of empty, no?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/rest/KeysUploadBody.kt
```diff
@@ -40,5 +40,12 @@ internal data class KeysUploadBody(
...
          */
         @Json(name = "one_time_keys")
-        val oneTimeKeys: JsonDict? = null
+        val oneTimeKeys: JsonDict? = null,
+
+        /**
+         * If the user had previously uploaded a fallback key for a given algorithm, it is replaced.
+         * The server will only keep one fallback key per algorithm for each user.
+         */
+        @Json(name = "fallback_keys")
```
bmarty, line 49: Use prefix `org.matrix.msc2732.`. This may explain the bug you are seeing

bmarty (changes requested):
Some remarks

bmarty:
Also, of course you will have to remove the binary and upgrade the dependency on Olm library once this one will be release.
Please ensure to git squash to avoid binaries in the git history.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
onurays: Your first suggestion is not the same. Because we shouldn't generate a key if it is already in the list.

bmarty: Yes, you are right. So for clarity, please use my second suggestion, to avoid 🤯

onurays: Lint is not happy with the clear one of (redundant) if/else. I prefer a happy medium:

```
val shouldGenerateFallbackKey = if (syncResponse.deviceUnusedFallbackKeyTypes != null) {
    // Generate a fallback key only if the server does not already have an unused fallback key.
    !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)
} else {
    // Server does not support fallbackKey
    false
}
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
onurays: Done.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
onurays: It was an easy solution for me. Couldn't see alternatives without refactoring multiple functions, but will check again.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
onurays: Done.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/rest/KeysUploadBody.kt
onurays: Thanks for the catch :)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
bmarty: Yes, using if/else just for comments was a bit ugly :)

bmarty:
> Also, of course you will have to remove the binary and upgrade the dependency on Olm library once this one will be release. Please ensure to git squash to avoid binaries in the git history.

Done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -136,6 +136,24 @@ internal class MXOlmDevice @Inject constructor(
...
+    fun getFallbackKey(): MutableMap<String, MutableMap<String, String>>? {
+        try {
+            return store.getOlmAccount().fallbackKey()
+        } catch (e: Exception) {
+            Timber.e("## getFallbackKey() : failed")
+        }
+        return null
+    }
+
+    fun generateFallbackKey() {
```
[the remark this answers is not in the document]
BillCarsonFr: This is a more global problem, would be fixed by some changes on lib olm. Will be handle in a next PR

BillCarsonFr (approved):

bmarty (approved):
LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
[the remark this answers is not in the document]
poljar: Why? If you want to merge this PR before libolm is fixed you can easily store the fact that a fallback key has been uploaded or not yourself:

```python
In [1]: import olm

In [2]: x = olm.Account()

In [3]: x.fallback_key
Out[3]: {'curve25519': {}}

In [4]: x.generate_fallback_key()

In [5]: x.fallback_key
Out[5]: {'curve25519': {'AAAAAQ': '2XcMkoLlaJrKeNxV7yH4lUlJKc87u4QSrjbs/Js2KxQ'}}

In [6]: # Store the key ID 'AAAAAQ` and a boolean if it has been uploaded or not
```

On the other hand, this PR has been open now for more than a month, I raised the issue 3 weeks ago without any significant movement on it until now. Waiting a day or two for the libolm fix wouldn't add a significant amount of delay here.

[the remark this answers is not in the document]
BillCarsonFr: Ok, can wait it will be on next EA release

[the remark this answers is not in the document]
BillCarsonFr: PR is updated, the java binding are now directly calling the new `unpublished_*` olm calls for fallback keys.
So if a new fallback key is needed but there is still an unpublished one locally we retry to upload it.
Also added simple mechanism to forget fallback key if a new one as been published more than 5mn ago.

Tested against https://gitlab.matrix.org/uhoreg/matrix-client-test-bot

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -431,6 +432,14 @@ internal class DefaultCryptoService @Inject constructor(
                 if (isStarted()) {
                     // Make sure we process to-device messages before generating new one-time-keys #2782
                     deviceListManager.refreshOutdatedDeviceLists()
+                    // The presence of device_unused_fallback_key_types indicates that the server supports fallback keys.
+                    // If there's no unused signed_curve25519 fallback key we need a new one.
+                    if (syncResponse.deviceUnusedFallbackKeyTypes != null &&
+                            // Generate a fallback key only if the server does not already have an unused fallback key.
+                            !syncResponse.deviceUnusedFallbackKeyTypes.contains(KEY_SIGNED_CURVE_25519_TYPE)) {
+                        oneTimeKeysUploader.setNeedsNewFallback()
```
poljar, line 440: Can't we just generate the key directly here? It seems to be a bit convoluted to set the flag here and then somewhere else in the code the flag may or may not trigger the generation of the fallback key.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
```diff
@@ -121,24 +143,64 @@ internal class OneTimeKeysUploader @Inject constructor(
...
-        olmDevice.generateOneTimeKeys(keysThisLoop)
+        var keysThisLoop = 0
+        if (keyLimit > keyCount) {
+            // Creating keys can be an expensive operation so we limit the
+            // number we generate in one go to avoid blocking the application
+            // for too long.
+            keysThisLoop = min(keyLimit - keyCount, ONE_TIME_KEY_GENERATION_MAX_NUMBER)
+            olmDevice.generateOneTimeKeys(keysThisLoop)
+        }
+        if (needNewFallbackKey && !hasUnpublishedFallbackKey()) {
```
poljar, line 158: I would move the condition `hasUnpublishedFallbackKey()` into the `generateFallbackKey()` method. If the code changes around someone could call `generateFallbackKey()` in another place, or this check may be forgotten.

This would lead to overrotation of the fallback key quite easily, if we instead only generate a fallback key if there isn't one already there, the chances of misuse are lowered.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
```diff
@@ -78,37 +100,37 @@ internal class OneTimeKeysUploader @Inject constructor(
...
+        // these factors.
+        tryOrNull("Unable to upload OTK") {
+            val uploadedKeys = uploadOTK(oneTimeKeyCountFromSync, keyLimit)
+            Timber.v("## uploadKeys() : success, $uploadedKeys key(s) sent")
         }
         oneTimeKeyCheckInProgress = false
+
+        // Check if we need to forget a fallback key
+        val latestPublishedTime = getLastFallbackKeyPublishTime()
+        if (latestPublishedTime != 0L && System.currentTimeMillis() - latestPublishedTime > FIVE_MINUTES) {
```
poljar, line 123: I would probably give this a bit more leeway, 5 minutes is quite the short amount of time in a federated world where your server might be down for a while.

Perhaps 1h?

poljar (changes requested):
Seems like we can simplify this a bit. Removing the flag and moving the hasUnpublishedFallbackKey() check would simplify the code quite a bit.

Otherwise looks good.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
BillCarsonFr: yes updated.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
BillCarsonFr: done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
BillCarsonFr: Done, but would need to sync with js-sdk implementation

poljar (changes requested):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
```diff
@@ -28,18 +29,27 @@ import javax.inject.Inject
...
+        private val uploadKeysTask: UploadKeysTask,
+        context: Context
 ) {
     // tell if there is a OTK check in progress
     private var oneTimeKeyCheckInProgress = false
 
     // last OTK check timestamp
     private var lastOneTimeKeyCheck: Long = 0
     private var oneTimeKeyCount: Int? = null
+    private var needNewFallbackKey: Boolean = false
```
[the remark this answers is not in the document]
BillCarsonFr: Updated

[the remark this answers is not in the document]
poljar: It is, but the spelling issues up above aren't addressed yet. I think we can merge after those are addressed as well.

poljar (approved):
Looks good.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
```diff
@@ -41,6 +47,9 @@ internal class OneTimeKeysUploader @Inject constructor(
     private var lastOneTimeKeyCheck: Long = 0
     private var oneTimeKeyCount: Int? = null
 
+    // Simple storage to remember when was uploaded the last fallback key
+    private val storage = context.getSharedPreferences("OneTimeKeysUploader_${olmDevice.deviceEd25519Key.hashCode()}", Context.MODE_PRIVATE)
```
bmarty, line 51: A bit dirty... A proper store should be injected in the constructor...

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
```diff
@@ -121,24 +149,47 @@ internal class OneTimeKeysUploader @Inject constructor(
...
+                    uploadOTK(response.oneTimeKeyCountsForAlgorithm(MXKey.KEY_SIGNED_CURVE_25519_TYPE), keyLimit) +
+                    (if (hadUnpublishedFallbackKey) 1 else 0)
         } else {
             Timber.e("## uploadOTK() : response for uploading keys does not contain one_time_key_counts.signed_curve25519")
             throw Exception("response for uploading keys does not contain one_time_key_counts.signed_curve25519")
         }
     }
 
+    private fun saveLastFallbackKeyPublishTime(timeMillis: Long) {
+        storage.edit().putLong("last_fb_key_publish", timeMillis).apply()
```
bmarty, line 186: Nit: use `edit {}`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OneTimeKeysUploader.kt
```diff
@@ -121,24 +149,47 @@ internal class OneTimeKeysUploader @Inject constructor(
...
+    }
+
+    private fun getLastFallbackKeyPublishTime(): Long {
+        return storage.getLong("last_fb_key_publish", 0)
```
bmarty, line 190: Defining a const should be better (in a store...)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -136,6 +136,51 @@ internal class MXOlmDevice @Inject constructor(
...
+                return true
+            }
+        } catch (e: Exception) {
+            Timber.e("## generateFallbackKey() : failed")
+        }
+        return false
+    }
+
+    internal fun hasUnpublishedFallbackKey(): Boolean {
+        return getFallbackKey()?.get(OlmAccount.JSON_KEY_ONE_TIME_KEY).orEmpty().isNotEmpty()
```
bmarty, line 172: `orEmpty().isNotEmpty()` is a bit weird :/

bmarty (commented):
Some minor remarks. Having a store could be nice though, to have unit test.

bmarty:
Latest remarks will be handled later... or not.

PR #3971, opened 2021-09-06
Room details: check power level before displaying actions in the timeline

### LINKED ISSUE #3959: [Room details] Hide options when the user is not allowed to use them
### Your use case

Currently in the room history just after the room title there are 2 options (see attached screenshot)
- change the room topic
- invite new members

image

These 2 options are displayed without checking the actual power level of the current user.
If the user don't have enough power to edit the room topic or invite people the user experience is very bad

We request here to check power level before displaying these options 

_No response_

### Additional context

_No response_

Florian14:
This PR fixes #3959 

User power-level is now checked before showing action as set a topic or invite people to a room.
Previously, an error dialog was shown **after** the forbidden action, making a bad user experience.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MergedRoomCreationItem.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MergedRoomCreationItem.kt
@@ -143,7 +143,8 @@ abstract class MergedRoomCreationItem : BasedMergedItem<MergedRoomCreationItem.H
         val topic = roomSummary?.topic
         if (topic.isNullOrBlank()) {
             // do not show hint for DMs or group DMs
-            if (!isDirect) {
+            val canSetTopic = attributes.canChangeTopic && !isDirect
+            if (canSetTopic) {
                 val addTopicLink = holder.view.resources.getString(R.string.add_a_topic_link_text)
                 val styledText = SpannableString(holder.view.resources.getString(R.string.room_created_summary_no_topic_creation_text, addTopicLink))
                 holder.roomTopicText.setTextOrHide(styledText.tappableMatchingText(addTopicLink, object : ClickableSpan() {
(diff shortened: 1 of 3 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MergedRoomCreationItem.kt
```diff
@@ -227,6 +229,7 @@ abstract class MergedRoomCreationItem : BasedMergedItem<MergedRoomCreationItem.H
             val hasEncryptionEvent: Boolean,
             val isEncryptionAlgorithmSecure: Boolean,
             val roomSummary: RoomSummary?,
+            val canAddPeople: Boolean = false,
```
bmarty, line 232: Please rename to `canInvite` as per the other similar variable name for this feature.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MergedRoomCreationItem.kt
```diff
@@ -189,8 +190,9 @@ abstract class MergedRoomCreationItem : BasedMergedItem<MergedRoomCreationItem.H
             }
         }
 
-        holder.addPeopleButton.isVisible = !isDirect
-        if (!isDirect) {
+        val shouldAddPeople = attributes.canAddPeople && !isDirect
```
bmarty, line 193: Please also rename `shouldAddPeople` to `canInvite`

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MergedRoomCreationItem.kt
```diff
@@ -143,7 +143,8 @@ abstract class MergedRoomCreationItem : BasedMergedItem<MergedRoomCreationItem.H
         val topic = roomSummary?.topic
         if (topic.isNullOrBlank()) {
             // do not show hint for DMs or group DMs
-            if (!isDirect) {
+            val shouldSetTopic = attributes.canChangeTopic && !isDirect
```
bmarty, line 146: Please rename to `canSetTopic`

bmarty (commented):
LGTM, thanks, just some remarks about naming

bmarty (approved):
Thanks for the update!

PR #3985, opened 2021-09-08
Build Time Optimizations

ariskotsomitopoulos:
- Optimize build time for local builds
- Enhance CI/CD workflows (7GB Ram, 2CPU) by adding custom gradle arguments to properly run
- Add unit test UI comment representation plugin

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,12 @@ on:
   push:
     branches: [ main, develop ]
 
+# Enrich gradle.properties for CI/CD
+env:
+  CI_GRADLE_ARG_PROPERTIES: >
+    -Porg.gradle.jvmargs=-Xmx2g
+    -Porg.gradle.parallel=false
+
 jobs:
   debug:
     name: Build debug APKs (${{ matrix.target }})
@@ -25,7 +31,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Assemble ${{ matrix.target }} debug apk
-        run: ./gradlew assemble${{ matrix.target }}Debug --stacktrace
+        run: ./gradlew assemble${{ matrix.target }}Debug $CI_GRADLE_ARG_PROPERTIES --stacktrace
       - name: Upload ${{ matrix.target }} debug APKs
         uses: actions/upload-artifact@v2
         with:
@@ -39,6 +45,8 @@ jobs:
     if: github.ref == 'refs/heads/main'
     steps:
       - uses: actions/checkout@v2
+#      - name: Generate CI gradle.properties
+#        run: echo "$CI_GRADLE_PROPERTIES" > gradle.properties
       - uses: actions/cache@v2
         with:
           path: |
@@ -48,7 +56,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Assemble GPlay unsigned apk
-        run: ./gradlew clean assembleGplayRelease --stacktrace
+        run: ./gradlew clean assembleGplayRelease $CI_GRADLE_ARG_PROPERTIES  --stacktrace
       - name: Upload Gplay unsigned APKs
         uses: actions/upload-artifact@v2
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,8 +45,6 @@ jobs:
     if: github.ref == 'refs/heads/main'
     steps:
       - uses: actions/checkout@v2
-#      - name: Generate CI gradle.properties
-#        run: echo "$CI_GRADLE_PROPERTIES" > gradle.properties
       - uses: actions/cache@v2
         with:
           path: |
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -32,13 +32,11 @@ jobs:
 
   integration-tests:
     name: Integration Tests (Synapse)
-#    env:
-#      JAVA_HOME: $JAVA_HOME_11_X64
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        api-level: [21,23, 28, 30]
+        api-level: [28]
     steps:
       - uses: actions/checkout@v2
       - uses: gradle/wrapper-validation-action@v1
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -5,6 +5,17 @@ on:
...
     branches: [ main, develop ]
 
+#Custom gradle.properties file only for CI/CD
+env:
+  CI_GRADLE_PROPERTIES: |
+    #org.gradle.jvmargs=-Xmx2g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8 -XX:+UseParallelGC
+    org.gradle.jvmargs=-Xmx2g
+    org.gradle.vfs.watch=true
+    android.enableJetifier=true
+    android.useAndroidX=true
```
bmarty, line 15: It strange to add again the value which are already in 'gradle.properties'. See my other remark to try to avoid that.

.github/workflows/build.yml
```diff
@@ -16,6 +27,8 @@ jobs:
         target: [ Gplay, Fdroid ]
     steps:
       - uses: actions/checkout@v2
+      - name: Generate CI gradle.properties
+        run: echo "$CI_GRADLE_PROPERTIES" > gradle.properties
```
bmarty, line 31: Use `>>` instead of `>` to append line in this file?

bmarty (commented):
Thinking about it `gradle.properties` file should not contain any configuration that depend on the host.
But it can contain commented properties that every developer can add to their local configuration.
For the CI, OK to append configuration to this file using `>>`

bmarty:
It also possible to add gradle parameters in the command line, which is maybe a bit nicer that appending to `gradle.properties`. But appending is fine for me

.github/workflows/build.yml
ariskotsomitopoulos: Yes we can do that, but the point was to be decoupled. So the whole configuration for each flow is unique with no dependencies. However, I can change it with append in order to reduce the amount of code lines

ariskotsomitopoulos: in addition with decoupled, if a developer change something (accidentally or not in gradle.properties) it will not interfere with our github action, while its sensitive with 7 GB ram. what do you think?

bmarty: It's not ideal, if developer add a required config in gradle.properties (for instance something similar to `android.useAndroidX`, it will be stripped by the CI if the developer forget to update the workflow. So I prefer to use `>>` or maybe better command line argument.

Also gradle.properties file should be well documented.

Reviewers take particular attention to changed in those files anyway.

ariskotsomitopoulos: Yes, it might be better from that point of view. I rebased the branch and changed the implementation with  command line arguments

gradle.properties
```diff
@@ -6,20 +6,20 @@
...
-android.useAndroidX=true
-org.gradle.jvmargs=-Xmx2048m
-# When configured, Gradle will run in incubating parallel mode.
-# This option should only be used with decoupled projects. More details, visit
-# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
 
-# Enable file system watch (https://docs.gradle.org/6.7/release-notes.html)
+# Build Time Optimizations
+org.gradle.jvmargs=-Xmx3g -Xms512M -XX:MaxPermSize=2048m -XX:MaxMetaspaceSize=1g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8 -XX:+UseParallelGC
```
bmarty, line 11: I guess those parameters (except for `-Dfile.encoding=UTF-8`)  really depend on the host, but those default values look sane.

.github/workflows/integration.yml
```diff
@@ -22,17 +28,24 @@ jobs:
...
   integration-tests:
     name: Integration Tests (Synapse)
+#    env:
+#      JAVA_HOME: $JAVA_HOME_11_X64
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        api-level: [21, 28, 30]
+        api-level: [21,23, 28, 30]
```
bmarty, line 41: Why adding 23 here, integration tests are still failing on all platforms so far :/. I'm afraid of using too much resources from GitHub, which could slow down the entire CI for all our projects

.github/workflows/integration.yml
```diff
@@ -22,17 +28,24 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Compile Android tests
-        run: ./gradlew clean assembleAndroidTest --stacktrace -PallWarningsAsErrors=false
+        run: ./gradlew clean assembleAndroidTest $CI_GRADLE_ARG_PROPERTIES --stacktrace -PallWarningsAsErrors=false
 
   integration-tests:
     name: Integration Tests (Synapse)
+#    env:
+#      JAVA_HOME: $JAVA_HOME_11_X64
```
bmarty, line 36: Clean up?

.github/workflows/tests.yml
```diff
@@ -20,4 +26,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Run unit tests
-        run: ./gradlew clean test --stacktrace -PallWarningsAsErrors=false
+        run: ./gradlew clean test $CI_GRADLE_ARG_PROPERTIES -PallWarningsAsErrors=false --stacktrace
+      - name: Publish Unit Test Results
```
bmarty, line 30: 👍

bmarty (commented):
Comparing with another recent PR (for instance #4008), it appears that the gain is 1 or 2 minutes per check, so this is not a big win, but it's better than before.
Thanks for the adding of the publication of test result.

Do you have some build time stat on your computer, to compare to previous build time?

.github/workflows/integration.yml
ariskotsomitopoulos: Yes it was for testing purposes, or even better maybe we can just let none or one api-level until we fix the integration tests not to slow down the entire CI

bmarty: Yes, lets keep only 28 for now, so the it will look like `[28]`, thanks.

bahur142:
By me the compilation process is 12 minutes with these corrections and 14 min with default configuration.

bmarty:
Thanks @bahur142 for the feedback and @ariskotsomitopoulos for the update. Let's merge this!

bmarty (approved):

PR #4000, opened 2021-09-10
Fix parent relation handling

### LINKED ISSUE #3947: Space | A removed room from a space can't be re-added as it won't be shown in add-room
### Steps to reproduce

1. Create a new space with 3 default rooms
2. Go to manage rooms and remove the room
3. Try to go to add room to add it back

### What did you expect?

Expected the removed room to show up

### What happened?

The removed room is not shown in the add candidates

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

No

BillCarsonFr:
Fixes #3947 

Handle properly m.space.parent.
Consider a m.space.parent as valid if and only if:
- There is a corresponding m.space.child the overway around
- If the m.space.parent sender is also able to send m.space.child in the parent (kind of secret sub room)

Also when removing a room from a space, if there is a pending m.space.parent relation clean it.

### DIFF
--- a/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
+++ b/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
@@ -19,6 +19,8 @@ package org.matrix.android.sdk.common
 import android.content.Context
 import android.net.Uri
 import androidx.lifecycle.Observer
+import androidx.test.internal.runner.junit4.statement.UiThreadStatement
+import androidx.test.internal.runner.junit4.statement.UiThreadStatement.runOnUiThread
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.delay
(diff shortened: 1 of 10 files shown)


BillCarsonFr:
Fixes #4000

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
```diff
@@ -19,6 +19,8 @@ package org.matrix.android.sdk.common
 import android.content.Context
 import android.net.Uri
 import androidx.lifecycle.Observer
+import androidx.test.internal.runner.junit4.statement.UiThreadStatement
+import androidx.test.internal.runner.junit4.statement.UiThreadStatement.runOnUiThread
```
bmarty, line 23: strange to use `internal` class here, but it appear that you fix the pb with UI thread like that...

bmarty (approved):
I have to admit I read this a bit fast, also the algorithm is quite complicated, but you added a test to cover your change, so I rely on the test :)

bmarty (changes requested):
CI is not happy:
`/workdir/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryUpdater.kt: (267, 59): Variable 'isValidRelation' initializer is redundant`

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/CommonTestHelper.kt
BillCarsonFr: Took that from message from @onurays in android rooms. But we might need to dig more and see how to make test more simple it's painful to have to switch thread like that

bmarty:
One error to fix please:
```
❌ Error: '[^─]{161}' detected 1 time, in file './matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryUpdater.kt', at line 260.
Rule: Line length is limited to 160 chars. Please split long lines
```

PR #4002, opened 2021-09-10
Fix annoying issue detected by lint (CheckResult)

bmarty:
And consider that's errors from now, because generally it is so.


BillCarsonFr (approved):
Great, thx

PR #4008, opened 2021-09-13
Feature/fga/small timeline optimisation

ganfra:
This PR simplifies layout on ReactionButton as animation was broken anyway.
Also uses JsonDict directly instead of Moshi on some methods used a lot for building timeline event in epoxy. (we might want to do that for other things later when it makes sense).


bmarty (approved):
LGTM.
RIP animated reactions

PR #4017, opened 2021-09-15
Optimize dependencies handling

ariskotsomitopoulos:
- Fix minSdk version to be aligned to all modules
- Fix & upgrade markwon dependency
- Create a file dependencies.gradle that will contain all the shared & common dependencies between modules in order to avoid inconsistencies

### DIFF
--- a/attachment-viewer/build.gradle
+++ b/attachment-viewer/build.gradle
@@ -18,11 +18,12 @@ apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
 android {
-    compileSdk 30
+
+    compileSdk versions.compileSdk
 
     defaultConfig {
-        minSdk 21
-        targetSdk 30
+        minSdk versions.minSdk
+        targetSdk versions.targetSdk
     }
 
     buildTypes {
@@ -32,8 +33,8 @@ android {
         }
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_11
-        targetCompatibility JavaVersion.VERSION_11
+        sourceCompatibility versions.sourceCompat
+        targetCompatibility versions.targetCompat
     }
     kotlinOptions {
         jvmTarget = "11"
@@ -49,12 +50,12 @@ dependencies {
 
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
 
-    implementation 'io.reactivex.rxjava2:rxkotlin:2.4.0'
-    implementation 'io.reactivex.rxjava2:rxandroid:2.1.1'
+    implementation libs.rx.rxKotlin
+    implementation libs.rx.rxAndroid
 
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}"
     implementation 'androidx.core:core-ktx:1.6.0'
-    implementation 'androidx.appcompat:appcompat:1.3.1'
+    implementation libs.androidx.appCompat
     implementation "androidx.recyclerview:recyclerview:1.2.1"
 
     implementation 'com.google.android.material:material:1.4.0'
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -1,22 +1,23 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    // Ref: https://kotlinlang.org/releases.html
```
bmarty, line 4: Can you move this comment to the file with all the dependencies please?

attachment-viewer/build.gradle
```diff
@@ -49,12 +50,12 @@ dependencies {
...
+    implementation libs.rx.rxAndroid
 
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}"
     implementation 'androidx.core:core-ktx:1.6.0'
-    implementation 'androidx.appcompat:appcompat:1.3.1'
+    implementation libs.androidx.appCompat
     implementation "androidx.recyclerview:recyclerview:1.2.1"
 
     implementation 'com.google.android.material:material:1.4.0'
```
bmarty, line 61: this lib is also on several modules

dependencies.gradle
```diff
@@ -0,0 +1,49 @@
...
+        'dagger'           : '2.38.1',
+        'timber'           : '5.0.1',
+        'work'             : '2.5.0',
+        'lifecycle'        : '2.2.0',
+        'moshi'            : '1.12.0',
+        'arch'             : '2.1.0',
+        'markwon'          : '4.6.2',
+        'arrow'            : '0.8.2',
+        'rxKotlin'         : '2.4.0',
+        'rxAndroid'        : '2.1.1',
```
bmarty, line 20: Will dependabot be able to understand and update those values?

vector/build.gradle
```diff
@@ -315,16 +318,9 @@ dependencies {
...
     def big_image_viewer_version = '1.8.1'
     def glide_version = '4.12.0'
-    def moshi_version = '1.12.0'
-    def daggerVersion = '2.38.1'
     def autofill_version = "1.1.0"
-    def work_version = '2.5.0'
-    def arch_version = '2.1.0'
-    def lifecycle_version = '2.2.0'
     def rxbinding_version = '3.1.0'
     def jjwt_version = '0.11.2'
```
bmarty, line 325: Why not moving all the `def` values?

bmarty (commented):
Thanks, some remarks.
Also owe will have to be careful when exporting the SDK, but I will take care of it.

attachment-viewer/build.gradle
ariskotsomitopoulos: Yes, there are a lot of dependencies that should be moved there, I just moved the basic ones for feedback. I will continue with the others.

vector/build.gradle
ariskotsomitopoulos: Sure, it will be done in the next commit

dependencies.gradle
ariskotsomitopoulos: Yes, it should work while it is the same as setting an **ext {}**  in root _build.gradle_

dependencies.gradle
```diff
@@ -5,45 +5,123 @@ ext.versions = [
...
-        'timber'           : '5.0.1',
-        'work'             : '2.5.0',
-        'lifecycle'        : '2.2.0',
-        'moshi'            : '1.12.0',
-        'arch'             : '2.1.0',
-        'markwon'          : '4.6.2',
-        'arrow'            : '0.8.2',
-        'rxKotlin'         : '2.4.0',
-        'rxAndroid'        : '2.1.1',
+// Ref: https://kotlinlang.org/releases.html
```
bmarty, line 10: This ref is use to ensure kotlin and kotlinCoroutines version are aligned, so can you move to the line below please?

bmarty: We will do that later, let see what happen for dependabot current PRs when this one will be merged

bmarty (approved):
Thanks for the cleanup!

PR #4018, opened 2021-09-15
feature/aris/issue_dinsic_618

ariskotsomitopoulos:
_Apps targeting Android 12 and higher are required to specify an explicit value for android:exported when the corresponding
component has an intent filter defined._

### android:exported=“true”

**Activities**

```
.features.MainActivity 
[intent-filter] —> Required to launch the app

.features.login.LoginActivity
[intent-filter] -> Redirection URL after SSO login in external browser

.features.login2.LoginActivity2
[intent-filter] -> Redirection URL after SSO login in external browser

.features.link.LinkHandlerActivity
[intent-filter] -> Deep Link Converter/Transformer

.features.permalink.PermalinkHandlerActivity
[intent-filter] -> Deep Links Handling

.features.share.IncomingShareActivity
[intent-filter] -> Share functionality

```
**Services**
```
.features.call.telecom.VectorConnectionService
[intent-filter] -> Call connection service.  [maybe false?]
```

**Receivers**
```
androidx.media.session.MediaButtonReceiver
[intent-filter] -> Translate hardware media playback buttons  [maybe false?]
```

### android:exported=“false”
```
All the other activities, services, receivers, providers
```

### DIFF
--- a/vector/src/main/AndroidManifest.xml
+++ b/vector/src/main/AndroidManifest.xml
@@ -92,8 +92,10 @@
             android:theme="@style/Theme.Vector.Launcher" />
 
         <!-- Activity alias for the launcher Activity (must be declared after the Activity it targets) -->
+        <!-- exported="true" is required to launch application -->
         <activity-alias
             android:name=".features.Alias"
+            android:exported="true"
             android:targetActivity=".features.MainActivity">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
@@ -107,9 +109,12 @@
         </activity-alias>
 
         <activity android:name=".features.home.HomeActivity" />
+
+        <!-- exported="true" is required to handle android.intent.action.VIEW for URL redirection-->
         <activity
             android:name=".features.login.LoginActivity"
             android:enabled="@bool/useLoginV1"
+            android:exported="true"
             android:launchMode="singleTask"
             android:windowSoftInputMode="adjustResize">
             <!-- Add intent filter to handle redirection URL after SSO login in external browser -->
@@ -124,9 +129,12 @@
                     android:scheme="element" />
             </intent-filter>
         </activity>
+
+        <!-- exported="true" is required to handle android.intent.action.VIEW for URL redirection-->
         <activity
             android:name=".features.login2.LoginActivity2"
             android:enabled="@bool/useLoginV2"
+            android:exported="true"
             android:launchMode="singleTask"
             android:windowSoftInputMode="adjustResize">
             <!-- Add intent filter to handle redirection URL after SSO login in external browser -->
@@ -147,7 +155,6 @@
             android:name=".features.media.VectorAttachmentViewerActivity"
             android:theme="@style/Theme.Vector.Black.Transparent"
             tools:ignore="Instantiatable" />
-
         <activity android:name=".features.media.BigImageViewerActivity" />
         <activity
             android:name=".features.rageshake.BugReportActivity"
@@ -165,7 +172,6 @@
         <activity
             android:name=".features.crypto.keysbackup.settings.KeysBackupManageActivity"
             android:label="@string/encryption_message_recovery" />
-
         <activity
             android:name=".features.reactions.EmojiReactionPickerActivity"
             android:label="@string/title_activity_emoji_reaction_picker" />
@@ -186,6 +192,7 @@
         <activity android:name=".features.webview.VectorWebViewActivity" />
 
         <!-- Activity to intercept links coming from a web instance -->
+        <!-- exported="true" is required to handle android.intent.action.VIEW for URL redirection-->
         <activity
             android:name=".features.link.LinkHandlerActivity"
             android:enabled="true"
@@ -206,6 +213,7 @@
         </activity>
 
         <!-- Activity alias for matrix.to or element permalinks -->
+        <!-- exported="true" is required to handle android.intent.action.VIEW for URL redirection-->
         <activity-alias
             android:name=".features.permalink.PermalinkHandlerActivity"
             android:enabled="true"
@@ -231,8 +239,10 @@
             </intent-filter>
         </activity-alias>
 
+        <!-- exported="true" is required for the share functionality-->
         <activity
             android:name=".features.share.IncomingShareActivity"
+            android:exported="true"
             android:parentActivityName=".features.home.HomeActivity">
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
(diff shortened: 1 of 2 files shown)


vector/src/main/AndroidManifest.xml
```diff
@@ -364,7 +439,9 @@
         A media button receiver receives and helps translate hardware media playback buttons,
         such as those found on wired and wireless headsets, into the appropriate callbacks in your app.
            -->
-        <receiver android:name="androidx.media.session.MediaButtonReceiver">
+        <receiver
+            android:name="androidx.media.session.MediaButtonReceiver"
+            android:exported="true" >
```
ariskotsomitopoulos, line 444: Maybe this will also work with `exported="false"` ?

vector/src/main/AndroidManifest.xml
```diff
@@ -338,7 +412,8 @@
 
         <service
             android:name=".features.call.telecom.VectorConnectionService"
-            android:permission="android.permission.BIND_TELECOM_CONNECTION_SERVICE">
+            android:permission="android.permission.BIND_TELECOM_CONNECTION_SERVICE"
+            android:exported="true">
```
ariskotsomitopoulos, line 416: Maybe this will also work with exported="false" ?

vector/src/main/AndroidManifest.xml
bmarty: Can you test it please?

vector/src/main/AndroidManifest.xml
bmarty: Can you test it please?

vector/src/main/AndroidManifest.xml
ariskotsomitopoulos: I tested with wireless earbuds if click on call events work with `exported="false"` and they seem to work (**onMediaButtonEvent** function is triggered). So I will update it with **false**. If there is an other usage other than that, that you can think of let me know

vector/src/main/AndroidManifest.xml
ariskotsomitopoulos: I think **VectorConnectionService** is not even working so there will be no problem to set `exported=false`.  I also test incoming and outgoing calls and they work properly.

vector/src/main/AndroidManifest.xml
bmarty: thanks!

vector/src/main/AndroidManifest.xml
bmarty: ok, thanks!

bmarty (commented):
Thanks for the update.
Can you add in the Manifest itself one line of comment to explain why exported="true" is mandatory? You can use the related comment form the PR itself.
Thanks

ariskotsomitopoulos:
> Thanks for the update.
> Can you add in the Manifest itself one line of comment to explain why exported="true" is mandatory? You can use the related comment form the PR itself.
> Thanks

done

bmarty:
> > Thanks for the update.
> > Can you add in the Manifest itself one line of comment to explain why exported="true" is mandatory? You can use the related comment form the PR itself.
> > Thanks
> 
> done

Sorry there have been a misunderstanding, your last commit was not what I was expecting. I wanted you to explicitly comment all Android Manifest element with `android:exported` set to "true" to justify why it is mandatory for the app to work properly.

We will take care of Android 12 support when we will be targeting those devices. In the mean time, I guess there is no pb to install on Android 12, since compatibility mode will occur on the device.

ariskotsomitopoulos:
> > > Thanks for the update.
> > > Can you add in the Manifest itself one line of comment to explain why exported="true" is mandatory? You can use the related comment form the PR itself.
> > > Thanks
> > 
> > 
> > done
> 
> Sorry there have been a misunderstanding, your last commit was not what I was expecting. I wanted you to explicitly comment all Android Manifest element with `android:exported` set to "true" to justify why it is mandatory for the app to work properly.
> 
> We will take care of Android 12 support when we will be targeting those devices. In the mean time, I guess there is no pb to install on Android 12, since compatibility mode will occur on the device.

Thanks, now its more clear, I will update accordingly !

vector/src/main/AndroidManifest.xml
```diff
@@ -348,11 +352,11 @@
         <!-- Add tools:ignore="Instantiatable" for the error reported only by Buildkite and for lintGplayRelease check :/ -->
         <service
             android:name=".core.services.VectorSyncService"
-            android:exported="false"
```
bmarty, line 351: it would be preferable to keep the existing `android:exported="false"`

vector/src/main/AndroidManifest.xml
```diff
@@ -361,25 +365,22 @@
 
         <!-- Receivers -->
 
-        <receiver
-            android:name=".features.call.service.CallHeadsUpActionReceiver"
-            android:exported="false" />
+        <receiver android:name=".features.call.service.CallHeadsUpActionReceiver" />
```
bmarty, line 368: it would be preferable to keep the existing `android:exported="false"`

vector/src/main/AndroidManifest.xml
```diff
@@ -361,25 +365,22 @@
...
 
-        <receiver
-            android:name=".features.settings.troubleshoot.TestNotificationReceiver"
-            android:exported="false" />
+        <receiver android:name=".features.settings.troubleshoot.TestNotificationReceiver" />
```
bmarty, line 370: it would be preferable to keep the existing `android:exported="false"`

vector/src/main/AndroidManifest.xml
```diff
@@ -361,25 +365,22 @@
...
 
         <!-- Exported false, should only be accessible from this app!! -->
         <receiver
             android:name=".features.notifications.NotificationBroadcastReceiver"
-            android:enabled="true"
-            android:exported="false" />
+            android:enabled="true" />
```
bmarty, line 375: it would be preferable to keep the existing `android:exported="false"`

vector/src/main/AndroidManifest.xml
```diff
@@ -402,7 +403,6 @@
         <provider
             android:name="androidx.core.content.FileProvider"
             android:authorities="${applicationId}.fileProvider"
-            android:exported="false"
```
bmarty, line 405: it would be preferable to keep the existing `android:exported="false"`

bmarty (commented):
Thanks for the update. See my remarks. Also can you squash all the commits please?

vector/src/main/AndroidManifest.xml
```diff
@@ -186,9 +192,9 @@
         <activity android:name=".features.webview.VectorWebViewActivity" />
 
         <!-- Activity to intercept links coming from a web instance -->
+        <!-- exported="true" is required to handle android.intent.action.VIEW for URL redirection-->
         <activity
             android:name=".features.link.LinkHandlerActivity"
-            android:enabled="true"
```
bmarty, line 191: Can we keep this line please? I know `true` is the default value, but it's here to help fork maintainers that it can be disabled here.

bmarty (commented):
Thanks for the update. One final comment.
Also can you add a file for the changelog please (filename 4018.misc would be fine)? As we are doing this for the forks, it will help to track this work.

bmarty (approved):
Thanks for the update!

PR #4027, opened 2021-09-16
Add client base url support for permalinks

Florian14:
This PR adds a field `clientPermalinkBaseUrl` into the `MatrixConfiguration` object.
If not null, this base url will be used to generate custom permalinks instead of usual `matrix.to` links.
eg. "https://www.example.org/#" for "https://www.example.org/#/room/#element-android:matrix.org"


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+ * Mapping of an input URI to a matrix.to compliant URI.
+ */
+object MatrixToMapper {
+
+    /**
+     * Try to convert a URL from an element web instance or from a client permalink to a matrix.to url.
+     * Examples:
+     * - https://riot.im/develop/#/room/#element-android:matrix.org  ->  https://matrix.to/#/#element-android:matrix.org
+     * - https://app.element.io/#/room/#element-android:matrix.org   ->  https://matrix.to/#/#element-android:matrix.org
+     * - https://www.example.org/#/room/#element-android:matrix.org  ->  https://matrix.to/#/#element-android:matrix.org
```
bmarty, line 31: What is the point to handle any host?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+            SUPPORTED_PATHS.any { it in uriString }                                 -> {
+                val path = SUPPORTED_PATHS.first { it in uriString }
+                Uri.parse(PermalinkService.MATRIX_TO_URL_BASE + uriString.substringAfter(path))
+            }
+            // Element custom scheme
+            uriString.startsWith(PermalinkService.MATRIX_TO_CUSTOM_SCHEME_URL_BASE) -> {
+                val permalinkId = when {
+                    uriString.startsWith(USER_LINK_PREFIX) -> uriString.substring(USER_LINK_PREFIX.length)
+                    uriString.startsWith(ROOM_LINK_PREFIX) -> uriString.substring(ROOM_LINK_PREFIX.length)
+                    else                                   -> null
```
bmarty, line 50: there is also `GROUP_LINK_PREFIX` to handle ("group"), even if they will probably disappear...

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkParser.kt
```diff
@@ -61,20 +64,14 @@ object PermalinkParser {
...
                 .take(2)
 
-        // the element-based domain permalinks (e.g. https://app.element.io/#/user/@chagai95:matrix.org) don't have the
-        // mxid in the first param (like matrix.to does - https://matrix.to/#/@chagai95:matrix.org) but rather in the second after /user/ so /user/mxid
-        var identifier = params.getOrNull(0)
-        if (identifier.equals("user")) {
-            identifier = params.getOrNull(1)
-        }
-
+        val identifier = params.getOrNull(0)
```
bmarty, line 67: @chagai95 heads up that this will remove your patch, but I think with the new code it will work. Just wanted you to have a look, since you have changed the value of "PermalinkService.MATRIX_TO_URL_BASE", I think this will actually break something on your side.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkService.kt
```diff
@@ -19,53 +19,59 @@ package org.matrix.android.sdk.api.session.permalinks
...
 
 /**
- * Useful methods to create Matrix permalink (matrix.to links).
+ * Useful methods to create permalink (like matrix.to links or client permalinks).
  */
 interface PermalinkService {
 
     companion object {
         const val MATRIX_TO_URL_BASE = "https://matrix.to/#/"
+        const val MATRIX_TO_CUSTOM_SCHEME_URL_BASE = "element://"
```
bmarty, line 28: Mentions of "element" should not occur in the Matrix SDK. Is there an other solution?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkService.kt
```diff
@@ -19,53 +19,59 @@ package org.matrix.android.sdk.api.session.permalinks
 import org.matrix.android.sdk.api.session.events.model.Event
 
 /**
- * Useful methods to create Matrix permalink (matrix.to links).
+ * Useful methods to create permalink (like matrix.to links or client permalinks).
```
bmarty, line 22: Maybe explain that to create client permalink, the setting is in MatrixConfiguration.

vector/src/main/AndroidManifest.xml
```diff
@@ -180,7 +180,9 @@
         <activity android:name=".features.createdirect.CreateDirectRoomActivity" />
         <activity android:name=".features.invite.InviteUsersToRoomActivity" />
         <activity android:name=".features.webview.VectorWebViewActivity" />
-        <activity android:name=".features.link.LinkHandlerActivity">
+        <activity
+            android:name=".features.link.LinkHandlerActivity"
```
bmarty, line 184: Can you add explicit `enable="true"` attribute? It will be nice for the forks who want to disable it.
Also maybe add a comment to explain which links type are handled here.

vector/src/main/AndroidManifest.xml
```diff
@@ -196,6 +198,30 @@
             </intent-filter>
         </activity>
 
+        <activity-alias
+            android:name=".features.permalink.PermalinkHandlerActivity"
+            android:exported="true"
+            android:launchMode="singleTask"
+            android:targetActivity=".features.link.LinkHandlerActivity">
```
bmarty, line 205: Can you add explicit `enable="true"` attribute? It will be nice for the forks who want to disable it.
Also maybe add a comment to explain which links type are handled here.

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
```diff
@@ -559,10 +546,6 @@ class HomeActivity :
                         putExtra(MvRx.KEY_ARG, args)
                     }
         }
-
-        private const val MATRIX_TO_CUSTOM_SCHEME_URL_BASE = "element://"
```
bmarty, line 563: "element://" was there at first place (in app side)

bmarty (changes requested):
Thanks for the PR, nice work!
I've made some remarks.
Also @ganfra can you make a review too please?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package org.matrix.android.sdk.api.session.permalinks
+
+import android.net.Uri
+
+/**
+ * Mapping of an input URI to a matrix.to compliant URI.
+ */
+object MatrixToMapper {
```
bmarty, line 24: Maybe MatrixToConverter will be a better name

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Try to convert a URL from an element web instance or from a client permalink to a matrix.to url.
+     * Examples:
+     * - https://riot.im/develop/#/room/#element-android:matrix.org  ->  https://matrix.to/#/#element-android:matrix.org
+     * - https://app.element.io/#/room/#element-android:matrix.org   ->  https://matrix.to/#/#element-android:matrix.org
+     * - https://www.example.org/#/room/#element-android:matrix.org  ->  https://matrix.to/#/#element-android:matrix.org
+     * - element://room/#element-android:matrix.org                  ->  https://matrix.to/#/#element-android:matrix.org
+     */
+    fun map(uri: Uri): Uri? {
```
bmarty, line 34: and `convert` instead of `map`, WDYT?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
Florian14: SDK does not know which hosts are supported, this check should be done in the application side.
Goal of this mapper is to convert any host which is correctly formated. Would you prefer I add a config in the SDK to setup supported hosts ? I'm not sure it is relevant to add this kind of logic in this mapper.
I just added the host verification inside the PermalinkHandler which is the only place where the parser is called.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
Florian14: groups were not handled before, this part of code came from the HomeActivity, I just rollback these changes which are "specific" to Element client

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkService.kt
Florian14: As I said, I rolled back this custom scheme management which was in the HomeActivity. That implies this kind of link will only work from those coming from a deep link, not for the other accessed from a room which are not passed to the HomeActivity (I kept the previous behaviour)

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
Florian14: Indeed, I observed you already moved this scheme from the service to the activity in a previous commit...

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
Florian14: It sounds like better after the class renaming

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkService.kt
```diff
@@ -32,40 +33,45 @@ interface PermalinkService {
      * Ex: "https://matrix.to/#/!nbzmcXAqpxBXjAdgoX:matrix.org/$1531497316352799BevdV:matrix.org"
      *
      * @param event the event
+     * @param forceMatrixTo whether we should force using matrix.to base URL
      *
      * @return the permalink, or null in case of error
      */
-    fun createPermalink(event: Event): String?
+    fun createPermalink(event: Event, forceMatrixTo: Boolean = false): String?
```
ganfra, line 40: We try to avoid default param on SDK, but I guess it's ok?

ganfra (approved):
Just one remark for me. Otherwise LGTM, but not tested

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixConfiguration.kt
```diff
@@ -32,8 +32,12 @@ data class MatrixConfiguration(
                 "https://scalar-staging.riot.im/scalar/api"
         ),
         /**
-         * Optional proxy to connect to the matrix servers
-         * You can create one using for instance Proxy(proxyType, InetSocketAddress.createUnresolved(hostname, port)
+         * Optional base url to create client permalinks instead of Matrix ones (matrix.to links).
```
giomfo, line 35: please add an example like "https://www.example.com/#/"

vector/src/main/java/im/vector/app/features/link/LinkHandlerActivity.kt
```diff
@@ -45,30 +44,38 @@ class LinkHandlerActivity : VectorBaseActivity<ActivityProgressBinding>() {
...
+        when {
+            uri == null                                                                    -> {
+                // Should not happen
+                Timber.w("Uri is null")
+                finish()
+            }
+            uri.getQueryParameter(LoginConfig.CONFIG_HS_PARAMETER) != null                 -> handleConfigUrl(uri)
+            uri.toString().startsWith(PermalinkService.MATRIX_TO_URL_BASE)                 -> handleSupportedHostUrl()
+            uri.toString().startsWith(PermalinkHandler.MATRIX_TO_CUSTOM_SCHEME_URL_BASE)   -> handleSupportedHostUrl()
+            resources.getStringArray(R.array.permalink_supported_hosts).contains(uri.host) -> handleSupportedHostUrl()
```
giomfo, line 73: Don't you think we should consider the case of "clientPermalinkBaseUrl" here too?

vector/src/main/java/im/vector/app/features/permalink/PermalinkHandler.kt
```diff
@@ -122,6 +123,13 @@ class PermalinkHandler @Inject constructor(private val activeSessionHolder: Acti
         }
     }
 
+    private fun isPermalinkSupported(context: Context, url: String): Boolean {
+        return url.startsWith(PermalinkService.MATRIX_TO_URL_BASE)
+                || context.resources.getStringArray(R.array.permalink_supported_hosts).any {
```
giomfo, line 128: idem "Don't you think we should consider the case of "clientPermalinkBaseUrl" here too?"

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkService.kt
ganfra: Yes but still, we should avoid having reference to element in the SDK

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/PermalinkService.kt
Florian14: Default behaviour is "do not force matrix.to", Benoit asked me to add this param but I'm not sure we really need it, or maybe a single method `createMatrixToPermalink(event: Event)` would be good enough?

vector/src/main/java/im/vector/app/features/link/LinkHandlerActivity.kt
Florian14: Yes it makes sense, thanks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
bmarty: Ok, I see, thanks!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToMapper.kt
bmarty: Ok, let's not support group here then.

vector/src/main/java/im/vector/app/features/permalink/PermalinkHandler.kt
giomfo: fixed by https://github.com/vector-im/element-android/pull/4027/files#diff-47e155bb67cd55047db56ba0b5d2853eb4dd79f5e1bb07b566b6b7ba31efd7f0R38

bmarty (approved):
Thanks for the update!

Issue #4028, opened 2021-09-16
[Voip] : Other call not always refreshed when ended


ganfra:
### Steps to reproduce

1. Launch 2 different calls in 2 rooms.
2. Let the opponent of the held call hangups

### What did you expect?
The held call should disappear completely from UI (Active call views, InCall screen)
### What happened?
The held call is still shown in Active call views and InCall screen. It disappears when you close the screens.

### Have you submitted a rageshake?

No

kittykat:
I'm going to close this issue for now as we were unable to reproduce it in a community testing session. Please re-open if it's still valid.

Issue #4026, opened 2021-09-16
Call screen stuck with some hanging up scenarios


ganfra:
### Steps to reproduce

1. Have Element Web open with Account A and Account B. Have Element Android open with Account A.
2. Start a call from Account B to Account A.
3. Open the call screen on Element Android by clicking on the notification
4. Pick up on Element Web.

### What did you expect?
The call screen should be dismissed
### What happened?
It's not dismissed.

### Have you submitted a rageshake?

No

bmarty:
Done in #4029

Issue #4019, opened 2021-09-15
Sticky end call notification


ganfra:
### Steps to reproduce

1. Start a 1 to 1 call through WebRtc
2. Hangs up.

### What did you expect?
No remaining notification 
### What happened?
"Ended call" notification got stuck

### Have you submitted a rageshake?

No

ouchadam:
fixed in 1.3.0

PR #4029, opened 2021-09-16
Feature/fga/fix voip issues

ganfra:
This PR fixes multiple issues on Voip : 
#4028 
#4026 
#4019

### DIFF
--- a/vector/src/main/java/im/vector/app/features/notifications/NotificationUtils.kt
+++ b/vector/src/main/java/im/vector/app/features/notifications/NotificationUtils.kt
@@ -468,7 +468,6 @@ class NotificationUtils @Inject constructor(private val context: Context,
                         setSmallIcon(R.drawable.ic_call_answer)
                     }
                 }
-                // This is a trick to make the previous notification with same id disappear as cancel notification is not working with Foreground Service.
                 .setTimeoutAfter(1)
                 .setColor(ThemeUtils.getColor(context, android.R.attr.colorPrimary))
                 .setCategory(NotificationCompat.CATEGORY_CALL)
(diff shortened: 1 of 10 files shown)


vector/src/main/java/im/vector/app/features/notifications/NotificationUtils.kt
```diff
@@ -468,7 +468,6 @@ class NotificationUtils @Inject constructor(private val context: Context,
                         setSmallIcon(R.drawable.ic_call_answer)
                     }
                 }
-                // This is a trick to make the previous notification with same id disappear as cancel notification is not working with Foreground Service.
```
bmarty, line 471: So also remove the trick itself?

bmarty (commented):
LGTM, just one question

vector/src/main/java/im/vector/app/features/notifications/NotificationUtils.kt
ganfra: Nope, it's not really a trick anymore, cancelNotification work, but only when stopForeground is called.

DoM1niC:
I found a strange issue, I use Element develop on Android / Web and use this PR but I dunno where the issue is but If I dismiss a Call so my Element Client still ringing the Caller still rejected but not the React Client (Element Web)

vector/src/main/java/im/vector/app/features/notifications/NotificationUtils.kt
bmarty: OK, thanks

bmarty (approved):

ganfra:
Thanks for commenting, I can't reproduce this behavior, not sure where it comes from, but shouldn't be from this PR.

PR #4033, opened 2021-09-17
Use in memory cache when adding inBoundGroupSession

### LINKED ISSUE #4011: RageShake | Messages are displayed as unabled to decrypt then decrypted a few seconds later
https://github.com/matrix-org/element-android-rageshakes/issues/26594

BillCarsonFr:
Fixes #4011

### DIFF
--- a/changelog.d/4011.bugfix
+++ b/changelog.d/4011.bugfix
@@ -1,0 +1,1 @@
+Messages are displayed as unable to decrypt then decrypted a few seconds later #4011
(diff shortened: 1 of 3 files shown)


changelog.d/4011.bugfix
```diff
@@ -0,0 +1 @@
+Messages are displayed as unable to decrypt then decrypted a few seconds later #4011
```
bmarty, line 1: remove issue id in the content please.

bmarty (commented):
What about changing also https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt#L648 which is called at keybackup restoration or at manual keys import?

changelog.d/4011.bugfix
BillCarsonFr: done

bmarty (approved):

bmarty:
Sorry, ktlint is not happy, can you fix it please?

PR #4034, opened 2021-09-17
Add a fallback for user displayName when this one is null or empty

### LINKED ISSUE #3732: Add a fallback for user displayName when this one is null or empty
Actually if user displayName is null or empty the function `getBestName()` return the matrixId.
It should be better to have a fallback as for the room name :`RoomDisplayNameFallbackProvider`

For example we have this fallback on Tchap : https://github.com/tchapgouv/tchap-android-v2/blob/27fd9730546e7f51734868e61433cb5a4bf4cafd/vector/src/main/java/fr/gouv/tchap/core/utils/TchapUtils.kt#L89

bmarty:
So that the application can customize the dispay name

Fixes #3732

### DIFF
--- a/vector/src/main/java/im/vector/app/features/displayname/Extension.kt
+++ b/vector/src/main/java/im/vector/app/features/displayname/Extension.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 The Matrix.org Foundation C.I.C.
+ * Copyright (c) 2021 New Vector Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,17 +14,17 @@
  * limitations under the License.
  */
 
-package org.matrix.android.sdk.api.session.user.model
+package im.vector.app.features.displayname
 
-/**
- * Data class which holds information about a user.
- * It can be retrieved with [org.matrix.android.sdk.api.session.user.UserService]
- */
-data class User(
-        val userId: String,
-        /**
-         * For usage in UI, consider converting to MatrixItem and call getBestName()
-         */
-        val displayName: String? = null,
-        val avatarUrl: String? = null
-)
+import org.matrix.android.sdk.api.util.MatrixItem
+
+fun MatrixItem.getBestName(): String {
+    return if (this is MatrixItem.GroupItem || this is MatrixItem.RoomAliasItem) {
+        // Best name is the id, and we keep the displayName of the room for the case we need the first letter
+        id
+    } else {
+        displayName
+                ?.takeIf { it.isNotBlank() }
+                ?: VectorMatrixItemDisplayNameFallbackProvider.getDefaultName(this)
+    }
+}
--- a/vector/src/main/java/im/vector/app/features/displayname/Extension.kt
+++ b/vector/src/main/java/im/vector/app/features/displayname/Extension.kt
@@ -19,6 +19,7 @@ package im.vector.app.features.displayname
 import org.matrix.android.sdk.api.util.MatrixItem
 
 fun MatrixItem.getBestName(): String {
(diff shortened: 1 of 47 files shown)


vector/src/main/java/im/vector/app/features/displayname/Extension.kt
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.features.displayname
+
+import org.matrix.android.sdk.api.util.MatrixItem
+
+fun MatrixItem.getBestName(): String {
```
ouchadam, line 21: QUESTION: what's the difference between `MatrixItem.getBestName()` and `DisplayNameResolver.getBestName()`, do we need both?

ouchadam:
small question around the different ways of fetching names, otherwise LGTM :+1:

vector/src/main/java/im/vector/app/features/displayname/Extension.kt
bmarty: That's a good question, and I do not like the code duplication too.
The idea was to create MatrixItem.getBestName() extension on the app side, to avoid touching the code everywhere.
If you find a way to get `displayNameResolver` here, it would be great.

ouchadam: I can think of two options but they both need more code changes, I'm not sure if they're worth it.

- extract all the `MatrixItem` creation to a dedicated factory, passing and creating the best name at the point of item creation

```kotlin
class MatrixItemFactory @Inject constructor() {
  fun createUser() = MatrixItem.User(bestName = "calculate best name")
  fun createGroup() = MatrixItem.Group(bestName = "calculate best name")
  //etc...
}
```

- expose the name resolver on the `Session` or a `Service` and use a `with` to access the extension

```kotlin
val bestName = with(session) {
  MatrixItem.getBestName()
}


interface Session/*Service {
  fun displayNameResolver()
  fun MatrixItem.getBestName() = displayNameResolver.getBestName(this)
}
```


I guess our only potential issue with the current implementation is the `matrixItemDisplayNameFallbackProvider` and `VectorMatrixItemDisplayNameFallbackProvider` becoming inconsistent

ouchadam (approved):

vector/src/main/java/im/vector/app/features/displayname/Extension.kt
bmarty: Thanks for the suggestions.

> I guess our only potential issue with the current implementation is the matrixItemDisplayNameFallbackProvider and VectorMatrixItemDisplayNameFallbackProvider becoming inconsistent

I agree, maybe just add a comment in the Vector side would be enough.

bmarty: https://github.com/vector-im/element-android/pull/4034/commits/b5f5f8cc73b135e8385abb6bf9911d0759a7e7d0

ouchadam: thanks!

yostyle:
LGTM

yostyle (approved):

bmarty:
CI is not happy:

```
e: /home/runner/work/element-android/element-android/vector/src/main/java/im/vector/app/features/userdirectory
/UserListController.kt: (129, 61): Unresolved reference: getBestName
```

But I do not understand:

image

bmarty:
Forced push, lets see what happen on the CI now :/

bmarty:
I guess my git configuration was a bit messy. I added the missing import, should be OK now

PR #4041, opened 2021-09-20
Fix switch to best space when tapping notification

### LINKED ISSUE #3839: Notification does not take me to the room when another space was last viewed
#### To Reproduce
For these steps, lets assume that the user has two spaces, Space1 and Space2
1. Go to Space1
2. Lock screen
3. Receive notification from a room in Space2
4. Click on the notification on the lockscreen

I see the list of rooms in Space1

#### Expected behavior
I would expect the app to open with a view of the room where the notification came from in Space2

If I last view Space2 before receiving the notification for Space2, the notification opens the correct room.

 - Device: Pixel 2
 - OS: Android 11

#### Additional context
 - App store, version 1.1.15 [40101152] (G-b2969)
 - Homeserver: matrix.org

BillCarsonFr:
Fixes #3839 (at least partially as couldn't reproduce all the issue)

There is a subtility, if I switch space when opening the room, the Lifecycle on resume update of the AppStateHandler will override the change by taking back the latest saved space. So I added a param to force persist the current space when coming from the notification


vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt
```diff
@@ -196,6 +199,24 @@ class RoomDetailViewModel @AssistedInject constructor(
...
+                val currentRoomSummary = room.roomSummary() ?: return@let
+                // nothing we are good
+                if (currentSpace == null || !currentRoomSummary.flattenParentIds.contains(currentSpace.roomId)) {
+                    // take first one or switch to home
+                    appStateHandler.setCurrentSpace(
+                            currentRoomSummary
+                                    .flattenParentIds.firstOrNull { it.isNotBlank() },
+                            // force persist, because if not on resume the AppStateHandler will resume
+                            // the current space from what was persisted on enter background
+                            persistNow = true)
```
bmarty, line 216: It looks like a weird workaround.
Do we want to update the notifications to group them by space instead, and/or include a spaceId in the notification intent itself?

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt
```diff
@@ -196,6 +199,24 @@ class RoomDetailViewModel @AssistedInject constructor(
...
             prepareForEncryption()
         }
+
+        if (initialState.switchToParentSpace) {
+            // We are coming from a notification, try to switch to the most relevant space
+            // so that when hitting back the room will appear in the list
+            appStateHandler.getCurrentRoomGroupingMethod()?.space().let { currentSpace ->
+                val currentRoomSummary = room.roomSummary() ?: return@let
+                // nothing we are good
+                if (currentSpace == null || !currentRoomSummary.flattenParentIds.contains(currentSpace.roomId)) {
```
bmarty, line 209: if `currentSpace` is null, there is nothing to do, no?

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt
```diff
@@ -196,6 +199,24 @@ class RoomDetailViewModel @AssistedInject constructor(
...
+                    // take first one or switch to home
+                    appStateHandler.setCurrentSpace(
+                            currentRoomSummary
+                                    .flattenParentIds.firstOrNull { it.isNotBlank() },
```
bmarty, line 213: Also why taking the first space here?

bmarty (commented):
Thanks. Some thoughs

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt
BillCarsonFr: It's null when the app was killed and you tap on the notification. The currentGrouping method is restored later in lifecycle :/

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt
BillCarsonFr: It's not that trivial to know which one to take. We could check and take the root parent (in case of part of a subspace)?

bmarty (approved):

PR #4057, opened 2021-09-22
Improve invite bottom sheet

BillCarsonFr:
Quick improvement of invite bottomsheet:
Whe showing an invite we try in background to query the room Summary API to get a bit more information (like the number of members). Also it could get a more up to date strippedState than the one in the room invite sync.

image

### DIFF
--- a/vector/src/main/java/im/vector/app/features/spaces/invite/SpaceInviteBottomSheetViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/spaces/invite/SpaceInviteBottomSheetViewModel.kt
@@ -61,38 +61,35 @@ class SpaceInviteBottomSheetViewModel @AssistedInject constructor(
                         peopleYouKnow = Success(peopleYouKnow)
                 )
             }
-            refreshInviteSummaryIfNeeded(roomSummary)
+            if (roomSummary.membership == Membership.INVITE) {
+                getLatestRoomSummary(roomSummary)
+            }
         }
     }
 
-    private fun refreshInviteSummaryIfNeeded(roomSummary: RoomSummary) {
-        if (roomSummary.membership == Membership.INVITE) {
-            // we can try to query the room summary api to get more info?
-            viewModelScope.launch(Dispatchers.IO) {
-                when (val peekResult = tryOrNull { session.peekRoom(roomSummary.roomId) }) {
-                    is PeekResult.Success -> {
-                        setState {
-                            copy(
-                                    summary = Success(
-                                            roomSummary.copy(
-                                                    joinedMembersCount = peekResult.numJoinedMembers,
-                                                    // it's also possible that the name/avatar did change since the invite..
-                                                    // if it's null keep the old one as summary API might not be available
-                                                    // and peek result could be null for other reasons (not peekable)
-                                                    avatarUrl = peekResult.avatarUrl ?: roomSummary.avatarUrl,
-                                                    displayName = peekResult.name ?: roomSummary.displayName,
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/spaces/invite/SpaceInviteBottomSheetViewModel.kt
```diff
@@ -57,6 +61,37 @@ class SpaceInviteBottomSheetViewModel @AssistedInject constructor(
                         peopleYouKnow = Success(peopleYouKnow)
                 )
             }
+
+            if (roomSummary.membership == Membership.INVITE) {
```
ganfra, line 65: Can you remove some nesting because it's hard to read here!

BillCarsonFr: Done, extracted a new private function do refresh summary in case of invite

vector/src/main/java/im/vector/app/features/spaces/invite/SpaceInviteBottomSheetViewModel.kt
```diff
@@ -57,6 +61,38 @@ class SpaceInviteBottomSheetViewModel @AssistedInject constructor(
                         peopleYouKnow = Success(peopleYouKnow)
                 )
             }
+            refreshInviteSummaryIfNeeded(roomSummary)
+        }
+    }
+
+    private fun refreshInviteSummaryIfNeeded(roomSummary: RoomSummary) {
+        if (roomSummary.membership == Membership.INVITE) {
```
bmarty, line 69: To follow @ganfra 's remark, I would have let this `if` test at the calling place (i.e. in the init {} block) to reduce indentation in the body of this method. Also the method could then be renamed to `fetchRoomSummary`

vector/src/main/java/im/vector/app/features/spaces/invite/SpaceInviteBottomSheetViewModel.kt
```diff
@@ -57,6 +61,38 @@ class SpaceInviteBottomSheetViewModel @AssistedInject constructor(
...
+            // we can try to query the room summary api to get more info?
```
bmarty, line 70: Remove the trailing `?`?

bmarty (commented):
One small remark, else LGTM

vector/src/main/java/im/vector/app/features/spaces/invite/SpaceInviteBottomSheetViewModel.kt
```diff
@@ -57,6 +61,38 @@ class SpaceInviteBottomSheetViewModel @AssistedInject constructor(
...
+            viewModelScope.launch(Dispatchers.IO) {
+                when (val peekResult = tryOrNull { session.peekRoom(roomSummary.roomId) }) {
+                    is PeekResult.Success -> {
+                        setState {
+                            copy(
+                                    summary = Success(
+                                            roomSummary.copy(
+                                                    joinedMembersCount = peekResult.numJoinedMembers,
```
ganfra, line 78: Maybe have a small method to map PeekResult to RoomSummary also?

bmarty: Here we are copying an existing roomSummary with info from peekResult, so I think this is not applicable

bmarty (approved):
LGTM, thanks

PR #4052, opened 2021-09-21
Add email notification toggle

ouchadam:
#2243 Email notification toggles

- Adds a `Email notifications` category to `settings/notifications` which allows for individually toggling email notifications for all associated emails
    - When the account has no associated emails, the category contains a link to the `Email and Phone numbers` screen 
    - `addEmailPusher` functionality is added to the `matrix-sdk PushersService` which allows an email address to be registered or unregistered for notifications  
- Adds a `REMOVE` button to the `settings/advanced settings/notification targets`
- Renames a few `AddHttpPusher` classes to `AddPusher` as they were able to be repurposed for emails 

* personal data has been manually overridden and replaced with [redacted]

| BEFORE | NO EMAILS AVAILABLE | EMAIL PRESENT | LIGHT - MULTIPLE EMAILS |
| --- | --- | --- | --- |
|account-before|account-no-emails-registered|account-emails-notification-disabled|account-light-multiple

| NAVIGATING EMPTY | REMOVING | ADDING | ERROR HANDLING |
| --- | --- | --- | --- |
|account-no-emails-clickthrough|account-removing-email-pusher|account-registering-email|account-error-handling

PR #4062, opened 2021-09-22
Improved /settings/notifications push toggle error handling

ouchadam:
Carrying on from #4052 Email notifications

Adds a separate synchronous (via suspend) api for registering pushers in order to provide instant feedback within the settings page when an error occurs. The background enqueuing version is still used for the FCM token regeneration.

- Splits the logic from the `WorkManager` based implementation of the `AddPusherWorker` into a `AddPusherTask` which the worker now delegates to.
- Adds unit tests around the extracted `AddPusherTask` which in turn meant extracting our a `RequestExecutor` along with other test _fakes_
- Updates the settings page to use the suspending versions of the pusher registration calls

Email notification toggles whilst offline 

| BEFORE | AFTER |
| --- | --- |
before-email-enable|after-email-errors

Session notification toggles whilst offline

| BEFORE | AFTER ENABLE | AFTER DISABLE |
| --- | --- | --- |
before-session-enable|after-disable-session|after-enable-session


matrix-sdk-android/build.gradle
```diff
@@ -130,6 +130,8 @@ dependencies {
 
     // Database
     implementation 'com.github.Zhuinden:realm-monarchy:0.7.1'
+    testImplementation libs.rx.rxKotlin
```
ouchadam, line 133: in order to mock the realm instance we need to include a test dependency on `rx` as [realm internally uses it](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/Realm.java#L49) and `monarchy` doesn't include `rx` in its transitive dependencies

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/pushers/AddPusherWorker.kt
```diff
@@ -43,9 +34,7 @@ internal class AddPusherWorker(context: Context, params: WorkerParameters)
             override val lastFailureMessage: String? = null
     ) : SessionWorkerParams
 
-    @Inject lateinit var pushersAPI: PushersAPI
-    @Inject @SessionDatabase lateinit var monarchy: Monarchy
-    @Inject lateinit var globalErrorReceiver: GlobalErrorReceiver
+    @Inject lateinit var addPusherTask: AddPusherTask
```
ouchadam, line 37: The worker now delegates to the extracted `AddPusherTask`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/pushers/AddPusherWorker.kt
```diff
@@ -58,50 +47,17 @@ internal class AddPusherWorker(context: Context, params: WorkerParameters)
             return Result.failure()
         }
         return try {
-            setPusher(pusher)
+            addPusherTask.execute(AddPusherTask.Params(pusher))
             Result.success()
         } catch (exception: Throwable) {
             when (exception) {
                 is Failure.NetworkConnection -> Result.retry()
-                else                         -> {
```
ouchadam, line 66: due to extracting the logic to a task there's now a slight change in behaviour, the `AddPusherTask` will always update any existing pusher state on failure instead of `exception != NetworkConnection` as the task can be used in non worker contexts (such as the settings page)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/pushers/DefaultPushersService.kt
```diff
@@ -107,13 +103,6 @@ internal class DefaultPushersService @Inject constructor(
         return request.id
     }
 
-    private fun JsonPusher.validateParameters() {
```
ouchadam, line 110: as we're always validating the payload I've lifted the logic as close of possible to the fields by doing it in the given class's `init`

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/pushers/DefaultAddPusherTaskTest.kt
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+internal class FakeMonarchy {
+
+    val instance = mockk<Monarchy>()
+    private val realm = mockk<Realm>(relaxed = true)
+
+    init {
+        mockkStatic("org.matrix.android.sdk.internal.util.MonarchyKt")
```
ouchadam, line 144: in its current form the transactions call chain isn't very test friendly so I've opted to mock the top level `awaitTransaction` extension

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/pushers/DefaultAddPusherTaskTest.kt
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+internal class FakePushersAPI : PushersAPI {
+
+    private var setRequestPayload: JsonPusher? = null
+    private var error: Throwable? = null
+
+    override suspend fun getPushers(): GetPushersResponse {
+        TODO("Not yet implemented")
```
ouchadam, line 121: the current tests don't need this yet, hence the throwing TODO, ideally this would be implemented when a test around fetching pushes is created

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/pushers/DefaultAddPusherTaskTest.kt
```diff
@@ -0,0 +1,178 @@
...
+        monarchy.givenWhereReturns<PusherEntity>(result = null)
+        pushersAPI.givenSetPusherErrors(SocketException())
+
+        assertFailsWith<SocketException> {
+            runBlocking { addPusherTask.execute(AddPusherTask.Params(A_JSON_PUSHER)) }
+        }
+    }
+}
+
+internal class FakePushersAPI : PushersAPI {
```
ouchadam, line 115: this PR introduces the concept of fakes [(here's a nice refresher!)](https://blog.pragmatists.com/test-doubles-fakes-mocks-and-stubs-1a7491dfa3da), the main purpose in this context is to avoid hitting the real api and database and to provide some reusable assertion/verification helpers

I've left the implementations within the test file for easier discussion/reviewing, will extract out to their own package if the team is happy with them :crossed_fingers:

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/pushers/DefaultAddPusherTaskTest.kt
```diff
@@ -0,0 +1,178 @@
...
+
+    @Test
+    fun `given no persisted pusher when adding Pusher then updates api and inserts result with Registered state`() {
+        monarchy.givenWhereReturns<PusherEntity>(result = null)
+
+        runBlocking { addPusherTask.execute(AddPusherTask.Params(A_JSON_PUSHER)) }
+
+        pushersAPI.verifySetPusher(A_JSON_PUSHER)
+        monarchy.verifyInsertOrUpdate<PusherEntity> {
+            withArg { actual ->
```
ouchadam, line 72: unfortunately the base entity models don't include any equality implementations so we have to manually check the fields

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/pushers/AddPusherTask.kt
```diff
@@ -0,0 +1,95 @@
...
+                echo.profileTag = pusher.profileTag
+                echo.data?.format = pusher.data?.format
+                echo.data?.url = pusher.data?.url
+                echo.state = PusherState.REGISTERED
+            }
+        }
+    }
+}
+
+internal interface RequestExecutor {
```
ouchadam, line 79: this should be extracted to its own file, which package would be best?

matrix-sdk-android/build.gradle
bmarty: Can you add the comment in the code? Also maybe move this line next to the other `testImplementation` items (I have no strong opinion on this last remark)?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushers/PushersService.kt
```diff
@@ -128,4 +107,62 @@ interface PushersService {
...
+    data class HttpPusher(
+
+            /**
+             * This is a unique identifier for this pusher. The value you should use for
+             * this is the routing or destination address information for the notification,
+             * for example, the APNS token for APNS or the Registration ID for GCM. If your
+             * notification client has no such concept, use any unique identifier. Max length, 512 chars.
+             * If the kind is "email", this is the email address to send notifications to.
+             */
+            val pushkey: String,
```
bmarty, line 120: Since the kind can be "email", this is strange that this class is called `HttpPusher`. Rename to `Pusher` ?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushers/PushersService.kt
```diff
@@ -128,4 +107,62 @@ interface PushersService {
...
     fun getPushers(): List<Pusher>
+
+    data class HttpPusher(
+
+            /**
+             * This is a unique identifier for this pusher. The value you should use for
+             * this is the routing or destination address information for the notification,
+             * for example, the APNS token for APNS or the Registration ID for GCM. If your
+             * notification client has no such concept, use any unique identifier. Max length, 512 chars.
+             * If the kind is "email", this is the email address to send notifications to.
```
bmarty, line 118: I think with this Service API change, the kind should not be "email", so maybe remove this line?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/RequestExecutor.kt
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.internal.network
+import org.matrix.android.sdk.internal.network.executeRequest as internalExecuteRequest
+
+internal interface RequestExecutor {
+    suspend fun <DATA> executeRequest(globalErrorReceiver: GlobalErrorReceiver?,
+                                      canRetry: Boolean = false,
+                                      maxDelayBeforeRetry: Long = 32_000L,
```
bmarty, line 23: Why this default value 32s ?

vector/src/main/java/im/vector/app/core/pushers/PushersManager.kt
```diff
@@ -44,24 +45,29 @@ class PushersManager @Inject constructor(
...
-                append = false,
-                withEventIdOnly = true
-        )
+    suspend fun registerPusherWithFcmKey(pushKey: String) {
+        val currentSession = activeSessionHolder.getActiveSession()
+        currentSession.addHttpPusher(httpPusher(pushKey))
     }
 
-    fun registerEmailForPush(email: String) {
+    private fun httpPusher(pushKey: String) = PushersService.HttpPusher(
```
bmarty, line 58: rename to `createHttpPusher`? functions should start with a verb

vector/src/main/java/im/vector/app/features/settings/notifications/VectorSettingsNotificationPreferenceFragment.kt
```diff
@@ -85,6 +85,21 @@ class VectorSettingsNotificationPreferenceFragment @Inject constructor(
             (pref as SwitchPreference).isChecked = areNotifEnabledAtAccountLevel
         }
 
+        findPreference<SwitchPreference>(VectorPreferences.SETTINGS_ENABLE_THIS_DEVICE_PREFERENCE_KEY)?.let {
+            it.setTransactionalSwitchChangeListener(lifecycleScope) { isChecked ->
+                if (isChecked) {
+                    FcmHelper.getFcmToken(requireContext())?.let {
+                        pushManager.registerPusherWithFcmKey(it)
```
bmarty, line 92: it's weird that in this case, we do not call `session.refreshPushers()` (I see this is not a change from this PR)

bmarty (commented):
Thanks a lot for improving the test framework.
Some minor remarks

PR #4070, opened 2021-09-23
Adding benchmarking script to start monitoring build times

ouchadam:
Adds a script to run the [gradle-profiler](https://github.com/gradle/gradle-profiler) so that we can start monitoring and quantifying build time improvements  

### Usage 

- Requires the gradle-profile to be installed
- running from the root of the repository `./tools/benchmark/run-benchmark "comma separated scenario names or blank to run all"`

The warm up and run iterations have been reduced from the defaults, makes the overall suite take less time but at the cost some reliability

Here's some early results...

#### Recompiling

```
assemble {
  tasks = [":vector:assembleGPlayDebug"]
}
```

2021-09-23T16:08:22,201186771+01:00

recompiling without any changes takes 1 second~ it's almost entirely gradle configuration, it shows our gradle setup isn't invaliding the configuration and causing extra processing

#### Clean builds 
assemble clean builds with and without the build cache, build cache is disabled by default
```
assemble {
  tasks = ["clean", ":vector:assembleGPlayDebug"]
}

assemble_build_cache {
  tasks = ["clean", ":vector:assembleGPlayDebug"]
  gradle-args = ["--build-cache"]
}
```

2021-09-23T15:49:46,263256186+01:00

From these results we can see that the build cache saves a massive amount of build time, 2m 30s~ down to 14~ seconds, not too unexpected as the build cache doesn't get cleared with a `clean` and since we're only doing a clean we should get lots of cache hits.

#### Incremental

Making a non ABI and ABI change in the matrix sdk module

2021-09-23T16:07:57,679358923+01:00

a 15 second~ incremental time shows us at a high level that the SDK module in correctly compiling incrementally, a clean assembly of the sdk module takes 1 minute~. Bencharking ABI changes will be more useful as we introduce more modules as gradle will force clients of modules with ABI to also recompile.  

More information can be found at https://developer.android.com/studio/build/profile-your-build

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .idea/*.xml
 .DS_Store
 /build
+/benchmark-out
 /captures
 .externalNativeBuild
 
--- a/tools/benchmark/benchmark.profile
+++ b/tools/benchmark/benchmark.profile
@@ -1,0 +1,39 @@
+#
+# Copyright 2021 New Vector Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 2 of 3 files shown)


bmarty (commented):
Really interesting thanks.
By monitoring, you mean let the CI run gradle-profiler?

ouchadam:
@bmarty it could be the CI if we want to automate build times and produce a dashboard (which would be pretty nice!) although to start with running locally periodically would be simpler

bmarty (approved):
Thanks!

PR #4071, opened 2021-09-23
Fixing no tests found when running vector:connectedAndroidTest

ouchadam:
Adds missing orchestrator test util, this was being set in the `matrix-android-sdk` module but not in `vector`.

This should hopefully get the CI sanity test step completing again

caught by @onurays  :pray:

### DIFF
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -508,4 +508,5 @@ dependencies {
     androidTestImplementation('com.adevinta.android:barista:4.1.0') {
         exclude group: 'org.jetbrains.kotlin'
     }
+    androidTestUtil 'androidx.test:orchestrator:1.4.0'
 }
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -508,5 +508,5 @@ dependencies {
     androidTestImplementation('com.adevinta.android:barista:4.1.0') {
         exclude group: 'org.jetbrains.kotlin'
     }
-    androidTestUtil 'androidx.test:orchestrator:1.4.0'
+    androidTestUtil libs.androidx.orchestrator
 }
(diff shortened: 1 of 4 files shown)


vector/build.gradle
```diff
@@ -508,4 +508,5 @@ dependencies {
     androidTestImplementation('com.adevinta.android:barista:4.1.0') {
         exclude group: 'org.jetbrains.kotlin'
     }
+    androidTestUtil 'androidx.test:orchestrator:1.4.0'
```
bmarty, line 511: For dependency common in several module, we should now use [this file](https://github.com/vector-im/element-android/blob/develop/dependencies.gradle), introduced by @ariskotsomitopoulos.

bmarty (commented):
LGTM, thanks, one remark though

vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
```diff
@@ -283,6 +283,7 @@ class UiAllScreensSanityTest {
         clickListItem(R.id.matrixProfileRecyclerView, 9)
         // File tab
         clickOn(R.string.uploads_files_title)
+        sleep(1000)
```
ouchadam, line 286: whilst running the tests locally I was seeing some flaky behaviour around closing dialogs and switching viewpagers tabs, the sleeps aren't ideally but the tests run too fast without them!

bmarty: OK, this is not the first occurrence (I think) in this test. I would like to find some active triggering to continue the test instead of waiting, but I have now idea how to do that today.
Adding too many sleep() will slow down the entire test suite execution, and we want tests to be fast. But of course you already know that :)

bmarty: But anyway I prefer the test to pass at first shot instead of trying again because of some flaky behavior :)

bmarty (approved):
Thanks for the update!

PR #4073, opened 2021-09-24
[ignore] Temporarily running the sanity tests on mac-osx VM

ouchadam:
Attempting to get the sanity tests to run to completion, the current theory is we need a hardware accelerated emulator in order to correctly render the UI


ouchadam:
the CI emulator really doesn't like our test suite... closing for the time being

PR #4083, opened 2021-09-27
Fix default encrypted for restricted

### LINKED ISSUE #4045: Align new room encryption default to Web
### Your use case

In a space, create a new room.
In the Create New Room dialog, Room settings -> Enable encryption should be on by default as it is in Web because we want to encourage secure communications

_No response_

### Additional context

_No response_

BillCarsonFr:
Fixes #4045 

Align default encrypted for restricted room as web is doing it, i.e use what is in well-known (default e2e).
Private and restricted encrypted switch is on if well-known e2e is by default.
=> Also if the user did manually change the encrypted toggled then always take that value and not the default.

Additional Fix:
When no space was selected, the option to set the room to restricted was visible, that would create a room restricted to nothing. So now if no space selected you'll only have a choice between Private or Public

### DIFF
--- a/vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomController.kt
+++ b/vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomController.kt
@@ -165,7 +165,11 @@ class CreateRoomController @Inject constructor(
                             host.stringProvider.getString(R.string.create_room_encryption_description)
                         }
                 )
-                switchChecked(viewState.isEncrypted)
+                if (viewState.isEncrypted != null) {
+                    switchChecked(viewState.isEncrypted)
+                } else {
+                    switchChecked(viewState.defaultEncrypted[viewState.roomJoinRules] ?: false)
+                }
 
                 listener { value ->
                     host.listener?.setIsEncrypted(value)
--- a/vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomController.kt
+++ b/vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomController.kt
@@ -29,6 +29,7 @@ import im.vector.app.features.form.formEditTextItem
 import im.vector.app.features.form.formEditableAvatarItem
 import im.vector.app.features.form.formSubmitButtonItem
 import im.vector.app.features.form.formSwitchItem
+import org.matrix.android.sdk.api.extensions.orFalse
 import org.matrix.android.sdk.api.session.room.failure.CreateRoomFailure
 import org.matrix.android.sdk.api.session.room.model.RoomJoinRules
 import javax.inject.Inject
@@ -165,11 +166,8 @@ class CreateRoomController @Inject constructor(
                             host.stringProvider.getString(R.string.create_room_encryption_description)
                         }
                 )
-                if (viewState.isEncrypted != null) {
-                    switchChecked(viewState.isEncrypted)
-                } else {
-                    switchChecked(viewState.defaultEncrypted[viewState.roomJoinRules] ?: false)
-                }
+
+                switchChecked(viewState.isEncrypted ?: viewState.defaultEncrypted[viewState.roomJoinRules].orFalse())
 
                 listener { value ->
                     host.listener?.setIsEncrypted(value)
--- a/vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomFragment.kt
@@ -163,8 +163,9 @@ class CreateRoomFragment @Inject constructor(
     }
 
     override fun selectVisibility() = withState(viewModel) { state ->
-
-        val allowed = if (state.supportsRestricted) {
+        // If restricted is supported and the user is in the context of a parent space
+        // then show restricted option.
+        val allowed = if (state.supportsRestricted && state.parentSpaceId != null) {
             listOf(RoomJoinRules.INVITE, RoomJoinRules.PUBLIC, RoomJoinRules.RESTRICTED)
         } else {
             listOf(RoomJoinRules.INVITE, RoomJoinRules.PUBLIC)
--- a/vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomViewModel.kt
@@ -36,6 +36,7 @@ import im.vector.app.features.settings.VectorPreferences
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import org.matrix.android.sdk.api.MatrixPatterns.getDomain
+import org.matrix.android.sdk.api.extensions.orFalse
 import org.matrix.android.sdk.api.extensions.tryOrNull
 import org.matrix.android.sdk.api.raw.RawService
 import org.matrix.android.sdk.api.session.Session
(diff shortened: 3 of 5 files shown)


vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomController.kt
```diff
@@ -165,7 +165,11 @@ class CreateRoomController @Inject constructor(
                             host.stringProvider.getString(R.string.create_room_encryption_description)
                         }
                 )
-                switchChecked(viewState.isEncrypted)
+                if (viewState.isEncrypted != null) {
```
ouchadam, line 168: NIT: Something I've noticed through the codebase is we tend to favour inverted if conditions, it might just be me but I find positive conditions much easier on the brain 

```kotlin
if (viewState.isEncrypted == null) {
  switchChecked(viewState.defaultEncrypted[viewState.roomJoinRules] ?: false)
} else {
  switchChecked(viewState.isEncrypted)
}
``` 

and whilst inverting the if ^^^ I also noticed we could simplify the checked state by chaining the fallbacks

```kotlin
val isSwitchChecked = viewState.isEncrypted ?: viewState.defaultEncrypted[viewState.roomJoinRules] ?: false
```

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomFragment.kt
```diff
@@ -163,8 +163,9 @@ class CreateRoomFragment @Inject constructor(
     }
 
     override fun selectVisibility() = withState(viewModel) { state ->
-
-        val allowed = if (state.supportsRestricted) {
+        // If restricted is supported and the user is in the context of a parent space
+        // then show restricted option.
+        val allowed = if (state.supportsRestricted && state.parentSpaceId != null) {
```
ouchadam, line 168: NIT: if we wanted to avoid using separate branches for appending to the list (and potentially be more kotlin idiomatic) we could use 

`listOfNotNull`

```kotlin
val list = listOfNotNull(
        INVITE,
        PUBLIC,
        RESTRICTED.takeIf { state.supportsRestricted && state.parentSpaceId != null }
)
```

or `buildList`

```kotlin
val list = buildList {
    add(INVITE)
    add(PUBLIC)
    if (state.supportsRestricted && state.parentSpaceId != null) {
        add(RESTRICTED)
    }
}
```

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomViewModel.kt
```diff
@@ -286,8 +291,17 @@ class CreateRoomViewModel @AssistedInject constructor(@Assisted private val init
                     disableFederation = state.disableFederation
 
                     // Encryption
-                    if (state.isEncrypted) {
-                        enableEncryption()
+                    // we ignore the isEncrypted for public room as the switch is hidden in this case
+                    if (state.roomJoinRules != RoomJoinRules.PUBLIC && state.isEncrypted != null) {
```
ouchadam, line 295: NIT: it might be clearer to convert this to a `when` and decouple `isEncrypted` creation from the `enableEncryption()`

```kotlin
val isEncrpyted = when {
  // public rooms use the default encrypted join rules as the switch is hidden
  state.roomJoinRules == RoomJoinRules.PUBLIC -> state.defaultEncrypted[state.roomJoinRules] ?: false
  else -> state.isEncrypted ?: false
}

if (isEncrpyted) {
  enableEncryption()
}
```

what do you think?

ouchadam:
awesome stuff! :100:  I'm attempting to use https://emmer.dev/blog/code-review-comment-prefixes/ to hopefully avoid confusion in my reviews :sweat_smile: 

`NIT` being opinionated optional comments

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomController.kt
bmarty: Generally I agree to avoid `if(!cond) {} else {}`, but sometimes it also nice to have the most common case first. I'm not saying this is the case here, just wanted to add my two pence :)

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomController.kt
```diff
@@ -165,7 +165,11 @@ class CreateRoomController @Inject constructor(
...
+                    switchChecked(viewState.isEncrypted)
+                } else {
+                    switchChecked(viewState.defaultEncrypted[viewState.roomJoinRules] ?: false)
+                }
```
bmarty, line 172: In this case, I would have probably write:
```kotlin
switchChecked(
    viewState.isEncrypted
    ?: viewState.defaultEncrypted[viewState.roomJoinRules]
    ?: false
)
```

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomViewModel.kt
bmarty: `?: false` can also get out the `when` branch. We also have a nice `.orFalse` extension.
@BillCarsonFr if you copy this code, please fix the typo in `isEncrpyted` 😛

bmarty (commented):
LGTM, just some comment on top on @adam comments :)

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomViewModel.kt
ouchadam: for some reason I can't type `isEncrypted` :facepalm:

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomFragment.kt
BillCarsonFr: AS is telling me that it's experimental

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomViewModel.kt
BillCarsonFr: ah ah I definitly failed to make it clearer :D

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomController.kt
BillCarsonFr: Yes I explicitly used the if/else like that thinking it will ease reading of the code (otherwise I would have gone for the chaining).
Looks like I failed  :D.

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomFragment.kt
ouchadam: ah my bad, I totally forgot that `buildList` is still experimental

vector/src/main/java/im/vector/app/features/roomdirectory/createroom/CreateRoomViewState.kt
```diff
@@ -38,7 +38,8 @@ data class CreateRoomViewState(
         val parentSpaceSummary: RoomSummary? = null,
         val supportsRestricted: Boolean = false,
         val aliasLocalPart: String? = null,
-        val isSubSpace: Boolean = false
+        val isSubSpace: Boolean = false,
+        val defaultEncrypted: Map<RoomJoinRules, Boolean> = emptyMap()
```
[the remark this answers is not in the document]
BillCarsonFr: done

ouchadam (approved):

bmarty (approved):

PR #4090, opened 2021-09-27
Feature/aris/presence

ariskotsomitopoulos:


DoM1niC:
Presence in RoomDetails and in Room would be nice to have :) I like this PR 💯

gradle.properties
```diff
@@ -19,7 +19,7 @@ android.useAndroidX=true
 
 #Project Settings
 # Change debugPrivateData to true for debugging
-vector.debugPrivateData=false
+vector.debugPrivateData=true
 # httpLogLevel values: NONE, BASIC, HEADERS, BODY
-vector.httpLogLevel=BASIC
+vector.httpLogLevel=BODY
```
bmarty, line 24: You should have kept the comment here: https://github.com/vector-im/element-android/commit/fe1d9aa589e248507555c4243361dccdb459a8c8#diff-3d103fc7c312a3e136f88e81cef592424b8af2464c468116545c4d22d6edcf19L23 to avoid editing that

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/di/PresenceModule.kt
```diff
@@ -0,0 +1,54 @@
...
+
+@Module
+internal abstract class PresenceModule {
+
+    @Module
+    companion object {
+        @Provides
+        @JvmStatic
+        @SessionScope
+        fun providesProfileAPI(retrofit: Retrofit): PresenceAPI {
```
ganfra, line 40: wrong typo on provides methods => providesPresenceAPI

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/ReadReceiptHandler.kt
```diff
@@ -1,11 +1,11 @@
 /*
- * Copyright 2020 The Matrix.org Foundation C.I.C.
+ * Copyright (c) 2021 New Vector Ltd
```
ganfra: This should use Matrix copyright

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 The Matrix.org Foundation C.I.C.
+ * Copyright (c) 2021 New Vector Ltd
```
ganfra: Matrix copyright

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/RoomSummaryEntity.kt
```diff
@@ -204,6 +205,11 @@ internal open class RoomSummaryEntity(
             if (value != field) field = value
         }
 
+    var presence: UserPresenceEntity? = null
```
ganfra, line 208: I guess this is only for DM rooms? Maybe the name should reflect

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/RoomSummary.kt
```diff
@@ -60,7 +61,8 @@ data class RoomSummary(
         val roomType: String? = null,
         val spaceParents: List<SpaceParentInfo>? = null,
         val spaceChildren: List<SpaceChildInfo>? = null,
-        val flattenParentIds: List<String> = emptyList()
+        val flattenParentIds: List<String> = emptyList(),
+        val userPresence: UserPresence? = null
```
ganfra, line 65: I guess this is only for DM rooms? Maybe the name should reflect

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryItem.kt
```diff
@@ -52,6 +54,8 @@ abstract class RoomSummaryItem : VectorEpoxyModel<RoomSummaryItem.Holder>() {
     @EpoxyAttribute(EpoxyAttribute.Option.DoNotHash) lateinit var lastFormattedEvent: CharSequence
     @EpoxyAttribute lateinit var lastEventTime: CharSequence
     @EpoxyAttribute var encryptionTrustLevel: RoomEncryptionTrustLevel? = null
+    @EpoxyAttribute var userPresence: UserPresence? = null
+    @EpoxyAttribute var izDirect: Boolean = false
```
ganfra, line 58: I would prefer having a showPresence flag here (even if it's the same result)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/presence/PresenceService.kt
```diff
@@ -0,0 +1,43 @@
...
+     * @param presence the new user presence
+     * @param message the new user status message
+     */
+    suspend fun setPresence(userId: String, presence: PresenceEnum, message: String? = null)
+
+    /**
+     * Returns the presence status for this user
+     * @param userId the userId to update the avatar of
+     */
+    suspend fun getPresence(userId: String): GetPresenceResponse
```
ganfra, line 41: Maybe rename to fetchPresence and adds a method to get the UserPresence from db?

ganfra (changes requested):
Some remarks, otherwise, good job (not tried)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/di/PresenceModule.kt
ariskotsomitopoulos: True, thanks!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/presence/PresenceService.kt
```diff
@@ -0,0 +1,43 @@
...
+
+interface PresenceService {
+
+    /**
+     * Update the presence status for this user
+     * @param userId the userId to update the display name of
+     * @param presence the new user presence
+     * @param message the new user status message
+     */
+    suspend fun setPresence(userId: String, presence: PresenceEnum, message: String? = null)
```
bmarty, line 35: As per the documentation `You cannot set the presence state of another user`, so maybe remove the `userId` parameter, and its wrong documentation :)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/presence/PresenceService.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Returns the presence status for this user
+     * @param userId the userId to update the avatar of
```
bmarty, line 39: bad doc

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/RoomSummary.kt
bmarty: Also please move it just after `directUserId`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/RealmSessionStoreMigration.kt
```diff
@@ -338,4 +341,26 @@ internal object RealmSessionStoreMigration : RealmMigration {
         realm.schema.get("EventInsertEntity")
                 ?.addField(EventInsertEntityFields.CAN_BE_PROCESSED, Boolean::class.java)
     }
+
+    private fun migrateTo18(realm: DynamicRealm) {
+        Timber.d("Step 17 -> 18")
+        realm.schema.create("UserPresenceEntity")?.apply {
```
bmarty, line 347: why not using the same format than the other migration?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/presence/UserPresenceEntity.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import io.realm.RealmObject
+import io.realm.annotations.PrimaryKey
+import org.matrix.android.sdk.internal.session.presence.messages.PresenceEnum
+
+internal open class UserPresenceEntity(@PrimaryKey var userId: String = "",
+                                       var lastActiveAgo: Long? = null,
+                                       var statusMessage: String? = null,
+                                       var isCurrentlyActive: Boolean? = null,
+                                       var avatarUrl: String? = null
```
bmarty, line 27: why not storing also the display name?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/RoomSummaryEntityQueries.kt
```diff
@@ -67,3 +68,13 @@ internal fun RoomSummaryEntity.Companion.isDirect(realm: Realm, roomId: String):
             .findAll()
             .isNotEmpty()
 }
+
+internal fun RoomSummaryEntity.Companion.updateDirectUserPresence(realm: Realm, directUserId: String, userPresenceEntity: UserPresenceEntity) {
+    RoomSummaryEntity.where(realm)
+            .equalTo(RoomSummaryEntityFields.IS_DIRECT, true)
+            .equalTo(RoomSummaryEntityFields.DIRECT_USER_ID, directUserId)
+            .findFirst()
+            ?.apply {
```
bmarty, line 77: why not writing `?.presence = userPresenceEntity`?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/presence/UserPresenceEntityQueries.kt
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.internal.database.query.presence
+
+import io.realm.Realm
+import org.matrix.android.sdk.internal.database.model.presence.UserPresenceEntity
+
+internal fun UserPresenceEntity.Companion.insertOrUpdate(realm: Realm, userPresenceEntity: UserPresenceEntity) : UserPresenceEntity? {
```
bmarty, line 22: I'm not a big fan of using the same name than a fun from the Realm library. Also this extension is maybe a bit useless, no?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/PresenceAPI.kt
```diff
@@ -0,0 +1,44 @@
...
+
+internal interface PresenceAPI {
+
+    /**
+     * Set the presence status of the current user
+     * Ref: https://matrix.org/docs/spec/client_server/latest#put-matrix-client-r0-presence-userid-status
+     */
+    @PUT(NetworkConstants.URI_API_PREFIX_PATH_R0 + "presence/{userId}/status")
+    suspend fun setPresence(@Path("userId") userId: String,
+                            @Body body: SetPresenceBody): Any
```
bmarty, line 35: I would prefer this to return nothing than `Any`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/PresenceAPI.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Get the presence status of the current user
```
bmarty, line 38: This is not only for the current user, I think

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/di/PresenceModule.kt
```diff
@@ -0,0 +1,54 @@
...
+        fun providesPresenceAPI(retrofit: Retrofit): PresenceAPI {
+            return retrofit.create(PresenceAPI::class.java)
+        }
+    }
+
+    @Binds
+    abstract fun bindSetPresenceTask(task: DefaultSetPresenceTask): SetPresenceTask
+
+    @Binds
+    abstract fun bindPresenceService(service: DefaultPresenceService): PresenceService
```
bmarty, line 49: very minor remark, but maybe reorder to put the service first and the tasks after.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/service/DefaultPresenceService.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.matrix.android.sdk.api.session.presence.PresenceService
+import org.matrix.android.sdk.internal.session.presence.messages.PresenceEnum
+import org.matrix.android.sdk.internal.session.presence.service.task.GetPresenceTask
+import org.matrix.android.sdk.internal.session.presence.service.task.SetPresenceTask
+import javax.inject.Inject
+
+internal class DefaultPresenceService @Inject constructor(private val setPresenceTask: SetPresenceTask,
+                                                          private val getPresenceTask: GetPresenceTask) : PresenceService {
+
+    override suspend fun setPresence(userId: String, presence: PresenceEnum, message: String?) {
```
bmarty, line 29: you can inject the current userId in the constructor using `@UserId userId: String`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/service/task/SetPresenceTask.kt
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2020 The Matrix.org Foundation C.I.C.
```
bmarty, line 2: All new file should have 2021 copyright

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/service/task/SetPresenceTask.kt
```diff
@@ -0,0 +1,47 @@
...
+            val message: String?
+    )
+}
+
+internal class DefaultSetPresenceTask @Inject constructor(private val presenceAPI: PresenceAPI,
+                                                          private val globalErrorReceiver: GlobalErrorReceiver) : SetPresenceTask() {
+
+    override suspend fun execute(params: Params): Any {
+        return executeRequest(globalErrorReceiver) {
+            params.apply {
```
bmarty, line 41: weird way to write it, generally we just use `params.` which is I think a bit more readable.
As a general rule, we avoid using `apply` when it's not strictly necessary, because it may lead to hard to find bug: in the apply block you've got 2 different `this`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/PresenceSyncHandler.kt
```diff
@@ -0,0 +1,66 @@
...
+        presenceSyncResponse?.events?.filter { event ->
+            event.type == EventType.PRESENCE
+        }?.forEach { event ->
+            val content = event.content ?: return@forEach
+            val userId = event.senderId ?: return@forEach
+            val presence = content["presence"]?.let { PresenceEnum.from(it.toString()) } ?: return@forEach
+            val lastActiveAgo = content["last_active_ago"] as? Long
+            val statusMessage = content["status_msg"]?.toString()
+            val isCurrentlyActive = content["currently_active"] as? Boolean
+            val avatarUrl = content["avatar_url"]?.toString()
```
bmarty, line 42: why parsing manually here? You can create a data class `PresenceContent`with all the specified field for such content

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/PresenceSyncHandler.kt
```diff
@@ -0,0 +1,66 @@
...
+
+            val userPresenceEntity = UserPresenceEntity(
+                    userId = userId,
+                    lastActiveAgo = lastActiveAgo,
+                    statusMessage = statusMessage,
+                    isCurrentlyActive = isCurrentlyActive,
+                    avatarUrl = avatarUrl
+            ).apply {
```
bmarty, line 50: maybe `also` is more appropriate here?

vector/src/main/res/layout/item_room.xml
```diff
@@ -84,6 +84,21 @@
...
         app:layout_constraintCircleRadius="28dp"
         tools:ignore="MissingConstraints"
+        tools:visibility="gone" />
+
+    <ImageView
+        android:id="@+id/roomAvatarPresenceImageView"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:background="@drawable/background_circle"
+        android:contentDescription="@string/a11y_public_room"
```
bmarty, line 94: bad

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/presence/PresenceService.kt
ariskotsomitopoulos: You are right, I thought it was the same logic with the set profile `displayName`, thanks

vector/src/main/res/layout/item_room.xml
```diff
@@ -84,6 +84,21 @@
...
+        android:padding="2dp"
+        android:src="@drawable/ic_presence_offline"
```
bmarty, line 96: maybe use `tools:` here, to avoid blink when creating the screen

bmarty (changes requested):
Good work, I know the PR is still Draft, I've just made first remarks (static review).

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/RealmSessionStoreMigration.kt
ariskotsomitopoulos: I think the code is clearer without so many:
```
?.
?.
?.
```
However, I will change it to be consistent with the other migrations

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/presence/PresenceService.kt
ariskotsomitopoulos: Yes, I agree with you I just see other services like in `ProfileService` `getDisplayName` and `getAvatarUrl` that they don't use **fetch** and use **get**. So lets stick with **fetch** from server and **get** from DB as a convention ?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/presence/UserPresenceEntityQueries.kt
ariskotsomitopoulos: I agree it's quite useless for that simple case, I will remove it. My initial thought was that all the queries transaction for `UserPresenceEntity` would be from the file `UserPresenceEntityQueries` and not from whenever the code needs.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/service/task/SetPresenceTask.kt
ariskotsomitopoulos: Nice, let's stick to that convention then. Furthermore, when we use `apply` we can explicitly use `this@ClassName` to be more clear.

vector/src/main/res/layout/item_room.xml
ariskotsomitopoulos: Can you be more specific? 
 - `wrap_content` is replaced by static `16dp`
 - `android:contentDescription` needs translations

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/presence/PresenceService.kt
ganfra: Yes I think we should change in other services too

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/service/task/SetPresenceTask.kt
bmarty: Yes, I know, but it's quite ugly to me.

I want to avoid mistake which can happen for instance in this simple case:

```kotlin
class B {
    var myVar: Int = 0
}

class A {
    var b = B()
    var myVar: Int = 0

    fun test() {
        myVar = 4
        b.apply {
            myVar = 5
        }
    }
}
```

This code is compiling fine, without any warning, but developer can make mistake and/or have doubt about the value of `A.myVar` once the function `test` is called.

vector/src/main/res/layout/item_room.xml
bmarty: Yes, sorry, `a11y_public_room` is not the correct value to use here.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/model/PresenceContent.kt
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.internal.session.presence.model
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+/**
+ * Class representing the EventType.STATE_ROOM_MEMBER state event content
```
bmarty, line 23: Please update this comment.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/service/task/SetPresenceTask.kt
ariskotsomitopoulos: Yes, I agree it might be confusing for someone so lets avoid it

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/service/DefaultPresenceService.kt
```diff
@@ -0,0 +1,36 @@
...
+
+internal class DefaultPresenceService @Inject constructor(@UserId private val userId: String,
+                                                          private val setPresenceTask: SetPresenceTask,
+                                                          private val getPresenceTask: GetPresenceTask) : PresenceService {
+
+    override suspend fun setPresence(presence: PresenceEnum, message: String?) {
+        setPresenceTask.execute(SetPresenceTask.Params(userId, presence, message))
+    }
+
+    override suspend fun fetchPresence(userId: String) = getPresenceTask.execute(GetPresenceTask.Params(userId))
```
bmarty, line 35: As said by @ganfra you should add a method to get LiveData of a user presence (or if you want to use Flow, it would be better!), to make it live in the UI.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/mapper/RoomSummaryMapper.kt
```diff
@@ -49,6 +49,7 @@ internal class RoomSummaryMapper @Inject constructor(private val timelineEventMa
                 joinRules = roomSummaryEntity.joinRules,
                 isDirect = roomSummaryEntity.isDirect,
                 directUserId = roomSummaryEntity.directUserId,
+                directUserPresence = roomSummaryEntity.directUserPresence?.let { UserPresence(it.lastActiveAgo, it.statusMessage, it.isCurrentlyActive, it.presence) },
```
ganfra: Maybe create a small map method?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/presence/PresenceService.kt
```diff
@@ -27,17 +27,15 @@ import org.matrix.android.sdk.internal.session.presence.messages.PresenceEnum
...
-     * Update the presence status for this user
-     * @param userId the userId to update the display name of
-     * @param presence the new user presence
-     * @param message the new user status message
+     * Update the presence status for the current user
+     * @param presence the new presence state
+     * @param message the status message to attach to this state
      */
-    suspend fun setPresence(userId: String, presence: PresenceEnum, message: String? = null)
+    suspend fun setPresence(presence: PresenceEnum, message: String? = null)
```
ganfra, line 34: setMyPresence?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/model/PresenceContent.kt
ganfra: Don't forget this one ^

ganfra (commented):
Some small remaining remarks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/RoomMemberSummaryEntity.kt
```diff
@@ -39,6 +40,11 @@ internal open class RoomMemberSummaryEntity(@PrimaryKey var primaryKey: String =
             membershipStr = value.name
         }
 
+    var userPresence: UserPresenceEntity? = null
+        set(value) {
+            if (value != field) field = value
+        }
```
bmarty, line 46: I'm not sure it make sense to test `!=` for Realm object, @BillCarsonFr WDYT?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/mapper/RoomMemberSummaryMapper.kt
```diff
@@ -18,12 +18,14 @@ package org.matrix.android.sdk.internal.database.mapper
...
 import org.matrix.android.sdk.api.session.room.model.RoomMemberSummary
 import org.matrix.android.sdk.internal.database.model.RoomMemberSummaryEntity
+import org.matrix.android.sdk.internal.database.model.presence.toUserPresence
 
 internal object RoomMemberSummaryMapper {
 
     fun map(roomMemberSummaryEntity: RoomMemberSummaryEntity): RoomMemberSummary {
         return RoomMemberSummary(
                 userId = roomMemberSummaryEntity.userId,
+                userPresence = roomMemberSummaryEntity.userPresence?.toUserPresence(),
```
bmarty, line 28: `userPresence?.toUserPresence()` is a bit strange to read. Maybe rename `roomMemberSummaryEntity.userPresence` to `roomMemberSummaryEntity.userPresenceEntity`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/RoomMemberEntityQueries.kt
```diff
@@ -32,3 +33,12 @@ internal fun RoomMemberSummaryEntity.Companion.where(realm: Realm, roomId: Strin
     }
     return query
 }
+
+internal fun RoomMemberSummaryEntity.Companion.updateUserPresence(realm: Realm, userId: String, userPresenceEntity: UserPresenceEntity) {
+    realm.where<RoomMemberSummaryEntity>()
+            .equalTo(RoomMemberSummaryEntityFields.USER_ID, userId)
+            .findAll()
+            .map {
+                it.userPresence = userPresenceEntity
```
bmarty, line 42: is it necessary to do this? The link between ` it.userPresence` and `userPresenceEntity` is always the same, no?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/model/UserPresence.kt
```diff
@@ -20,4 +20,16 @@ data class UserPresence(
         val statusMessage: String? = null,
         val isCurrentlyActive: Boolean? = null,
         val presence: PresenceEnum = PresenceEnum.OFFLINE
-)
+) {
+    /**
+     * Transforms the presence to a human readable string from the given multilingual value
+     * Currently we support only online/offline (unavailable will be marked as offline)
+     */
+    fun toHumanReadable(onlineLabel: String, offlineLabel: String): String {
```
bmarty, line 28: this is weird to have this fun SDK side.
It should be better to have an Extension app side.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
```diff
@@ -47,7 +49,13 @@ internal class RoomMemberEventHandler @Inject constructor(
         if (roomMember == null) {
             return false
         }
-        val roomMemberEntity = RoomMemberEntityFactory.create(roomId, userId, roomMember)
+        val roomMemberEntity = RoomMemberEntityFactory.create(
+                roomId,
+                userId,
+                roomMember,
+                // To resolve an edge cases like, on initial sync RoomMemberEventHandler is called after PresenceSyncHandler,
```
bmarty, line 56: maybe we can change and call PresenceSyncHandler after RoomMemberEventHandler?

vector/src/main/java/im/vector/app/core/ui/views/PresenceStateImageView.kt
```diff
@@ -0,0 +1,48 @@
...
+        context: Context,
+        attrs: AttributeSet? = null,
+        defStyleAttr: Int = 0
+) : AppCompatImageView(context, attrs, defStyleAttr) {
+
+    fun render(showPresence: Boolean = true, userPresence: UserPresence?) {
+        if (showPresence) {
+            isVisible = true
+            if (userPresence?.presence == PresenceEnum.ONLINE) {
+                setImageResource(R.drawable.ic_presence_online)
```
bmarty, line 40: Please also update the content description for a11y

vector/src/main/res/layout/fragment_room_detail.xml
```diff
@@ -45,38 +45,56 @@
...
+                    app:layout_constraintStart_toEndOf="@+id/roomToolbarAvatarImageView"
+                    app:layout_constraintTop_toTopOf="@+id/roomToolbarTitleView" />
+
+
+                <im.vector.app.core.ui.views.PresenceStateImageView
+                    android:id="@+id/roomToolbarPresenceImageView"
+                    android:layout_width="12dp"
+                    android:layout_height="12dp"
+                    android:background="@drawable/background_circle"
+                    android:importantForAccessibility="no"
```
bmarty, line 62: there is no text for presence now, so we should remove that line

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/RoomMemberSummaryEntity.kt
ariskotsomitopoulos: Yeah, I believe it does not make sense, but I there are other implementations like this in` RoomSummaryEntity`:
`    var latestPreviewableEvent: TimelineEventEntity? = null
        set(value) {
            if (value != field) field = value
        }
`
`    var userDrafts: UserDraftsEntity? = null
        set(value) {
            if (value != field) field = value
        }
`
So if we decide to remove it let's remove it from the other places too, if you agree

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
ariskotsomitopoulos: Well, I deep dived into the problem about what is happening. I will also add a comment in the codebase explaining that. Let me explain, when we use `insertOrUpdate` then we have two results:

1. **Insert** : A brand new object is created with a primary key etc and is inserted in the DB
2. **Update**: A primary key is already found in that table so the framework **replace** the existing object with the new one. 

The problem is when there is an **update**! With An example will be more clear. Let's say we have the following record:
[ **MyPrimaryKey**: 123, catName: "Anna", dogName:"John" , .... etc]

and now lets say server has an update with [ **MyPrimaryKey**: 123, catName: "Ketty"]

So now if we do `insertOrUpdate` with an object containing only **MyPrimaryKey** and **catName**. All the other records will be replaced by null, so after the **insertOrUpdate** our record will be updated to:
[ **MyPrimaryKey**: 123, catName: "Anna", dogName:**NULL** , .... etc]

So to avoid this there are two solutions:

1.  Manually check if the record exists, and update only the fields we want to update
2. Pick the old value and pass it to new object creation (Faster if the select is with **@indexed** fields)

While **userId** and **roomId** are **@indexed**, I think the second one is clearer and more efficient. What do you think?

DoM1niC:
Hey Aris,
can you fix the conflicts plz :) I wanne use it in my fork.... 

THX for all your Work!!!!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/RoomMemberEntityQueries.kt
ariskotsomitopoulos: Well, I think you are partially right, we need that only for the first linking, after that it will auto update the changes, so I simply add in the query` .isNull(RoomMemberSummaryEntityFields.USER_PRESENCE.`$`)` so now it will only do that for the first linking. You think there is a better way  for the initial linking?

ariskotsomitopoulos:
> Circle of offline presence is only transparent instead of filled color
> 
> image

hmm strange, what android do you have device/version? is the same with light mode too?

DoM1niC:
> > Circle of offline presence is only transparent instead of filled color
> > image
> 
> hmm strange, what android do you have device/version? is the same with light mode too?

Android 11 on Android Studio

DoM1niC:
image

DoM1niC:
Some Users are fine and some not lol 😂

ariskotsomitopoulos:
> Some Users are fine and some not lol 😂
> 
> 

hmm I think your colors are altered, do you use any custom Theme? Maybe thats why as I can see and other colors are altered

ariskotsomitopoulos:
> Circle of offline presence is only transparent instead of filled color
> 
> image

Can you check again? It seems like I had to disable it in servers that do not support presence. So now I guess it will work!

Issue #3833, opened 2021-08-14
[Bug|Crash] Voice Messages sending need more robustfully


DoM1niC:
Hey Element Devs,

The Issue cause by  low or unstable internet connectivity, the Voice Message temporarily dismiss and try to resend but the tmp file is broken or seems to have invalid data, the 2nd problem after some retries the App crash instantly by opening this Room again. 

**Code:** Develop
**Server:** Synapse last Stable 1.43.0 (messenger.3dns.eu)

https://user-images.githubusercontent.com/1436958/129432354-cc1987a1-02e4-4e38-a340-4cea24f6ceea.mp4

DoM1niC:
Fixed by #4527

PR #4105, opened 2021-09-28
Fix crash to render failed message warning stub

### LINKED ISSUE #4110: Crash while rendering failed message warning
### Steps to reproduce

- Enable flight mode
- Send a message (easy to repro with media messages)
- Restart app
- Open the room which has failed messages

### What did you expect?
To see the timeline with failed messages warning

### What happened?
App crashed.

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

No

onurays:
Fixes #4110

### DIFF
--- a/changelog.d/3833.bugfix
+++ b/changelog.d/3833.bugfix
@@ -1,0 +1,1 @@
+[Bug|Crash] Voice Messages sending need more robustfully
--- a/changelog.d/3833.bugfix
+++ b/changelog.d/3833.bugfix
@@ -1,1 +1,0 @@
-[Bug|Crash] Voice Messages sending need more robustfully
--- a/vector/src/main/res/layout/fragment_room_detail.xml
+++ b/vector/src/main/res/layout/fragment_room_detail.xml
@@ -98,7 +98,7 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         app:barrierDirection="top"
-        app:constraint_referenced_ids="composerLayout,notificationAreaView,failedMessagesWarningView" />
+        app:constraint_referenced_ids="composerLayout,notificationAreaView,failedMessagesWarningStub" />
 
     <im.vector.app.features.sync.widget.SyncStateView
         android:id="@+id/syncStateView"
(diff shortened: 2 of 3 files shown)


changelog.d/3833.bugfix
```diff
@@ -0,0 +1 @@
+[Bug|Crash] Voice Messages sending need more robustfully
```
bmarty, line 1: This is not really related to voice message, and/or issue 3833.

ganfra:
> Fixes #3833

I don't see how it's fixing 3833?
Your PR is fixing a crash introduced after the issue was created

vector/src/main/res/layout/fragment_room_detail.xml
```diff
@@ -98,7 +98,7 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         app:barrierDirection="top"
-        app:constraint_referenced_ids="composerLayout,notificationAreaView,failedMessagesWarningView" />
+        app:constraint_referenced_ids="composerLayout,notificationAreaView,failedMessagesWarningStub" />
```
ganfra, line 101: I think the other Barrier needs to update too

ganfra (changes requested):
See my remarks

bmarty (changes requested):
Thanks for the fix, crash reproduced and fix tested OK, but the changelog file is not correct.

Can you update it please?
It should be file `4105.bugfix` with content "Fix a crash (regression in 1.3.0) in the timeline when a message has failed to be sent"

vector/src/main/res/layout/fragment_room_detail.xml
bmarty: It's already `failedMessagesWarningStub`, do you confirm it's wrong?
https://github.com/vector-im/element-android/blob/develop/vector/src/main/res/layout/fragment_room_detail.xml#L207

onurays: Yes, it is.

bmarty: it is... wrong?

bmarty (approved):

PR #4115, opened 2021-09-29
Skipping passphrase screen when no passphrase is available

### LINKED ISSUE #3898: Wrong navigation after login
### Steps to reproduce

- Login to an existing account
- Click on the banner to verify the session
- Click on "Use a Recovery Passphrase or Key

### What did you expect?

I should see the passphrase form first, which include an option to open the Recovery Key form.

### What happened?

Recovery Key form is displayed first and if I press back I see the passphrase form

### Your phone model

Samsung S9

### Operating system version

Android 10

1.2.0

No

ouchadam:
Fixes #3898

When an key info response doesn't contain a passphrase the login verification flow would still attempt to visit the passphrase screen when _going back_

Fixed by simply checking if a passphrase is available when _going back_ and dismissing if no passphrase is available

- Adds a suite of tests around the `SharedSecureStorageViewModelTest`, these tests also found an issue where we're posting `ViewEvents` as part of the `ViewModel` init flow, this is problematic because we're unable to register an observer before `init` is called (will raise a separate issue)

| BEFORE | AFTER |
| --- | --- |
|before-no-passphrase|after-no-passphrase

** logging into an account which doesn't have a passphrase set

### DIFF
--- a/vector/src/main/java/im/vector/app/features/crypto/quads/SharedSecureStorageViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/crypto/quads/SharedSecureStorageViewModel.kt
@@ -76,7 +76,6 @@ class SharedSecureStorageViewModel @AssistedInject constructor(
     }
 
     init {
-
         setState {
             copy(userId = session.myUserId)
         }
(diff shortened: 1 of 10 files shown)


vector/src/main/java/im/vector/app/features/crypto/quads/SharedSecureStorageViewModel.kt
```diff
@@ -167,10 +166,14 @@ class SharedSecureStorageViewModel @AssistedInject constructor(
         if (state.checkingSSSSAction is Loading) return@withState // ignore
         when (state.step) {
             SharedSecureStorageViewState.Step.EnterKey -> {
-                setState {
-                    copy(
-                            step = SharedSecureStorageViewState.Step.EnterPassphrase
-                    )
+                if (state.hasPassphrase) {
```
ouchadam, line 169: this block is the fix, everything else in this PR is for the unit tests

bmarty: perfect, thanks!

bmarty (approved):
LGTM, thanks for improving / setting up a better test environment.

Issue #523, opened 2019-08-27
Drop notification when messages are read on another device


jeschmidt:
I searched the database for duplicates before posting this but regardless, my apologies if it is duplicate.  I use multiple Matrix clients, one of which is Riot/Android and I just tried installing RiotX/Android.  The first thing I noticed is that Riot notifications will disappear when I read the messages on another device (like Riot/Desktop), but for RiotX they persist.  To be clear, the original Riot/Android behaviour is what I recommend.  Unread messages should trigger the notification, and when they are all marked as read (including via other devices) the notification should automatically clear itself.

bmarty:
Yes, I agree with that.
The limit is that you can hear your phone ring for a notification and when you check it the notification is gone, which can be a bit disturbing I you don't realized it's because you've read the message on another client.

But anyway let's do this as it's done on Riot-Android

hpdeifel:
It seems to work for some users:

![Screenshot_20200217-215301](https://user-images.githubusercontent.com/77743/74685900-3146ea80-51d0-11ea-841c-6a0c6478a76d.png)

bmarty:
Fix by #4129 , closed manually since GitHub grammar is not clever enough to fully understand #4129 description :)

In Element Android 1.3.2

PR #4129, opened 2021-09-30
Fixing notifications not being dismissed when read from other devices

### LINKED ISSUE #3347: Notification does not vanish if the message is read on another session
- Bob has 2 sessions: one on Element Android, one on Element Web, and share a room with Alice.
- Bob is viewing the room list on Element Android
- Alice sends a message to Bob in room A
- The notification appears on Bob's Android device.
- Bob read the room in his Element Web session
- The notification is still visible to Bob's Android device -> KO

The notification should be cancel on Bob's Android device.

It used to work, so this is regression.

During the while process Element Android is in foreground (so is syncing permanently). If Element Android app goes to background, I'm not sure we will receive a Push to update the read marker of Bob, but it is acceptable in this case.

ouchadam:
Fixes #3347 and #523 

Our notification dismissing works by matching the currently held notification events to the latest timeline chunk for the room, this presents a problem when the next chunk is created which doesn't reference those original notification events, which then presents itself as notifications not being dismissed when read from other sessions.

The fix is to check the existence of missing events against the entire chunk history if there's also a read receipt event in the latest chunk.

- I was able to to reproduce the issue by spamming messages to force the timeline to create a new chunk

Here's the worlds least exciting gif... notice the element icon in the top
| |
| --- |
|after-new-chunk

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt
@@ -78,7 +78,6 @@ private fun Realm.hasReadReceiptInLatestChunk(latestChunkEntity: ChunkEntity, ro
     } != null
 }
 
-
 internal fun isReadMarkerMoreRecent(realmConfiguration: RealmConfiguration,
                                     roomId: String?,
                                     eventId: String?): Boolean {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt
@@ -23,9 +23,6 @@ import org.matrix.android.sdk.internal.database.model.ReadMarkerEntity
 import org.matrix.android.sdk.internal.database.model.ReadReceiptEntity
 import org.matrix.android.sdk.internal.database.model.TimelineEventEntity
 
-private const val MARK_OLD_EVENT_AS_READ = true
-private const val MARK_UNREAD_DUE_TO_FASTLANE = false
-
 internal fun isEventRead(realmConfiguration: RealmConfiguration,
                          userId: String?,
                          roomId: String?,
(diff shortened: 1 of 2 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt
```diff
@@ -23,6 +23,9 @@ import org.matrix.android.sdk.internal.database.model.TimelineEventEntity
 import io.realm.Realm
 import io.realm.RealmConfiguration
 
+private const val MARK_OLD_EVENT_AS_READ = true
+private const val MARK_UNREAD_DUE_TO_FASTLANE = false
```
bmarty, line 27: I'm not sure declaring const for boolean returned value increases the readability of the code. Maybe a good comment is better?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt
```diff
@@ -59,6 +59,25 @@ internal fun isEventRead(realmConfiguration: RealmConfiguration,
...
+        MARK_UNREAD_DUE_TO_FASTLANE
+    }
+}
+
+private fun Realm.doesEventExistInChunkHistory(eventId: String): Boolean {
+    return ChunkEntity.findIncludingEvent(this, eventId) != null
+}
+
+private fun Realm.hasReadReceiptInLatestChunk(latestChunkEntity: ChunkEntity, roomId: String, userId: String) : Boolean {
+    return ReadReceiptEntity.where(this, roomId, userId).findFirst()?.let {
```
bmarty, line 76: Nit: The code is OK, but maybe use named arguments for roomId and userId to avoid confusion, since the types are the same (String)

bmarty (approved):
That a really good catch, thanks!
I have not tested it, but I understand the idea.
See my 2 small remarks.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt
ouchadam: :100: will do

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt
ouchadam: :+1: I think this is a sign that the `handleMissingEvent` is the wrong name, renaming to `hasReadMissingEvent` allows us to infer the return result is `has read` or `has not read`

ouchadam: 4a51583a7608e8a83add318332d0f3cbab244595

bmarty (approved):
Thanks for the update! It's more readable to me now!

PR #4132, opened 2021-10-01
Ensure initial sync progress dialog is hidden when the initial sync is over

bmarty:
Add robustness, if Idle state is lost for any reason, the dialog will be hidden also on incr sync Events.

It does not fix the pb with foreground service notification not displayed.


ouchadam (approved):

bmarty:
Ref: #983

Issue #4109, opened 2021-09-29
Inactive "Open settings" button


bmarty:
- disable notification for the app in the system setting
<img width="315" alt="image" src="https://user-images.githubusercontent.com/3940906/135219681-4a1ded70-a6af-4024-8573-62b337b80a92.png">

- run the in app notification troubleshoot
- This screen is displayed:
<img width="323" alt="image" src="https://user-images.githubusercontent.com/3940906/135219822-e947c825-995a-4239-adea-77e5b9903bef.png">
and at the bottom:
<img width="328" alt="image" src="https://user-images.githubusercontent.com/3940906/135219877-90856be8-d7ec-4612-919b-f41f815a2d9f.png">

- The first "OPEN SETTINGS" button is not working since [we wait for the diagnostic to be over](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/settings/troubleshoot/TestSystemSettings.kt#L43), which is not possible because of the last test waiting to be passed, which is not possible since the notification can't be displayed.

The line [`if (manager?.diagStatus == TestStatus.RUNNING) return // wait before all is finished`](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/settings/troubleshoot/TestSystemSettings.kt#L43) is a copy paste from another test and can be removed

bmarty:
Fixed in EA 1.3.2

PR #4154, opened 2021-10-04
Troubleshoot notification: Fix button not clickable

bmarty:
See details in #4109


vector/src/main/java/im/vector/app/features/settings/troubleshoot/TestSystemSettings.kt
```diff
@@ -40,7 +40,6 @@ class TestSystemSettings @Inject constructor(private val context: Context,
             description = stringProvider.getString(R.string.settings_troubleshoot_test_system_settings_failed)
             quickFix = object : TroubleshootQuickFix(R.string.open_settings) {
                 override fun doFix() {
-                    if (manager?.diagStatus == TestStatus.RUNNING) return // wait before all is finished
```
ouchadam, line 43: ah, in this case the `doFix` is the click listener for the button? (which won't be shown until the correct step)

ouchadam (approved):

vector/src/main/java/im/vector/app/features/settings/troubleshoot/TestSystemSettings.kt
bmarty: > ah, in this case the doFix is the click listener for the button?

yes

Issue #2893, opened 2021-02-21
When creating a DM with someone you already have a DM with, element should give a warning or automatically switch to the preexisting DM


michaelharmonart:
In a large number of chat apps (and the one I'm specifically thinking of, hangouts)  when the new conversation button is used, and you search and "create" a conversation with someone you already have a conversation with, it simply pulls up that conversation. Element should do something similar, or at least give a warning that if you continue, you'll be creating a new conversation/room and inviting the recipient to it. Currently, if you accidentally create a new DM with someone, there is no way to withdraw the invitation or cancel/delete the duplicate thread (at least that I could find).
All of this is just very annoying in practice if you don't know the way element handles rooms.

Ideal Solution: Automatically open the correct DM when the new conversation button is used to create one with someone you already have a DM with.

Otherwise, at least display some sort of warning that if the user continues, they'll be creating and inviting the recipient to a new room.

jwsp1:
it's on their roadmap https://github.com/vector-im/roadmap/issues/19

bmarty:
It looks like a regression, it should not create a new DM, but open the existing one. We will check

chagai95:
@bmarty did you get around to checking? 😉

strypey:
It would be equally useful if the app also avoided cloning private rooms with the same set of users. I presume this is a tricky problem, because even well-funded proprietary apps like WeChat don't do it yet. The UX would look like:
* I have a private room with 3 users in it
* I want to post a message in that room but I've lost track of where it is in the UI
* I try to find it by going to a DM with one of the other 2 users and inviting the third one it
* The app identifies that I already have a room with that set of users and navigates to it, and *does not* bring the third person into our DM, or create a new room with the 3 of us in it.

chagai95:
@strypey very valid points I agree with all of them, could you create separate tickets for each one?

I will then comment on those tickets 🙃

chagai95:
@bmarty just wondering if this is really a regression like you mentioned half a year ago...

chagai95:
@strypey let's have this discussion on matrix, maybe in the HQ matrix? Otherwise you can write me a pm @me:chagai.website 

If I understand correctly the way the Element team works we would need a ticket for every platform and since web is the most active I would start there... In general don't be afraid of opening tickets, worst case scenario they will just be closed 😉

chagai95:
I can confirm this is fixed.

PR #4157, opened 2021-10-04
Load existing DM instead of creating a new one

Florian14:
In previous implementation, existing DM was checked only for QR code invites.
This PR moves this check in the ViewModel side (which makes more sense) and applies it for 1:1 in both cases (QR and user invites).

### DIFF
--- a/vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomViewModel.kt
@@ -71,10 +71,13 @@ class CreateDirectRoomViewModel @AssistedInject constructor(@Assisted
      * If users already have a DM room then navigate to it instead of creating a new room.
      */
     private fun onSubmitInvitees(action: CreateDirectRoomAction.CreateRoomAndInviteSelectedUsers) {
-        if (action.existingDmRoomId != null) {
+        val existingRoomId = action.selections.singleOrNull()?.getMxId()?.let { userId ->
+            session.getExistingDirectRoomWithUser(userId)
+        }
+        if (existingRoomId != null) {
             // Do not create a new DM, just tell that the creation is successful by passing the existing roomId
             setState {
-                copy(createAndInviteState = Success(action.existingDmRoomId))
+                copy(createAndInviteState = Success(existingRoomId))
             }
         } else {
             // Create the DM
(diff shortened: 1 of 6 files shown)


bmarty (approved):
LGTM

bmarty:
@Florian14 do you thing this can close #2893?

Florian14:
> @Florian14 do you thing this can close #2893?

@bmarty I think so, just tested before and after my PR, work fine now.

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomViewModel.kt
```diff
@@ -71,10 +71,13 @@ class CreateDirectRoomViewModel @AssistedInject constructor(@Assisted
      * If users already have a DM room then navigate to it instead of creating a new room.
      */
     private fun onSubmitInvitees(action: CreateDirectRoomAction.CreateRoomAndInviteSelectedUsers) {
-        if (action.existingDmRoomId != null) {
+        val existingRoomId = action.selections.singleOrNull()?.getMxId()?.let { userId ->
```
BillCarsonFr, line 74: This does not look right. If it's a request for a grouped DM this will just open the DM with the first user?

vector/src/main/java/im/vector/app/features/userdirectory/PendingSelection.kt
```diff
@@ -29,4 +29,11 @@ sealed class PendingSelection {
             is ThreePidPendingSelection -> threePid.value
         }
     }
+
+    fun getMxId(): String {
+        return when (this) {
+            is UserPendingSelection     -> user.userId
+            is ThreePidPendingSelection -> threePid.value
```
BillCarsonFr, line 36: this would be a mail or Pn but not a mxid

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomViewModel.kt
bmarty: singleOrNull() ≠ firstOrNull()

vector/src/main/java/im/vector/app/features/userdirectory/PendingSelection.kt
bmarty: Yes, but I think it will work, until the other user has not joined the room. @Florian14 do you confirm?

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomViewModel.kt
Florian14: Indeed, agree to check for existing grouped DM, requires some more changes

Issue #3904, opened 2021-08-26
Cannot record a voice message


532910:
### Steps to reproduce

I dunno how to reproduce this.
https://user-images.githubusercontent.com/14617699/130969769-e454062e-1672-4c30-bb42-7bc612eeea82.mp4

All permissions are granted and voice/video calls works as expected.

### What did you expect?

A recorded voice message.

### What happened?

Unable to record a voice message.

### Have you submitted a rageshake?

Yes

532910:
@takiainen, @teldra, @bigCrash could you create a screenshot for more visibility?

bigCrash:
Your screen video is pretty on point. I do not know anything to add with a screenshot.

takiainen:
Exactly, the video above already shows everything there is to see. Mine behaves just like that.

bmarty:
The MediaRecorder fails to initialize. Can you send a rageshake just after reproducing the issue please? You can mention this issue number (#3904) in the bug report description.

532910:
The rageshake was sent before this issue,
I can give you matrix account ID which it was sent from.

vsatmydynipnet:
+ 1 LOS 18.1 - 1.2.2 F-Droid; all permissions set

lfom:
Same problem here, v1.3.9 from GitHub, Android 12 GSI with MicroG, rageshake reports sent.
App crashes when recording is locked (no waveform shown) and the message field is tapped.

lfom:
@bmarty and other devs
I got some logcats, hopefully it can hel fixing it soon.

Command to grab log:
`adb logcat | grep -i -e element -e vector -e media`

Cannot record a voice message
```
12-07 22:22:07.523  1095  1949 E MediaProfiles: The given audio encoder 7 is not found
12-07 22:22:07.523  1095  1949 I chatty  : uid=1013(media) Binder:1095_2 identical 4 lines
12-07 22:22:07.523  1095  1949 E MediaProfiles: The given audio encoder 7 is not found
12-07 22:22:07.549 23665 23665 E MediaRecorder: start failed: -2147483648
12-07 22:22:08.708 23665 23665 W im.vector.app: type=1400 audit(0.0:130462): avc: denied { search } for name="/" dev="cgroup2" ino=1 scontext=u:r:untrusted_app:s0:c104,c257,c512,c768 tcontext=u:object_r:cgroup_bpf:s0 tclass=dir permissive=0 app=im.vector.app
```


[...]

PR #4165, opened 2021-10-05
Add a log to get info when an error occurred when recording / playing voice message

bmarty:
Will help debugging issues like #3904

No changelog required for a such small change

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
@@ -60,6 +60,7 @@ class VoiceMessageHelper @Inject constructor(
         try {
             voiceRecorder.startRecord()
         } catch (failure: Throwable) {
+            Timber.e(failure, "Unable to start recording")
             throw VoiceFailure.UnableToRecord(failure)
         }
         startRecordingAmplitudes()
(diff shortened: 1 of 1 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
```diff
@@ -60,6 +60,7 @@ class VoiceMessageHelper @Inject constructor(
         try {
             voiceRecorder.startRecord()
         } catch (failure: Throwable) {
+            Timber.e(failure, "Unable to start recording")
```
ouchadam, line 63: do these logs end up in the rageshakes?

ouchadam (approved):

vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
bmarty: Yes, the Tree is planted [here](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/VectorApplication.kt#L133)

ouchadam: good to know, thanks!

PR #4182, opened 2021-10-07
Immutable NotifiableEvent models

ouchadam:
A little bit of clean up to attempt to reduce the notification redesign PR size...

Updates the `NotifiableEvent` models to favour immutable state, this allows us to the simplify the mutation flows 

- Gives the `NotifiableEventResolver` full control over the creation of the events (when I get time I'll revisit and add tests)
- `isPushGatewayEvent` -> `canBeReplaced` 
- `isNoisy` is calculated at the point of creation rather than appended later on
- Ensures events can't mutate whilst we're iterating over them*
- Removes unused `handleNotificationWithoutSyncingMode` and `lockScreenVisibility` 

Notice how the summary sneaks through in the reading gif, this will also be fixed by the redesign.
| RECEIVING | READING |
| --- | --- |
|after-notification-messages|after-reading-messages

*this doesn't include updating `hasBeenDisplayed` as this will be removed entirely 
*against a feature branch

### DIFF
--- a/changelog.d/582.feature
+++ b/changelog.d/582.feature
@@ -1,0 +1,1 @@
+Adding the room name to the invitation notification (if the room summary is available)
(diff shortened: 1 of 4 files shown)


ouchadam:
@bmarty thanks! this PR is probably the lowest risk compared to the big creation/display split and the diffing changes

I could create a separate PR will all the changes merged purely for testing if you think that would help?

PR #4186, opened 2021-10-08
Device dehydration and rehydration

onurays:


bmarty:
Will be handled later. We keep the branch.

Issue #4007, opened 2021-09-13
Crash on Logout


yostyle:
### Steps to reproduce

In debug mode start the app. Login on matrix.org
After the initial sync, click on logout button.

### What happened?

Sometimes Element app crashed on logout

The stacktrace : 
2021-09-13 15:38:24.908 9369-1929/im.vector.app.debug E/DefaultGetHomeServerCapabilitiesTask: Exception when executing request
2021-09-13 15:38:24.912 9369-1929/im.vector.app.debug E/DefaultGetWellknownTask: Exception when executing request
2021-09-13 15:38:24.922 9369-9369/im.vector.app.debug E/CleanupSession: All realm instance for session has not been closed (3)
2021-09-13 15:38:24.922 9369-28634/im.vector.app.debug E/DefaultDownloadKeysForUsers: Exception when executing request
2021-09-13 15:38:24.923 9369-28634/im.vector.app.debug E/DeviceListManager: ## CRYPTO | doKeyDownloadForUsers(): error
    java.util.concurrent.CancellationException: Closing crypto module
        at org.matrix.android.sdk.internal.crypto.DefaultCryptoService$close$1.invokeSuspend(DefaultCryptoService.kt:391)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
        at android.os.Looper.loop(Looper.java:237)
        at android.os.HandlerThread.run(HandlerThread.java:67)
2021-09-13 15:38:24.923 9369-28634/im.vector.app.debug E/DeviceListManager: ## CRYPTO | refreshOutdatedDeviceLists() : ERROR updating device keys for users [@*****:matrix.org]
    java.util.concurrent.CancellationException: Closing crypto module
        at org.matrix.android.sdk.internal.crypto.DefaultCryptoService$close$1.invokeSuspend(DefaultCryptoService.kt:391)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at android.os.Handler.handleCallback(Handler.java:883)
        at android.os.Handler.dispatchMessage(Handler.java:100)
[...]

PR #4191, opened 2021-10-08
Feature/aris/issue 4007 logout crash

ariskotsomitopoulos:
The crash is really strange maybe its a problem within the library, I tried a lot of different approaches to overcome this. The only thing that worked was removing `.close() ` from realm (I don't like it, it's the only thing worked). 

Tested around 25 times without a crash, more tests would be welcome.

original issue [here](https://github.com/vector-im/element-android/issues/4007)


bmarty:
Following our discussion, I think I have a better fix for this issue #4007. Will open a PR.

PR #4216, opened 2021-10-11
Feature/aris/issue_3902_add_scripts_to_ci

### LINKED ISSUE #3902: Let GitHub works for us and create PR for changes which comes after running internal scripts
### Your use case

During the release we run some scripts to update the codebase from external resources, but it has to be done manually.

The [two steps](https://github.com/vector-im/element-android/blob/main/.github/ISSUE_TEMPLATE/release.md#L21) (so far) are:
- Run `./tools/import_emojis.py` and commit the change if any.
- Run `./tools/import_sas_strings.py` and commit the change if any.

It should be better to set up a GitHub cron or something similar to let GitHub run the script and create PR in case of any changes.

Maybe one cron for each script, hence the changes are not frequent, so maybe only one generic cron can be enough, but commit will have to be done after each script.

No

### Additional context

Ideally, if any PR exists and has been created by the cron, no new PR should be created with the same change. If there are new change, existing PR should be deleted and replaced by the new one. A bit similar to what dependabot is doing to upgrade the gradle dependencies.

Once this is done, we could remove the 2 steps from the release template.

ariskotsomitopoulos:
Fixes #3902

Implement a new github workflow to run :
```
./tools/import_emojis.py
./tools/import_sas_strings.py
```
every Monday at 00:00 UTC and create two PRs:

1. Sync Emojis
2. Sync SAS Strings

How the automated cron action behaves:

- If there are changes (i.e. a diff exists with the checked-out base branch), the changes will be pushed to a new branch and a pull request created.

- If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.

- If a pull request already exists and there are no further changes (i.e. no diff with the current pull request branch) then the action exits silently.

- **If a pull request exists and new changes on the base branch make the pull request unnecessary (i.e. there is no longer a diff between the pull request branch and the base), the pull request is automatically closed. Additionally, if delete-branch is set to true the branch will be deleted.**

_Further workflows(Like Unit tests, Builds APK etc) on the generated PRs are disabled, we can enable them with the appropriate configuration if needed_


ariskotsomitopoulos:
> I agree with @ouchadam , every PRs should be checked by the regular CI. Looking at #4217, this is not the case, only Buildkite has been run... But this is probably fine for what the 2 tools are doing. Anyway we will still review the change before merging such PR

I agree with both approaches! If we want all tests to run I can enhance it anytime, no worries

bmarty:
> Thanks for the PR.
> 
> Can I ask you to remove the two lines referencing the 2 scripts [in this file](https://github.com/vector-im/element-android/blob/develop/.github/ISSUE_TEMPLATE/release.yml)?

It has been done directly on the develop branch IFAIU in d26340993f0b34df94d77b4c31a18d8113d736a2. It could have been part of this PR.

Issue #4214, opened 2021-10-11
[ANR] NotificationDrawerManager.setCurrentRoom 


ouchadam:
This looks like the main thread is stuck waiting for the the synchronised `eventList` to unlock

```
  at im.vector.app.features.notifications.NotificationDrawerManager.setCurrentRoom (NotificationDrawerManager.kt:1)
  at im.vector.app.features.home.room.detail.RoomDetailFragment.onPause (RoomDetailFragment.kt:2)
  at androidx.fragment.app.Fragment.performPause (Fragment.java:8)
  at androidx.fragment.app.FragmentStateManager.pause (FragmentStateManager.java:3)
  at androidx.fragment.app.FragmentStateManager.moveToExpectedState (FragmentStateManager.java:30)
  at androidx.fragment.app.FragmentManager.moveToState (FragmentManager.java:73)
  at androidx.fragment.app.FragmentManager.dispatchStateChange (FragmentManager.java:5)
  at androidx.fragment.app.FragmentActivity.onPause (FragmentActivity.java:5)
  at im.vector.app.core.platform.VectorBaseActivity.onPause (VectorBaseActivity.kt:1)
  at android.app.Activity.performPause (Activity.java:8294)
  at android.app.Instrumentation.callActivityOnPause (Instrumentation.java:1510)
  at android.app.ActivityThread.performPauseActivityIfNeeded (ActivityThread.java:4839)
  at android.app.ActivityThread.performPauseActivity (ActivityThread.java:4800)
  at android.app.ActivityThread.handlePauseActivity (ActivityThread.java:4748)
  at android.app.servertransaction.PauseActivityItem.execute (PauseActivityItem.java:46)
  at android.app.servertransaction.TransactionExecutor.executeLifecycleState (TransactionExecutor.java:176)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:97)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2142)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:236)
  at android.app.ActivityThread.main (ActivityThread.java:7864)
```

ouchadam:
was fixed by #4235

PR #4220, opened 2021-10-11
Notification diffing

ouchadam:
- Fixes notification related ANR  #4214 
  - Splits the event list processing from the notification rendering, this allows the synchronised block to lock for less time (only whilst we process the events) 
  
- Fixes the serialised event list becoming out of sync with the currently rendered notifications, eg receiving multiple room invites
  - Adds `event list` and `currently rendered events` diffing, this allows us to remove out of sync notifications  

| BEFORE | AFTER | 
| --- | --- |
before-accept-multiple|after-multiple-joins

### DIFF
--- a/vector/src/main/java/im/vector/app/features/notifications/NotificationFactory.kt
+++ b/vector/src/main/java/im/vector/app/features/notifications/NotificationFactory.kt
@@ -25,16 +25,21 @@ class NotificationFactory @Inject constructor(
         private val summaryGroupMessageCreator: SummaryGroupMessageCreator
 ) {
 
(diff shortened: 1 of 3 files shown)


bmarty:
There is a lot PR to be reviewed, I'm wondering if it's worth creating sub PR like that. Also I'm afraid I will have to review the same code when reviewing the PR which want to be merged on develop.
Maybe a big single PR with all the changes regarding the notification management would be better, to be revied and tested only once, but I'm not sure.
WDYT?

ouchadam:
closing to avoid ordering confusion

Issue #4152, opened 2021-10-03
Doubled notifications


waclaw66:
### Steps to reproduce

Just receive a message.

https://user-images.githubusercontent.com/15554561/135765548-a0284d91-fc22-40b4-a986-020245f7be76.mp4

### What did you expect?
Single notification for a message.

### What happened?
When single message is received two same notifications are shown for a little while in a row.
I tried to re-register pusher, create a new session, optimized init sync. Nothing helped.

### Your phone model

Xiaomi Mi 9T

### Operating system version

Android 11

### Application version and app store

1.3.1

### Homeserver

own

### Have you submitted a rageshake?

Yes

ouchadam:
Hi @waclaw66 thanks for raising this, unfortunately I don't have a android 11 MIUI device to test on and I'm not familiar with their notification UI changes, would you be able to provide some more information?

- do you have the same issue when using a matrix.org account
- has this issue recently started happening, if so was it tied to an application update/homeserver upgrade
- do the messages stack as list with notification actions
- "notifications are shown for a little while in a row." does this mean the duplicate notification disappears 

| MIUI 12.0.3/Android 10 | Stock~ (Samsung One UI) Android 11 | 
| --- | --- |
![2021-10-04T13:01:44,528885888+01:00](https://user-images.githubusercontent.com/1848238/135847982-e7b98777-d8c4-410e-b388-7615ce93284b.png)|![2021-10-04T13:07:28,372840475+01:00](https://user-images.githubusercontent.com/1848238/135848668-3dbb6340-b276-418a-a939-ac194bf818b4.png)

waclaw66:
@ouchadam:
some feedback...
- yes it does also using matrix.org account
- it happens a while, now I've tried version 1.2.2, it behaves the same
- messages stack same as on your screenshot, but for two messages there is count (4)
- yes, I see only single notification for a message, but it does a sound and counts twice

ouchadam:
the video really helps clarify, thanks!

It's interesting to see the final expanded notification is 

```
test 1 
test 2
```

rather than 

```
test 1
test 1
test 2 
test 2
``` 

I'll keep investigating and see if I can reproduce on my side and hopefully fix

ouchadam:
one last thing to confirm, does the issue happen for both notification styles? Settings -> Notifications -> Notifications Shade -> android / MIUI

[...]

PR #4235, opened 2021-10-13
Splitting notifications into separate creation and displaying passes

ouchadam:
Previously opened as #4185 and #4220 

This entire change is the result of extracting, breaking down and unit testing 1 function [NotificationDrawerManager.refreshNotificationDrawerBg](https://github.com/vector-im/element-android/blob/b71e2de9acf611d4a50703aed458d1e022d2b105/vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt#L226)

There are 3 main objectives/changes to the notification design

1. The notification flow becomes a 2 pass system. All notifications are created and then all notifications are notified/cancelled, instead of looping through all the events notifying/cancelling as we go. 

- This has the benefit of reducing the time between posting individual message and summary group notification changes to the 
system UI message queue, which in turn should address our double notification problem :crossed_fingers: #4152 

2. The `queueEvents`  _(previously called `eventList`)_ is the sole source of truth, we now simply render it as android notifications, when a notification is dismissed it triggers a delete action which causes the event to be reflected in the `eventList`,  which in turn renders the newly missing event as a notification removal. 

-  This avoids redundant notification `cancel -> show` flows, we only show or cancel event changes

3. Testable, immutable and simpler to reason, we create all the notifications and then display all the notifications 

- Creates small classes with single responsibilities, tested where possible. The notification creation helpers `RoomGroupMessageCreator` & `SummaryGroupMessageCreator` are not particularly friendly to unit tests due to calling into android builders (although we could abstract that and test...)

-----

- Tentatively fixes the double notifications #4152 

- Fixes notification related ANR  #4214 
  - Splits the event list processing from the notification rendering, this allows the synchronised block to lock for less time (only whilst we process the events) 
  
- Fixes the serialised event list becoming out of sync with the currently rendered notifications, eg receiving multiple room invites
  - Adds `event list` and `currently rendered events` diffing, this allows us to remove out of sync notifications  

| BEFORE JONING ROOM EVENT | AFTER JOINING ROOMS EVENT | 
| --- | --- |
before-accept-multiple|after-multiple-joins

| INCOMING MESSAGES | ROOM INVITE | ACCEPTING / READING |
| --- | --- | --- |
after-incoming-messages|after-room-invite|after-accepting-reading

### DIFF
--- a/vector/src/main/java/im/vector/app/features/notifications/NotifiableEventProcessor.kt
+++ b/vector/src/main/java/im/vector/app/features/notifications/NotifiableEventProcessor.kt
@@ -28,8 +28,8 @@ class NotifiableEventProcessor @Inject constructor(
         private val autoAcceptInvites: AutoAcceptInvites
 ) {
 
-    fun process(eventList: List<NotifiableEvent>, currentRoomId: String?, renderedEventsList: ProcessedEvents): ProcessedEvents {
-        val processedEventList = eventList.map {
+    fun process(queuedEvents: List<NotifiableEvent>, currentRoomId: String?, renderedEvents: ProcessedEvents): ProcessedEvents {
+        val processedEvents = queuedEvents.map {
             val type = when (it) {
                 is InviteNotifiableEvent  -> if (autoAcceptInvites.hideInvites) REMOVE else KEEP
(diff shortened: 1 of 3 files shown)


bmarty:
This is OK to me to merge this PR into the branch `feature/adm/notification-redesign`, but just to confirm, will you create a PR to merge `feature/adm/notification-redesign` into `develop`?
This will be my opportunity to test the change (I did not test it yet 🙈), and have a final (quick) review

PR #4237, opened 2021-10-13
Voice Message - Draft support

### LINKED ISSUE #3922: Voice messages: Persist drafts of voice messages when navigating between rooms
- Leaving a room stops recording but saves the draft.
- On going to any other room, the user can send messages as normal. But on returning to the room where they were recording a message, the draft will remain in place ready to be sent.
- The draft should look and behave the same as an unsent voice message if I had stopped recording but stayed in the room. 
- Screenshot 2021-08-27 at 09 55 58
- I cannot edit or continue recording the voice message at this stage.
- The only choices are delete, play/preview or send
- It is also not possible to send another message in this room until you have either deleted or sent the draft voice message.
- It is possible to have multiple draft voice messages in multiple rooms. The draft only appears when I go to the relevant room.

onurays:
If a user leaves the room detail screen without sending or deleting the ongoing voice message record then we need to save this record as a draft message and render it again in playback mode when the user enters the room detail screen again.

- [x] Extend the draft service to support multiple message types
- [x] Listen onPause of the RoomDetailFragment and save the VM draft if the recording is active
- [x] Show voice message in playback mode
- [x] Initialize voice message view with the saved draft
- [x] Delete the cache file if the user decides to delete or send the recording 

EDIT
Fixes #3922


vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -1003,11 +1019,20 @@ class RoomDetailFragment @Inject constructor(
                 .show()
     }
 
-    private fun renderRegularMode(text: String) {
-        autoCompleter.exitSpecialMode()
-        views.composerLayout.collapse()
-        views.composerLayout.setTextIfDifferent(text)
-        views.composerLayout.views.sendButton.contentDescription = getString(R.string.send)
+    private fun renderRegularMode(content: String, messageType: String) {
```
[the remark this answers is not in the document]
onurays: Done.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/RealmSessionStoreMigration.kt
```diff
@@ -381,6 +383,13 @@ internal class RealmSessionStoreMigration @Inject constructor(
 
     private fun migrateTo19(realm: DynamicRealm) {
         Timber.d("Step 18 -> 19")
+        realm.schema.get("DraftEntity")
+                ?.addField(DraftEntityFields.MESSAGE_TYPE, String::class.java)
+                ?.setRequired(DraftEntityFields.MESSAGE_TYPE, true)
+                ?.transform {
+                    it.setString(DraftEntityFields.MESSAGE_TYPE, MessageType.MSGTYPE_TEXT)
+                }
+
```
bmarty, line 392: This will be a problem for those who are already on version 19. In this case you should create a version 20.

bmarty (changes requested):
I'm sorry I think this has to be modified.
I would like the voice messages to be saved as draft by the SDK, using `DraftService`.
It will simplify the code I think and make it more robust, by following our current architecture.
What do you think about it?

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt
```diff
@@ -657,6 +665,16 @@ class RoomDetailViewModel @AssistedInject constructor(
         voiceMessageHelper.stopAllVoiceActions(deleteRecord)
     }
 
+    private fun handleRoomDetailEntersBackground(isVoiceMessageActive: Boolean) {
+        if (isVoiceMessageActive) {
+            val audioType = voiceMessageHelper.stopAllVoiceActions(deleteRecord = false)
+            val audioJsonString = audioType?.toContentAttachmentData()?.toJsonString()
+            _viewEvents.post(RoomDetailViewEvents.SaveDraft(audioJsonString, MessageType.MSGTYPE_AUDIO))
+        } else {
+            _viewEvents.post(RoomDetailViewEvents.SaveDraft(null, MessageType.MSGTYPE_TEXT))
```
bmarty, line 674: weird arch

ouchadam:
going to close this PR whilst working on it

Issue #3395, opened 2021-05-22
Marking message as read clears all other messages from notification bar


chagai95:
Describe the bug
Clicking on mark read of one message in the notification bar clears all notifications from Element

To Reproduce

- get a message
- click on mark read in the notification bar

Expected behavior
It should clear only that notification and not clear all others

Screenshots
I'll send a screen cap soon and logs

Smartphone (please complete the following information):
Device: Redmi Note 8 Pro
OS: MIUI Global 12.0.4 (Android Version 10)
![IMG_20210522_102701](https://user-images.githubusercontent.com/31655082/119223144-114f5680-baf8-11eb-9bd7-06e04feeba6a.jpg)

Additional context
App version and store: 1.1.7 Google Play beta tester
Homeserver: matrix.org

I think this is specific to Xiaomi but it does work on Telegram - I did notice it seems to clear everything but pop up a second later, so maybe they added specific code for Xiaomi

ouchadam:
not a Xiaomi device but a close relative~ the gif above is from a Pocofone f1 running MIUI 12.0.3 global

Issue #1491, opened 2020-06-16
Do not notify about m.room.redaction


Dragnucs:
If someone redacts a message from a room I have notifications on for, RiotX send a notification with the sole "m.room.redaction" text.

If I "see" the redaction from another client, on the laptop for example, RiotX will still keep the notification on while other are retracted.

witchent:
Also, in encrypted rooms this message is being shown with a red shield, as it is obviously an non-encrypted message in an encrypted room. I guess this could lead to some confusion.

Issue #582, opened 2019-09-24
Notification of invitation does not display the room name


bmarty:
Here is the current rendering:

<img width="440" alt="image" src="https://user-images.githubusercontent.com/3940906/65488571-bb04bd00-dea9-11e9-93a7-6eb483d21e8d.png">

The second "RiotX dbg" should be the room name, if available

The `NotificationDrawerManager` has only the invitation event and not the whole `RoomInviteState`

bmarty:
It should be possible to do better, maybe using room peeking?

PR #4274, opened 2021-10-20
Notifications redesign

ouchadam:
Notification redesign feature PR

Addresses an assortment of issues

- #4152 tentatively fixed (needs confirmation)
- #3395 
- #582 
- #1491

The notifications logic has been redesigned in order to split the creation of the notifications and the displaying/rendering of them. 
This aims to eliminate any race conditions between updating the child and group summary notifications which can cause duplicated notifications, sounds, vibration.

Another benefit of the redesign is the simplification of the flow allowing us to diff new events and rendered events which means we're able to stay better _in sync_ with which notifications to display  

Included reviewed PRs
- #4182 
- #4207
- #4230
- #4286
- #4235  
    - #4185 Splitting all creation and all displaying into separate passes 
    - #4220 Diffiing notification to avoid redundant updates and keep the currently displayed and queued events in sync

| JOINING ROOMS EVENT | MARKING READ |
| --- | --- |
after-multiple-joins|after-mark-read

| INCOMING MESSAGES | ROOM INVITE | ACCEPTING / READING |
| --- | --- | --- |
after-incoming-messages|after-room-invite|after-accepting-reading

### DIFF
--- a/changelog.d/1491.bugfix
+++ b/changelog.d/1491.bugfix
@@ -1,0 +1,1 @@
+Stops showing a dedicated redacted event notification, the message notifications will update accordingly


ouchadam:
TODO - create changelogs

bmarty:
Doing some tests.
- I have pushed 2 commits to stop using a room member avatar when there is no room avatar.
- There is a regression with message redaction:
1. send a message -> a notif is displayed
2. redact the message -> the notif is still displayed, because I think there is no PUSH when message are redacted
3. send another message -> PUSH; sync; the second message is displayed, and the first one is removed -> This is where the regression is seen, see the diff between prod (Element) and this branch (Element dbg):
image

ouchadam:
> Another remark, I think `fun loadEventInfo()` will failed after an upgrade of the application, since the model has changed, so the file could not be deserialized (Not tested though). The failure will be caught, so no crash will occur. This is maybe acceptable. Also I have noticed that after an upgrade of the application, the notification are removed (by the system I guess), but I have never tested that a new notification would restore all the previously displayed notification.

that's a very good point that I completely missed

it's not ideal but at least it will only happen once for this upgrade

------

for the fix above I'm waiting for some test devices to charge so I can take a screenshot/gif of the redaction in action

Issue #4297, opened 2021-10-21
Invite by email doesn't work


kittykat:
### Steps to reproduce

1. In the left panel, ⋮ -> Invite people
2. By username or email
3. Enter valid email

#### What did you expect?
Be able to invite by email

#### What happened instead?
Cannot proceed

See also #3820

### Application version and app store

Element 1.3.4 (RC)

### Will you send logs?

No

bmarty:
Fixed in EA 1.3.8

PR #4339, opened 2021-10-26
Missing ability to invite users by email

ouchadam:
Fixes the email invites not being available when searching for users #4297

- `onIdentityServerChange` is using exceptions to control the flow so we can't use `tryEmit` in this case. This caused the identity server steps to be missed
- `MutableStateFlow` automatically deduplicates emissions so we have to provide some more data in order to re-emit identical search terms (which the identity server steps rely on in order to refresh the page and move to the next step) 

| BEFORE | AFTER | GIF |
| --- | --- | --- | 
Screenshot_20211026_132047|Screenshot_20211026_131814|after-invite-email

Copy updates 

| BEFORE | AFTER |
| --- | --- |
before-you-have-not|Screenshot_20211104_115135
before-give-consent|Screenshot_20211104_115438

### DIFF
--- a/vector/src/main/java/im/vector/app/features/userdirectory/UserListViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/userdirectory/UserListViewModel.kt
@@ -17,6 +17,7 @@
 package im.vector.app.features.userdirectory
 
 import androidx.lifecycle.asFlow
+import com.airbnb.mvrx.Fail
 import com.airbnb.mvrx.MavericksViewModelFactory
 import com.airbnb.mvrx.Uninitialized
 import dagger.assisted.Assisted
@@ -40,6 +41,7 @@ import kotlinx.coroutines.flow.sample
 import org.matrix.android.sdk.api.MatrixPatterns
 import org.matrix.android.sdk.api.extensions.tryOrNull
 import org.matrix.android.sdk.api.session.Session
+import org.matrix.android.sdk.api.session.identity.IdentityServiceError
 import org.matrix.android.sdk.api.session.identity.IdentityServiceListener
 import org.matrix.android.sdk.api.session.identity.ThreePid
 import org.matrix.android.sdk.api.session.profile.ProfileService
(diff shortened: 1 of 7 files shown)


daniellekirkwood:
Based on the feedback from @bmarty change the copy to...

"Your contacts are private. You'll need to share contact info to find people you know. 
Give consent"

and

"Your contacts can't find you, and you can't find them. In order to see each other you must share contact info with your identity server (<name>). 
Your data will be sent securely."

bmarty:
@daniellekirkwood are you sure you have copied the latest version of the strings? This is still wrong to me.

daniellekirkwood:
Screenshot 2021-11-04 at 11 23 59


updated copy...

bmarty:
@daniellekirkwood is it possible that you copy paste the Strings to the PR so we can easily import them in the code? We cannot extract strings from the image. Thanks!

daniellekirkwood:
Your contacts are private. To discover users from your contacts, we need your permission to send contact info to your identity server.
 <Give consent> 

---

Send emails and phone numbers to <name of identity server>?

To discover existing contacts, you need to send contact info to your identity server. 

We hash your data before sending for privacy. Do you consent to send this info?

Policy | No | Yes

ouchadam:
https://github.com/vector-im/element-android/pull/4339/commits/85f8f217b515f6cdf61650e010cb126096ac2cc6 scxreenshots updated

vector/src/main/java/im/vector/app/features/userdirectory/UserListViewModel.kt
```diff
@@ -241,3 +254,10 @@ class UserListViewModel @AssistedInject constructor(@Assisted initialState: User
         setState { copy(pendingSelections = selections) }
     }
 }
+
+private fun UserListViewState.hasNoIdentityServerConfigured() = matchingEmail is Fail && matchingEmail.error == IdentityServiceError.NoIdentityServerConfigured
+
+/**
+ * Wrapper class to allow identical search terms to be re-emitted
+ */
+private data class UserSearch(val searchTerm: String, val cacheBuster: Long = 0)
```
bmarty, line 263: I guess the cacheBuster param is useful behind the scene? (I do not see any `distinct()` call or similar)

ouchadam: yep, it's because `SharedFlow` is distinct by default

bmarty (approved):
Thanks for the update!

PR #4360, opened 2021-10-28
Poll Feature - Create

onurays:
[Figma](https://www.figma.com/file/tepA3ELWBE5v4O2WB8PqWn/Polls?node-id=247%3A60002)

<table align="center">
  <tr>
    <td valign="top"></td>
    <td valign="top"></td>
  </tr>
  <tr>
    <td valign="top"></td>
    <td valign="top"></td>
  </tr>
  <tr>
    <td valign="top"></td>
    <td valign="top"></td>
  </tr>

</table>


vector/src/main/java/im/vector/app/features/navigation/DefaultNavigator.kt
```diff
@@ -501,7 +502,10 @@ class DefaultNavigator @Inject constructor(
     }
 
     override fun openCreatePoll(context: Context, roomId: String) {
-        val intent = CreatePollActivity.getIntent(context, CreatePollArgs(roomId = roomId))
+        val intent = CreatePollActivity.getIntent(
+                context,
+                CreatePollArgs(roomId = roomId, minOptionsCount = CreatePollViewModel.MIN_OPTIONS_COUNT)
```
bmarty, line 507: What is the point to pass `minOptionsCount` as an Activity parameter?

vector/src/main/java/im/vector/app/features/poll/create/CreatePollViewModel.kt
```diff
@@ -42,16 +39,25 @@ class CreatePollViewModel @AssistedInject constructor(
...
+        onEach(
+                CreatePollViewState::question,
+                CreatePollViewState::options
+        ) { question, options ->
+            setState {
+                copy(
+                        canCreatePoll = canCreatePoll(question, options),
+                        canAddMoreOptions = options.size < MAX_OPTIONS_COUNT
+                )
+            }
```
bmarty, line 60: 👍

vector/src/main/java/im/vector/app/features/poll/create/CreatePollViewState.kt
```diff
@@ -14,19 +14,21 @@
...
         val options: List<String> = emptyList(),
-        val canCreatePoll: Boolean = false
+        val canCreatePoll: Boolean = false,
+        val canAddMoreOptions: Boolean = true
 ) : MavericksState {
 
     constructor(args: CreatePollArgs) : this(
-            roomId = args.roomId
+            roomId = args.roomId,
+            options = List(args.minOptionsCount) { "" }
```
bmarty, line 31: This is not the best place to init `options`. I would have done it on line 24 instead

bmarty (commented):
Thanks for the update.
A few more comments for you. Also can you please check the CI status?

gaelledel:
@onurays I'm not able to see the polls feature in the test build you sent

vector/src/main/res/layout/view_attachment_type_selector.xml
```diff
@@ -163,5 +163,36 @@
...
         </LinearLayout>
+
+
+        <LinearLayout
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:layout_margin="16dp"
+            android:baselineAligned="false"
+            android:orientation="horizontal"
+            android:visibility="gone"
```
bmarty, line 174:
```suggestion
            android:visibility="visible"
```

vector/src/main/java/im/vector/app/features/navigation/DefaultNavigator.kt
onurays: I thought it was necessary to use it in `CreatePollViewState`. Fixed by using const value in ViewModel directly.

vector/src/main/java/im/vector/app/features/poll/create/CreatePollViewState.kt
onurays: Done.

gaelledel (changes requested):
- [x] We need to left align the close button with the content below please ( cross aligned with title under)
- [x] Change the colour of the text field frames so it matches Figma colours --> Quinary
- [x] The option text fields should not grow (this behaviour should only happen for the poll question)
- [x] The delete Option button should be centre aligned with the text field
- [ ] Weird animation when no options fields are displayed and you press on Add option. The add option text animates downwards. Could you remove this?
- [x] On user input in the text field: the Return key should direct users to the next field automatically (if other text field is created)
- [ ] On user input in a text field: We need to make sure that the "add option" text button remains visible above the composer so users can easily create a new option without having to scroll down.
- [ ] The Create Poll button disabled state visual does not match Figma's disabled state. For more info on buttons and core components please check Compound UI kit.

ouchadam:
how does the screen look in dark mode?

ouchadam (approved):
small remark about double checking dark mode, otherwise looks good!

vector/src/main/java/im/vector/app/features/form/FormEditTextWithDeleteItem.kt
```diff
@@ -0,0 +1,99 @@
...
+    override fun bind(holder: Holder) {
+        super.bind(holder)
+        holder.textInputLayout.isEnabled = enabled
+        holder.textInputLayout.hint = hint
+
+        holder.textInputEditText.setTextIfDifferent(value)
+
+        holder.textInputEditText.isEnabled = enabled
+        if (singleLine) {
+            holder.textInputEditText.setSingleLine()
```
bmarty, line 74: When doing binding, we have to take care of the recycling of views. So a else block is needed here I think

vector/src/main/java/im/vector/app/features/form/FormEditTextWithDeleteItem.kt
```diff
@@ -0,0 +1,99 @@
...
+        }
+        imeOptions?.let {
+            holder.textInputEditText.imeOptions = it
+        }
```
bmarty, line 78: same remark here

vector/src/main/java/im/vector/app/features/poll/create/CreatePollActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+import im.vector.app.core.extensions.addFragment
+import im.vector.app.core.platform.SimpleFragmentActivity
+import dagger.hilt.android.AndroidEntryPoint
+
+@AndroidEntryPoint
+class CreatePollActivity : SimpleFragmentActivity() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        views.toolbar.visibility = View.GONE
```
bmarty, line 33: Strange to hide the toolbar here. I guess the toolbar is part of the fragment?

vector/src/main/java/im/vector/app/features/poll/create/CreatePollFragment.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    private fun renderToast(message: String) {
+        views.createPollToast.removeCallbacks(hideToastRunnable)
+        views.createPollToast.text = message
+        views.createPollToast.isVisible = true
+        views.createPollToast.postDelayed(hideToastRunnable, 2_000)
+    }
+
+    private val hideToastRunnable = Runnable {
+        views.createPollToast.isVisible = false
```
bmarty, line 130: I think this can crash if the Fragment has been destroyed

bmarty (commented):
Just a few more questions / remark.
Also please fix errors reported by the CI

vector/src/main/java/im/vector/app/features/form/FormEditTextWithDeleteItem.kt
onurays: Done.

vector/src/main/java/im/vector/app/features/form/FormEditTextWithDeleteItem.kt
onurays: Done.

vector/src/main/java/im/vector/app/features/poll/create/CreatePollActivity.kt
onurays: Otherwise, it shows the back button on Toolbar. We are doing the same for `CreateDirectRoomActivity` and `InviteUsersToRoomActivity`.

vector/src/main/java/im/vector/app/features/poll/create/CreatePollFragment.kt
onurays: We are using the same mechanism while recording voice messages. It is not crashing because it uses Handler inside. Also, we will never show toast on this screen, since we disable the button.

But I want to keep this to protect against post improvements.

vector/src/main/res/values/strings.xml
```diff
@@ -3626,4 +3627,18 @@
     <string name="link_this_email_settings_link">Link this email with your account</string>
     <!-- %s will be replaced by the value of link_this_email_settings_link and styled as a link -->
     <string name="link_this_email_with_your_account">%s in Settings to receive invites directly in Element.</string>
+
+    <!-- Poll -->
+    <string name="create_poll_title">Create Poll</string>
+    <string name="create_poll_question_title">Poll question or topic</string>
+    <string name="create_poll_question_hint">Question or topic</string>
+    <string name="create_poll_options_title">Create options</string>
```
bmarty, line 3635: We discussed about that wording in a Matrix room. Maybe "Options" is better here. What is the status?
CC @gaelledel

bmarty (changes requested):
The CI is not happy, can you have a look please?

vector/src/main/res/values/strings.xml
onurays: @gaelledel wants to use "Create options".

bmarty: OK, thanks

bmarty (approved):
Thanks for the update.
I will do a quick change on the style, sth I did not see before, but not blocking.

PR #4380, opened 2021-11-01
Key request screen crash on launch

ouchadam:
Fixes crash when launching the key request dev screen

| BEFORE | AFTER | 
| --- | --- |
|before-key-request|after-key-request

### DIFF
--- a/changelog.d/4380.bugfix
+++ b/changelog.d/4380.bugfix
@@ -1,1 +1,0 @@
-Fixes crash when launching the key request dev tools screen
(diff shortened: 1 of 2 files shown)


bmarty:
Strangely, I do not repro the crash on 1.3.6, but can on 1.3.7-dev.
I'm wondering if it deserves a changelog entry, since the crash has not been released? Or am I wrong here?

ouchadam:
happy to remove the changelog if the issue never made it to users :+1:

bmarty (approved):

PR #4381, opened 2021-11-01
Force call ringtones to loop

### LINKED ISSUE #4047: [VoIP] Ring tone plays only once
### Steps to reproduce

Called a phone with Element Android v1.2.2 installed. This phone has the default OS ring tone.

### What did you expect?

The ring tone should have played the entire time the call was being made.

### What happened?

The ring tone played once fully, and then went silent. The call banner stayed, though.

### Your phone model

Nokia 6.1

### Operating system version

Android 10

Element Android 1.2.2, Play Store

### Homeserver

matrix.org

No

ouchadam:
Fixes #4047 

Out of the box android is relying on audio file metadata to determine if the ringtone should loop

- Android 9.0 introduced an api to programmatically enable looping https://developer.android.com/reference/android/media/Ringtone#setLooping(boolean)
- Users below Android 9.0 will need to select other ringtones or edit them to include a `ANDROID_LOOP true` tag

### DIFF
--- a/changelog.d/4047.bugfix
+++ b/changelog.d/4047.bugfix
@@ -1,0 +1,1 @@
+Fixing call ringtones only playing once when the ringtone doesn't contain looping metadata (android 9.0 and above)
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Looks nice, thanks!

opusforlife2:
Thank you, @ouchadam!

PR #4391, opened 2021-11-01
create fallback turn server option

### LINKED ISSUE #2329: allow to use fallbackcall assist server (turn.matrix.org)
it seems that element android does not allow to use the fallback call assist server (turn.matrix.org), while riot android did, and element web does.

this option is useful for people running their own homeserver without a turn server. ideally, each homeserver should have its turn server, but it is not always possible to run one. it can be because of network traffic or computing power limitations, or because of network setup constraints. for example, when the homeserver is running in a home network behind nat and has a different ip address inside and outside the home network, running a turn server on it does not seem to work (or if it is, its configuration seems pretty challenging).

will you please add this option?

### LINKED ISSUE #3063: ability to switch on using turn server turn.matrix.org from Android device
I urgently needed to set up a video call with someone. I was using element on linux on a PC and he was using element on an android tablet on WiFi installed from fdroid. We could text message but when you tried to make a video or audio call it refused to connect. 

On my linux element I switched on connect to matrix turn server:

> 
> Einstellungen > Sprach- & Videoanruf > 
> „Auf den Server turn.matrix.org zurückgreifen, falls deine Heimserver keine Anruf-Assistenz anbietet (deine IP-Adresse wird während eines Anrufs geteilt)“ 

Settings - video and voice calls > use turn.matrix.org if the homeserver offers no call assist (your ip address will be shared during the call)

That was not enough. 

I asked him to do the same and he said it was not there. 

Now I installed element myself on an Android device and note that indeed it definitely is NOT there!!

Could you please add it? It is very important. If it is hiding somewhere please tell me where it is..

CicadaCinema:
Fixes #2329 
Fixes #3063
You need to enable the fallback option in `Settings > Voice & Video` to use this feature:

### Pull Request Checklist


vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
```diff
@@ -269,17 +271,28 @@ class WebRtcCall(
     private fun createPeerConnection(turnServerResponse: TurnServerResponse?) {
         val peerConnectionFactory = peerConnectionFactoryProvider.get() ?: return
         val iceServers = mutableListOf<PeerConnection.IceServer>().apply {
-            turnServerResponse?.let { server ->
-                server.uris?.forEach { uri ->
+            if (vectorPreferences.useFallbackTurnServer()) {
```
bmarty, line 274: I'm not a specialist on this part of the code, but maybe use the fallback if and only if `turnServerResponse?.uris is null or empty, and the user has enable the fallback from the setting?

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
```diff
@@ -269,17 +271,28 @@ class WebRtcCall(
...
+                listOf("turn:turn.matrix.org:3478?transport=udp", "turn:turn.matrix.org:3478?transport=tcp", "turns:turn.matrix.org:443?transport=tcp").forEach {
```
bmarty, line 275: Maybe extract this list to the file https://github.com/vector-im/element-android/blob/main/vector/src/main/res/values/config.xml#L2 ?

bmarty (commented):
Thanks for the PR.
Please see my first remarks.
Also can you add a file for the changelog as per https://github.com/vector-im/element-android/blob/main/CONTRIBUTING.md#changelog ?

Thanks

CicadaCinema:
Thank you for reviewing this! I actually don't know if this change fixes #2329 at all, I've asked people there to test out my builds. I'll finish up work on this PR if I get positive feedback on the changes.

bmarty:
Ok, thanks. I have enabled the CI for this PR so that APKs will be built and can be downloaded from here to be tested.

CicadaCinema:
Thank you!

DoM1niC:
Nice PR @CicadaCinema but I prefer first to use the TURN from Homeserver instead of general Matrix to use.... I my mind you can read this from API or you switch the Server URLs to gradle vars to find and setup faster the own TURN Server like me....

But anyway good work 👍🏻

CicadaCinema:
> Nice PR @CicadaCinema but I prefer to use the TURN first from Homeserver instead of general Matrix to use.... I think, you can read this from API or you switch the Server URLs to gradle vars to find and setup faster the own TURN Server like me....
> 
> But anyway good work 👍🏻 

If this setting is enabled, I plan to _only_ use turn.matrix.org when the user's homeserver does not have one that they can use (similarly to the way the setting in Element Web works).

CicadaCinema:
Signed-off-by: Arseny Uskov <arsenyuskov@gmail.com>

ouchadam:
awesome change! with the fallback only being used when no turn server uris are returned by the server **and** the preference is enabled this looks like it ticks everyone's boxes

ouchadam (approved):

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
```diff
@@ -280,13 +285,15 @@ class WebRtcCall(
         val peerConnectionFactory = peerConnectionFactoryProvider.get() ?: return
         val iceServers = mutableListOf<PeerConnection.IceServer>().apply {
             turnServerResponse?.let { server ->
-                server.uris?.forEach { uri ->
+                val useFallback = server.uris.isNullOrEmpty() && vectorPreferences.useFallbackTurnServer()
+                val serverList = if (useFallback) stringArrayProvider.getStringArray(R.array.fallback_ice_servers).toList() else server.uris
+                serverList?.forEach { uri ->
```
bmarty, line 290: Maybe a stupid question, but do we want to use the fallback if `turnServerResponse` is null? Here this will not be the case.

BillCarsonFr:
Notice that using a fallback while revealing your IP to the relay will also hide it to your contact.
In other app this option can be presented as always relay Calls (to avoid revealing your IP adress to your contact).
So the current text copy is maybe a bit too negative

BillCarsonFr:
Also the old android app was prompting the user to enable this option when you have a call that failed to connect and that your home server had no ice server config. Maybe a nice to have feature

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
BillCarsonFr: Yes we should only use the fallback if the home server does not define any ice servers (consitent with web).

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
```diff
@@ -280,13 +285,15 @@ class WebRtcCall(
...
                     add(
                             PeerConnection
                                     .IceServer
                                     .builder(uri)
-                                    .setUsername(server.username)
-                                    .setPassword(server.password)
+                                    .setUsername(if (useFallback) "xxxxx" else server.username)
```
BillCarsonFr, line 295: The fallback is stun, should do (from old sdk)
````
 if (!defaultIceServerUri.startsWith("stun:")) {
                    defaultIceServerUri = "stun:" + defaultIceServerUri;
                }
                defaultIceServer = PeerConnection.IceServer.builder(defaultIceServerUri).createIceServer();

File: vector/src/main/res/values/config.xml [hunk 5]
```diff
@@ -7,6 +7,12 @@
     <string name="matrix_org_server_url" translatable="false">https://matrix.org</string>
     <string name="piwik_server_url" translatable="false">https://piwik.riot.im</string>
     <string name="bug_report_url" translatable="false">https://riot.im/bugreports/submit</string>
+    <string name="fallback_stun_server_url" translatable="false">turn.matrix.org</string>
+    <string-array name="fallback_ice_servers" translatable="false">
+        <item>turn:turn.matrix.org:3478?transport=udp</item>
```
BillCarsonFr, line 12: I don't think this will work, turn.matrix.org requires auth.
Just use stun: would work

vector/src/main/res/values/strings.xml
```diff
@@ -724,7 +724,7 @@
     <!-- Note to translators: the translation MUST contain the string "${app_name}", which will be replaced by the application name -->
     <string name="template_settings_call_ringtone_use_app_ringtone">Use default ${app_name} ringtone for incoming calls</string>
     <string name="settings_call_ringtone_use_default_stun">Allow fallback call assist server</string>
-    <string name="settings_call_ringtone_use_default_stun_sum">Will use "%s" as assist when your homeserver does not offer one (your IP address will be shared during a call)</string>
+    <string name="settings_call_ringtone_use_default_stun_sum">Will use %s as assist when your homeserver does not offer one (your IP address will be shared during a call)</string>
```
BillCarsonFr, line 727: I think this copy is missleading. IP will be shared with who? the peer yes, the stun yes.
Not using it will still share your IP with the peer.
So maybe your IP will be seen by the stun server? (not sure how valuable/actionable this information is for the user, given that not doing it will make calls fails depending on your network)

BillCarsonFr (changes requested):
Hello,
The default server will only do stun, please see my comments.

CicadaCinema:
@BillCarsonFr 

I adjusted the behaviour of the fallback to use the stun server instead, and I adjusted the option description.

> Also the old android app was prompting the user to enable this option when you have a call that failed to connect and that your home server had no ice server config. Maybe a nice to have feature

I'm not sure how to even begin implementing this unfortunately - I only took a look at this issue because I noticed I couldn't make calls myself.

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
bmarty: Ok, so the code `turnServerResponse?.let { server ->` has to be updated IIUC.

CicadaCinema: there, I've handled that now

vector/src/main/res/values/config.xml
bmarty: `fallback_ice_servers` can just be removed now I think

CicadaCinema: yep, I've removed it

CicadaCinema:
# PR Moved to https://github.com/vector-im/element-android/pull/5781

PR #4396, opened 2021-11-03
Feature/aris/thread aware

### LINKED ISSUE #4246: Make Element Android Thread aware

To take a few steps towards understanding Threads in Android before a full implementation

To ensure maximum backwards compatibility and continuity in conversations when threads will be available to the masses.

[MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) recommends two things for clients that do not fully support threads yet

- [x] To treat `m.thread` the same way that they treat `m.in_reply_to`
- [x] To change the event references from `m.in_reply_to` to `m.thread` when replying to an event that has an `m.thread` relation (to ensure that the message will be displayed correctly in the thread when viewed on thread's ready clients)

_No response_

### Additional context

Threads might be using a `m.thread` relation over `m.in_reply_to`

ariskotsomitopoulos:
## Make Element Android Thread aware

This feature temporarily until the main threads release and will be deleted. The idea is to make smoother the thread transition release for the clients
Closes #4246

**What was the main challenge?**
Too old **root** thread messages, are not fetched in the device so is impossible to show the thread as a reply appropriately. So a custom fetch root thread event mechanism was implemented in order to fetch those **root thread** events prior the transformation to **replies**

- To treat m.thread the same way that they treat m.in_reply_to
-  To change the event references from m.in_reply_to to m.thread when replying to an event that has an m.thread relation _( no need for that while it is already supported and displayed correctly in the web clients)_


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/EventEntity.kt
```diff
@@ -56,10 +57,10 @@ internal open class EventEntity(@Index var eventId: String = "",
 
     companion object
 
-    fun setDecryptionResult(result: MXEventDecryptionResult) {
+    fun setDecryptionResult(result: MXEventDecryptionResult, clearEvent: JsonDict? = null) {
```
bmarty, line 60: This change is strange, let me see the rest of the changes to understand why it is done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
```diff
@@ -601,6 +609,20 @@ internal class DefaultTimeline(
         return offsetResults.size
     }
 
+    /**
+     * This function is responsible to fetch and store the root event of a thread event
+     * in order to be able to display the event to the user appropriately
+     */
+    private fun fetchRootThreadEventsIfNeeded(offsetResults: RealmResults<TimelineEventEntity>) = runBlocking {
```
bmarty, line 616: We should not use `runBlocking`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
```diff
@@ -601,6 +609,20 @@ internal class DefaultTimeline(
...
+        val eventEntityList = offsetResults
+                .mapNotNull {
+                    it?.root
+                }.map {
+                    EventMapper.map(it)
+                }.toList()
```
bmarty, line 622: is `toList()` mandatory here?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineEventDecryptor.kt
```diff
@@ -106,10 +108,19 @@ internal class TimelineEventDecryptor @Inject constructor(
             val result = cryptoService.decryptEvent(request.event, timelineId)
             Timber.v("Successfully decrypted event ${event.eventId}")
             realm.executeTransaction {
-                val eventId = event.eventId ?: ""
-                EventEntity.where(it, eventId = eventId)
+                val eventId = event.eventId ?: return@executeTransaction
```
bmarty, line 111: Thanks, this is better to return if eventId is null (it should never happen, but...)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/ThreadsAwarenessHandler.kt
```diff
@@ -0,0 +1,264 @@
...
+        val eventEntity = EventEntity.where(realm, eventId = eventId).findFirst() ?: return null
+        return EventMapper.map(eventEntity)
+    }
+
+    /**
+     * Returns True if the event is a thread
+     * @param event
+     */
+    private fun isThreadEvent(event: Event): Boolean =
+            event.content.toModel<MessageRelationContent>()?.relatesTo?.type == "io.element.thread"
```
bmarty, line 250: Nit: "io.element.thread" should be defined in `RelationType`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
```diff
@@ -94,6 +96,10 @@ internal class SyncResponseHandler @Inject constructor(
             Timber.v("Finish handling toDevice in $it ms")
         }
         val aggregator = SyncResponsePostTreatmentAggregator()
+
+        // Prerequisite for thread events handling in RoomSyncHandler
+        threadsAwarenessHandler.fetchRootThreadEventsIfNeeded(syncResponse)
```
bmarty, line 101: Do we want to do that even for the initial sync? It can take time if there are lots of events to retrieve

bmarty (commented):
I have some remarks. I think it would be valuable if @ganfra review this PR too.
I'm wondering why it's mandatory to retrieve the root event during the sync response processing. Can we do it only if the timeline is rendered?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
ariskotsomitopoulos: Correct, but we are already in a thread (and the UI will not freeze) and it's the only way to ensure that the next code block will have the required info, if you can suggest an alternative to that it's welcome. Furthermore keep in mind that this is a temporary fix and will be removed after the thread release

ariskotsomitopoulos:
> I have some remarks. I think it would be valuable if @ganfra review this PR too. I'm wondering why it's mandatory to retrieve the root event during the sync response processing. Can we do it only if the timeline is rendered?

Later in the handler you would need that info, especially for the 10 latest messages.
It's not a big deal while:
- User will fetch the root event **ONLY** once and never again
- User will fetch them only if there is a thread  response in the last messages

The majority of the fetching will be done when the timeline is rendered and decrypted.

Furthermore, if we go with a different approach we should add loaders on messages and the complexity of that is not worth it while its just a temporarily solution, and we should not spent that much time

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
ariskotsomitopoulos: Yes, without this it will not be rendered appropriately. I think there will be never more than max 1/2 fetches per room worst case. Ofc this will be removed upon the first thread release.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt
ganfra: Yes thats not really great but it's ok I guess for now. Note that my rework of the timeline authorise direct usage of suspend methods

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/ThreadsAwarenessHandler.kt
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Fetch root thread events if they are missing from the local storage
+     * @param eventList a list with the events to examine
+     */
+    suspend fun fetchRootThreadEventsIfNeeded(eventList: List<Event>) {
+        if (eventList.isNullOrEmpty()) return
+
+        val threadsToFetch = emptyMap<String, String>().toMutableMap()
+        monarchy.awaitTransaction { realm ->
```
ganfra, line 88: Why are u using awaitTransaction here?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/ThreadsAwarenessHandler.kt
```diff
@@ -0,0 +1,264 @@
...
+/**
+ * This handler is responsible for a smooth threads migration. It will map all incoming
+ * threads as replies. So a device without threads enabled/updated will be able to view
+ * threads response as replies to the original message
+ */
+internal class ThreadsAwarenessHandler @Inject constructor(
+        private val permalinkFactory: PermalinkFactory,
+        private val cryptoService: CryptoService,
+        @SessionDatabase private val monarchy: Monarchy,
+        private val session: Session
```
ganfra, line 58: We shouldn't use session in the SDK this is only intended to be used by SDK user

ganfra (changes requested):
Some remarks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/ThreadsAwarenessHandler.kt
ariskotsomitopoulos: Updated accordingly

bmarty (approved):
LGTM, @ganfra ?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/ThreadsAwarenessHandler.kt
```diff
@@ -0,0 +1,269 @@
...
+/**
+ * This handler is responsible for a smooth threads migration. It will map all incoming
+ * threads as replies. So a device without threads enabled/updated will be able to view
+ * threads response as replies to the original message
+ */
+internal class ThreadsAwarenessHandler @Inject constructor(
+        private val permalinkFactory: PermalinkFactory,
+        private val cryptoService: CryptoService,
+        @SessionDatabase private val realmConfiguration: RealmConfiguration,
+        @SessionDatabase private val monarchy: Monarchy,
```
ganfra, line 59: If you already have Monarchy object you can use the monarchy.realmConfiguration

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/ThreadsAwarenessHandler.kt
```diff
@@ -0,0 +1,269 @@
...
+     * This function will fetch the event from the homeserver, this is mandatory when the
+     * initial thread message is too old and is not saved in the device, so in order to
+     * construct the "reply to" format we need to know the event thread.
+     * @return the Event or null otherwise
+     */
+    private suspend fun fetchEvent(eventId: String, roomId: String): Event? {
+        return runCatching {
+            Timber.i("------> Fetching event[$eventId]....")
+            getEventTask.execute(GetEventTask.Params(roomId = roomId, eventId = eventId)).apply {
+                unsignedData?.age?.let { System.currentTimeMillis() - it }
```
ganfra, line 135: Maybe this should be added to the task?

ganfra (approved):
Thanks for the changes, ok for me now

PR #4401, opened 2021-11-03
Single `PushTriggerListener` dispatch

ouchadam:
Some preparation/refactors around the `Push triggers` in order to support _suspending_ `NotifiableEventResolver.resolveEvent` which is required for including images in the notifications

*points to a feature branch

- `PushRuleListener` functions are replaced with a single `onEvents(PushEvents)` in order to allow any logic within the implementation to be asynchronous 
- All notification event changes are handled via the `notificationDrawerManager.updateEvents`, this ensures a single point of synchronisation and notification render pass (reduces the amount of render passes and synchronisation locks)
- Notification queue mutations are extracted to a dedicated class along with unit tests 
- Lifts the `NotifiableEvent` fixtures to a dedicated fixtures test package

PR #4402, opened 2021-11-03
Supporting images in notifications 

### LINKED ISSUE #1697: Render image in notification
Using https://developer.android.com/reference/android/app/Notification.BigPictureStyle

ouchadam:
Fixes #1697, relies on #4401 refactor 

Adds image support to the message notifications

- Downloads the image as part of the notification event resolving chain
- Temporarily exposes the image via `FileProvider Uri` in order to allow the notifications to read and display the image

| EXPANDED GROUP WITH TEXT | EXPANDED GROUP | GROUP COLLAPSED  |  GROUP SUMMARY | SINGLE | ROOM LIST |
| --- | --- | --- | --- | --- | --- |
|Screenshot_20211103_154653|Screenshot_20211103_154629|Screenshot_20211103_154620|Screenshot_20211103_154612|Screenshot_20211103_154705|after-room-list

### DIFF
--- a/vector/src/main/java/im/vector/app/features/notifications/NotifiableEventResolver.kt
+++ b/vector/src/main/java/im/vector/app/features/notifications/NotifiableEventResolver.kt
@@ -15,8 +15,10 @@
  */
 package im.vector.app.features.notifications
 
+import android.net.Uri
 import im.vector.app.BuildConfig
 import im.vector.app.R
+import im.vector.app.core.extensions.takeAs
 import im.vector.app.core.resources.StringProvider
 import im.vector.app.features.displayname.getBestName
 import im.vector.app.features.home.room.detail.timeline.format.DisplayableEventFormatter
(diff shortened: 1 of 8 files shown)


vector/src/main/java/im/vector/app/features/notifications/NotifiableEventResolver.kt
```diff
@@ -192,6 +199,26 @@ class NotifiableEventResolver @Inject constructor(
         }
     }
 
+    private suspend fun TimelineEvent.fetchImageIfPresent(session: Session): Uri? {
+        return when {
+            root.isEncrypted() && root.mxDecryptionResult == null -> null
+            root.getClearType() == EventType.MESSAGE              -> downloadAndExportImage(session)
```
bmarty, line 205: I think we should filter on `msgtype` here, since the attachment is not necessarily an image, be can be anything else. `Event.isImageMessage()` can help to do that.

bmarty (commented):
One remark, plus:
- You want to merge that PR into `feature/adm/single-transaction-push` (and not develop)?
- Changelog please :)
- `BigPictureStyle` does not need to be used (see #1697)?

vector/src/main/java/im/vector/app/features/notifications/NotifiableEventResolver.kt
ouchadam: :+1:  will do!

ouchadam:
> * You want to merge that PR into `feature/adm/single-transaction-push` (and not develop)?

It would be great to merge this after `feature/adm/single-transaction-push`, I split the refactor from the implementation to help with the diff 

> * Changelog please :)
:+1:  will add (was originally planning to add as part of the feature PR

> * `BigPictureStyle` does not need to be used (see #1697)?
correct, not needed because we're already using the `MessageStyle` which supports inline images

`BigPictureStyle` would only work for 1 off notifications meaning we would lose the conversation aspect of the notification

bmarty:
Thanks for the explanations about `BigPictureStyle`.

I still think there is something strange in your branches since I see this on the other PR:
image
And here:
image
So it looks a bit "crossy"

bmarty:
Ah there are 3 branches, my bad

bmarty (approved):
LGTM, thanks!

PR #4404, opened 2021-11-03
Adding `View source` to sanity check

### LINKED ISSUE #4308: Add action View source to the sanity test
This is not the first time we have a crash on this action, since it's using an external library.
It should be nice to detect such issue before the release

ouchadam:
Fixes #4308 

Adds `View Source` checking in the sanity test suite
- Also includes some extra sleeps to make the login flow less flaky when running locally (might be just my machine?)

*Added extra sleep for the gif to show the View Source screen for longer

| IN ACTION |
| --- |
|view-source-sanity

### DIFF
--- a/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
+++ b/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
@@ -83,6 +83,7 @@ class UiAllScreensSanityTest {
     // 2021-04-08 Testing 429 change
     @Test
     fun allScreensTest() {
+        uiTestBase.enableDeveloperMode()
         // Create an account
         val userId = "UiTest_" + UUID.randomUUID().toString()
         uiTestBase.createAccount(userId = userId)
@@ -211,6 +212,14 @@ class UiAllScreensSanityTest {
     private fun longClickOnMessageTest() {
         // Test quick reaction
         longClickOnMessage()
+
+        // Check View Source library
+        clickOn(R.string.view_source)
+        sleep(1000)
+        pressBack()
+
+        longClickOnMessage()
+
         // Add quick reaction
         clickOn("\uD83D\uDC4D️") // 👍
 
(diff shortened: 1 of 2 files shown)


ouchadam:
this feels like it doesn't need a changelog entry as it's not user facing

vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
```diff
@@ -83,6 +83,7 @@ class UiAllScreensSanityTest {
     // 2021-04-08 Testing 429 change
     @Test
     fun allScreensTest() {
+        uiTestBase.enableDeveloperMode()
```
bmarty, line 86: I would prefer to not do that at the beginning of the main test, but rather add a dedicated `fun` to test all developer mode features at the end of this test (there are other screens we could navigate to, devtools, account data, key requests (and we have seen recently that this last one can crash, you fixed it 💯 )), and eventually optionally at the end of this `fun` disable the developer mode.

vector/src/androidTest/java/im/vector/app/ui/UiTestBase.kt
```diff
@@ -29,8 +31,17 @@ import im.vector.app.R
 import im.vector.app.espresso.tools.waitUntilActivityVisible
 import im.vector.app.features.home.HomeActivity
 import im.vector.app.waitForView
+import java.lang.Thread.sleep
 
 class UiTestBase {
+
+    fun enableDeveloperMode() {
+        PreferenceManager.getDefaultSharedPreferences(InstrumentationRegistry.getInstrumentation().targetContext)
+                .edit()
```
bmarty, line 40: Nit: can't we use `edit {}` extension from tests?
By the way this is a nice way to do it, I was thinking of navigating to the settings with the UI (in allScreensTest()) to change this pref :):):)

bmarty (commented):
Thanks for this PR, just 2 small remarks

vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
ouchadam: a fair point, I definitely took the shortcut here by editing the preferences directly 

will close and reopen with the tests interacting with the real dev mode button :+1:

bmarty: Thanks!

PR #4418, opened 2021-11-05
Hotfix/1.3.7 oom

bmarty:
Try to fix an OOM on big accounts.
Not sure about the fix though, but a stacktrace (that I've lost) has shown that `pushDynamicShortcut` can throw excpetion if too many shortcut are pushed.
I've read the doc https://developer.android.com/guide/topics/ui/shortcuts, all paragraphs, and I think we were good :/


ouchadam:
nice catch! :100: 

even if this doesn't fix the OOM it still looks like a good change to havef

ouchadam (approved):

bmarty:
Thanks @ouchadam . There is maybe a lead in the class ShortcutCreator, with Bitmap. I will not block the release with that, but it definitely needs more investigation.
Hotfix has been manually merged.

PR #4420, opened 2021-11-05
Sanity tests robot pattern and reliability refactor

ouchadam:
- Ports the sanity test to the `robot` pattern (extracting the UI navigation and manipulation to dedicated classes which return themselves back to the original state), this enables fine grain reusable navigation 
- Swaps out timeouts for view waits where possible
- Adds missing view waits to make the tests less flaky (they now consistently pass for me with an emulator)
- Disables the sign out flow as the tests currently hit a bug

| GIF | SIGN OUT BUG |
| --- | --- |
after-sanity|Screenshot_20211104_163044


vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
```diff
@@ -85,63 +55,49 @@ class UiAllScreensSanityTest {
...
+        elementRobot.roomList {
+            verifyCreatedRoom()
+        }
 
-        uiTestBase.signout()
-        clickDialogPositiveButton()
+// Disable until the "you don't have a session for id %d" sign out bug is fixed
+//        elementRobot.signout()
+////        Login again on the same account
+//        elementRobot.login(userId)
```
bmarty, line 95: We do not test regular login then... This is a bit annoying

vector/src/androidTest/java/im/vector/app/ui/robot/CreateNewRoomRobot.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.adevinta.android.barista.assertion.BaristaListAssertions
+import com.adevinta.android.barista.interaction.BaristaClickInteractions
+import com.adevinta.android.barista.interaction.BaristaListInteractions
+import im.vector.app.R
+import im.vector.app.espresso.tools.waitUntilActivityVisible
+import im.vector.app.espresso.tools.waitUntilViewVisible
+import im.vector.app.features.home.room.detail.RoomDetailActivity
+
+class CreateNewRoomRobot(
+        var createdRoom: Boolean = false
```
bmarty, line 30: What is the goal of this boolean? (I do not know the Robot pattern)

vector/src/androidTest/java/im/vector/app/EspressoExt.kt
```diff
@@ -70,6 +73,8 @@ fun waitForView(viewMatcher: Matcher<View>, timeout: Long = 10_000, waitForDispl
             val endTime = startTime + timeout
             val visibleMatcher = isDisplayed()
 
+            uiController.loopMainThreadForAtLeast(100)
```
ouchadam, line 76: I really wanted to avoid this but this single change made the biggest difference for me when it came to reliability. 

Dialogs and bottomsheets would be `visible` but not interactable, back actions or button presses would be ignored half the time~.

vector/src/androidTest/java/im/vector/app/ui/robot/NewRoomRobot.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.adevinta.android.barista.interaction.BaristaClickInteractions
+import im.vector.app.R
+
+class NewRoomRobot {
+
+    fun createNewRoom(block: CreateNewRoomRobot.() -> Unit) {
+        BaristaClickInteractions.clickOn(R.string.create_new_room)
+        val createNewRoomRobot = CreateNewRoomRobot()
+        block(createNewRoomRobot)
+        if (!createNewRoomRobot.createdRoom) {
```
bmarty, line 29: the test should fail in this case, no?

bmarty (commented):
Thanks for this nice rework.
A few feedbacks

vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
ouchadam: agreed! I'd like to look into the bug and reenable this ASAP

ouchadam:
some observations during my testing

- Sometimes I see a concurrent exception during the login flow, seems to be coming from the testing library

```
2021-11-05 14:45:28.209 12441-12441/im.vector.app.debug E/MonitoringInstr: Exception encountered by: im.vector.app.features.login.LoginActivity@c5555e9. Dumping thread state to outputs and pining for the fjords.
    java.util.ConcurrentModificationException
        at java.util.ArrayList$Itr.next(ArrayList.java:860)
        at androidx.test.internal.runner.lifecycle.ActivityLifecycleMonitorImpl.signalLifecycleChange(ActivityLifecycleMonitorImpl.java:159)
        at androidx.test.runner.MonitoringInstrumentation.callActivityOnStop(MonitoringInstrumentation.java:754)
        at android.app.Activity.performStop(Activity.java:8018)
        at android.app.ActivityThread.callActivityOnStop(ActivityThread.java:4616)
        at android.app.ActivityThread.performStopActivityInner(ActivityThread.java:4594)
        at android.app.ActivityThread.handleStopActivity(ActivityThread.java:4669)
        at android.app.servertransaction.TransactionExecutor.performLifecycleSequence(TransactionExecutor.java:233)
        at android.app.servertransaction.TransactionExecutor.cycleToPath(TransactionExecutor.java:201)
        at android.app.servertransaction.TransactionExecutor.executeLifecycleState(TransactionExecutor.java:173)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:97)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2016)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
```

- On my emulator the `EditTextPreference` causes an OS tombstone, this happens in the display name dialog but also when viewing the System android build version :thinking:

vector/src/androidTest/java/im/vector/app/ui/robot/CreateNewRoomRobot.kt
ouchadam: it's to avoid a `pressBack` if we create a new room as navigating to the new room finishes the current room creation activity

vector/src/androidTest/java/im/vector/app/ui/robot/NewRoomRobot.kt
ouchadam: without this the next steps will fail as every robot assumes the start and end position are the same

in the case of creating a new room the room creation screen is automatically finished so there's no need to press back, if we did then we'd exit the app as the `RoomDetails` 


```kotlin
// move to room list tab and press #+
elementRobot.newRoom {
    // press create new room launches CreateRoomActivity
    createNewRoom {
        //  press create, navigates to the RoomDetailsActivity, CreateRoomActivity automatically closes
        createRoom {
            // goes through all the room detail checks
            crawl()             
            // on createRoom finish pressBack to take us back to root activity
        }
        // avoid pressing back because we created a room
    }
    // avoid pressing back because we created a room
}
```

vector/src/androidTest/java/im/vector/app/ui/robot/CreateNewRoomRobot.kt
bmarty: OK, I see. But this test should always create a room, so this is still strange to me.

bmarty (approved):
Thanks for the update.
I'll trust your judgment for the boolean var in the room creation robot.

bmarty:
ktlint issues fixed on develop in 2b58c0e5abe9cce6a59cd264859d5060e5ebf24d

PR #4421, opened 2021-11-05
add viewbinding compileOnly dependency to gradle files

Bubu:
I was exploring the codebase and noticed the following error in the `attachement-viewer` and `ui-styles` modules:

```
Cannot access 'android.viewbinding.ViewBinding' which is a supertype of 'im.vector.lib.attachmentviewer.databinding.ItemVideoAttachmentBinding'. Check your module classpath for missing or conflicting dependencies
```

Adding an explicit compile time only dependency on the viewbinding dependency them solves this
problem.

See https://stackoverflow.com/a/63137481/1634837

### Pull Request Checklist

- [x] If you have modified the screen flow, or added new screens to the application, you have updated the test [UiAllScreensSanityTest.allScreensTest()](https://github.com/vector-im/element-android/blob/main/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt#L73)

### DIFF
--- a/attachment-viewer/build.gradle
+++ b/attachment-viewer/build.gradle
@@ -49,7 +49,7 @@ dependencies {
     implementation project(":library:ui-styles")
 
     // make viewBinding classes accessible to Android Studio
-    compileOnly 'com.android.databinding:viewbinding:7.0.3'
+    compileOnly libs.android.viewbinding
 
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
 
--- a/attachment-viewer/build.gradle
+++ b/attachment-viewer/build.gradle
@@ -48,6 +48,9 @@ android {
 dependencies {
     implementation project(":library:ui-styles")
 
+    // make viewBinding classes accessible to Android Studio
+    compileOnly 'com.android.databinding:viewbinding:7.0.3'
+
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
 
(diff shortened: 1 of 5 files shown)


Bubu:
Mh, I also needed to add the dep to the `vector` module now... which is confusing as then this problem doesn't seem to be occurring for others?

bmarty:
Hello @Bubu,
Which issue are you seeing and fixing here? We have no problem on our side.
Maybe we do not use the same AS version?

Bubu:
Hi,
I'm using AS version 2020.3.1 (which afaik is the latest version as this auto-updates):

image

The problem I'm seeing without this patch is the following:
image

I don't think there's any *functional* limitation with this. But having the majority of every UI related file marked as red and thus having most files show up as containing errors somewhat irritating. :D

The stackoverflow page I liked as pasrt of the MR suggests that the way to fix this is by adding an explicit dependency on the viewbinding library.

ouchadam:
Usually a `invalid caches and restart` fixes these types of problems. Using Bumblebee 2021.1.1 Beta 2 and haven't seen this issue for a while~

Bubu:
Let me try the beta.

Bubu:
Well, that seems to work fine, so either this is some residual caching issue on my stable AS install or some bug they fixed in the upcoming version.

Thanks for the help and sorry for the noise. :-)

PR #4427, opened 2021-11-08
Add an explicit reason when shortcut are disabled.

bmarty:
Also prefer using ShortcutManagerCompat to disable shortcuts

After a room for which a shortcut has been pinned is left, when clicking on the disabled shortcut, we can see a Toast with:

### Before

Default message from the system

image

### After

Explicit message

image

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/ShortcutsHandler.kt
+++ b/vector/src/main/java/im/vector/app/features/home/ShortcutsHandler.kt
@@ -21,8 +21,10 @@ import android.content.pm.ShortcutManager
 import android.os.Build
 import androidx.core.content.getSystemService
 import androidx.core.content.pm.ShortcutManagerCompat
+import im.vector.app.R
 import im.vector.app.core.di.ActiveSessionHolder
 import im.vector.app.core.dispatchers.CoroutineDispatchers
+import im.vector.app.core.resources.StringProvider
 import im.vector.app.features.pin.PinCodeStore
 import im.vector.app.features.pin.PinCodeStoreListener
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/home/ShortcutsHandler.kt
```diff
@@ -130,8 +137,15 @@ class ShortcutsHandler @Inject constructor(
...
-                        ?.let {
-                            it.disableShortcuts(it.pinnedShortcuts.map { pinnedShortcut -> pinnedShortcut.id })
+                        ?.pinnedShortcuts
+                        ?.takeIf { it.isNotEmpty() }
+                        ?.map { pinnedShortcut -> pinnedShortcut.id }
+                        ?.let { shortcutIdsToDisable ->
+                            ShortcutManagerCompat.disableShortcuts(
+                                    context,
+                                    shortcutIdsToDisable,
+                                    stringProvider.getString(R.string.shortcut_disabled_reason_sign_out)
```
ouchadam, line 147: to double check, is `session has been signed out` the right message or should it be something to do with the room pin?

ouchadam (approved):
nice catch! didn't realise we could provide a custom reason

small question around the copy, otherwise looking good :100:

vector/src/main/java/im/vector/app/features/home/ShortcutsHandler.kt
bmarty: this fun `clearShortcuts()` is called during sign out so I think the message is OK.
When user click on the (greyed out) pin, this message will be displayed

PR #4428, opened 2021-11-08
Fix potential NPE on Optional objects

Florian14:
`toOptional()` function was safe called (prefixed by `?.`) causing the resulting Optional object to be null instead of encapsulating the null value inside the Optional object (which is the purpose of this object).


bmarty (approved):
Good catch, thanks!

PR #4429, opened 2021-11-08
Sign out - Sign in sanity check & nightly sanity check

ouchadam:
- Reintroduces the sign out/re-signin flow to the sanity tests 
- Fixes multiple sign outs occurring whilst signing out during a sync task execution 
- Changes the Sanity check to execute on a nightly basis instead of on each pull request push
  - Updates the action runner to the use hardware accelerated OSX runner 
  - Fixes bad argument placement for the `--no-rate-limit` option

| GIF | 
| --- | 
|after-sanity-2

### DIFF
--- a/.github/workflows/sanity_test.yml
+++ b/.github/workflows/sanity_test.yml
@@ -14,7 +14,7 @@ env:
 jobs:
   integration-tests:
     name: Sanity Tests (Synapse)
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     strategy:
       fail-fast: false
       matrix:
@@ -50,6 +50,10 @@ jobs:
           pip install synapse matrix-synapse
           curl -sL https://raw.githubusercontent.com/matrix-org/synapse/develop/demo/start.sh --no-rate-limit \
             | sed s/127.0.0.1/0.0.0.0/g | bash
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'adopt'
+          java-version: '11'
       - name: Run sanity tests on API ${{ matrix.api-level }}
         uses: reactivecircus/android-emulator-runner@v2
         with:
--- a/.github/workflows/sanity_test.yml
+++ b/.github/workflows/sanity_test.yml
@@ -1,9 +1,9 @@
 name: Sanity Test
 
 on:
-  pull_request: { }
-  push:
-    branches: [ main, develop ]
+  schedule:
+    # At 20:00 every day UTC
+    - cron: '0 20 * * *'
 
 # Enrich gradle.properties for CI/CD
 env:
@@ -21,6 +21,8 @@ jobs:
         api-level: [28]
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: develop
       - name: Set up Python 3.8
         uses: actions/setup-python@v2
         with:
--- a/.github/workflows/sanity_test.yml
+++ b/.github/workflows/sanity_test.yml
@@ -48,8 +48,8 @@ jobs:
           python3 -m venv .synapse
           source .synapse/bin/activate
           pip install synapse matrix-synapse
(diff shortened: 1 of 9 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/SessionManager.kt
```diff
@@ -51,6 +51,13 @@ internal class SessionManager @Inject constructor(private val matrixComponent: M
         }
     }
 
+    fun stopSession(sessionId: String) {
+        if (sessionComponents.containsKey(sessionId).not()) {
+            throw RuntimeException("You don't have a session for id $sessionId")
+        }
+        sessionComponents[sessionId]!!.session().stopSync()
```
bmarty, line 58: To avoid usage of `!!`, maybe store `sessionComponents[sessionId]` in a val and test for nullability?

vector/src/androidTest/java/im/vector/app/ui/robot/ElementRobot.kt
```diff
@@ -88,9 +96,42 @@ class ElementRobot {
     }
 
     fun signout() {
-        OnboardingRobot().signout()
+        clickOn(R.id.groupToolbarAvatarImageView)
+        clickOn(R.id.homeDrawerHeaderSignoutView)
+
+        val hasSentMessages = kotlin.runCatching {
+            waitUntilViewVisible(withId(R.id.exitAnywayButton))
+        }.isSuccess
```
bmarty, line 104: That's a clever way to do it, but which will never make the test fail if it should fail. Maybe add a param `expectSignOutWarning: Boolean` to compare with?

.github/workflows/sanity_test.yml
```diff
@@ -46,8 +48,12 @@ jobs:
           python3 -m venv .synapse
           source .synapse/bin/activate
           pip install synapse matrix-synapse
-          curl -sL https://raw.githubusercontent.com/matrix-org/synapse/develop/demo/start.sh --no-rate-limit \
-            | sed s/127.0.0.1/0.0.0.0/g | bash
+          curl -sL https://raw.githubusercontent.com/matrix-org/synapse/develop/demo/start.sh \
+            | sed s/127.0.0.1/0.0.0.0/g | sed 's/http:\/\/localhost/http:\/\/10.0.2.2/g' | bash -s -- --no-rate-limit
```
bmarty, line 52: The issue `--no-rate-limit` has already been fixed somewhere, probably in a not-merged PR... Good catch anyway

bmarty (approved):
Thanks, a few minor remarks, not blocking the merge

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/SessionManager.kt
ouchadam: good point :+1:

vector/src/androidTest/java/im/vector/app/ui/robot/ElementRobot.kt
ouchadam: I originally had the `hasSentMessages` as a parameter but the verification popup and exit flow isn't always consistent 

Adding a warning log sounds like the best of both worlds, will add :+1:

.github/workflows/sanity_test.yml
ouchadam: yeahh it was fixed here but it caused 3/4 hour+ runs so I ended up not merging

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/SessionManager.kt
ouchadam: done 9c1d6e0484c33d6a471790a9017db57954babb40

ouchadam:
@bmarty once the CI is happen I'll merge and hopefully we'll have some results tomorrow morning :crossed_fingers:

PR #4433, opened 2021-11-09
Android12

### LINKED ISSUE #4262: Support Android 12 (API 31)
https://developer.android.com/about/versions/12

### Things to do

Make the required changed

- [x] Study all the changes from https://developer.android.com/about/versions/12/summary
- [x] Migration recipe: https://developer.android.com/about/versions/12/migration
- [x] Change compileSdk to 31
- [x] Change targetSdk to 31
- [x] Handle change in Bluetooth permission. Related: #3901
- [ ] Handle change regarding deprecated API usage if any
- [ ] List to be continued (or to edit when implementing the required changes)
- [ ] Test the app on a Android 12 emulator
- [ ] Update / check the back press behavior https://developer.android.com/about/versions/12/behavior-changes-all#back-press
- [x] Update the config to sign the APK [here](https://github.com/vector-im/element-android/blob/main/tools/release/sign_apk.sh#L20)...
- [x] ... and [here](https://github.com/vector-im/element-android/blob/main/tools/release/sign_apk_unsafe.sh#L26)

### Effect

Making those change will allow some PRs to be merged, for instance #4241

### To handle Later

All the possible changes to use [new capabilities](url) will be handled in dedicated issues.

bmarty:
Ok, sanity test is passing (a problem at the end, will run it again), but see 0ecf1a4 to be able to run it.

The PR can be reviewed and CI should be happy now.

Notification are not working on the emulator, I'm testing F-Droid version. Will check on GPlay version, but weird. Quick reply from notification has to be tested because the Pending Intent should be MUTABLE in this case (but now it's IMMUTABLE), see https://developer.android.com/guide/components/intents-filters#CreateImmutablePendingIntents . I wanted to test it before applying the fix.

**Edit**: This is probably a pb with the emulator (or with my test account?).
The notification troubleshoot is working fine, the PUSH test is received, but take about 20s to come to the app...
Push tested OK on real device with Android 10.
We should merge the PR and see what happen during the next release.

**Edit 2**: room notification are now working fine thanks to dddcbfbe7d61b29967ee8c8468baea2d27bb49ee

Fixes #4262


vector/src/main/AndroidManifest.xml
```diff
@@ -417,6 +420,22 @@
                 android:name="android.support.FILE_PROVIDER_PATHS"
                 android:resource="@xml/sdk_provider_paths" />
         </provider>
+
+        <!-- Temporary fix for Android 12. android:exported has to be explicitly set when targeting Android 12
+        Do it for services coming from dependencies - BEGIN -->
```
ouchadam, line 425: :100: hopefully the libraries will update their targets soonish~ (especially androidx)

multipicker/src/main/java/im/vector/lib/multipicker/ContactPicker.kt
```diff
@@ -72,10 +73,13 @@ class ContactPicker : Picker<MultiPickerContactType>() {
                                 selection,
                                 selectionArgs,
                                 null
-                        )?.use { cursor ->
-                            while (cursor.moveToNext()) {
-                                val mimeType = cursor.getString(cursor.getColumnIndex(ContactsContract.Data.MIMETYPE))
-                                val contactData = cursor.getString(cursor.getColumnIndex(ContactsContract.Data.DATA1))
+                        )?.use inner@{ innerCursor ->
```
ouchadam, line 76: this is my first time seeing `inner@{`  what does it do?

vector/src/main/java/im/vector/app/AppStateHandler.kt
```diff
@@ -57,7 +56,7 @@ class AppStateHandler @Inject constructor(
         private val sessionDataSource: ActiveSessionDataSource,
         private val uiStateRepository: UiStateRepository,
         private val activeSessionHolder: ActiveSessionHolder
-) : LifecycleObserver {
+) : DefaultLifecycleObserver {
```
ouchadam, line 59: nice clean up! might speed up the lifecycle KAPT processing a tiny bit as well :crossed_fingers:

vector/src/main/java/im/vector/app/features/rageshake/RageShake.kt
```diff
@@ -46,7 +46,7 @@ class RageShake @Inject constructor(private val activity: FragmentActivity,
 
         shakeDetector = ShakeDetector(this).apply {
             setSensitivity(vectorPreferences.getRageshakeSensitivity())
-            start(sensorManager)
+            start(sensorManager, SensorManager.SENSOR_DELAY_GAME)
```
ouchadam, line 49: I don't think we'll have any issues but wanted to highlight this means we've changed from`SENSOR_DELAY_FASTEST` to `SENSOR_DELAY_GAME`

I wonder if `SENSOR_DELAY_UI` could be more suited? haven't tested, assuming from the docs...

```
/** get sensor data as fast as possible */
public static final int SENSOR_DELAY_FASTEST = 0;
/** rate suitable for games */
public static final int SENSOR_DELAY_GAME = 1;
/** rate suitable for the user interface  */
public static final int SENSOR_DELAY_UI = **2;**
```

ouchadam (approved):
awesome stuff! :100:  thanks for updating the blocked dependencies as well

some small comments but no blockers

vector/src/main/java/im/vector/app/features/rageshake/RageShake.kt
bmarty: TBH I haven't checked SENSOR_DELAY_UI. The change was necessary because when using SENSOR_DELAY_FASTEST we need to add a permission in the manifest.
Rageshake still works with SENSOR_DELAY_GAME, I did not investigate more.

multipicker/src/main/java/im/vector/lib/multipicker/ContactPicker.kt
bmarty: "inner" is just a label, it could be anything else. It allows to use `return@inner` below (lines 77 and 78). Using classical `return@use` is not possible there since we are in 2 nested `use` blocks, so is ambiguous.

ouchadam: ah thanks for explaining

vector/src/main/AndroidManifest.xml
bmarty: Yes. Jitsi lib has not been upgrading since a while, and it requires some manual steps. See https://github.com/vector-im/element-android/blob/main/docs/jitsi.md if you are curious (I know you are :) )

PR #4442, opened 2021-11-10
Fix incorrect cropping of conversation icons 

ouchadam:
#4424 with unit tests updated


bmarty (approved):
Thanks for the fix in test compilation

PR #4447, opened 2021-11-10
Fix fallback otk support

### LINKED ISSUE #4446: Unable to establish Olm outbound session from fallback key (BAD_MAC error when verifying signature)
### Steps to reproduce

Try to establish an olm session from an OTK of type fallback

### Intended result and actual result

Signature will fail, because the signable json is only using 'key' where it should use the all json and only remove unsigned and signatures.

Only happens with fallback key as a new `fallback` field is present in the dict:
````
{
    "fallback": true,
    "key": "dQp6YuFhLbZ4...wBq2deKNmdpewSzc",
    "signatures": {
        "@XXXX:matrix.org": {
            "ed25519:EXRRRRRR": "R1NiKUQd+ZhvxMPZiFgcG/6Lm5.....h6RkF2+mdopOw+W8UB4YRBA"
        }
    }
}
````

Current:
````
fun signalableJSONDictionary(): Map<String, Any> {
        return mapOf("key" to value)
    }
   ````
   

Expected from JS sdk:
````
    // prepare the canonical json: remove unsigned and signatures, and stringify with anotherjson
    const mangledObj = Object.assign({}, obj);
    if ("unsigned" in mangledObj) {
        delete mangledObj.unsigned;
    }
    delete mangledObj.signatures;
    const json = anotherjson.stringify(mangledObj);
````

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

### PR DESCRIPTION
**Pull request by BillCarsonFr**
Fixes #4446

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/MXKey.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/MXKey.kt
@@ -38,14 +38,18 @@ data class MXKey(
         /**
          * signature user Id to [deviceid][signature]
          */
-        private val signatures: Map<String, Map<String, String>>
+        private val signatures: Map<String, Map<String, String>>,
+
+        private val rawMap: JsonDict
 ) {
 
     /**
      * @return the signed data map
      */
     fun signalableJSONDictionary(): Map<String, Any> {
-        return mapOf("key" to value)
+        return rawMap.filter {
+            it.key != "signatures" && it.key != "unsigned"
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/MXKey.kt [hunk 1]
```diff
@@ -38,14 +38,18 @@ data class MXKey(
         /**
          * signature user Id to [deviceid][signature]
          */
-        private val signatures: Map<String, Map<String, String>>
+        private val signatures: Map<String, Map<String, String>>,
+
+        private val rawMap: JsonDict
```
bmarty, line 43: Can you add doc like other fields please?

bmarty (approved):
LGTM on the impl side (not sure what it is doing/fixing though). Just a remark about doc, which could help me to get some context actually

BillCarsonFr:
> LGTM on the impl side (not sure what it is doing/fixing though). Just a remark about doc, which could help me to get some context actually


There are more details in the linked issue.
But basically when we need to share a megolm key and that we don't have an existing olm session with a target device we need to create one.

This is done by claiming a one time key to establish outbound session.
But before creating the olm session we first check the otk signature. And in the case of the fallback key the signature check will currently fail because we don't include the correct fileds in the json that is signed

The code that gets the signable json is wrong `signalableJSONDictionary`it was adding only the `key`hardcoded. Instead of removing unsigned and signature from the raw otk json

bmarty (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/MXKey.kt
BillCarsonFr: added

PR #4370, opened 2021-10-29
Background Crypto store usage

ouchadam:
Fixes some parts of the crypto store flow being on the main thread, this will help address a bunch of `RealmCryptoStore` related ANRs as we'll no longer be calling `Realm.getInstance` on the main thread when interacting with the crypto store.

- Fixed by lazily initialising any `init` logic
- Moving view model `init` logic to `Start` events

This could break all sorts of things, needs a bunch more testing!

<details>
<summary>Before - Sign in</summary>

```
getOrCreateOlmAccount main
getDeviceTrackingStatuses Crypto_Thread
getPendingIncomingGossipingRequests main
getMyCrossSigningInfo main
getCrossSigningInfo main
open Crypto_Thread
areDeviceKeysUploaded Crypto_Thread
areDeviceKeysUploaded Crypto_Thread
getMyCrossSigningInfo Crypto_Thread
getCrossSigningInfo Crypto_Thread
getUserDevices Crypto_Thread
storeUserDevices Crypto_Thread
getCrossSigningInfo main
getUserDeviceList main
getCrossSigningInfo DefaultDispatcher-worker-6
getCrossSigningInfo DefaultDispatcher-worker-4
getCrossSigningPrivateKeys DefaultDispatcher-worker-9
getUserDeviceList main
getUserDeviceList DefaultDispatcher-worker-8
getMyDevicesInfo DefaultDispatcher-worker-5
getCrossSigningPrivateKeys DefaultDispatcher-worker-3
getUserDeviceList DefaultDispatcher-worker-6
setDeviceKeysUploaded Crypto_Thread
saveOlmAccount Crypto_Thread
saveMyDevicesInfo main
saveOlmAccount Crypto_Thread
saveOlmAccount Crypto_Thread
saveOlmAccount Crypto_Thread
saveOlmAccount Crypto_Thread
saveOlmAccount Crypto_Thread
saveOlmAccount Crypto_Thread
saveOlmAccount Crypto_Thread
saveOlmAccount Crypto_Thread
open Crypto_Thread
tidyUpDataBase Crypto_Thread
getDeviceTrackingStatuses Crypto_Thread
getDeviceTrackingStatuses Crypto_Thread
saveDeviceTrackingStatuses Crypto_Thread
getDeviceTrackingStatuses Crypto_Thread
getDeviceTrackingStatuses Crypto_Thread
saveDeviceTrackingStatuses Crypto_Thread
shouldEncryptForInvitedMembers Crypto_Thread
[...]

PR #4459, opened 2021-11-12
Clean group key share + logs + robustness

BillCarsonFr:
Some quick improvment for key sharing.

1) Some cleaning on key share by slice, stop using reccursion use chunks
2) More reliable key delivery by using a fallback worker in case the initial send to device failed
2) Added a bit more logs, and more consistent with what js sdk is logging to improve RS handling.
3) ~~Also added some basic @Synchronized when using olm lib as it's not thread safe (account access should be locked, inboud session also)...~~

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
@@ -119,7 +119,7 @@ internal class MXOlmDevice @Inject constructor(
     /**
      * @return The current (unused, unpublished) one-time keys for this account.
      */
-    fun getOneTimeKeys(): Map<String, Map<String, String>>? {
+    @Synchronized fun getOneTimeKeys(): Map<String, Map<String, String>>? {
         try {
             return store.getOlmAccount().oneTimeKeys()
         } catch (e: Exception) {
@@ -132,7 +132,7 @@ internal class MXOlmDevice @Inject constructor(
     /**
      * @return The maximum number of one-time keys the olm account can store.
      */
-    fun getMaxNumberOfOneTimeKeys(): Long {
+    @Synchronized fun getMaxNumberOfOneTimeKeys(): Long {
         return store.getOlmAccount().maxOneTimeKeys()
     }
 
@@ -153,7 +153,7 @@ internal class MXOlmDevice @Inject constructor(
      * @param message the message to be signed.
      * @return the base64-encoded signature.
      */
-    fun signMessage(message: String): String? {
+    @Synchronized fun signMessage(message: String): String? {
         try {
             return store.getOlmAccount().signMessage(message)
         } catch (e: Exception) {
@@ -166,7 +166,7 @@ internal class MXOlmDevice @Inject constructor(
     /**
      * Marks all of the one-time keys as published.
      */
-    fun markKeysAsPublished() {
+    @Synchronized fun markKeysAsPublished() {
         try {
             store.getOlmAccount().markOneTimeKeysAsPublished()
             store.saveOlmAccount()
@@ -180,7 +180,7 @@ internal class MXOlmDevice @Inject constructor(
      *
      * @param numKeys number of keys to generate
      */
-    fun generateOneTimeKeys(numKeys: Int) {
+    @Synchronized fun generateOneTimeKeys(numKeys: Int) {
         try {
             store.getOlmAccount().generateOneTimeKeys(numKeys)
             store.saveOlmAccount()
@@ -197,7 +197,7 @@ internal class MXOlmDevice @Inject constructor(
      * @param theirOneTimeKey  the remote user's one-time Curve25519 key
      * @return the session id for the outbound session.
      */
-    fun createOutboundSession(theirIdentityKey: String, theirOneTimeKey: String): String? {
+    @Synchronized fun createOutboundSession(theirIdentityKey: String, theirOneTimeKey: String): String? {
         Timber.v("## createOutboundSession() ; theirIdentityKey $theirIdentityKey theirOneTimeKey $theirOneTimeKey")
         var olmSession: OlmSession? = null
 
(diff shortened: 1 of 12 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/EnsureOlmSessionsForDevicesAction.kt
```diff
@@ -33,29 +35,39 @@ internal class EnsureOlmSessionsForDevicesAction @Inject constructor(
 
     suspend fun handle(devicesByUser: Map<String, List<CryptoDeviceInfo>>, force: Boolean = false): MXUsersDevicesMap<MXOlmSessionResult> {
         val devicesWithoutSession = ArrayList<CryptoDeviceInfo>()
+        val devicesWithSession = MXUsersDevicesMap<MXOlmSessionResult>()
 
-        val results = MXUsersDevicesMap<MXOlmSessionResult>()
+//        val results = MXUsersDevicesMap<MXOlmSessionResult>()
```
bmarty, line 40: can you just remove this line?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/EnsureOlmSessionsForDevicesAction.kt
```diff
@@ -33,29 +35,39 @@ internal class EnsureOlmSessionsForDevicesAction @Inject constructor(
...
 
         for ((userId, deviceInfos) in devicesByUser) {
             for (deviceInfo in deviceInfos) {
                 val deviceId = deviceInfo.deviceId
                 val key = deviceInfo.identityKey()
+                if (key == null) {
+                    Timber.w("## CRYPTO | Ignoring device (${deviceInfo.userId}|$deviceId) without identity key ")
```
bmarty, line 47: (minor) can you trim this string please? (extra space at the end)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/EnsureOlmSessionsForDevicesAction.kt
```diff
@@ -68,41 +80,80 @@ internal class EnsureOlmSessionsForDevicesAction @Inject constructor(
...
-                            continue
-                        }
-                        // Update the result for this device in results
-                        olmSessionResult.sessionId = verifyKeyAndStartSession(oneTimeKey, userId, deviceInfo)
+
+        // We iterate through claimed info to log missing otks
+        usersDevicesToClaim.forEach { userId, deviceId, _ ->
+            val foundOtk = oneTimeKeys.getObject(userId, deviceId)
+            if (foundOtk == null) {
+                Timber.d("## CRYPTO: No one time key for $userId|$deviceId")
```
bmarty, line 94: Maybe for another PR. Instead of using "## CRYPTO", you should use the .tag facility, see https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/logger/LoggerTag.kt#L25

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/EnsureOlmSessionsForDevicesAction.kt
```diff
@@ -68,41 +80,80 @@ internal class EnsureOlmSessionsForDevicesAction @Inject constructor(
...
+//                                    " for device " + userId + "|" + deviceId)
+//                            continue
+//                        }
+//                        // Update the result for this device in results
+//                        olmSessionResult.sessionId = verifyKeyAndStartSession(oneTimeKey, userId, deviceInfo)
+//                    }
+//                }
+//            }
+//        }
+//        return results
```
bmarty, line 156: Remove this commented out code?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/MXUsersDevicesMap.kt
```diff
@@ -129,3 +129,11 @@ inline fun <T> MXUsersDevicesMap<T>.forEach(action: (String, String, T) -> Unit)
         }
     }
 }
+
+fun <T> MXUsersDevicesMap<T>.toDebugString() =
+        map.entries.joinToString { "${it.key} [${it.value.keys.joinToString { it }}]" }
+
+fun <T> MXUsersDevicesMap<T>.toDebugCount() =
```
bmarty, line 136: Those two extensions should be kept internal I think

bmarty (commented):
LGTM, a few remarks on the form

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -119,7 +119,7 @@ internal class MXOlmDevice @Inject constructor(
     /**
      * @return The current (unused, unpublished) one-time keys for this account.
      */
-    fun getOneTimeKeys(): Map<String, Map<String, String>>? {
+    @Synchronized fun getOneTimeKeys(): Map<String, Map<String, String>>? {
```
ouchadam, line 122: I'm a little weary of making all these functions synchronised vs ensuring they're only called from the crypto thread, at the moment some of these are called from the main thread although https://github.com/vector-im/element-android/pull/4370 should avoid ANRs at least :crossed_fingers: 

each time we synchronise we increase the risk of a deadlock :fearful:

BillCarsonFr: Ok I revert it, I don't want to spend too much time on that as we are moving to rust.
But I wonder if it's not what is causing the strange BAD_MAC when decrypting

BillCarsonFr:
Sorry @bmarty , I'll close this PR and cut it into separate PR

PR #4470, opened 2021-11-15
Clean up styles

bmarty:
Small cleanup after #4360


onurays (approved):
Nice clean up, thanks!

PR #4480, opened 2021-11-16
Sign out crash - Realm configuration mismatch 

ouchadam:
Fixes an intermittent crash on sign out caused by a race condition where we attempt to access an already closed session causing it to be recreated which in turn causes realm to crash because the new session is slightly different.

the race condition boils down to...

```kotlin
sessionComponents.remove(sessionId)

runOnUiThread {
  if (sessionParamsStore[sessionId] != null) {
    val session = sessionComponents.getOrCreate(sessionId).session()
    onStopped(session)
  }
} 

sessionParamsStore.delete(sessionId)
```

- The solution is to avoid using the `SessionManager` within the `SessionListeners` and instead pass the current session directly (to avoid creating a new one), which we were already doing in most cases

| BEFORE | AFTER |
| --- | --- |
|before-signout|after-signout-crash

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
@@ -174,8 +174,8 @@ internal class DefaultSession @Inject constructor(
             lifecycleObservers.forEach {
                 it.onSessionStarted(this)
             }
-            sessionListeners.dispatch { _, listener ->
-                listener.onSessionStarted(this)
+            dispatchTo(sessionListeners) { session, listener ->
+                listener.onSessionStarted(session)
             }
         }
     }
@@ -217,8 +217,8 @@ internal class DefaultSession @Inject constructor(
         // timelineEventDecryptor.destroy()
         uiHandler.post {
             lifecycleObservers.forEach { it.onSessionStopped(this) }
-            sessionListeners.dispatch { _, listener ->
-                listener.onSessionStopped(this)
+            dispatchTo(sessionListeners) { session, listener ->
+                listener.onSessionStopped(session)
             }
         }
         cryptoService.get().close()
@@ -249,8 +249,8 @@ internal class DefaultSession @Inject constructor(
             lifecycleObservers.forEach {
                 it.onClearCache(this)
             }
-            sessionListeners.dispatch { _, listener ->
-                listener.onClearCache(this)
+            dispatchTo(sessionListeners) { session, listener ->
+                listener.onClearCache(session)
             }
         }
         withContext(NonCancellable) {
@@ -260,8 +260,8 @@ internal class DefaultSession @Inject constructor(
     }
 
     override fun onGlobalError(globalError: GlobalError) {
-        sessionListeners.dispatch { _, listener ->
-            listener.onGlobalError(this, globalError)
+        dispatchTo(sessionListeners) { session, listener ->
+            listener.onGlobalError(session, globalError)
         }
     }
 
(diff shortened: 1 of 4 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/DefaultSession.kt
```diff
@@ -174,8 +174,8 @@ internal class DefaultSession @Inject constructor(
             lifecycleObservers.forEach {
                 it.onSessionStarted(this)
             }
-            sessionListeners.dispatch { _, listener ->
-                listener.onSessionStarted(this)
```
ouchadam, line 178: as we were already avoiding the `session` parameter in the callback it felt like the `SessionListeners` didn't actually need to hold the session itself

this also means that the `close` flow can safely use the session instance without worrying about it dissapearing from the `SessionManager` which would cause `onSessionStopped` to not be called

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
```diff
@@ -50,7 +51,7 @@ private const val GET_GROUP_DATA_WORKER = "GET_GROUP_DATA_WORKER"
 
 internal class SyncResponseHandler @Inject constructor(
         @SessionDatabase private val monarchy: Monarchy,
-        @SessionId private val sessionId: String,
+        private val session: Session,
```
ouchadam, line 54: I wasn't sure about this change, technically this works as the `SyncThread` uses a `Dagger Provider` so the instance is always correct but it feels a little overkill~

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionListeners.kt
```diff
@@ -42,18 +38,19 @@ internal class SessionListeners @Inject constructor(
         }
     }
 
-    fun dispatch(block: (Session, Session.Listener) -> Unit) {
+    fun dispatch(session: Session, block: (Session, Session.Listener) -> Unit) {
```
ouchadam, line 41: this is the fix, instead of relying on `sessionManager.getSessionComponent(sessionId)?.session()` which may create a new session if one doesn't exist, the callers provide a known session instance

bmarty:
This black icon is strange, maybe due to the video recording/compression?

image

ouchadam:
> This black icon is strange, maybe due to the video recording/compression?
> 
> image

I'm not actually sure, I thought it was meant to be like this :sweat_smile: 

going frame by frame I can see it transition from white to black


| | | | | 
| --- | --- | --- | --- |
2021-11-16T15:34:13,389376883+00:00|2021-11-16T15:34:22,919342615+00:00|2021-11-16T15:34:33,926603071+00:00|2021-11-16T15:34:45,238805858+00:00

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
ouchadam: we need a session instance in order to call 

```kotlin
sessionListeners.dispatch(session) { session, listener ->
    listener.onNewInvitedRoom(session, roomId)
}
```

I could pass the `SessionManager` to the `SyncResponseHandler` like the `SessionListeners` was previously doing

ouchadam: https://github.com/vector-im/element-android/pull/4480/commits/405cda7a2f2aaf4368a474a9a4620d893fe934f9 it's an inlined version of what the `SessionListeners` was doing

ganfra: I think we could also replace onNewInvitedRoom by listening to the Flow<SyncResponse> instead

bmarty: @ganfra yes, please do that in another PR :p

bmarty (approved):
Thanks for the update. I will squash the commit to cleanup the git history.

Issue #2299, opened 2020-10-26
Two elements in the task switcher


turt2live:
**Describe the bug**
![image](https://user-images.githubusercontent.com/1190097/97225268-0e75c880-1798-11eb-9f07-62ee7a3ce873.png)

**To Reproduce**
Unclear - I opened the app to log in and noticed there were two elements in the task/app switcher. The green one (left) just redirects me to the second (right).

**Expected behavior**
Only one element.

**Screenshots**
See above.

**Smartphone (please complete the following information):**
 - Device: Pixel 2
 - OS: Android 11

**Additional context**
 - App version and store: 1.0.9 playstore
 - Homeserver: t2l.io

T-bond:
I can create new instances of the application, if I have one already running, and after that I share something from another app.
The sharing will create a new instance, and will leave it in the recents.

BloodyIron:
I can easily get multiple instances of Element running, but neither of them ask me to login, so I don't think it's tied to logins. I'm on android 7.0.1 by the way. And I think this is also really messing up alerts for Element on my phone too.

notramo:
Calling, or using PIN code will also create new instance.

Bubu:
This doesn't seem quite fixed yet. I still get multiple elements in recents when sharing things to element and others have reported the same happens when opening it from a notification.

BloodyIron:
Yeah I still get multiple instances of Element on Android too. Please re-open @bmarty

anoadragon453:
Note that this still seems to be a problem: https://github.com/vector-im/element-android/issues/2569

Issue #842, opened 2020-01-14
Ensure builds are reproducible


bmarty:
Using full git commit hash as per comment in https://github.com/vector-im/riot-android/issues/3131

PR #4498, opened 2021-11-18
StrandHogg detection

yostyle:


vector/src/main/AndroidManifest.xml
```diff
@@ -111,7 +112,9 @@
                 android:resource="@xml/shortcuts" />
         </activity-alias>
 
-        <activity android:name=".features.home.HomeActivity" />
+        <activity
+            android:name=".features.home.HomeActivity"
+            android:launchMode="singleTask" />
```
ouchadam, line 117: I tried adding `singleTask` to this activity for a deeplink issue and it caused a bunch of odd behaviours (skipping deeplinks, activity not fully loading), have you noticed anything strange after the change?

yostyle: I didn't see any problem on HomeActivity after this change but I didn't stress it. I would prefer to make more tests.

vector/src/main/java/im/vector/app/features/lifecycle/VectorActivityLifecycleCallbacks.kt
```diff
@@ -42,5 +58,81 @@ class VectorActivityLifecycleCallbacks constructor(private val popupAlertManager
     }
 
     override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
+        // restart the app if the task contains an unknown activity
+        if (isTaskCorrupted(activity)) {
```
ouchadam, line 62: would be interesting to know if there's a performance impact for this, might be worth making the check async to avoid slowing down the activity start flow

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
```diff
@@ -520,7 +521,16 @@ class HomeActivity :
         if (views.drawerLayout.isDrawerOpen(GravityCompat.START)) {
             views.drawerLayout.closeDrawer(GravityCompat.START)
         } else {
-            super.onBackPressed()
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R && supportFragmentManager.backStackEntryCount == 0) {
```
ouchadam, line 524: looks like we could extract this logic to reduce the duplication, either extension or base class function 

something like 

```kotlin
if (views.drawerLayout.isDrawerOpen(GravityCompat.START)) { 
  views.drawerLayout.closeDrawer(GravityCompat.START)
} else {
  validateBackPressed { super.onBackPressed() }
}

...

fun Activity.validateBackPressed(onBackPressed: () -> Unit) {
    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R && supportFragmentManager.backStackEntryCount == 0) {
        if (isTaskRoot) {
            onBackPressed()
        } else {
            Timber.e("Application is potentially corrupted by an unknown activity")
            finishAffinity()
        }
    } else {
        onBackPressed()
    }
}

```

vector/src/main/AndroidManifest.xml
bmarty: Indeed, clicking on a link like https://app.element.io/#/room/%23element-web-announcements:matrix.org

creates 2 tasks if Element was already running, so we can see in the task switcher:

image

See https://github.com/vector-im/element-android/issues/2299 and the fix https://github.com/vector-im/element-android/pull/2341 :)

bmarty (changes requested):
See https://github.com/vector-im/element-android/pull/4498/files#r754454956

It's worth noting that the app seems robust to the StrandHogg vulnerability with the fix. I have the error log when I try the attack.

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
yostyle: Done

vector/src/main/java/im/vector/app/features/lifecycle/VectorActivityLifecycleCallbacks.kt
yostyle: Done

vector/src/main/java/im/vector/app/features/lifecycle/VectorActivityLifecycleCallbacks.kt
```diff
@@ -16,31 +16,127 @@
...
+     * @return true if the activity is potentially malicious
+     */
+    private fun isPotentialMaliciousActivity(context: Context, activity: ComponentName): Boolean {
+        return try {
+            val taskAffinity = context.packageManager.getActivityInfo(activity, 0).taskAffinity
+
+            // Check whether the task affinity matches with the package name
+            if (!taskAffinity.isNullOrEmpty() && taskAffinity == context.packageName) {
+                // Check whether the activity is legitimate (declared in the manifest)
+                !activitiesInfo.any { it.name == activity.className }
```
Florian14, line 132: Just a small improvement
```suggestion
                activitiesInfo.none { it.name == activity.className }
```

vector/src/main/java/im/vector/app/core/extensions/Activity.kt
```diff
@@ -114,6 +116,19 @@ fun AppCompatActivity.hideKeyboard() {
     currentFocus?.hideKeyboard()
 }
 
+fun AppCompatActivity.validateBackPressed(onBackPressed: () -> Unit) {
```
Florian14, line 119: Please add documentation to explain that this should be only used for top-level activities. A developer which is not aware of this could think that it's a good practice to use it for all activities.

yostyle: Done

vector/src/main/java/im/vector/app/features/lifecycle/VectorActivityLifecycleCallbacks.kt
yostyle: Done

vector/src/main/AndroidManifest.xml
ouchadam: is this a blocker for the fix? maybe `singleTop` could provide similar protection without allowing multiple instances

yostyle: No more needed to change launchMode since we generate a random task affinity on each build.

bmarty (commented):
Maybe add the suffix to the taskAffinity of `VectorCallActivity`?

vector/build.gradle
```diff
@@ -91,6 +91,17 @@ static def getFdroidVersionSuffix() {
...
 }
 
+static def getTaskAffinitySuffix() {
+    String alphaCharset = (('A'..'Z') + ('a'..'z')).join()
+    String alphaNumCharset = (('A'..'Z') + ('0'..'9') + ('a'..'z')).join()
+    Random rnd = new Random()
+
+    return alphaCharset[rnd.nextInt(alphaCharset.length())] +
+            rnd.with {
+                (1..7).collect { alphaNumCharset[nextInt(alphaNumCharset.length())] }.join()
```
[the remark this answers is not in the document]
yostyle: done

vector/build.gradle
```diff
@@ -91,6 +91,17 @@ static def getFdroidVersionSuffix() {
     }
 }
 
+static def getTaskAffinitySuffix() {
```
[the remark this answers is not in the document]
yostyle: done

ouchadam (approved):

vector/src/main/java/im/vector/app/core/extensions/Activity.kt
bmarty: Funny: I had the same mental path "A developer which is not aware of this could think that it's a good practice to use it for all activities". And now I see this comment in GitHub, and the comment in the code :)

vector/build.gradle
```diff
@@ -131,6 +136,9 @@ android {
         // Required for sonar analysis
         versionName "${versionMajor}.${versionMinor}.${versionPatch}-sonar"
 
+        // Generate a random app task affinity
+        manifestPlaceholders = [appTaskAffinitySuffix:"${generateTaskAffinitySuffix()}"]
```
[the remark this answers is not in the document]
bmarty: Ah, something else I did not think about.
We want to have stable builds, i.e. the same code generates the same binary (related: #842, and context in https://github.com/vector-im/riot-android/issues/3131). It's important for F-Droid.
With this change it will not be the case anymore.
I have not idea how to handle that now.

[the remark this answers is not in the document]
ouchadam: could we use the a short git hash for the suffix? :thinking:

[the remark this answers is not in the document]
yostyle: We should propably compute the task affinity suffix from tags or commit hash ?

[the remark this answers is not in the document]
yostyle: > could we use the short git hash for the suffix? 🤔

We can't use it as is. taskAffinity has same format namespace : 
- It must have at least two segments (one or more dots).
- Each segment must start with a letter.
- All characters must be alphanumeric or an underscore [a-zA-Z0-9_].

https://developer.android.com/studio/build/configure-app-module#set-application-id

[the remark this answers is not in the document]
ouchadam: we could use the git hash as a seed for the `random()`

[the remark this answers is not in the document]
bmarty: That's a good id to compute a suffix using git hash and just remove forbidden chars (but is there any?).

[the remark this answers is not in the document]
bmarty: So many having a `appTaskAffinitySuffix` like `a${git_hash_8_digits}` with a `a` prefix to ensure it's always starting with a letter would fit all our needs. WDYT @yostyle ?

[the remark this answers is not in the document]
yostyle: we already have `gitRevision`. I'm going to reuse it.

bmarty (approved):
Thanks for the update. As per my test, it's working well.
I will squash commits to clean up the history of this branch.

PR #4506, opened 2021-11-18
Non dismissing notifications

### LINKED ISSUE #4505: Notifications are not dismissed when app is killed by the system are then interacted with
### Steps to reproduce

1. Receive a notification
2. Stop the app by swiping it away from the recents menu (or have a battery/memory optimisation stop the app) 
3. Interact with the notification (mark as read, or tapping to enter the room)
4. The action happens but the notification is still shown

#### What did you expect?
The notification to be automatically dismissed

#### What happened instead?
The notification is still show

### Your phone model

_No response_

### Operating system version

_No response_

1.3.6+

### Homeserver

_No response_

### Will you send logs?

No

ouchadam:
Fixes #4505 Notifications not automatically dismissing, this is caused by the in memory tracking being out of sync.

The most likely trigger of this issue is the android OS reclaiming memory and killing the Element app whilst it's idle in the background (which is perfectly normal)  

- The loading of the persisted events was happening before the current session was active which meant we weren't able to read any previously persisted events.
- Preloads the in memory cache of rendered events with the last known persisted events (this fixes the original issue)
- Persists the event queue every time we receive a new event instead of when the app moves to the background, this allows us to handle crash cashes where the app doesn't gracefully move to the background 
- Also extracts out the `NotificationState` and persistence to their own classes/files

| BEFORE | AFTER | 
| --- | --- |
|before-notifications-staying|after-dismiss-notification

### DIFF
--- a/vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
+++ b/vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
@@ -38,12 +38,15 @@ import javax.inject.Singleton
  * Events can be grouped into the same notification, old (already read) events can be removed to do some cleaning.
  */
 @Singleton
-class NotificationDrawerManager @Inject constructor(private val context: Context,
-                                                    private val notificationDisplayer: NotificationDisplayer,
-                                                    private val vectorPreferences: VectorPreferences,
-                                                    private val activeSessionDataSource: ActiveSessionDataSource,
-                                                    private val notifiableEventProcessor: NotifiableEventProcessor,
-                                                    private val notificationRenderer: NotificationRenderer) {
+class NotificationDrawerManager @Inject constructor(
+        private val context: Context,
+        private val notificationDisplayer: NotificationDisplayer,
+        private val vectorPreferences: VectorPreferences,
+        private val activeSessionDataSource: ActiveSessionDataSource,
+        private val notifiableEventProcessor: NotifiableEventProcessor,
+        private val notificationRenderer: NotificationRenderer,
+        private val notificationEventPersistence: NotificationEventPersistence
+) {
 
     private val handlerThread: HandlerThread = HandlerThread("NotificationDrawerManager", Thread.MIN_PRIORITY)
     private var backgroundHandler: Handler
@@ -55,7 +58,7 @@ class NotificationDrawerManager @Inject constructor(private val context: Context
     /**
      * Lazily initializes the NotificationState as we rely on having a current session in order to fetch the persisted queue of events
      */
-    private val notificationState by lazy { NotificationState.createInitialNotificationState(context, currentSession) }
+    private val notificationState by lazy { createInitialNotificationState() }
     private val avatarSize = context.resources.getDimensionPixelSize(R.dimen.profile_avatar_size)
     private var currentRoomId: String? = null
     private val firstThrottler = FirstThrottler(200)
@@ -65,6 +68,14 @@ class NotificationDrawerManager @Inject constructor(private val context: Context
     init {
         handlerThread.start()
         backgroundHandler = Handler(handlerThread.looper)
+    }
+
+    private fun createInitialNotificationState(): NotificationState {
+        val queuedEvents = notificationEventPersistence.loadEvents(currentSession, factory = { rawEvents ->
+            NotificationEventQueue(rawEvents.toMutableList(), seenEventIds = CircularCache.create(cacheSize = 25))
+        })
+        val renderedEvents = queuedEvents.rawEvents().map { ProcessedEvent(ProcessedEvent.Type.KEEP, it) }.toMutableList()
+        return NotificationState(queuedEvents, renderedEvents)
     }
 
     /**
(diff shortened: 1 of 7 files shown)


vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
```diff
@@ -171,85 +149,43 @@ class NotificationDrawerManager @Inject constructor(private val context: Context
...
-            val myUserDisplayName = user?.toMatrixItem()?.getBestName() ?: session.myUserId
-            val myUserAvatarUrl = session.contentUrlResolver().resolveThumbnail(
-                    contentUrl = user?.avatarUrl,
-                    width = avatarSize,
-                    height = avatarSize,
-                    method = ContentUrlResolver.ThumbnailMethod.SCALE
-            )
-            notificationRenderer.render(session.myUserId, myUserDisplayName, myUserAvatarUrl, useCompleteNotificationFormat, eventsToRender)
+            renderEvents(session, eventsToRender)
+            notificationState.persist(context, session)
```
ouchadam, line 164: whilst on the worker thread we persist the queue each time we process new events

vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
```diff
@@ -50,42 +48,24 @@ class NotificationDrawerManager @Inject constructor(private val context: Context
...
+        get() = activeSessionDataSource.currentValue?.orNull()
 
     /**
-     * The last known rendered notifiable events
-     * we keep track of them in order to know which events have been removed from the eventList
-     * allowing us to cancel any notifications previous displayed by now removed events
+     * Lazily initializes the NotificationState as we rely on having a current session in order to fetch the persisted queue of events
      */
-    private var renderedEvents = emptyList<ProcessedEvent<NotifiableEvent>>()
+    private val notificationState by lazy { NotificationState.createInitialNotificationState(context, currentSession) }
```
ouchadam, line 58: not sure if the preloading was working in the past :thinking:  we have to wait for an active session before we can query for the previously persisted events

vector/src/main/java/im/vector/app/features/notifications/NotificationEventPersistence.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.matrix.android.sdk.api.session.Session
+import timber.log.Timber
+import java.io.File
+import java.io.FileOutputStream
+
+// TODO Multi-account
+private const val ROOMS_NOTIFICATIONS_FILE_NAME = "im.vector.notifications.cache"
+private const val KEY_ALIAS_SECRET_STORAGE = "notificationMgr"
+
+object NotificationEventPersistence {
```
bmarty, line 29: Is is possible to inject the constructor with the context to simplify the API?

vector/src/main/java/im/vector/app/features/notifications/NotificationEventQueue.kt
```diff
@@ -19,7 +19,14 @@ package im.vector.app.features.notifications
 import timber.log.Timber
 
 class NotificationEventQueue(
-        private val queue: MutableList<NotifiableEvent> = mutableListOf()
+        private val queue: MutableList<NotifiableEvent> = mutableListOf(),
```
bmarty, line 22: I'm wondering why `NotificationEventQueue` is not a data class, also why `queue` has default value and `seenEventIds` does not.

bmarty (commented):
LGTM, did not test it though. Some small remarks.

vector/src/main/java/im/vector/app/features/notifications/NotificationEventPersistence.kt
ouchadam: will do :+1: it will mean passing the `persistence` into the `NotificationState` methods that use it or lifting those interactions back to the `NotificationManagerDrawer`

vector/src/main/java/im/vector/app/features/notifications/NotificationEventQueue.kt
ouchadam: my original thinking was because the `NotificationEventQueue` is acting as a wrapper around the `List<NotifiableEvent>` and that consumers would use the `rawEvents` for exporting the content instead of comparing equality against the  `NotificationEventQueue` directly

now that you've mentioned it... I can't think any reason not to make it a data class, will change and will also remove the default value to keep all the initialisation in one place :+1:

vector/src/main/java/im/vector/app/features/notifications/NotificationEventPersistence.kt
ouchadam: done https://github.com/vector-im/element-android/pull/4506/commits/7693f3bc6c7607ade8015838e0cbfaf3bdc1160f

vector/src/main/java/im/vector/app/features/notifications/NotificationEventQueue.kt
ouchadam: done https://github.com/vector-im/element-android/pull/4506/commits/55fd362b3d7fbc8d3806c5e4bafaab99bb14c703

bmarty (approved):
Thanks for the update. Looks great!

PR #4512, opened 2021-11-18
Disable sync data jobs in forks

yostyle:
Sync data jobs should not run in forked projects

### DIFF
--- a/.github/workflows/sync-from-external-sources.yml
+++ b/.github/workflows/sync-from-external-sources.yml
@@ -7,6 +7,8 @@ on:
 jobs:
   sync-emojis:
     runs-on: ubuntu-latest
+    # Skip in forks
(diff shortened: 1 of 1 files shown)


ariskotsomitopoulos (approved):

.github/workflows/sync-from-external-sources.yml
```diff
@@ -7,6 +7,8 @@ on:
 jobs:
   sync-emojis:
     runs-on: ubuntu-latest
+    # Skip in forks
+    if: github.repository == 'vector-im/element-android'
```
ariskotsomitopoulos, line 11: Maybe you can also create an `env` variable containing the `vector-im/element-android` at the top of the file not to duplicate the repository name

ariskotsomitopoulos (approved):

PR #4516, opened 2021-11-18
UI test CI tweaks

ouchadam:
- Makes the Sanity github action fail on error and use a Pixel 5 profile
- Adds waits for bottom sheets and dialogs
- Increases default espresso timeouts from 60 seconds to 120
- Increases text length in the edit text check as the runner tends to tap the "(edited)" part which opens a different bottom sheet


bmarty (approved):
LGTM thanks a lot for all this improvements!

PR #4521, opened 2021-11-19
Fix "@JvmStatic used for @Provides function in an object class" Lint warning

bmarty:
See https://github.com/google/dagger/releases/tag/dagger-2.25.2 point 2.ii. : @Module object classes no longer need @JvmStatic on the provides methods.
Ensure this warning does not appear again.

### DIFF
--- a/vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
+++ b/vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
@@ -73,69 +73,58 @@ abstract class VectorBindModule {
 object VectorStaticModule {
 
     @Provides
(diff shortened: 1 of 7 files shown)


vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
```diff
@@ -73,69 +73,58 @@ abstract class VectorBindModule {
 object VectorStaticModule {
 
     @Provides
-    @JvmStatic
```
ouchadam, line 76: :wave: :wave:

ouchadam (approved):

PR #4527, opened 2021-11-19
Voice Draft support

### LINKED ISSUE #3833: [Bug|Crash] Voice Messages sending need more robustfully

ouchadam:
Continuation of @onurays #4237 rebased on all of the refactors around the `VoiceMessageRecorderView` and `MessageComposerViewModel` refactors

- This doesn't include moving any logic to the SDK yet but all the Text + Voice draft logic sits in one place under the `MessageComposerViewModel`
- Also fixes #3833 which was caused by only using a single file for all voice recordings, the fix is to simply use a unique file per voice recording 

| BEFORE | AFTER | 
| --- | --- |
|before-voice-draft|after-voice-draft

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
@@ -23,11 +23,14 @@ package org.matrix.android.sdk.api.session.room.send
  * EDIT: draft of an edition of a message
  * REPLY: draft of a reply of another message
  */
-sealed class UserDraft(open val content: String, open val messageType: String) {
-    data class REGULAR(override val content: String, override val messageType: String) : UserDraft(content, messageType)
-    data class QUOTE(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft(content, messageType)
-    data class EDIT(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft(content, messageType)
-    data class REPLY(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft(content, messageType)
+sealed interface UserDraft {
+    val content: String
+    val messageType: String
+
+    data class REGULAR(override val content: String, override val messageType: String) : UserDraft
+    data class QUOTE(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft
+    data class EDIT(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft
+    data class REPLY(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft
 
     fun isValid(): Boolean {
         return when (this) {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
@@ -24,17 +24,14 @@ package org.matrix.android.sdk.api.session.room.send
  * REPLY: draft of a reply of another message
  */
 sealed interface UserDraft {
-    val content: String
-    val messageType: String
-
-    data class REGULAR(override val content: String, override val messageType: String) : UserDraft
-    data class QUOTE(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft
-    data class EDIT(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft
-    data class REPLY(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft
+    data class Regular(val content: String, val messageType: String) : UserDraft
+    data class Quote(val linkedEventId: String, val content: String, val messageType: String) : UserDraft
+    data class Edit(val linkedEventId: String, val content: String, val messageType: String) : UserDraft
+    data class Reply(val linkedEventId: String, val content: String, val messageType: String) : UserDraft
 
     fun isValid(): Boolean {
         return when (this) {
-            is REGULAR -> content.isNotBlank()
+            is Regular -> content.isNotBlank()
             else       -> true
         }
     }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
@@ -24,7 +24,7 @@ package org.matrix.android.sdk.api.session.room.send
  * REPLY: draft of a reply of another message
  */
 sealed interface UserDraft {
-    data class Regular(val content: String, val messageType: String) : UserDraft
+    data class Regular(val content: String) : UserDraft
     data class Quote(val linkedEventId: String, val content: String) : UserDraft
     data class Edit(val linkedEventId: String, val content: String) : UserDraft
     data class Reply(val linkedEventId: String, val content: String) : UserDraft
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
@@ -25,9 +25,9 @@ package org.matrix.android.sdk.api.session.room.send
  */
 sealed interface UserDraft {
     data class Regular(val content: String, val messageType: String) : UserDraft
-    data class Quote(val linkedEventId: String, val content: String, val messageType: String) : UserDraft
-    data class Edit(val linkedEventId: String, val content: String, val messageType: String) : UserDraft
-    data class Reply(val linkedEventId: String, val content: String, val messageType: String) : UserDraft
+    data class Quote(val linkedEventId: String, val content: String) : UserDraft
+    data class Edit(val linkedEventId: String, val content: String) : UserDraft
+    data class Reply(val linkedEventId: String, val content: String) : UserDraft
 
     fun isValid(): Boolean {
         return when (this) {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
@@ -28,6 +28,7 @@ sealed interface UserDraft {
     data class Quote(val linkedEventId: String, val content: String) : UserDraft
     data class Edit(val linkedEventId: String, val content: String) : UserDraft
     data class Reply(val linkedEventId: String, val content: String) : UserDraft
+    data class Voice(val content: String) : UserDraft
 
     fun isValid(): Boolean {
         return when (this) {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
@@ -23,15 +23,15 @@ package org.matrix.android.sdk.api.session.room.send
  * EDIT: draft of an edition of a message
  * REPLY: draft of a reply of another message
  */
-sealed class UserDraft(open val text: String) {
-    data class REGULAR(override val text: String) : UserDraft(text)
-    data class QUOTE(val linkedEventId: String, override val text: String) : UserDraft(text)
-    data class EDIT(val linkedEventId: String, override val text: String) : UserDraft(text)
-    data class REPLY(val linkedEventId: String, override val text: String) : UserDraft(text)
+sealed class UserDraft(open val content: String, open val messageType: String) {
+    data class REGULAR(override val content: String, override val messageType: String) : UserDraft(content, messageType)
+    data class QUOTE(val linkedEventId: String, override val content: String, override val messageType: String) : UserDraft(content, messageType)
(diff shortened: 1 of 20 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -741,14 +745,32 @@ class MessageComposerViewModel @AssistedInject constructor(
...
     }
 
     private fun handlePauseRecordingVoiceMessage() {
         voiceMessageHelper.pauseRecording()
     }
 
+    private fun handleEntersBackground(composerText: String) {
+        withState {
+            if (it.isVoiceRecording) {
+                handleEndAllVoiceActions(deleteRecord = false)?.toContentAttachmentData()?.let { voiceDraft ->
```
ouchadam, line 764: in order to move all of the voice draft logic to the SDK, we'd also need to move the voice helpers, is that what you had in mind @bmarty ?

DoM1niC:
This will fix @ouchadam ? https://github.com/vector-im/element-android/issues/3833 

in my case this could be reproduce in flight mode and more than one sendet Voice Messages after back online only one will be success send and the other show resend error state but couldn't be resend .... and have to dismiss and new recorded...

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
bmarty: Not really, I would like the SDK be able to manage the draft file storage itself, but not the logic. We can discuss this on Monday.

ouchadam: I could imagine something like... 

```kotlin
override fun initializeRecord(roomId: String) {
  outputFile = draftService.readFile(key = roomId)
}

override fun startRecord(roomId: String) {
  outputFile = draftService.createFile(key = roomId)
}
```

ouchadam:
> This will fix @ouchadam ? #3833
> 
> in my case this could be reproduce in flight mode and more than one sendet Voice Messages after back online only one will be success send and the other show resend error state but couldn't be resend .... and have to dismiss and new recorded...


Not yet, but I have been able to reproduce the issue locally thanks to your steps, I'm look into the cause :crossed_fingers: 

2021-11-22T12:56:54,356286228+00:00


> **Bug in voice-composer Commit:** voice-composer changes brick the Message edit Stuff here with the Info you cant edit Messages while Recording a Voice Message :D ^^

great catch! https://github.com/vector-im/element-android/pull/4527/commits/5687c36c65d488cd48c0ac6e19fa524ad610be97

| BEFORE | AFTER |
| --- | --- |
|before-edit-message|after-edit-message

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
```diff
@@ -23,15 +23,18 @@ package org.matrix.android.sdk.api.session.room.send
...
-sealed class UserDraft(open val text: String) {
-    data class REGULAR(override val text: String) : UserDraft(text)
-    data class QUOTE(val linkedEventId: String, override val text: String) : UserDraft(text)
-    data class EDIT(val linkedEventId: String, override val text: String) : UserDraft(text)
-    data class REPLY(val linkedEventId: String, override val text: String) : UserDraft(text)
+sealed interface UserDraft {
+    val content: String
+    val messageType: String
+
+    data class REGULAR(override val content: String, override val messageType: String) : UserDraft
```
ouchadam, line 30: I'm wondering if `Voice` should be a dedicated draft type rather than a `messageType` sub type :thinking:

bmarty: I think the code would be clearer with a VOICE data class (they are allcaps...). But I do not know if we will end up with many other cases. @onurays WDYT?

vector/src/main/java/im/vector/app/features/voice/AbstractVoiceRecorder.kt
```diff
@@ -57,9 +61,22 @@ abstract class AbstractVoiceRecorder(
...
+        getFilenameFromUri(context, attachmentData.queryUri)?.let {
+            val voiceMessageFolder = File(outputDirectory, roomId.md5())
+            outputFile = File(voiceMessageFolder, it)
+        }
+    }
+
+    override fun startRecord(roomId: String) {
         init()
-        outputFile = File(outputDirectory, "Voice message.$filenameExt")
+        val fileName = "Voice message-${UUID.randomUUID()}.$filenameExt"
```
ouchadam, line 73: using unique files for the recording fixes #3833

bmarty: The filename can in some circumstance be visible to the user, is it possible to use a subfolder instead, and keep a human readable filename?

bmarty: image

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
```diff
@@ -392,7 +393,13 @@ class RoomDetailFragment @Inject constructor(
                 return@onEach
             }
             when (mode) {
-                is SendMode.REGULAR -> renderRegularMode(mode.text)
+                is SendMode.REGULAR -> {
+                    if (mode.messageType == MessageType.MSGTYPE_AUDIO) {
```
bmarty, line 397: This is where the code is not typed enough IMO, MSGTYPE_AUDIO is not necessarily a voice message. But for draft this is the only possible case I think...

vector/src/main/java/im/vector/app/features/voice/AbstractVoiceRecorder.kt
ouchadam: I'll have a look, out of interest do we need to display the real file name here, could the label be hardcoded?

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -741,14 +745,32 @@ class MessageComposerViewModel @AssistedInject constructor(
...
+                    handleSaveDraft(draft = voiceDraft.toJsonString(), messageType = MessageType.MSGTYPE_AUDIO)
+                }
+                setState { it.copy(voiceRecordingUiState = VoiceMessageRecorderView.RecordingUiState.None) }
```
bmarty, line 767: Is it worth (re-)setting the state when entering Background? What about screen rotation?

vector/src/main/java/im/vector/app/features/voice/AbstractVoiceRecorder.kt
```diff
@@ -57,9 +61,22 @@ abstract class AbstractVoiceRecorder(
...
+        val outputDirectoryForRoom = File(outputDirectory, roomId.md5()).apply {
+            if (!exists()) {
+                mkdirs()
```
bmarty, line 76: `mkdirs()` do the existing check, do you remember :)

bmarty (commented):
Nice work! Still some open question about who is responsible to save the draft voice file.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -741,14 +745,32 @@ class MessageComposerViewModel @AssistedInject constructor(
...
 
     private fun handlePauseRecordingVoiceMessage() {
         voiceMessageHelper.pauseRecording()
     }
 
+    private fun handleEntersBackground(composerText: String) {
+        withState {
+            if (it.isVoiceRecording) {
+                handleEndAllVoiceActions(deleteRecord = false)?.toContentAttachmentData()?.let { voiceDraft ->
+                    handleSaveDraft(draft = voiceDraft.toJsonString(), messageType = MessageType.MSGTYPE_AUDIO)
```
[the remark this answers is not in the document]
ouchadam: I'll give it a go :crossed_fingers:

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
ouchadam: will check

vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailFragment.kt
ouchadam: I think this is a sign to move away from the `MSGTYPE` and have a separate `Voice` send mode

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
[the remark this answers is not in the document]
ouchadam: if we pass the file it means the SDK will have parse the vector Uri https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt#L78 which includes interacting with the `multipicker` 

the current recording and draft file are the same instance, the draft itself is the audio file metadata and the uri reference back to the file, other than creating the initial file and reopening I'm not sure we can move much else to the SDK

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
ouchadam: this state update isn't needed, will remove and for rotation the recording stops and the playback composer is displayed until sent or deleted

vector/src/main/java/im/vector/app/features/voice/AbstractVoiceRecorder.kt
ouchadam: https://github.com/vector-im/element-android/pull/4527/commits/cd4e9916ff076c8dcef60d3b619f362e28c7abff

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
ouchadam: it would also mean we can avoid any database entity changes

vector/src/main/java/im/vector/app/features/voice/AbstractVoiceRecorder.kt
bmarty: Yes, this is another way to fix the display issue. But also the file can be downloaded, etc. so a human readable filename should still be nice.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/UserDraft.kt
onurays: Good catch. That make sense and yes we are preserving the `draftMode` in `DraftEntity`. So no migration will be needed.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
bmarty: Nice to have: the recording should not stop on screen rotation... (maybe handle that in another PR)

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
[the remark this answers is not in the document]
bmarty: OK

ouchadam:
closing whilst addressing comments

vector/src/main/java/im/vector/app/features/voice/AbstractVoiceRecorder.kt
ouchadam: https://github.com/vector-im/element-android/pull/4527/commits/c2ec15726cb52fd85f3bbaf08c8c2d39f9752965

We can use the file provider to provide a human readable name, downloads automatically append a timestamp

| NAME | DOWNLOAD | 
| --- | --- |
|after-voice-message|after-downloaded-vm

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
ouchadam: Not stopping the recording will probably mean moving the recording/playback to a service (it would also enable recording/playing whilst the screen is off or in other screens), would need its own ticket/issue

The rotation is already a bit funny in develop~

| ROTATION |
| --- | 
vm-rotation

ouchadam:
noticed some more oddities with rotation, closing again :(

ouchadam:
closing again in favour of breaking this down into smaller PRs #4548 #4547

PR #4537, opened 2021-11-22
Feature/bma/more cleanup

bmarty:
Add a way to force the display of the DialPad (only in debug mode)


vector/src/main/java/im/vector/app/features/home/HomeDetailViewModel.kt
```diff
@@ -100,16 +100,19 @@ class HomeDetailViewModel @AssistedInject constructor(@Assisted initialState: Ho
...
-                }
-            }
+        vectorDataStore.pushCounterFlow.setOnEach { nbOfPush ->
+            copy(
+                    pushCounter = nbOfPush
+            )
         }
+
+        vectorDataStore.forceDialPadDisplayFlow.onEach { force ->
+            forceDialPad = force
```
ganfra, line 113: Why do we have a separate flag boolean?

ganfra (approved):
One remark, otherwise ok!

vector/src/main/java/im/vector/app/features/home/HomeDetailViewModel.kt
bmarty: OK

ganfra (approved):

Issue #153, opened 2019-05-27
Bug / Reply format broken when too many new lines|ws? in <mx-reply>


BillCarsonFr:
This displays correctly

`<mx-reply><blockquote><a href=\"https://matrix.to/#/!ybTPiZFJfZqybWGFeU:matrix.org/$1558973971435611GBlYs:matrix.org\">In reply to</a><a href=\"https://matrix.to/#/@valere35:matrix.org\">@valere35:matrix.org</a><br /><p>copy cou</p>\n</blockquote></mx-reply>ok
`

But not this:

`<mx-reply>\n              <blockquote>\n                <a href=\"https://matrix.to/#/!ybTPiZFJfZqybWGFeU:matrix.org/$1558973683434084MSYOS:matrix.org\">In reply to</a>\n                <a href=\"https://matrix.to/#/@valere35:matrix.org\">@valere35:matrix.org</a>\n                <br />\n                <p>Bonjour</p>\n\n              </blockquote>\n            </mx-reply>\n            salut
`
<img width="234" alt="image" src="https://user-images.githubusercontent.com/9841565/58431167-dc48ce80-80ac-11e9-9794-78b6366fd135.png">

tulir:
Caused by #350 (RiotX uses a markdown parser for HTML rendering) and would technically be fixed by #174 (RiotX doesn't render replies properly, but instead just displays the fallback)

dud1337:
This can be closed

kittykat:
@dud1337 thank you for going through and commenting on issues which have been resolved. If you come across any more that need closing, please feel welcome to tag me in your comment so that I get notifications!

PR #4541, opened 2021-11-22
Fix reply and quote newlines

### LINKED ISSUE #4540: Fix newlines in replies and quotes.
Steps:
1. Reply to or Quote a message
2. In the text include new lines

Expected: Newlines are  respected in the output, as rendered by each client.
Observed: newlines in both the new body and the content of the referenced/quoted message were missing the newlines.

langleyd:
Fixes #4540 

Updated in this PR:
- Adds `force` param to MarkdownParser to skip the check to detect markdown.
- Users the force flag for replies where were know we always want to send the formatted body html.
- Updated the Quote action to use the Markdown parser. I added the quote action to the `DefaultSendService` rather than the `DefaultRelationService` as quotes don't actually create a relationship, just take the original content for a new message. Happy to update that though if preferred the other way.
- I've added a "simple" commonmark parser that only enables BlockQuotes but not other block parsers(lists etc). It will also enables inline parsing such as bold italic(which are not easy to disable). This simple parser gets used specifically for replies and quotes wether the Markdown preference is enabled by the user or not. This will mean that bold, italic, etc markdown will get rendered for quotes and replies, but I think this is probably a lesser evil than quotes and replies not rendering correctly.


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/SendService.kt
```diff
@@ -56,6 +56,16 @@ interface SendService {
...
 
+    /**
+     * Method to quote an events content.
+     * @param quotedEvent The event to which we will quote it's content.
+     * @param text the text message to send
+     * @param msgType the message type: MessageType.MSGTYPE_TEXT (default) or MessageType.MSGTYPE_EMOTE
+     * @param autoMarkdown If true, the SDK will generate a formatted HTML message from the body text if markdown syntax is present
+     * @return a [Cancelable]
+     */
+    fun sendQuotedTextMessage(quotedEvent: TimelineEvent, text: String, autoMarkdown: Boolean = false): Cancelable
```
bmarty, line 67: There is a problem with our current "Quote" action on Element Android. User should be able to edit the quoted text, as per EleWeb.
I should create an issue about it, if it does not exist yet.

So I'm not sure that this new API is necessary.

bmarty (commented):
Thanks for the PR. I do not see any problem, will do some test once the CI has built the APK

langleyd:
Moved the PR back to draft as some of my test cases failed also. Will probably update again.

ouchadam:
@langleyd will this also fix https://github.com/vector-im/element-android/issues/153 ?

langleyd:
Replies and quotes need some markdown parsing to render correctly. I've added a "simple" commonmark parser that only enables `BlockQuotes` but not other block parsers(lists etc). It will also enables inline parsing such as bold italic(which are not easy to disable). This simple parser gets used specifically for replies and quotes wether the Markdown preference is enabled by the user or not. This will mean that bold, italic, etc markdown will get rendered for quotes and replies, but I think this is probably a lesser evil than quotes and replies not rendering correctly.

In future we could invest more in either making improvements to commonmark or exploring alternatives to have more fine-grained control over markdown parsing.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/send/SendService.kt
langleyd: Editable quotes won't really work with the markdown preference disabled. As it's kind of like any other message at that point and we won't parse the block quote. Maybe we could further explore having a simple/advanced mode and explore further can we can more fine grained control over commonmark or another parser.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/send/LocalEchoEventFactory.kt
```diff
@@ -198,20 +198,23 @@ internal class LocalEchoEventFactory @Inject constructor(
...
                                  newBodyText: String,
-                                 newBodyAutoMarkdown: Boolean,
                                  msgType: String,
                                  compatibilityText: String): Event {
         val permalink = permalinkFactory.createPermalink(roomId, originalEvent.root.eventId ?: "", false)
         val userLink = originalEvent.root.senderId?.let { permalinkFactory.createPermalink(it, false) } ?: ""
 
         val body = bodyForReply(originalEvent.getLastMessageContent(), originalEvent.isReply())
+        // As we always supply formatted body for replies we should force the MarkdownParser to produce html.
+        val newBodyFormatted = markdownParser.parse(newBodyText, force = true, advanced = false).takeFormatted()
```
ouchadam, line 208: do you think it's worth extracting the `replyFormatted` creation to avoid the little bits of duplication?https://github.com/vector-im/element-android/pull/4541/files#diff-1576cc22a9d563668ce8071b61b51151ea136025431174a7195872e8e52b4d75R398

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -408,29 +406,13 @@ class MessageComposerViewModel @AssistedInject constructor(
                     popDraft()
                 }
                 is SendMode.Quote   -> {
-                    val messageContent = state.sendMode.timelineEvent.getLastMessageContent()
```
ouchadam, line 411: :100: for lifting this out!

ouchadam (approved):
LGTM :100: (small comment about extracting some duplication but not a blocker)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/send/MarkdownParser.kt
```diff
@@ -27,7 +28,8 @@ import javax.inject.Inject
  * If any change is required, please add a test covering the problem and make sure all the tests are still passing.
  */
 internal class MarkdownParser @Inject constructor(
-        private val parser: Parser,
+        private val advancedParser: Parser,
+        @SimpleCommonmarkParser private val simpleParser: Parser,
```
bmarty, line 32: For parity do you think it's worth creating a AdvancedCommonmarkParser annotation class?

bmarty (approved):
LGTM

bmarty (changes requested):
The tests are not compiling, can you handle that please @langleyd ?

```
 e: /home/runner/work/element-android/element-android/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/session/room/send/MarkdownParserTest.kt: (53, 13): Type mismatch: inferred type is HtmlRenderer! but Parser was expected

e: /home/runner/work/element-android/element-android/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/session/room/send/MarkdownParserTest.kt: (54, 13): Type mismatch: inferred type is TextPillsUtils but HtmlRenderer was expected
e: /home/runner/work/element-android/element-android/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/session/room/send/MarkdownParserTest.kt: (62, 5): No value passed for parameter 'textPillsUtils'
```

Thanks

bmarty (approved):
LGTM

langleyd:
Just came back to fix this and found you cleaned up my mess 😳. Thanks @bmarty !

PR #4548, opened 2021-11-23
Voice related sealed class cleanup

ouchadam:
Extracted from #4527 

- Renames the `UserDraft` and `SendMode` child class names to `CamelCase`
- Removes unused base properties from the sealed classes
- Switches sealed classes to interfaces to help enforce no base properties


bmarty (approved):
LGTM, thanks!

PR #4556, opened 2021-11-24
Supporting rotation during voice recordings

### LINKED ISSUE #4067: Voice Message stops recording on Phone Rotation
### Steps to reproduce

1. Starting a voice Message
2. Rotate the device

### What did you expect?
Voice Message continues
### What happened?
Voice Message stops to record

### Your phone model

Xiaomi Redmi Not 7

### Operating system version

Android 11

Element 1.2.2

### Homeserver

hs.obco.pro

No

ouchadam:
Fixes #4067 Allows the device to rotate whilst recording a voice message

- Makes the `VoiceMessagePlaybackTracker` a singleton to avoid losing any state on activity destruction (means we need to be extra careful with releasing the listeners)
- Holds onto the recording start time in order to maintain the current recording length after rotation
- Checks if the device is actively recording a voice message and going through a configuration change, if so we don't stop the recording or save any drafts

| BEFORE | AFTER |
| --- | --- |
|before-vm-rotation|after-vm-rotation

### DIFF
--- a/vector/src/main/java/im/vector/app/core/time/Clock.kt
+++ b/vector/src/main/java/im/vector/app/core/time/Clock.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) 2021 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.core.time
+
+import javax.inject.Inject
+
+interface Clock {
+    fun epochMillis(): Long
+}
+
+class DefaultClock @Inject constructor() : Clock {
+    override fun epochMillis(): Long {
+        return System.currentTimeMillis()
+    }
+}
--- a/vector/src/main/java/im/vector/app/core/time/Clock.kt
+++ b/vector/src/main/java/im/vector/app/core/time/Clock.kt
@@ -23,6 +23,13 @@ interface Clock {
 }
 
 class DefaultClock @Inject constructor() : Clock {
+
+    /**
+     * Provides a UTC epoch in milliseconds
+     *
+     * This value is not guaranteed to be correct with reality
+     * as a User can override the system time and date to any values.
+     */
     override fun epochMillis(): Long {
         return System.currentTimeMillis()
     }
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/composer/voice/VoiceMessageRecorderView.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/composer/voice/VoiceMessageRecorderView.kt
@@ -105,32 +105,35 @@ class VoiceMessageRecorderView @JvmOverloads constructor(
 
     fun render(recordingState: RecordingUiState) {
         if (lastKnownState == recordingState) return
-        lastKnownState = recordingState
         when (recordingState) {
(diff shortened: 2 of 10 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/composer/voice/VoiceMessageRecorderView.kt
```diff
@@ -158,22 +162,23 @@ class VoiceMessageRecorderView @JvmOverloads constructor(
         dragState = newDragState
     }
 
-    private fun startRecordingTicker() {
+    private fun startRecordingTicker(startFromLocked: Boolean, startAt: Long) {
+        val startMs = ((System.currentTimeMillis() - startAt)).coerceAtLeast(0)
```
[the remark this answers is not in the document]
bmarty: Good idea!

bmarty (approved):
LGTM, I did not test it though.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/voice/VoiceMessageRecorderView.kt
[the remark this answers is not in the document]
ouchadam: added https://github.com/vector-im/element-android/pull/4556/commits/7d2c8e687392c06111c8a527eaf38ccf1a8af455

vector/src/main/java/im/vector/app/core/time/Clock.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import javax.inject.Inject
+
+interface Clock {
+    fun epochMillis(): Long
+}
+
+class DefaultClock @Inject constructor() : Clock {
+    override fun epochMillis(): Long {
+        return System.currentTimeMillis()
```
bmarty, line 27: A bit out of scope of this PR, but we will probably want to replace all `System.currentTimeMillis()` usages by this new Clock?

vector/src/main/java/im/vector/app/core/time/Clock.kt
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.core.time
+
+import javax.inject.Inject
+
+interface Clock {
+    fun epochMillis(): Long
```
bmarty, line 22: It's worth noting that the current impl rely on the system date, which can be set by the user (and at any time).
This can lead to bugs...
Also when the clock change automatically (Winter?Summer time) we can have bugs (my experience is talking here).

bmarty (approved):
Thanks for the update. 2 more remarks but you can merge the PR whenever you want.

vector/src/main/java/im/vector/app/core/time/Clock.kt
ouchadam: unless we use a server time is there a way to avoid this? analytics SDKs will rely on a mix of server _received at_ timestamps and client _sent at_

the `currentTimeMillis` epoch is UTC, unless a user sets a custom time using their local time (eg UTC+11) but with UTC+0 timezone :sweat_smile:   

it's becoming rarer for users to misconfigure their time due to automatic network times and a lot apps/websites start complaining about SSL errors if the current time/date is wildly wrong

vector/src/main/java/im/vector/app/core/time/Clock.kt
ouchadam: :100: that would be the dream! I'll create an issue

vector/src/main/java/im/vector/app/core/time/Clock.kt
ouchadam: https://github.com/vector-im/element-android/pull/4556/commits/fdfac8d20acad014090c90413476e3d7328ce1c7

vector/src/main/java/im/vector/app/core/time/Clock.kt
ouchadam: https://github.com/vector-im/element-android/issues/4562

marclaporte:
I have this issue on Element v1.4.32 (2022-08-10) from F-Droid.

PR #4559, opened 2021-11-24
Analytics

bmarty:
### Content

Analytics framework is set up.
No event is tracked yet, but could by nice to handle that in another PR to let other developer add their analytics Events.

Analytics Optin screen will be displayed for every users when the home screen is displayed for the first time (so either after the first sigin/signup, or after an upgrade of the application).

### Opt in screen

|Dark|Light|
|-|-|
|image|image|

### New section in settings

image

### This PR also contain a debug screen, only accessible on debug build:

image

### TODO

- [x] Iterate on the design for opt in (waiting for @amshakal)
- [ ] Update the policy link (waiting for @deniseal)
- [x] Show the opt in screen from the setting, rather than just having a switch (I am on it).


.github/workflows/sync-from-external-sources.yml
```diff
@@ -70,4 +70,27 @@ jobs:
...
+  sync-analytics-plan:
+    runs-on: ubuntu-latest
+    # Skip in forks
+    if: github.repository == 'vector-im/element-android'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run analytics import script
+        run: ./tools/import_analytic_plan.sh
+      - name: Create Pull Request for analytics plan
+        uses: peter-evans/create-pull-request@v3
```
ouchadam, line 84: another possibility is to `.jar` the generated output as part of the schema repository and upload to github packages, that way dependabot can handle updates and we can ensure the generated artifact compiles in isolation

vector/src/main/java/im/vector/app/features/analytics/AnalyticsConfig.kt
```diff
@@ -0,0 +1,51 @@
...
+
+data class AnalyticsConfig(
+        val postHogHost: String,
+        val postHogApiKey: String
+) {
+    companion object {
+        /**
+         * Read the analytics config from the Build config
+         */
+        fun getConfig(): AnalyticsConfig? {
```
ouchadam, line 31: not for this PR, eventually we might want to wrap and inject the `BuildConfig` rather than directly accessing it (will enable the config to be shared across modules)

.github/workflows/sync-from-external-sources.yml
bmarty: Yes, @pixlwave told me about that possibility. It's probably a cleaner solution, but I have never done that before 🙈

ouchadam (approved):
from what I can tell, this looks good to me (assuming the tracking coming through!)

vector/src/main/java/im/vector/app/features/analytics/AnalyticsConfig.kt
bmarty: I have updated the way developer can configure the analytics in c1438f0a65758d79581bd999e367113b8001cf1a

bmarty:
Updated opt-in screen layout

image

amshakal (approved):

bmarty:
@amshakal 

|7'' tablets|10'' tablets|
|-|-|
|image|image|

We can iterate on the max width for CTA buttons and text block in a dedicated PR, since it will impact other screens of the app.

amshakal:
Thanks for sharing these. This is looking good, I just have two comments:
1. For larger screens, can we increase the padding on left and right of the text, since we have all this vertical space. This will make it easier to read.
2. For the button, would it be more helpful to have % of screen it should cover or pixels? I think capping to 600px for portrait and 800px for landscape mode might work. This works out to be around 60% i.e. the button width is 60% of the screen width. There doesn't seem to be a set guidelines for buttons on tablet.

bmarty:
@amshakal I have added code to limit the width of the content on tablet. We can iterate later for a proper way to support tablet across the whole code base.

amshakal:
Thanks Benoit! I had a discussion with the design team about tablet designs and the plan is to take it up in the new year and have it be included in the design system.

Issue #4488, opened 2021-11-17
Listening for notification still displayed after sign out, just clicked on it


bmarty:
https://github.com/matrix-org/element-android-rageshakes/issues/27987

bmarty:
<img width="487" alt="image" src="https://user-images.githubusercontent.com/3940906/142685835-c36244fd-6b66-4061-9b89-4e763e91ba18.png">

bmarty:
Fixed in 1.3.9

PR #4565, opened 2021-11-26
Dismissing Fdroid variant `Listening for notifications` on sign out

ouchadam:
Fixes https://github.com/vector-im/element-android/issues/4488

- Stops the guard service on clearActiveSession (sign out), the service is already tied to `setActiveSession` 

| BEFORE | AFTER |
| --- | --- |
|before-signout-notifications|after-signout-notification


bmarty (approved):
LGTM, thanks!

PR #4566, opened 2021-11-26
Emoji insertion leftover characters 

### LINKED ISSUE #3449: Emoji autocompleter sometimes leaves search string after replacement
#### Describe the bug
When searching for an emoji via `:` autocompletion sometimes it leaves the search text after inserting the emoji. It only happens occasionally but I've found one reliable way to reproduce it, see below:

#### To Reproduce
1. Type `:tada` (it's important to type exactly this)
2. Select the :tada:  emoji from the completer
3. The message composer will now say :tada:`tada`

#### Expected behavior
Tada text is removed

 - Device: Nexus 5X
 - OS: Android 8.1 (Lineage)

#### Additional context
 - App version and store Schildichat `1.1.7sc33` (I *think* this reproduces with EA as well)
 - Using AOSP keyboard.

ouchadam:
Fixes #3449 Leftover characters when inserting emojis by text.

This was a really strange issue, it seemed to only affect single character look ups eg `:a` but didn't affect all emojis equally, `:z` worked without issue. There's no obvious place where we're incorrectly manipulating the message text content and replacing the `replace` with a `delete + insert` also fixed the issue... perhaps some lingering spans in some cases?  

- Also fixed the emoji replacement always using the last `:` by taking into account the current cursor position

| BEFORE INSERT EMOJI | AFTER INSERT EMOJI |
| --- | --- |
|before-emoji-text|after-emoji-text

| BEFORE POSITION | AFTER POSITION |
| --- | --- |
before-emoji-position|after-emoji-position

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/AutoCompleter.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/AutoCompleter.kt
@@ -182,7 +182,7 @@ class AutoCompleter @AssistedInject constructor(
                 .with(object : AutocompleteCallback<String> {
                     override fun onPopupItemClicked(editable: Editable, item: String): Boolean {
                         // Detect last ":" and remove it
-                        var startIndex = editable.lastIndexOf(":")
+                        var startIndex = editable.subSequence(0, editText.selectionStart).lastIndexOf(":")
                         if (startIndex == -1) {
                             startIndex = 0
                         }
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/AutoCompleter.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/AutoCompleter.kt
@@ -181,7 +181,7 @@ class AutoCompleter @AssistedInject constructor(
                 .with(backgroundDrawable)
                 .with(object : AutocompleteCallback<String> {
                     override fun onPopupItemClicked(editable: Editable, item: String): Boolean {
-                        // Detect last ":" and remove it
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/AutoCompleter.kt
```diff
@@ -182,7 +182,7 @@ class AutoCompleter @AssistedInject constructor(
                 .with(object : AutocompleteCallback<String> {
                     override fun onPopupItemClicked(editable: Editable, item: String): Boolean {
                         // Detect last ":" and remove it
-                        var startIndex = editable.lastIndexOf(":")
+                        var startIndex = editable.subSequence(0, editText.selectionStart).lastIndexOf(":")
```
bmarty, line 185: Maybe update the comment above?

bmarty (approved):
Thanks for those 2 fixes. Just a small comment about a comment :)

vector/src/main/java/im/vector/app/features/home/room/detail/AutoCompleter.kt
ouchadam: https://github.com/vector-im/element-android/pull/4566/files#diff-b7f049abc21b0d7948c0e51a423be888d10aa33cf02111be634b4cd4f741cf8b

Bubu:
Thank you so much for fixing this :heart:!

This has been driving me nuts for such a long time (and I never really could find reliable reproduction steps)... :sweat_smile:

Issue #4144, opened 2021-10-02
Sometimes call stuck with no buttons or a way to hang up


532910:
### Steps to reproduce

I'm not sure this is always-reproducible.

1. Start a voice call (to ios, but not sure)
2. Answer a call on ios
3. Hangup a call on ios -- android will show the call is ended
4. Start a new call:
![image](https://user-images.githubusercontent.com/14617699/135716632-11c6ce2b-71cd-4613-b5fe-f92ef3deacf7.png)
![image](https://user-images.githubusercontent.com/14617699/135716673-a465c0f8-47c6-4791-a1c5-2c5ee624f82b.png)

### Operating system version

Android 11

### Application version and app store

The latest beta from GPlay

### Have you submitted a rageshake?

Yes

532910:
This is a very annoying issue, as it's not trivial how to kill this activity, and phone reboot is the only way for most people.

kittykat:
@532910 which version of Element are you using?

532910:
Sorry for forgetting to specify it, I'm on GPlay Beta channel, thus it was the latest at the time the issue was opened.
I'll try to reproduce it as soon as v1.3.8 becomes available.

kittykat:
I got it on 1.3.8: 
![Screenshot_20211121-221120.png](https://user-images.githubusercontent.com/51663/143217876-40ae4658-3cd0-4440-b93f-b748c9fae7a9.png)

ouchadam:
I'm not entirely sure why this happens yet but the screenshots are showing the call stuck in the `Idle` state. Element isn't actually ringing at this point. 

I can add UI to hang up but will investigate a little more to try and understand why we're getting stuck

ouchadam:
(symptoms) fixed in 1.3.9

PR #4567, opened 2021-11-26
Always showing hang up button in VOIP calls

ouchadam:
Fixes the symptom of #4144 

- Updates the `VectorCallActivity` to always show the hang up button during the initial phases of a VOIP call (`Idle, Create Offer`, matching the `Dialing` behaviour) 

At least from the logs submitted in #4144 there was a native media player error which we're not able to propagate up to our `CallActivity` which means we end up stuck in the `Idle` call state. I can't seem to find a way to hook into player errors via the `RemoteAudioTrack` :thinking:  

The only way I've been able to reproduce this issue is by instantly recreating the call (`Activity.recreate`) when it starts. 

| BEFORE | AFTER |
| --- | --- |
|before-stuck-voip|after-stuck-voip

### DIFF
--- a/changelog.d/4144.bugfix
+++ b/changelog.d/4144.bugfix
@@ -1,0 +1,1 @@
+Allowing users to hang up VOIP calls during the initialisation phase (avoids getting stuck in the call screen if something goes wrong)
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, but maybe @ganfra can double check here?

ganfra (approved):
Yes I guess it's ok like that, thanks!

Issue #4577, opened 2021-11-30
Iterate on the user consent dialog about identity server look up


bmarty:
### What we have now

<img width="332" alt="image" src="https://user-images.githubusercontent.com/3940906/144019952-ba612b79-07c0-47b4-9779-1a8f2ea74476.png">

### Changes

- [x] Wording:

>To discover existing contacts, you need to send contact info **(emails and phone numbers)** to your identity server.*(new lines removed)* We hash your data before sending for privacy.
>*(new lines added)*
>Do you <del>consent</del> **agree** to send this info?

- [x] "POLICY" action is removed, and the policy link(s) are added to the dialog itself, above the question "Do you agree...":

>Privacy policy
>
> - *link*

- [x] "NO" action is replaced by "Not now"
- [x] "YES" action is replaced by "Agree"

### Ref

https://support.google.com/googleplay/android-developer/answer/11150561 - section **User interface (UI)** and **Content**

deniseal:
lgtm @bmarty. Updates to the privacy policy to clarify this are with Design to get added to the website, this is what will be included in clause 1.2:

> Your relationship with identity servers: you might chose to use an identity server, to allow other Matrix users to discover you via the Element app. This is optional and requires your explicit consent to discover users from your personal contacts. For the identity server made available by us, please see the Identity Server Privacy Notice.

@AmandineLP since this technically a policy update, can I get your approval please?

bmarty:
First proposal:

<img width="335" alt="image" src="https://user-images.githubusercontent.com/3940906/144068041-86a0701d-76bd-4dfb-9283-28c0f9201a34.png">

bmarty:
In EA 1.3.9

PR #4587, opened 2021-11-30
Iterate on user consent dialog for identity server

bmarty:
Implements #4577 

New dialogs: https://github.com/vector-im/element-android/issues/4577#issuecomment-982701319

We need to retrieved the T&C, that's why many changes are necessary. Sadly there is some code duplication between the ViewModels, we may find a way to reduce that later.

### DIFF
--- a/vector/src/main/java/im/vector/app/core/utils/Dialogs.kt
+++ b/vector/src/main/java/im/vector/app/core/utils/Dialogs.kt
@@ -17,10 +17,15 @@
 package im.vector.app.core.utils
 
 import android.content.Context
+import android.text.method.LinkMovementMethod
 import android.webkit.WebView
 import android.webkit.WebViewClient
+import android.widget.TextView
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import im.vector.app.R
+import im.vector.app.features.discovery.IdentityServerPolicy
+import me.gujun.android.span.link
+import me.gujun.android.span.span
 
 /**
  * Open a web view above the current activity.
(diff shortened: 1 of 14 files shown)


vector/src/main/java/im/vector/app/features/discovery/Extensions.kt
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package im.vector.app.features.discovery
+
+import im.vector.app.core.utils.ensureProtocol
+import org.matrix.android.sdk.api.session.Session
+import org.matrix.android.sdk.api.session.terms.TermsService
+
+suspend fun Session.fetchIdentityServerWithTerms(userLanguage: String): IdentityServerWithTerms? {
```
ouchadam, line 23: :100: for lifting out of the view model

vector/src/main/java/im/vector/app/features/contactsbook/ContactsBookViewEvents.kt
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.features.contactsbook
+
+import im.vector.app.core.platform.VectorViewEvents
+import im.vector.app.features.discovery.IdentityServerWithTerms
+
+sealed class ContactsBookViewEvents : VectorViewEvents {
```
ouchadam, line 22: worlds smallest nit: could be a sealed interface, saves having to write `()` in the extends :sweat_smile:

vector/src/main/java/im/vector/app/core/utils/Dialogs.kt
```diff
@@ -40,16 +45,36 @@ fun Context.displayInWebView(url: String) {
             .show()
 }
 
-fun Context.showIdentityServerConsentDialog(configuredIdentityServer: String?, policyLinkCallback: () -> Unit, consentCallBack: (() -> Unit)) {
+fun Context.showIdentityServerConsentDialog(identityServerWithTerms: IdentityServerWithTerms?,
```
ouchadam, line 48: general question about identity servers, are privacy policies enforced?

ouchadam (approved):
LGTM :+1:

vector/src/main/java/im/vector/app/features/contactsbook/ContactsBookViewEvents.kt
bmarty: I think at the beginning of the project sealed interface was not a thing, but now, yes. I prefer to perform a global replace on a dedicated PR

vector/src/main/java/im/vector/app/core/utils/Dialogs.kt
bmarty: You mean can we have an IS without any privacy policy? I do not think so, but the code should be robust to that.
By the way I will create some issues (defect, but no P1) I have seen in the area

vector/src/main/java/im/vector/app/features/contactsbook/ContactsBookViewEvents.kt
bmarty: (created https://github.com/vector-im/element-android/issues/4588 about it)

bmarty:
Just pushed a small commit

Issue #4591, opened 2021-11-30
Remove requestModelBuild() from epoxy Controllers init block


ariskotsomitopoulos:
### Steps to reproduce

According to Epoxy documentation, requestModelBuild() should be called from the main thread:

`Note: The first time an EpoxyController builds models it will always run on the main thread, even if you have specified an async handler. This is necessary to allow views to synchronously restore saved state when first created. See Javadoc on EpoxyController#requestModelBuild for more details.`

Invoking `requestModelBuild()` from within epoxy Controller `init{}` block prevents `buildModels()` to run in UI thread for the first time.

### Outcome

Remove `requestModelBuild()` from epoxy Controllers `init{}` block

### Will you send logs?

No

PR #4592, opened 2021-11-30
Feature/aris/epoxy controller init improvement

ariskotsomitopoulos:
Remove `requestModelBuild()` from epoxy Controllers `init{}` block #4591

### DIFF
--- a/changelog.d/4591.misc
+++ b/changelog.d/4591.misc
@@ -1,0 +1,1 @@
+Remove requestModelBuild() from epoxy Controllers init{} block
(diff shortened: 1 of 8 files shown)


ouchadam:
I'm assuming removing this init doesn't break the pages? :crossed_fingers: do we know why it was originally added?

ariskotsomitopoulos:
> I'm assuming removing this init doesn't break the pages? 🤞 do we know why it was originally added?

Yes, the app seems fine without them. I only didn't test `SoftLogoutController.kt`. No idea why they where added, I guess it started on timeline Controller that is an extreme case and we should use that there. And maybe after that it was just copy paste from this controller. Any extra thoughts are welcome

bmarty (approved):
LGTM, thanks!

ouchadam (approved):

Issue #4600, opened 2021-12-01
The very first search in a room sometimes fails.


ariskotsomitopoulos:
### Steps to reproduce

Sometimes there are no displayable results in a room search. When the keyboard is open the results are displayed.
This is happening at the first search after opening the app

### Outcome

Data should be displayed as expected

### Will you send logs?

No

bmarty:
In EA 1.3.10

PR #4601, opened 2021-12-01
Fix issue 4600 that search results sometimes are not displayed properly

ariskotsomitopoulos:
Fixing issue #4600

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/search/SearchResultController.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/search/SearchResultController.kt
@@ -54,10 +54,6 @@ class SearchResultController @Inject constructor(
         fun loadMore()
     }
 
-    init {
-        setData(null)
-    }
-
     override fun buildModels(data: SearchViewState?) {
         data ?: return
 
--- a/changelog.d/4600.misc
+++ b/changelog.d/4600.misc
@@ -1,0 +1,1 @@
+At the very first room search after opening the app sometimes no results are displayed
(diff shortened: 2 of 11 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/search/SearchResultController.kt
```diff
@@ -54,10 +54,6 @@ class SearchResultController @Inject constructor(
         fun loadMore()
     }
 
-    init {
-        setData(null)
-    }
```
bmarty, line 59: I count 9 other controllers with the same code, do you think the same fix should be applied?

image

ariskotsomitopoulos: I think yes, they should be removed. I will update the PR and test those screens

ouchadam (approved):

bmarty:
After a discussion with @ganfra this code was added for performance reason (it should have deserve some more  comments), to not block the UI thread if the controller has many items to compute. It is maybe not necessary anymore, but  how can we be certain of that?

ariskotsomitopoulos:
> After a discussion with @ganfra this code was added for performance reason (it should have deserve some more comments), to not block the UI thread if the controller has many items to compute. It is maybe not necessary anymore, but how can we be certain of that?

Yes that will make the `buildModels` run on background. However, this is equivalent to that #4591. We are certain that the first model build should be in UI thread otherwise the views will not be properly restored. Both the documentation and the code doc explain that. Also, this search result bug is a demonstration that this can produce problems

Issue #4602, opened 2021-12-01
Remove not needed Job canceling in viewModel onCleared()  function


ariskotsomitopoulos:
There is no need to call job.cancel() when we are using viewModelScope()

bmarty:
In EA 1.3.10

PR #4603, opened 2021-12-01
There is no need to call job.cancel() when we are using viewModelScope()

ariskotsomitopoulos:
Fixing issue #4602

### DIFF
--- a/vector/src/main/java/im/vector/app/features/roomdirectory/RoomDirectoryViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/roomdirectory/RoomDirectoryViewModel.kt
@@ -233,7 +233,6 @@ class RoomDirectoryViewModel @AssistedInject constructor(
     }
 
     override fun onCleared() {
-        currentJob?.cancel()
         super.onCleared()
     }
 }
--- a/changelog.d/4602.misc
+++ b/changelog.d/4602.misc
@@ -1,0 +1,1 @@
+There is no need to call job.cancel() when we are using viewModelScope()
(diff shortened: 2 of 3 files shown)


vector/src/main/java/im/vector/app/features/roomdirectory/RoomDirectoryViewModel.kt
```diff
@@ -233,7 +233,6 @@ class RoomDirectoryViewModel @AssistedInject constructor(
     }
 
     override fun onCleared() {
-        currentJob?.cancel()
         super.onCleared()
```
bmarty, line 236: I think we can do more cleanup by removing the fun (at both places)

ouchadam: :+1: for removing the field as well

bmarty: the field is still useful I think

ariskotsomitopoulos: Yes, `override fun onCleared` can be deleted. The field is used in another line

ouchadam: ah I see this directory viewmodel is also using it to dedupe the requests 

```
if (currentJob == null) {
 // load more
}
```

bmarty: I would say this line is more important: https://github.com/vector-im/element-android/blob/main/vector/src/main/java/im/vector/app/features/roomdirectory/RoomDirectoryViewModel.kt#L119

PR #4608, opened 2021-12-01
Fixed voice messages on Android 28 and below

ouchadam:
- Fixing voice messages from _erroring_ on android 28 and below due to attempting to convert already converted audio files. This was caused by the draft flow converting the audio files to `.ogg` but expecting them to still be `.mp4` 

Fixed by only converting to .ogg as part of the sending process

*nexus 4 5.1.1

| AFTER DRAFT | AFTER PLAYBACK |
| --- | --- | 
|Screenshot_20211201_152808|Screenshot_20211201_152823

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
@@ -76,13 +76,17 @@ class VoiceMessageHelper @Inject constructor(
         startRecordingAmplitudes()
     }
 
-    fun stopRecording(): MultiPickerAudioType? {
+    fun stopRecording(convertForSending: Boolean): MultiPickerAudioType? {
         tryOrNull("Cannot stop media recording amplitude") {
             stopRecordingAmplitudes()
         }
         val voiceMessageFile = tryOrNull("Cannot stop media recorder!") {
             voiceRecorder.stopRecord()
-            voiceRecorder.getVoiceMessageFile()
+            if (convertForSending) {
+                voiceRecorder.getVoiceMessageFile()
+            } else {
+                voiceRecorder.getCurrentRecord()
+            }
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
```diff
@@ -76,13 +76,17 @@ class VoiceMessageHelper @Inject constructor(
...
         tryOrNull("Cannot stop media recording amplitude") {
             stopRecordingAmplitudes()
         }
         val voiceMessageFile = tryOrNull("Cannot stop media recorder!") {
             voiceRecorder.stopRecord()
-            voiceRecorder.getVoiceMessageFile()
+            if (convertForSending) {
+                voiceRecorder.getVoiceMessageFile()
+            } else {
+                voiceRecorder.getCurrentRecord()
```
bmarty, line 88: I'm wondering if in this case the MultiPickerAudioType that we are creating below will still be OK once the file is actually converted.

bmarty (approved):
LGTM, will do some test

vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
bmarty: Or we will create a new one. Let me do some test :)

ouchadam: if we continue to use the original uri it'll be invalid as the file will not longer exist, will do some extra testing

bmarty: Tested ok on my API 28 emulator. This fun is called when we send the event so I think this is OK.

PR #4611, opened 2021-12-01
Devx view ids

### LINKED ISSUE #4604: Layout View Ids: use `@id` instead of `@+id` for referenced ids
Using `@id` instead of `@+id` for referenced ids is better since there will be only one declaration for each id, so code navigation will be simplified. For instance the dialogs:

image

and in the code

image

will disappear and the navigation will directly goes to the view which has the searched id.

Initially reported by @Florian14 , but I do not think he has created an issue about it.

A global replacement could be done, also some rules could be added to this file: https://github.com/vector-im/element-android/blob/main/tools/check/forbidden_strings_in_layout.txt to avoid the issue to come back

bmarty:
Boring PR about cleaning up the layouts.

Fixes #4604


tools/check/forbidden_strings_in_layout.txt
```diff
@@ -25,3 +25,8 @@
 
 ### Use style="@style/Widget.Vector.TextView.*" instead of textSize attribute
 android:textSize===9
+
+### Use `@id` and not `@+id` when referencing ids in layouts
+layout_(.*)="@\+id
+accessibilityTraversal(.*)="@\+id
+toolbarId="@\+id
```
bmarty, line 32: Those mistakes should not reappear thanks to that new rules

vector/src/debug/res/layout/demo_theme_sample.xml
```diff
@@ -1,9 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-
 <!--
- The aim of this file is to test the different themes of Riot
+ The aim of this file is to test the different themes of Element
```
bmarty, line 3: RIP Riot

vector/src/main/res/layout/dialog_import_e2e_keys.xml
```diff
@@ -27,10 +26,7 @@
         android:id="@+id/importDialogTil"
         style="@style/Widget.Vector.TextInputLayout.Password"
         android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        app:layout_constraintEnd_toStartOf="@+id/importDialogShowPassword"
-        app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@+id/dialog_e2e_keys_passphrase_filename">
```
bmarty, line 33: The parent is not anymore in a ConstraintLayout

vector/src/debug/res/layout/demo_theme_sample.xml
ouchadam: :wave:

ouchadam (approved):
CI is happy and I did a quick local smoke test LGTM :100:

PR #4614, opened 2021-12-01
Workaround to fetch all the pending toDevice events from a Synapse homeserver

### LINKED ISSUE #4612: Workaround to fetch all the pending toDevice events from a Synapse homeserver
Due to [a bug on Synapse](https://github.com/matrix-org/synapse/issues/11457), the Android Matrix SDK can receive toDevice events with a long delay: 30s max if nothing else (new message, etc.) triggers a sync response to be sent.

The workaround implemented on some clients is to set the timeout parameter of the sync request to 0 if the previous sync response contains some toDevice events, to force Synapse to send all the pending toDevice events, if any, immediately.

Once https://github.com/matrix-org/synapse/issues/11457 will be fixed, the workaround could be removed. But I think it's still worth to implement it, because once fixed it can take time for the Synapse fix to be deployed on all the Synapse instances in the wild. Also we do not have any due date for the Synapse fix.

_Note_: the Android Matrix SDK is managing the sync requests, so implementing this will still be useful once we will use the Crypto Rust SDK.

bmarty:
Fixes #4612

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/job/SyncWorker.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/job/SyncWorker.kt
@@ -75,10 +75,20 @@ internal class SyncWorker(context: Context,
                     Result.success().also {
                         if (params.periodic) {
                             // we want to schedule another one after a delay, or immediately if hasToDeviceEvents
-                            automaticallyBackgroundSync(workManagerProvider, params.sessionId, params.timeout, params.delay, forceImmediate = hasToDeviceEvents)
+                            automaticallyBackgroundSync(
+                                    workManagerProvider = workManagerProvider,
+                                    sessionId = params.sessionId,
+                                    serverTimeout = params.timeout,
+                                    delayInSeconds = params.delay,
+                                    forceImmediate = hasToDeviceEvents
+                            )
                         } else if (hasToDeviceEvents) {
                             // Previous response has toDevice events, request an immediate sync request
-                            requireBackgroundSync(workManagerProvider, params.sessionId, 0)
+                            requireBackgroundSync(
+                                    workManagerProvider = workManagerProvider,
+                                    sessionId = params.sessionId,
+                                    serverTimeout = 0
+                            )
                         }
                     }
                 },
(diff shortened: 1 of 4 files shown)


BillCarsonFr (commented):
I am not sure this will work for the SyncWorker (after a push?)
You might need to schedule a subsequent background sync in this case

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/job/SyncWorker.kt
```diff
@@ -67,13 +69,16 @@ internal class SyncWorker(context: Context,
...
+                { hasToDeviceEvents ->
                     Result.success().also {
                         if (params.periodic) {
-                            // we want to schedule another one after delay
-                            automaticallyBackgroundSync(workManagerProvider, params.sessionId, params.timeout, params.delay)
+                            // we want to schedule another one after a delay, or immediately if hasToDeviceEvents
+                            automaticallyBackgroundSync(workManagerProvider, params.sessionId, params.timeout, params.delay, forceImmediate = hasToDeviceEvents)
+                        } else if (hasToDeviceEvents) {
+                            // Previous response has toDevice events, request an immediate sync request
+                            requireBackgroundSync(workManagerProvider, params.sessionId, 0)
```
bmarty, line 81: This is not a workaround actually, and we should always to that, even it the bug is fixed on Synapse, WDYT @BillCarsonFr ?

BillCarsonFr: Yes (workaround is to set timeout to 0, which is probably default here ?)

We should check if we try to redecrypt existing notifications after the catchup sync?

bmarty: Ok, thanks.
> We should check if we try to redecrypt existing notifications after the catchup sync?

To be honest I did not do any test, I'm just coding it. Not sure how I can reproduce the test conditions

BillCarsonFr (changes requested):
Just a last comment, I noticed that on iOS when catching up we don't try ti upload new otk or to handle room key request until the catchup sync (to get all to devices) is done https://github.com/matrix-org/matrix-ios-sdk/blob/4e8db1ee5d47e7cb3bcfc4b02508c9f56996fad0/MatrixSDK/Crypto/MXCrypto.m#L931

issue ref: https://github.com/vector-im/riot-web/issues/2782

bmarty:
LGTM, @BillCarsonFr can you approve this PR please?

BillCarsonFr (approved):
LGTM

PR #4620, opened 2021-12-02
Avoiding duplicate message long press events 

### LINKED ISSUE #4324: Don't open menu when long clicking a link
### Steps to reproduce

1. long clicking a link to copy it opens the menu  meant for the long click on a message

#### What did you expect?
The menu should not open
#### What happened instead?
The menu opened

### Your phone model

Galaxy A51

### Operating system version

One UI 3.1, Android 11

1.3.3 [40103032] (G-b3874)

### Homeserver

matrix.org

### Will you send logs?

No

ouchadam:
Fixes #4324 URL long presses opening menu and copying

- Infers that a Clickable span has been pressed by the presence of a selection (from this handy SO https://stackoverflow.com/a/35696744)

| BEFORE | AFTER |
| --- | --- |
|before|after

This does pose the question of whether we should have link copying at all as the message menu already contains a `Copy` item cc @daniellekirkwood

### DIFF
--- a/changelog.d/4324.bugfix
+++ b/changelog.d/4324.bugfix
@@ -1,0 +1,1 @@
+Fixes message menu showing when copying message urls
--- a/vector/src/main/java/im/vector/app/core/epoxy/Listener.kt
+++ b/vector/src/main/java/im/vector/app/core/epoxy/Listener.kt
@@ -17,6 +17,7 @@
 package im.vector.app.core.epoxy
 
 import android.view.View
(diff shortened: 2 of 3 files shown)


bmarty:
> This does pose the question of whether we should have link copying at all as the message menu already contains a Copy item

The link can be a part of the message and not the message, so I think it should be nice to try to keep this copy link feature

bmarty (approved):
LGTM!

PR #4625, opened 2021-12-03
select answer should be used for outgoing calls.

### LINKED ISSUE #4621: Outgoing voip call through sip bridge fails
1. Make an outgoing call through sip bridge
2. Client appears to connect but then call dies after 1 minute.

Expected: Call stays connected.

langleyd:
Fixes #4621 

Use case is:
- alice calls bob 
- bob answers alice
- alice selects bob's answer

The reason being that if bob tries to accept on two devices, it's clear which one alice is actually trying to connect to.

I believe this problem might have just surfaced recently due to changes at the sip bridge adding a timeout for the `selectAnswer`


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
```diff
@@ -203,16 +203,19 @@ internal class MxCallImpl(
...
-        state = CallState.Answering
-        CallSelectAnswerContent(
-                callId = callId,
-                partyId = ourPartyId,
-                selectedPartyId = opponentPartyId?.getOrNull(),
-                version = MxCall.VOIP_PROTO_VERSION.toString()
-        )
-                .let { createEventAndLocalEcho(type = EventType.CALL_SELECT_ANSWER, roomId = roomId, content = it.toContent()) }
-                .also { eventSenderProcessor.postEvent(it) }
+        if (isOutgoing) {
```
SimonBrandner, line 206: Would it be simpler to do
```
if (!isOutgoing) return
```
or whatever the Kotlin equivalent is?

langleyd: Yea you're right I could have kept the guard

SimonBrandner (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
```diff
@@ -203,8 +203,10 @@ internal class MxCallImpl(
 
     override fun selectAnswer() {
         Timber.tag(loggerTag.value).v("select answer $callId")
-        if (isOutgoing) return
-        state = CallState.Answering
```
ouchadam, line 207: do we still set this state somewhere else or does `CallState.Answering` no longer need to exist?

ouchadam (approved):
small question around removing the state, otherwise looking good! :100:

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
langleyd: It still exists and is set for incoming calls [here](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt#L179) as described in it's definition [here](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/call/CallState.kt#L37).

This `selectAnswer` function is actually used for outgoing calls the current state at this point is `Dialing` and stays as that until the `Connected` state.

ganfra (approved):
Thanks for the fix, my fault for that :/

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
ouchadam: thanks for explaining

bmarty: I guess the initial issue was due to a quick copy paste of the `accept` fun (https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt#L179)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
```diff
@@ -203,8 +203,10 @@ internal class MxCallImpl(
...
+        if (!isOutgoing) return
+        
+        // This is an outgoing call, select the remote client that answered.
+        // state is still DIALING.
```
bmarty, line 209: Do we want to actively reinforce that assumption at line 206?

bmarty (approved):
LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
```diff
@@ -203,8 +203,10 @@ internal class MxCallImpl(
 
     override fun selectAnswer() {
         Timber.tag(loggerTag.value).v("select answer $callId")
-        if (isOutgoing) return
-        state = CallState.Answering
+        if (!isOutgoing) return
+
```
bmarty, line 207:
```suggestion

```

Removing trailing space

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
langleyd: You think just remove the line about `state`?

bmarty: Your comment says that `state is still DIALING`
So I am wondering if it was worth  adding
```kotlin
if (state != CallState.Dialing) return
```

bmarty: (Or log a warning)

bmarty: Thanks

PR #4626, opened 2021-12-03
Feature flags base

ouchadam:
- Separates the SSO redirection from the `LoginActivity` + `LoginActivity2` by using a dedicated invisible `SSORedirectRouterActivity`, this allows both activities to be enabled at the same time and not exported

- Introduces a `VectorFeatures` abstraction to start grouping our feature flags, this will also enable overriding at runtime and provides an entry point for forks to extend to help avoid conflicts (debug overrides to come in the next PR)

- Ports the `LoginVersion` and `NotificationSettingsVersion` to the `VectorFeatures`

| BEFORE SSO REDIRECT | AFTER SSO REDIRECT |
| --- | --- | 
|before-sso|after-sso


vector/src/androidTest/java/im/vector/app/ui/robot/settings/SettingsNotificationsRobot.kt
```diff
@@ -21,22 +21,27 @@ import com.adevinta.android.barista.interaction.BaristaClickInteractions.clickOn
...
-            pressBack()
-            clickOn(R.string.settings_notification_mentions_and_keywords)
-            // TODO Test adding a keyword?
-            pressBack()
-            clickOn(R.string.settings_notification_other)
-            pressBack()
-        } else {
-            clickOn(R.string.settings_notification_advanced)
-            pressBack()
+        when (BuildConfig.NOTIFICATION_SETTINGS_VERSION!!) {
```
bmarty, line 29: Why not using a `DefaultVectorFeatures` instance here? It would prevent the usage of `!!` (I think)

vector/src/main/java/im/vector/app/features/settings/notifications/VectorSettingsNotificationPreferenceFragment.kt
```diff
@@ -145,6 +147,7 @@ class VectorSettingsNotificationPreferenceFragment @Inject constructor(
         refreshBackgroundSyncPrefs()
 
         handleSystemPreference()
+        handleVersionedSettings()
```
bmarty, line 150: Can we observe some blink with this new way to hide preference? Maybe hide both pref by default in the XML

bmarty (approved):
LGTM, thanks

vector/src/main/java/im/vector/app/features/settings/notifications/VectorSettingsNotificationPreferenceFragment.kt
ouchadam: there's a blink if the visibility is updated in `onResume` but `onBindPref` is instant (as long as we stay on the main thread)

the programmatic visibility is also removed in #4627 :tada:

vector/src/androidTest/java/im/vector/app/ui/robot/settings/SettingsNotificationsRobot.kt
ouchadam: good point! I wasn't sure how the injections would work in the `SanityTests`, I would invest more time but this is removed in #4627

vector/src/main/java/im/vector/app/features/settings/notifications/VectorSettingsNotificationPreferenceFragment.kt
```diff
@@ -309,6 +312,15 @@ class VectorSettingsNotificationPreferenceFragment @Inject constructor(
         }
     }
 
+    private fun handleVersionedSettings() {
```
ouchadam, line 315: this change is temporary as V2 is live, V1 is removed in #4627

PR #4628, opened 2021-12-03
Add script to compress video and convert to gif file

bmarty:
Small script to compress videos.
Gif is not always smaller than the mp4, but we can try to iterate later.

### DIFF
--- a/tools/compressVideo.sh
+++ b/tools/compressVideo.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+#
+# Copyright (c) 2021 New Vector Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -e
+echo "Converting file $1"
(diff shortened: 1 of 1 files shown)


tools/compressVideo.sh
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -e
+echo "Converting file $1"
+file=$(echo $1 | sed 's/\.[^.]*$//')
+ffmpeg -i $1 -filter_complex "[0:v] fps=12,scale=480:-1,split [a][b];[a] palettegen [p];[b][p] paletteuse" $file-tmp.gif
```
ouchadam, line 22: I ended up with two versions of this script, one with scale=480 and a small version with 240 for when the gif ends up bigger than 10mb :sob:

tools/compressVideo.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "Converting to Gif"
+gifsicle -O3 --lossy=80 -o $file.gif $file-tmp.gif
```
ouchadam, line 24: do you think it's worth mentioning the dependencies on `ffmpeg` and `gifsicle`?

ouchadam (approved):

tools/compressVideo.sh
bmarty: If missing users will find out fast enough (I had to install `gifsicle` on my side). This script is just for us. No doc, no comment, no robustness, nothing :). Ah yes, some echoing!

tools/compressVideo.sh
bmarty: There is room for improvement, for instance test the output file size and start again with 240.

PR #4637, opened 2021-12-06
Feature/dla/outgoing pstn call fails

### LINKED ISSUE #4636: Fix logging for callSelect
`callSelect` can also be called when the call is in connected state. 

As [outlined in the spec](https://github.com/matrix-org/matrix-doc/pull/2746/files#diff-4d65532ebdb47ba1fd5ea93342395abfeaab0e911360c0bd8d8839be90afd7dfR73) media can be flowing before call selection happens. I have also seen this in my testing.

We should update the logging warning accordingly.

Follow up issue for https://github.com/vector-im/element-android/issues/4621

langleyd:
Fixes #4636

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
@@ -205,8 +205,8 @@ internal class MxCallImpl(
         Timber.tag(loggerTag.value).v("select answer $callId")
         if (!isOutgoing) return
         // This is an outgoing call, select the remote client that answered.
-        if (state != CallState.Dialing) {
-            Timber.tag(loggerTag.value).w("Expected state is CallState.Dialing got $state.")
+        if (state is CallState.Dialing || state is CallState.Connected) {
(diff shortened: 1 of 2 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/call/model/MxCallImpl.kt
```diff
@@ -205,8 +205,8 @@ internal class MxCallImpl(
         Timber.tag(loggerTag.value).v("select answer $callId")
         if (!isOutgoing) return
         // This is an outgoing call, select the remote client that answered.
-        if (state != CallState.Dialing) {
-            Timber.tag(loggerTag.value).w("Expected state is CallState.Dialing got $state.")
+        if (state !is CallState.Dialing && state !is CallState.Connected) {
```
ouchadam, line 208: `!= Dialing` is still valid here as `Dialing` is an `object`, or was this done for consistency? I lean towards using `!=` where possible as it provides extra context, but not a strong opinion

langleyd: Yea no strong opinion here either, happy to change.

ouchadam (approved):
LGTM :100:

bmarty (approved):
LGTM, thanks

PR #4645, opened 2021-12-06
Cleanup onclick

bmarty:
See commit comments for more details


vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
```diff
@@ -121,7 +120,6 @@ abstract class VectorBaseActivity<VB : ViewBinding> : AppCompatActivity(), Maver
 
     protected fun View.debouncedClicks(onClicked: () -> Unit) {
         clicks()
-                .throttleFirst(300)
```
ouchadam, line 124: :100: this should help make the app feel a little snappier (if I've understood the current usages)

ouchadam (approved):
LGTM :+1:

vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
bmarty: No, it is not delaying the first click hopefully, but just ignore the next ones in a 300ms window

PR #4650, opened 2021-12-07
Cleanup id ref. Use type views instead.

bmarty:
Also use extension facility


ouchadam:
something else to look into could be using a non transitive R file https://blog.blundellapps.co.uk/speed-up-your-build-non-transitive-r-files/ although this will mostly benefit us when we have more modules

ouchadam (approved):

PR #4652, opened 2021-12-07
Use viewBindingIgnore="true" when we do not need the ViewBinding.

bmarty:
We are using Epoxy and do not need the view binding for the vast majority of list item layouts. Disable their generation and compilation and finally Proguard shrink to have a faster compilation. Let's see if the gain worth it on the CI.

See https://developer.android.com/topic/libraries/view-binding for more details


ouchadam (approved):
compiled and works locally fine for me :100: (benchmarking in progress :crossed_fingers: )

bmarty:
I am now wondering if we could use ViewBinding in our VectorEpoxyHolder, or change the bit our arch to be able to do so.

ouchadam:
> I am now wondering if we could use ViewBinding in our VectorEpoxyHolder, or change the bit our arch to be able to do so.

depends if we want to start putting our eggs in the compose basket :thinking:

bmarty:
Ok, thanks for the measures 💯

PR #4656, opened 2021-12-08
Feature based FTUE flows

ouchadam:
Depends on #4635 Part of #4584

- Creates a base for the iterative FTUE work (copied from `LoginActivity`) and allows it to co-exist with the Login2 (FTUE_WIP) 
- Ports the `LoginActivity2` to a `FTUEActivity` which dynamically routes to the iterative FTUE or FTUE_WIP variants
- Renames the `LoginVersion` to `LoginVariant` to allow for better naming
- Removes `SoftLogoutActivity2` as it wasn't exposed by the manifest and the implementation ended up using the legacy login flows 

My aim is to iteratively merge the changes from into FTUE_WIP into FTUE, eventually deleting the WIP

| FTUE | FTUE_WIP |
| --- | --- |
|ftue|ftue-wip

- Next steps, merge https://github.com/vector-im/element-android/pull/4590 into the `DefaultFTUEVariant`
- Add a way to access the features menu from the onboarding

Issue #4584, opened 2021-11-30
Build FTUE Splash screens


daniellekirkwood:
### Your use case

When the app opens, and the user is not logged in, we show a Splash screen. 
- [_link to Figma splash screens_](https://www.figma.com/file/xdV4PuI3DlzA1EiBvbrggz/Login-Flow-v2?node-id=1628%3A31003)

**The Splash screens include:**
- ~A loading page~ Descoped 
- A landing page carousel that automatically rotates 4 images + text combinations every 5 seconds. The user can also swipe between screens. Once a user has manually swiped, the automatic transitions are disabled.
- 2 buttons, Get Started (create new account) & I already have an account (login)
  - While we are still building the FTUE experience Get Started will lead to … and I already have an account will lead to...

--- 
**Tasks**
- [x] Jordan to review the visuals used on screen
- [ ] Marketing to approve the copy used on screens
- Build
  - [x] I already have an account #4668 
  - [x] Carousel with 4 images, text combinations, I already have an account flow #4727 
  - [x] Automatic carousel rotation #4740
  - [x] Disable automatic carousel rotation after user interaction #4914 
  - [x] Update with final images/copy/dark mode gradients, styled title full stops #4880 
  - [x] Locking phones to portrait during FTUE auth #4918 
  - [ ] Enable feature #4964 
- [x] Design review before merge

---
_This issue is considered complete when the tasks are complete, the screen matches the Figma design, it's been reviewed (by Product, Design & Engineering), and merged._

daniellekirkwood:
I'm testing something. Please ignore.

<img width="763" alt="Screenshot 2022-02-16 at 16 05 39" src="https://user-images.githubusercontent.com/89144281/154306382-4145b7ff-e3a9-4ed1-9cc9-dce595a33f4c.png">

PR #4657, opened 2021-12-08
Cloning the `LoginViewModel` domain to `Onboarding` 

ouchadam:
Depends on #4656 Part of #4584

Nothing too exciting here, clones the `LoginViewModel`, state, actions and view events so that the Onboarding domain can override the behaviour and start to merge with `Login2`, will help reduce for the diff for the https://github.com/vector-im/element-android/pull/4668


vector/src/main/java/im/vector/app/features/ftue/FTUEAction.kt
```diff
@@ -0,0 +1,79 @@
...
+import im.vector.app.core.platform.VectorViewModelAction
+import im.vector.app.features.login.LoginConfig
+import im.vector.app.features.login.ServerType
+import im.vector.app.features.login.SignMode
+import org.matrix.android.sdk.api.auth.data.Credentials
+import org.matrix.android.sdk.api.auth.data.SsoIdentityProvider
+import org.matrix.android.sdk.api.auth.registration.RegisterThreePid
+import org.matrix.android.sdk.internal.network.ssl.Fingerprint
+
+sealed class FTUEAction : VectorViewModelAction {
```
bmarty, line 28: I think there is a small misunderstanding here, FTUE stands for First Time User Experience, and it does not only cover the login flow, but also the first room creation, the first message sent, etc. So to me the FTUE is the user experience during the (2 or 3) first days using the app.

Also user can create an account or login even if they already know the app (so it will not be the first time)

So using FTUE to name classes related to the signin/signup flow is a bit strange to me.

ouchadam: good point, I was considering the FTUE to be the new variant of the onboarding but product/design is calling this the `FTUE authentication flow`

what do you think of using `onboarding` as the package/activity/domain which would then contain `Login, Login2, FTUE Auth`?

ouchadam: renamed ftue to onboarding https://github.com/vector-im/element-android/pull/4657/commits/119140b80e9ae61f80fc7eb7242cf8d246d06ce0

we have a onboarding domain with variants `legacy`, `login 2` and `ftue auth`

bmarty: Yes, Onboarding is better, also good for de-jargoning.
By the way, the Java recommendation for acronyms in naming is to capitalize only the first letter of them, so for instance we would have `FtueAction` here, instead of `FTUEAction`.

vector/src/main/java/im/vector/app/core/di/MavericksViewModelModule.kt
```diff
@@ -447,6 +448,11 @@ interface MavericksViewModelModule {
     @MavericksViewModelKey(AccountCreatedViewModel::class)
     fun accountCreatedViewModelFactory(factory: AccountCreatedViewModel.Factory): MavericksAssistedViewModelFactory<*, *>
 
+    @Binds
+    @IntoMap
+    @MavericksViewModelKey(OnboardingViewModel::class)
+    fun ftueViewModelFactory(factory: OnboardingViewModel.Factory): MavericksAssistedViewModelFactory<*, *>
```
bmarty, line 454: Maybe rename the fun to `onBoardingViewModelFactory`

bmarty (approved):
LGTM, just a small remark, but feel free to fix in another PR.
You can merge this one if you want to.

vector/src/main/java/im/vector/app/core/di/MavericksViewModelModule.kt
ouchadam: will update in the next PR to avoid having to wait for the CI to finish twice :smile:

ouchadam: 9ccad11642ec00b67fcf47ac383bc4f9b5705ff1

Issue #4594, opened 2021-11-30
Improve the display of privacy policy and T&C


bmarty:
Policy and T&C are a bit specific on a Matrix client. We have the policies of the client itself, the policies of the homeserver and the policy of the identity server, if there is one configured.

Desired changes:

- [x] Replace the actions `Settings/Help & About/Copyright`, `Settings/Help & About/Terms & conditions` and `Settings/Help & About/Privacy policy` to a single items `Legals` which will open a new screen to group all the policies from Element, the policies from the homeserver and the policies from the identity server if there is one configured.

<img width="384" alt="image" src="https://user-images.githubusercontent.com/3940906/144104232-3dbd12ad-9ce4-4d5b-b8f4-f7323b4255fc.png">

- [ ] Improve the way we display identity server policy. Remove the expandability of the section.

<img width="389" alt="image" src="https://user-images.githubusercontent.com/3940906/144104303-94ee5b94-71f6-496b-998e-d19c53f4ec0f.png">

- [ ] Add the homeserver policies to the homeserver setting page `Settings/General/Homeserver`. Currently those policies are only visible when the user create an account.

<img width="390" alt="image" src="https://user-images.githubusercontent.com/3940906/144104345-6da9e905-43b3-41d4-bfcd-a3a678b46fb0.png">

bmarty:
@deniseal I have assigned this to you, please let me know what you think about this plan ^. Thanks!

deniseal:
@bmarty this is a great idea. Should we also involve other teams to ensure consistency across all products?

deniseal:
@nadonomy I think it would be good to have product involved in this.

nadonomy:
Thx for flagging. @daniellekirkwood (product) & @amshakal (design) are best positioned for this via the mobile app teams.

bmarty:
Done

bmarty:
(at least the first point)

bmarty:
Will open other issues for the remaining parts

PR #4660, opened 2021-12-08
Legals

### LINKED ISSUE #4638: When a user clicks "Help & About" they only see About
### Problem

When a user selects Help & About, the next screen only shows About info. 

### Solution

There should be a button/link on available that takes users to the Element io Help page (similar to iOS)

Add a Help link to the list that leads here: https://element.io/help

bmarty:
Implement the first item of #4594 

### New `Legals` section in the main setting menu:

image

@amshakal which icon can I used for this item?
**EDIT**: icon updated

### Legals screen

- App policies (was previously on Help & about screen)
- homeserver policies (new)
- identity server policies, if there is one configured (else this section is not displayed). Same info displayed than on the discovery screen (which will have to be updated later, it's quite confusing today, there is a mix between identity server configuration and discovery settings)
- third party libraries (was previously on Help & about screen)

image

### Help and About screen updated

With a new help section. Fixes #4638

|Before|After|When help is clicked|
|-|-|-|
|image|image|image|


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/terms/DefaultTermsService.kt
```diff
@@ -48,11 +51,49 @@ internal class DefaultTermsService @Inject constructor(
...
+            executeRequest(null) {
+                termsAPI.register(baseUrl + NetworkConstants.URI_API_PREFIX_PATH_R0 + "register")
+            }
+            // Return empty result if it succeed, but it should never happen
+            buildEmptyGetTermsResponse()
+        } catch (throwable: Throwable) {
+            @Suppress("UNCHECKED_CAST")
+            ((throwable.toRegistrationFlowResponse()
+                    ?.params?.get(LoginFlowTypes.TERMS) as? JsonDict)
+                    ?.get("policies") as? JsonDict)
```
bmarty, line 79: A bit dirty but it works...

amshakal:
@bmarty I think we can use this icon for it. 
Legal_Files

amshakal (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/terms/DefaultTermsService.kt
```diff
@@ -48,11 +51,49 @@ internal class DefaultTermsService @Inject constructor(
...
+        }
+    }
+
+    /**
+     * We use a trick here to get the homeserver T&C, we use the register API
+     */
+    private suspend fun getHomeserverTerms(baseUrl: String): GetTermsResponse {
+        return try {
+            executeRequest(null) {
+                termsAPI.register(baseUrl + NetworkConstants.URI_API_PREFIX_PATH_R0 + "register")
```
ouchadam, line 71: is this something we should consider adding to the spec? it looks like we're relying on the `register` endpoint to fail

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/terms/TermsService.kt
```diff
@@ -19,7 +19,12 @@ package org.matrix.android.sdk.api.session.terms
 interface TermsService {
     enum class ServiceType {
         IntegrationManager,
-        IdentityService
+        IdentityService,
+
+        /**
+         * It's only possible to use this value for [getTerms]
+         */
+        Homeserver
```
ouchadam, line 27: do you think it's worth having a separate `getHomeserverTerms` to avoid potential misunderstandings with this api?

vector/src/main/java/im/vector/app/features/settings/legals/ElementLegals.kt
```diff
@@ -0,0 +1,38 @@
...
+import im.vector.app.R
+import im.vector.app.core.resources.StringProvider
+import im.vector.app.features.discovery.ServerPolicy
+import im.vector.app.features.settings.VectorSettingsUrls
+import javax.inject.Inject
+
+class ElementLegals @Inject constructor(
+        private val stringProvider: StringProvider
+) {
+    /**
```
ouchadam, line 28: small nit: I don't think this comment/doc adds much, do you think it's needed? renaming `getData` to `getServerPolicies` would avoid the need for a comment

bmarty: Probably but a bit more confusing, since there is no server involved here...

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/terms/TermsService.kt
bmarty: I though about it, will iterate.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/terms/DefaultTermsService.kt
bmarty: the register endpoint will fail for sure, we are not providing any data to the request. Also if registration is disabled, I'm not sure what will happen.
This is a temporary workaround waiting for https://github.com/matrix-org/matrix-doc/pull/3012 to land

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/terms/TermsService.kt
bmarty: 49721ed14d19c2dd86276e16e3c1bcf3a5d79884

ouchadam (approved):
awesome stuff! :100: hopefully this helps reduce our google play rejections

bmarty:
Force-pushed to fix merge conflict in VectorPreferences

PR #4668, opened 2021-12-09
FTUE Auth - I already have an account

### LINKED ISSUE #4382: Add "Sign in" to Splash screen
### Your use case

When a user opens the app they see a splash screen. Right now the only option is to "Get Started". 
We want to add the ability to Sign in, as well as Create an account.

_This is the Basic Splash screen and is a small change to the existing flow so that we might learn from users_

_No response_

### Additional context

_No response_

ouchadam:
Fixes #4382 Part of #4584

- Adds debug menu opening from the splash screen debug info
- Adds boolean types to the debug features 
- Adds the `I already have an account` button and flow behind the FTUE Auth feature flag

| GET STARTED | ALREADY HAVE ACCOUNT | FEATURE SWITCHING | 
| --- | --- | --- |
|after-get-started|after-already-have-account|onboarding-features


ouchadam:
I'm not super happy~ about having to the copy all of the login fragments, unfortunately there's a lot of inheritance in the area and the fragments are coupled directly to view models 

would recommend reviewing by commits!

ouchadam:
closing to extract the fragment copying to its own PR

vector/src/debug/java/im/vector/app/features/debug/features/DebugFeaturesStateFactory.kt
```diff
@@ -28,16 +28,25 @@ class DebugFeaturesStateFactory @Inject constructor(
         return FeaturesState(listOf(
                 createEnumFeature(
                         label = "Onboarding variant",
-                        selection = debugFeatures.onboardingVariant(),
+                        override = debugFeatures.onboardingVariant(),
```
bmarty, line 31: I am surprised `override` is not a reserved Kotlin keyword. Maybe a bit confusing to use it anyway.

vector/src/main/java/im/vector/app/features/VectorFeatures.kt
```diff
@@ -36,4 +38,5 @@ interface VectorFeatures {
 
 class DefaultVectorFeatures : VectorFeatures {
     override fun onboardingVariant(): VectorFeatures.OnboardingVariant = BuildConfig.ONBOARDING_VARIANT
+    override fun isAlreadyHaveAccountSplashEnabled() = true
```
ouchadam, line 41: the `I already have an account` is enabled by default when the onboarding variant is set to  `FTUE_AUTH`, it's currently set to `LEGACY`, will raise a dedicated PR to enable the `FTUE_AUTH` onboarding as the default variant

bmarty (approved):
LGTM

vector/src/debug/java/im/vector/app/features/debug/features/DebugFeaturesStateFactory.kt
ouchadam: yeah I was surprised as well, will prefix the names to avoid confusion with the language keywords

ouchadam: https://github.com/vector-im/element-android/pull/4668/commits/41f931e4663ee3d58cbf5772b5bcee45de10015d

PR #4670, opened 2021-12-09
Safer dependencies

bmarty:
Ref:
- https://docs.gradle.org/5.1.1/userguide/declaring_repositories.html#sec::matching_repositories_to_dependencies
- https://jebware.com/blog/?p=573


build.gradle
```diff
@@ -37,45 +37,198 @@ allprojects {
...
+                includeGroupByRegex 'com\\.android\\.tools\\..*'
+                includeGroupByRegex "com\\.google\\.android\\..*"
+                includeGroup 'com.google.firebase'
+                includeGroup 'com.android'
+                includeGroup 'com.android.tools'
+            }
+        }
+        mavenCentral {
+            content {
+                includeGroup 'com.adevinta.android'
```
bmarty, line 84: if only we had a way to externalize those list into another file...
`includeGroup` does not support taking a list as param

ariskotsomitopoulos: let me check if that is possible

bmarty:
@ariskotsomitopoulos thanks for your proposal solution. I have updated it a bit. Please let me know what you think.

ariskotsomitopoulos (approved):

ariskotsomitopoulos:
> @ariskotsomitopoulos thanks for your proposal solution. I have updated it a bit. Please let me know what you think.

Yes +1, nice changes! I agree

vector/build.gradle
```diff
@@ -317,11 +317,6 @@ android {
     }
 }
 
-configurations {
-    // videocache includes a sl4j logger which causes mockk to attempt to call the static android Log
-    testImplementation.exclude group: 'org.slf4j', module: 'slf4j-android'
-}
```
bmarty, line 323: As I removed `videocache`, I thought I could remove this as well. But maybe double check with @ouchadam first. And I just see that it's still needed: https://github.com/vector-im/element-android/pull/4670/files#diff-f6dc61a00090b6fc204667d7934e656dbb78c8bb0dc14f95468dd57b4300a8caR177

build.gradle
```diff
@@ -37,45 +36,50 @@ allprojects {
...
-                includeGroupByRegex 'com\\.github\\.jetradarmobile'
-                includeGroupByRegex 'nl\\.dionsegijn'
-
-                // Voice RecordView
-                includeGroupByRegex 'com\\.github\\.Armen101'
+                groups.jitpack.regex.each { includeGroupByRegex it }
+                groups.jitpack.group.each { includeGroup it }
             }
         }
-        maven { url 'https://oss.sonatype.org/content/repositories/snapshots/' }
```
bmarty, line 69: I am not sure why it has been added at first place, but it seems not necessary anymore 🤞

vector/build.gradle
ouchadam: if `videocache` is removed then removing this is also fine :+1:  (the unit tests will fail if not)

ouchadam (approved):
great change! :100: (might even have a little bit of a build speed improvement :crossed_fingers: )

PR #4689, opened 2021-12-10
Feature/aris/dev tools fix 4592

ariskotsomitopoulos:
Revert requestModelBuild() init block in dev tools controller for PR #4591

### DIFF
--- a/changelog.d/4592.bugfix
+++ b/changelog.d/4592.bugfix
@@ -1,0 +1,1 @@
+Revert requestModelBuild() init block in dev tools controller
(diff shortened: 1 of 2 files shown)


changelog.d/4592.bugfix
```diff
@@ -0,0 +1 @@
+Revert requestModelBuild() init block in dev tools controller
```
bmarty, line 1: This is correct, but the content here should reflect what will be visible to the user, when applicable.
I suggest:
```suggestion
Fix empty Dev Tools screen issue.
```

bmarty (commented):
Thanks, just a remark about the changelog content.

bmarty (approved):

PR #4703, opened 2021-12-13
Use only non empty option to create a poll

bmarty:
Quick fix to avoid creating poll with empty answers


onurays (approved):

bmarty:
(no changelog since poll is still under labs and not rendered yet in the timeline in production)

PR #4708, opened 2021-12-14
Disable automatic opt-in screen display

bmarty:
Still possible to opt-in from the settings.
Also add changelog for the previous PR.

Tested OK using the Analytics debug screen.


vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -83,9 +83,12 @@ class HomeActivityViewModel @AssistedInject constructor(
         observeInitialSync()
         checkSessionPushIsOn()
         observeCrossSigningReset()
-        observeAnalytics()
+        // Disable Analytics opt-in automatic display
```
ouchadam, line 86: in the future this would be a good candidate for a `VectorFeature` flag

PR #4728, opened 2021-12-15
Proximity sensor not releasing

### LINKED ISSUE #2467: Proximity sensor works after call finished
**Describe the bug**
Proximity sensor is still enabled after the call ends. 

**To Reproduce**
1. End VoIP call
2. Hide sensor with your hand 
3. Screen gets black
4. Have to uninstall app (killing process is not enough) to dis ale proximity sensor 

**Expected behavior**
Disable proximity sensor after call finished

ouchadam:
Fixes #2467 Proximity sensor not releasing

Wake locks are reference counted which means we would need to release as many times as we acquire in order to full release the proximity sensor. It's possible that the `onProximityNear` / `onProximityFar` may not trigger as cleanly as we expect on some devices which can cause multiple wakelock acquires 

- Guards the proximity wake lock acquiring by checking if we already have one active.

----

was able to reproduce the original issue by commenting out the `onProximityFar` and triggering the proximity sensor multiple times whilst in a call

### DIFF
--- a/changelog.d/2467.bugfix
+++ b/changelog.d/2467.bugfix
@@ -1,0 +1,1 @@
+Fixing proximity sensor still being active after a call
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks!

PR #4734, opened 2021-12-16
Implement analytics plan

bmarty:
- Introduce `AnalyticsTracker` interface which is now injected to be able to send event (only if user has opt-in of course)
- Implement part of the existing plan.

=> Could be nice to merge this PR, which has just been rebased, and iterate later for the remaining Events.


ouchadam:
is the plan to have future generated events pulled in automatically? https://github.com/vector-im/element-android/pull/4956

bmarty:
> is the plan to have future generated events pulled in automatically? #4956

Yes. It's already working, but since I called the script in this PR, I did not want to merge #4956 (and the previous ones)

vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
```diff
@@ -324,7 +334,7 @@ abstract class VectorBaseActivity<VB : ViewBinding> : AppCompatActivity(), Maver
     override fun onResume() {
         super.onResume()
         Timber.i("onResume Activity ${javaClass.simpleName}")
-
+        screenEvent = analyticsScreenName?.let { ScreenEvent(it) }
```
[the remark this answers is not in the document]
bmarty: We track on exit because we want to report the duration, and this is managed by `ScreenEvent`. It's a bit weird, I know, generally the duration could be managed by the analytics tools (since a screen replaces another, duration can be deduced). But as we do not track every screens it probably better like that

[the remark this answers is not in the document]
ouchadam: makes sense, thanks for explaining!

vector/src/main/java/im/vector/app/core/di/SingletonEntryPoint.kt
```diff
@@ -56,7 +56,7 @@ interface SingletonEntryPoint {
 
     fun pinLocker(): PinLocker
 
-    fun analytics(): VectorAnalytics
+    fun analyticsTracker(): AnalyticsTracker
```
ouchadam, line 59: :+1:

vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
```diff
@@ -363,6 +373,7 @@ abstract class VectorBaseActivity<VB : ViewBinding> : AppCompatActivity(), Maver
 
     override fun onPause() {
         super.onPause()
+        screenEvent?.send(analyticsTracker, analyticsScreenName)
```
ouchadam, line 376: do we need to be able to override the screen name in the `send`, can screens dynamically change at runtime?

if not we could use a `val` instead of `var` and drop the extra `screenName` parameter 

```kotlin
// or non null and abstract to force all screens to supply a name
protected open val name: ScreenName? = null  

screenEvent = screenName?.let { ScreenEvent(it) }
...

screenEvent?.send(analyticsTracker)
```

vector/src/main/java/im/vector/app/features/analytics/screen/ScreenEvent.kt
```diff
@@ -0,0 +1,50 @@
...
+
+import android.os.SystemClock
+import im.vector.app.features.analytics.AnalyticsTracker
+import im.vector.app.features.analytics.plan.Screen
+import timber.log.Timber
+
+/**
+ * Track a screen display. Unique usage.
+ */
+class ScreenEvent(val screenName: Screen.ScreenName) {
```
ouchadam, line 27: small nit - mainly thinking out loud, tying the start time to the object creation is a little bit hidden and means where the instance is created is implicitly tied to the duration

another way to model the behaviour could be with a separate `ScreenTracker` (which could have its own injected dependencies for clocks, trackers etc)

```kotlin
data class TimedScreen(val startTime: Long, val screenName: ScreenName)

val timedScreen: TimedScreen = screenTracker.trackEntry(screenName)

screenTracker.trackExit(trackedScreen)
```

vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
bmarty: Yes, the screen name can change during the screen life. I made this for [this particular case - LoginActivity](https://github.com/vector-im/element-android/pull/4734/files#diff-4a938f906dc61c1b687841b80edc339b631e4b1a695c425d9f2ee458ee268149R207)
This is not ideal, I admit and a bit not clear...

vector/src/main/java/im/vector/app/features/analytics/screen/ScreenEvent.kt
bmarty: Yes, it's probably a little better. Happy to iterate on a dedicated PR later.

PR #4740, opened 2021-12-16
Automatic carousel rotation

ouchadam:
Part of #4584

- Adds automatic 5 second rotation to the Ftue auth carousel
- Also included a super handy extension from [SO](https://stackoverflow.com/questions/57505875/change-viewpager2-scroll-speed-when-sliding-programmatically/59235979#59235979) to control the view pager page transition speed as it's too fast out of the box
- Fixes the carousel direction for RTL locales

| GIF | RTL
| --- | --- |
|device-2021-12-16-144427|after-rtl

### DIFF
--- a/library/ui-styles/src/main/res/drawable/bg_carousel_page_1.xml
+++ b/library/ui-styles/src/main/res/drawable/bg_carousel_page_1.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android">
     <gradient
+        android:angle="@integer/rtl_mirror_flip"
         android:endColor="#3372C7DA"
         android:startColor="#33BBE7CF" />
 </shape>
--- a/vector/src/main/java/im/vector/app/core/extensions/Integer.kt
+++ b/vector/src/main/java/im/vector/app/core/extensions/Integer.kt
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) 2021 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.core.extensions
+
+fun Int.incrementAndWrap(max: Int, min: Int = 0, amount: Int = 1) = if (this == max) min else this + amount
--- a/vector/src/main/java/im/vector/app/core/extensions/Integer.kt
+++ b/vector/src/main/java/im/vector/app/core/extensions/Integer.kt
@@ -16,4 +16,11 @@
 
 package im.vector.app.core.extensions
 
-fun Int.incrementAndWrap(max: Int, min: Int = 0, amount: Int = 1) = if (this == max) min else this + amount
+fun Int.incrementByOneAndWrap(max: Int, min: Int = 0): Int {
+    val incrementedValue = this + 1
+    return if (incrementedValue > max) {
+        min
+    } else {
+        incrementedValue
+    }
+}
--- a/vector/src/main/java/im/vector/app/core/extensions/ViewPager2.kt
+++ b/vector/src/main/java/im/vector/app/core/extensions/ViewPager2.kt
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2021 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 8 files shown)


bmarty (approved):
LGTM.
Is it working fine for RTL languages (I did not test it)?

vector/src/main/java/im/vector/app/core/extensions/Integer.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.core.extensions
+
+fun Int.incrementAndWrap(max: Int, min: Int = 0, amount: Int = 1) = if (this == max) min else this + amount
```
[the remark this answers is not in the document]
ouchadam: https://github.com/vector-im/element-android/pull/4740/commits/8663714de4e69da218fbc969f25129962b2ae90a made the helper specific to increments of 1 to keep things simpler

vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthSplashCarouselFragment.kt
```diff
@@ -110,7 +110,7 @@ class FtueAuthSplashCarouselFragment @Inject constructor(
         views.splashCarousel.apply {
             val itemCount = carouselAdapter.itemCount
             tickerFlow(lifecycleScope, delayMillis = CAROUSEL_ROTATION_DELAY_MS)
-                    .onEach { setCurrentItem(currentItem.incrementAndWrap(max = itemCount - 1), duration = CAROUSEL_TRANSITION_TIME_MS) }
+                    .onEach { setCurrentItem(currentItem.incrementByOneAndWrap(max = itemCount - 1), duration = CAROUSEL_TRANSITION_TIME_MS) }
```
bmarty, line 113: I also like the simple form `setCurrentItem((currentItem + 1) % itemCount)`

bmarty:
Just a small remark while testing it:
Maybe disable the automatic pagination once the user had manually change the page.

library/ui-styles/src/main/res/drawable/bg_carousel_page_1.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android">
     <gradient
+        android:angle="@integer/rtl_mirror_flip"
```
ouchadam, line 4: @bmarty the rtl gradient is fixed by making use of the `rtl_mirror_flip`, ltr = 0 rtl = 180

vector/src/main/java/im/vector/app/core/extensions/ViewPager2.kt
```diff
@@ -0,0 +1,63 @@
...
+    val pxToDrag: Int = pagePxWidth * (item - currentItem)
+    val animator = ValueAnimator.ofInt(0, pxToDrag)
+    var previousValue = 0
+    val isRtl = this.layoutDirection == View.LAYOUT_DIRECTION_RTL
+
+    animator.addUpdateListener { valueAnimator ->
+        val currentValue = valueAnimator.animatedValue as Int
+        val currentPxToDrag = (currentValue - previousValue).toFloat()
+        kotlin.runCatching {
+            when {
```
ouchadam, line 41: @bmarty also updated the fake scrolling to take into account if we need to fake scroll ltr or rtl

ouchadam:
> Just a small remark while testing it: Maybe disable the automatic pagination once the user had manually change the page.

checking in with product to see what they think, I agree it could be quite annoying to manually select a page only for it to keep changing

bmarty:
Ok, thanks. On my side I really want to stop animation on user interaction. Especially the timer is not reset so the user can go back to the previous page to finish reading it and the page change again automatically just after, this is not nice.

bmarty:
Another note to discuss with product/design: some of the images contains wording, which is not ideal for localisation. Can we avoid that? FTR it is also possible to add images per languages, but I would prefer to avoid doing that if possible

image

image

CC @daniellekirkwood

ouchadam:
@bmarty we've already done it! :sunglasses:  https://www.figma.com/file/xdV4PuI3DlzA1EiBvbrggz/Login-Flow-v2?node-id=2235%3A41935 

just need to update the temporary images with the updated versions

bmarty:
Update for RTL tested OK 💯 (build 5364).

But I had a crash when fast clicking on the TabLayout `carouselIndicator`:

```bash
    Thread: main, Exception: java.lang.IllegalStateException: Cannot change current item when ViewPager2 is fake dragging
        at androidx.viewpager2.widget.ViewPager2.setCurrentItem(ViewPager2.java:604)
        at com.google.android.material.tabs.TabLayoutMediator$ViewPagerOnTabSelectedListener.onTabSelected(TabLayoutMediator.java:268)
```

ouchadam:
fix for the crash whilst our fake dragging is taking place https://github.com/vector-im/element-android/pull/4740/commits/a0bda0282498daeb2ee731dba36a6a982accedc4

updated the automatic transition behaviour to only schedule once the page settings, rather than on a constant ticker, matches the iOS behaviour https://github.com/vector-im/element-android/pull/4740/commits/70c82443ee235811b4e29a08f7fe20ddf2f42f15 (whilst we wait for confirmation about whether the automatic transitions should be fully disabled on user interaction)

bmarty (approved):
Thanks for the update

PR #4742, opened 2021-12-16
Fix crash ImmutableModelException

ariskotsomitopoulos:
This PR fixes the long press crash on a reply event on develop branch. Somehow in production works.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/tools/EventRenderingTools.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/tools/EventRenderingTools.kt
@@ -44,7 +44,7 @@ fun CharSequence.findPillsAndProcess(scope: CoroutineScope, processBlock: (PillI
 }
 
 fun CharSequence.linkify(callback: TimelineEventController.UrlClickCallback?): CharSequence {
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/tools/EventRenderingTools.kt
```diff
@@ -44,7 +45,7 @@ fun CharSequence.findPillsAndProcess(scope: CoroutineScope, processBlock: (PillI
 
 fun CharSequence.linkify(callback: TimelineEventController.UrlClickCallback?): CharSequence {
     val text = this.toString()
-    val spannable = toSpannable()
+    val spannable = SpannableStringBuilder(this)
```
bmarty, line 48: Strange...
Maybe worth adding a comment, to ensure that this (recent - ab70ee5ae29b1809754b89dee0fd155a0d6cb83a) change will not be repeated.

ouchadam: b73bae452a0be27b29848dd087a53f1591c55d97

ganfra (approved):
LGTM

PR #4698, opened 2021-12-13
Emoji fix

bmarty:
The main issue (#4691) is fixed, but now the app is crashing for another reason:

```
[...]

PR #4744, opened 2021-12-16
Emojis in emotes causing crashes on Android 12+

### LINKED ISSUE #4743: Certain emojis crash on Android 12+ when used in notices
### Steps to reproduce

1. Be on a android 12+ device
2. Post a `/me :tada: `  
3. The room now crashes on entry

#### What did you expect?
No crash!
#### What happened instead?
```
2021-12-16 18:25:59.939 14077-14077/im.vector.app.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: im.vector.app.debug, PID: 14077
    com.airbnb.epoxy.ImmutableModelException: The model was changed during the bind call. Position: 9 Model: MessageTextItem_{searchForPills=false, message=* 🎉, bindingOptions=null,

    Epoxy attribute fields on a model cannot be changed once the model is added to a controller. Check that these fields are not updated, or that the assigned objects are not mutated, outside of the buildModels method. The only exception is if the change is made inside an Interceptor callback. Consider using an interceptor if you need to change a model after it is added to the controller and before it is set on the adapter. If the model is already set on the adapter then you must call `requestModelBuild` instead to recreate all models.
        at com.airbnb.epoxy.EpoxyModel.validateStateHasNotChangedSinceAdded(EpoxyModel.java:466)
        at im.vector.app.features.home.room.detail.timeline.item.MessageTextItem_.handlePostBind(MessageTextItem_.java:59)
```

The is the same cause as #4691 but in another location 

### Your phone model

_No response_

### Operating system version

12+

_No response_

### Homeserver

matrix.org

### Will you send logs?

No

ouchadam:
Fixes #4743 

A follow up to the original emoji fix #4698, we had a missing case (the emote section eg `/me :tada:`) 

| BEFORE | AFTER |
| --- | --- |
|before-emoji-crash|after-emoji-crash

### DIFF
--- a/changelog.d/4743.bugfix
+++ b/changelog.d/4743.bugfix
@@ -1,0 +1,1 @@
+Fixes crash when launching rooms which contain emojis in the notice content on android 12+
--- a/changelog.d/4743.bugfix
+++ b/changelog.d/4743.bugfix
@@ -1,1 +1,1 @@
-Fixes crash when launching rooms which contain emojis in the notice content on android 12+
+Fixes crash when launching rooms which contain emojis in the emote content on android 12+
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
@@ -613,7 +613,7 @@ class MessageItemFactory @Inject constructor(
         val formattedBody = SpannableStringBuilder()
         formattedBody.append("* ${informationData.memberName} ")
         formattedBody.append(messageContent.getHtmlBody())
-
+        val bindingOptions = spanUtils.getBindingOptions(formattedBody)
         val message = formattedBody.linkify(callback)
 
         return MessageTextItem_()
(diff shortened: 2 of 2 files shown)


ariskotsomitopoulos (approved):

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -625,6 +625,7 @@ class MessageItemFactory @Inject constructor(
                         message(message)
                     }
                 }
+                .bindingOptions(bindingOptions)
```
bmarty, line 628: If only I had checked all the instance of `MessageTextItem_()` in this class...

changelog.d/4743.bugfix
```diff
@@ -0,0 +1 @@
+Fixes crash when launching rooms which contain emojis in the notice content on android 12+
```
bmarty, line 1:
```suggestion
Fixes crash when launching rooms which contain emojis in the emote content on android 12+
```

bmarty (approved):
LGTM, thanks!

ganfra (approved):
Thanks for the catch, are we sure its ok on all items now?

ouchadam:
> Thanks for the catch, are we sure its ok on all items now?

I'm not 100% as any expoy attribute which can contain an emoji is at risk on android 12+

Issue #4474, opened 2021-11-16
Review list of commands to display in the Thread's composer


daniellekirkwood:
There are items in the **/command** list that are only applicable to the room and so it does not make sense to have them appear in the Thread.

[Here is a list of commands and the proposal for which to remove from the Threads composer](https://docs.google.com/spreadsheets/d/16BTyouFCqqBYhA1rnDPOvRb7wlJACcmeqaq0td8Mehc/edit?usp=sharing)

- [ ] Remove the commands from the list (When a user types "/" a list appears, this should only contain the actions that the user can take)
- [ ] When a user tries to use a command that is not available in the Thread, show an error message that says: 

> **Command error**
> This command is not permitted in a thread
> OK

daniellekirkwood:
@janogarcia I've added a solution into the issue info here, please let us know if there's anything that should change

janogarcia:
I'd suggest tweaking the copy to make it a bit more helpful, by telling the users how to move forward.

```
**Unavailable command**
This command is not available in threads. Try running it in the room instead.
OK
```

...or something along the lines. What do you think, @daniellekirkwood?

Issue #4434, opened 2021-11-09
Disable/Enable thread feature from build configuration 


ariskotsomitopoulos:
We would like to have the ability to disable threads for a specific android build. The main reason for this is that some clients may not want that feature enabled.

Issue #4389, opened 2021-11-01
Message action bar for events in Thread


daniellekirkwood:
### Your use case

  - [Design](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1433%3A238070)

  - Add "View in Room" option to Root Thread event 
  - Do not allow user to create a Thread from a Thread

daniellekirkwood:
Removing Jano's assignment as Designs have been added to the issue

Issue #4387, opened 2021-11-01
Search results from Threads


daniellekirkwood:
- When I search for a message, return results from a Thread
  - Design (pending)
  - Unencrypted Rooms & ~~Encrypted Rooms~~

Designs are here: https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=2858%3A290566

EDIT: In testing on January 20 we found that tapping on the Thread root from a search result causes the app to crash.

daniellekirkwood:
@janogarcia are we expecting any changes to the way that threaded messages show up when a user searches for them? Is today's behaviour acceptable?

janogarcia:
@daniellekirkwood Thanks for flagging this one. It seems I couldn't update the issue description, so that it points to the proper Figma frame, and forgot to ping you on that.

https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=2858%3A290566

daniellekirkwood:
Thanks, @janogarcia ! I have added the link to the description and removed your assignment :)

daniellekirkwood:
We discussed this issue in standup today and have updated the issue description to remove encrypted rooms as this is not possible.

PR #4746, opened 2021-12-17
Threads P0 Release

ariskotsomitopoulos:
### Local Threads implementation for P0 release

- [x] https://github.com/vector-im/element-android/issues/4328
  - Using the Message action bar, choose a message from the Timeline to start a Thread by sending a response (that does not show in the Timeline)
  - [Design: Ready - Long press on message](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1304%3A244131)

- [x] https://github.com/vector-im/element-android/issues/4329
  - See a Thread, and send a message in response (that does not show in the Timeline)
  - [Design: Ready](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1304%3A244141)
  - https://github.com/vector-im/element-android/issues/4474

- [x] **See/Read previous responses in a Thread in the right order (Thread history)**
  - When I click to read a Thread, I should be able to see previous responses in the correct order
  - *This is acceptance criteria and does not need a design*
  - https://github.com/vector-im/element-android/issues/4331

- [x] **See/Read all Threads in a Room & Thread List view**
  - If I'm in a Room, I need to be able to scroll up through the Timeline and click to read a Thread
  - From the Timeline, I can use the Thread icon to view the Thread List view
  - [Design](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1304%3A244141)
  - [Another design...](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1304%3A244161)
  - https://github.com/vector-im/element-android/issues/4384

- [x] **See Thread Summary from Timeline**
  - From the Timeline I should be able to see a summary of the Thread
  - [Design](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1304%3A244141)
  - https://github.com/vector-im/element-android/issues/4385

- [x] **Send a link to a Thread**
  - Share a Thread message, in the same way I can Share a Room message including a Matrix.to link
  - [Design](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1430%3A237753)
  - https://github.com/vector-im/element-android/issues/4386

- [x] **Search results from Threads**
  - When I search for a message, return results from a Thread
  - [Design](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=2858%3A290566)
  - Unencrypted Rooms & Encrypted Rooms 
  - https://github.com/vector-im/element-android/issues/4387

- [x] **Message action bar for events in Thread**
  - _Design_ https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1433%3A238070
  - [x] Add "View in Room" option to Root Thread event 
  - Do not allow user to create a Thread from a Thread
  - https://github.com/vector-im/element-android/issues/4389

- [x] **Create Thread deleted state UI for when root message removed**
  - [Design](https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=1304%3A244192)
  - vector-im/element-ios#5473

- [x] **See Threads as Reply if don’t have Threads** 
  - If a client/bridge does not have Threads enabled, Thread responses show as Replies in the Timeline
  - https://github.com/vector-im/element-android/issues/4246

- [x] **Disable/Enable thread feature** from build configuration while some customers may not want to have threads
    - https://github.com/vector-im/element-android/issues/4434


bmarty:
First remark, after a first test.
The message ("Yo") send in a thread also appear on the main timeline:
image

ariskotsomitopoulos:
> First remark, after a first test. The message ("Yo") send in a thread also appear on the main timeline: image

Wow. that is really strange, are you sure you didn't accidentally wrote "Yo" again in the main timeline? You can verify that in the web client and check what it displays there. If thats not the case please describe how to reproduce, while it never happened to me or to the testing session.

bmarty:
@ariskotsomitopoulos My bad, I had the "Show hidden events in timeline" developer option enabled. However in this case maybe we could show `Debug: event type "m.room.message"` as it is done for message edition.

bmarty:
@ariskotsomitopoulos when displaying all events in the timeline you can inspired from [this code](https://github.com/vector-im/element-android/blob/main/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt#L149) to also filter Thread message when displaying the main timeline. Please check that it will not also filter on the thread timeline itself :)

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
```diff
@@ -61,6 +62,20 @@ object CommandParser {
                 return ParsedCommand.ErrorEmptySlashCommand
             }
 
+            // If the command is not supported by threads return error
+
+            if (BuildConfig.THREADING_ENABLED && isInThreadTimeline) {
+                val slashCommand = messageParts.first()
+                val notSupportedCommandsInThreads = Command.values().filter {
```
bmarty, line 69: Maybe move this val as a lazy class member to avoid computing it each time a message is sent in a thread

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewEvents.kt
```diff
@@ -32,6 +32,8 @@ sealed class MessageComposerViewEvents : VectorViewEvents {
     data class JoinRoomCommandSuccess(val roomId: String) : SendMessageResult()
     class SlashCommandError(val command: Command) : SendMessageResult()
     class SlashCommandUnknown(val command: String) : SendMessageResult()
+    class SlashCommandNotSupportedInThreads(val command: String) : SendMessageResult()
```
bmarty, line 35: Maybe `command` could be of type `Command` here, since this is a valid command.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -183,135 +183,176 @@ class MessageComposerViewModel @AssistedInject constructor(
...
+                        is ParsedCommand.ErrorCommandNotSupportedInThreads -> {
+                            _viewEvents.post(MessageComposerViewEvents.SlashCommandNotSupportedInThreads(slashCommandResult.slashCommand))
+                        }
+                        is ParsedCommand.SendPlainText                     -> {
                             // Send the text message to the room, without markdown
-                            room.sendTextMessage(slashCommandResult.message, autoMarkdown = false)
+                            if (state.rootThreadEventId != null) {
+                                room.replyInThread(
+                                        rootThreadEventId = state.rootThreadEventId,
+                                        replyInThreadText = action.text.toString(),
```
bmarty, line 218: why you use `action.text.toString()` and not `slashCommandResult.message` as on line 221?

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -183,135 +183,176 @@ class MessageComposerViewModel @AssistedInject constructor(
...
-                        is ParsedCommand.PartRoom                 -> {
+                        is ParsedCommand.PartRoom                          -> {
                             handlePartSlashCommand(slashCommandResult)
                         }
-                        is ParsedCommand.SendEmote                -> {
-                            room.sendTextMessage(slashCommandResult.message, msgType = MessageType.MSGTYPE_EMOTE, autoMarkdown = action.autoMarkdown)
+                        is ParsedCommand.SendEmote                         -> {
+                            state.rootThreadEventId?.let {
+                                room.replyInThread(
+                                        rootThreadEventId = state.rootThreadEventId,
```
bmarty, line 273: Use `it`?

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -183,135 +183,176 @@ class MessageComposerViewModel @AssistedInject constructor(
...
                         }
-                        is ParsedCommand.SendRainbowEmote         -> {
-                            slashCommandResult.message.toString().let {
-                                room.sendFormattedTextMessage(it, rainbowGenerator.generate(it), MessageType.MSGTYPE_EMOTE)
-                            }
+                        is ParsedCommand.SendRainbowEmote                  -> {
+                            val message = slashCommandResult.message.toString()
+                            state.rootThreadEventId?.let {
+                                room.replyInThread(
+                                        rootThreadEventId = state.rootThreadEventId,
```
bmarty, line 296: same

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -183,135 +183,176 @@ class MessageComposerViewModel @AssistedInject constructor(
...
-                        is ParsedCommand.SendSpoiler              -> {
-                            room.sendFormattedTextMessage(
-                                    "[${stringProvider.getString(R.string.spoiler)}](${slashCommandResult.message})",
-                                    "<span data-mx-spoiler>${slashCommandResult.message}</span>"
+                        is ParsedCommand.SendSpoiler                       -> {
+                            val text = "[${stringProvider.getString(R.string.spoiler)}](${slashCommandResult.message})"
+                            val formattedText = "<span data-mx-spoiler>${slashCommandResult.message}</span>"
+                            state.rootThreadEventId?.let {
+                                room.replyInThread(
+                                        rootThreadEventId = state.rootThreadEventId,
```
bmarty, line 310: same. or maybe using if/else will be clearer for all those cases

vector/src/main/java/im/vector/app/features/home/room/threads/arguments/ThreadTimelineArgs.kt
```diff
@@ -0,0 +1,28 @@
...
+package im.vector.app.features.home.room.threads.arguments
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+data class ThreadTimelineArgs(
+        val roomId: String,
+        val displayName: String?,
+        val avatarUrl: String?,
```
bmarty, line 26: same remark

vector/src/main/java/im/vector/app/features/home/room/threads/list/model/ThreadListModel.kt
```diff
@@ -0,0 +1,106 @@
...
+import im.vector.app.core.extensions.clearDrawables
+import im.vector.app.core.extensions.setLeftDrawable
+import im.vector.app.core.utils.DimensionConverter
+import im.vector.app.features.displayname.getBestName
+import im.vector.app.features.home.AvatarRenderer
+import org.matrix.android.sdk.api.session.threads.ThreadNotificationState
+import org.matrix.android.sdk.api.util.MatrixItem
+
+@EpoxyModelClass(layout = R.layout.item_thread_list)
+abstract class ThreadListModel : VectorEpoxyModel<ThreadListModel.Holder>() {
```
bmarty, line 40: We call those classes with `Item` suffix in the codebase. So maybe `ThreadItem` here.

vector/src/main/res/drawable/ic_reply_in_thread.xml
```diff
@@ -0,0 +1,8 @@
...
+    <path android:fillColor="#00000000" android:pathData="M7,9H17"
```
bmarty, line 4: Looks like a big vector drawable.

vector/src/main/res/layout/fragment_room_detail.xml
```diff
@@ -18,109 +18,21 @@
             android:layout_height="wrap_content"
             android:minHeight="48dp"
             android:visibility="gone"
-            tools:visibility="visible" />
+            tools:visibility="gone" />
```
bmarty, line 21: Rename the file to `fragment_timeline`?

vector/src/main/res/layout/fragment_thread_list.xml
```diff
@@ -0,0 +1,115 @@
...
+        android:background="?android:colorBackground"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/threadListAppBarLayout"
+        tools:listitem="@layout/item_thread_list"
+        tools:visibility="gone" />
+
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/threadListEmptyConstraintLayout"
```
bmarty, line 41: We also have a `StateView` which could be reused.

vector/src/main/res/layout/view_room_detail_thread_toolbar.xml
```diff
@@ -0,0 +1,53 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="4dp"
+        android:ellipsize="end"
+        android:maxLines="1"
+        android:textColor="?vctr_content_secondary"
+        app:layout_constraintBottom_toBottomOf="@id/roomToolbarThreadImageView"
+        app:layout_constraintStart_toEndOf="@id/roomToolbarThreadImageView"
+        app:layout_constraintTop_toTopOf="@id/roomToolbarThreadImageView"
+        tools:text="RoomName"
```
bmarty, line 50: Could use `@sample/rooms.json/data/name` or create new sample data dedicated to thread?

vector/src/main/res/layout/view_thread_room_summary.xml
```diff
@@ -0,0 +1,59 @@
...
+        android:ellipsize="end"
+        android:maxLines="1"
+        android:textColor="?vctr_content_secondary"
+        app:layout_constrainedWidth="true"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintHorizontal_bias="0"
+        app:layout_constraintStart_toEndOf="@id/messageThreadSummaryAvatarImageView"
+        app:layout_constraintTop_toTopOf="parent"
+        tools:text="Hello There, whats up! Its a large sentence whats up! Its a large centence" />
```
bmarty, line 58: Could use `@sample/messages.json/data/message`

vector/src/main/res/values/strings.xml
```diff
@@ -1031,6 +1037,19 @@
...
+    <string name="thread_timeline_title">Thread</string>
+    <string name="thread_list_title">Threads</string>
+    <string name="thread_list_modal_title">Filter</string>
+    <string name="thread_list_modal_all_threads_title">All Threads</string>
+    <string name="thread_list_modal_all_threads_subtitle">Shows all threads from current room</string>
+    <string name="thread_list_modal_my_threads_title">My Threads</string>
+    <string name="thread_list_modal_my_threads_subtitle">Shows all threads you’ve participated in</string>
+    <string name="thread_list_empty_title">Keep discussions organised with threads</string>
+    <string name="thread_list_empty_subtitle">Threads help keep your conversations on-topic and easy to track.</string>
+    <string name="thread_list_empty_notice">Tip: Long tap a message and use “Reply in thread”.</string>
```
bmarty, line 1051: Maybe better to use %s instead of "Reply in thread" to avoid mistake in translation. You can add a comment above saying that the %s will be replaced by the value of `reply_in_thread`. I see that there is already a diff with the case :)

bmarty (commented):
Nice huge work!
I have made a static review, here are my 50 comments :)
Thanks!

matrix-sdk-android-flow/src/main/java/org/matrix/android/sdk/flow/FlowRoom.kt
```diff
@@ -98,6 +98,20 @@ class FlowRoom(private val room: Room) {
     fun liveNotificationState(): Flow<RoomNotificationState> {
         return room.getLiveRoomNotificationState().asFlow()
     }
+
+    fun liveThreadList(): Flow<List<TimelineEvent>> {
```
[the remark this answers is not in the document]
ariskotsomitopoulos: Yes, threading is actual just a relation between the existing models. So there is not need for an extra model in order to reuse existing features. More specific when a `TimelineEvent` is a root of a thread that means that all thread events **relatesTo** that `TimelineEvent`. Yes, I plan to document how threads work/architecture in detail

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/timeline/TimelineService.kt
```diff
@@ -55,4 +55,40 @@ interface TimelineService {
      * Returns a snapshot list of TimelineEvent with EventType.MESSAGE and MessageType.MSGTYPE_IMAGE or MessageType.MSGTYPE_VIDEO.
      */
     fun getAttachmentMessages(): List<TimelineEvent>
+
+    /**
+     * Get a live list of all the thread for the specified roomId
+     * @return the [LiveData] of [TimelineEvent]
+     */
+    fun getAllThreadsLive(): LiveData<List<TimelineEvent>>
```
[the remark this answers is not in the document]
ariskotsomitopoulos: Yes, maybe it would be good, however some of the functions are for the main timeline like:
`getNumberOfLocalThreadNotificationsLive`
`getNumberOfLocalThreadNotifications`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/timeline/TimelineSettings.kt
```diff
@@ -27,5 +27,14 @@ data class TimelineSettings(
...
+        val buildReadReceipts: Boolean = true,
+        /**
+         * The root thread eventId if this is a thread timeline, or null if this is NOT a thread timeline
+         */
+        val rootThreadEventId: String? = null) {
+
+    /**
+     * Returns true if this is a thread timeline or false otherwise
+     */
+    fun isThreadTimeline() = rootThreadEventId != null
```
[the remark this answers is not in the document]
ariskotsomitopoulos: Why do you think is a good idea to add more properties in a data class and not simply parse the existing `rootThreadEventId` when we need it. This function is to avoid repeating this check only when is needed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/threads/ThreadTimelineEvent.kt
```diff
@@ -0,0 +1,28 @@
...
+
+package org.matrix.android.sdk.api.session.threads
+
+import org.matrix.android.sdk.api.session.room.timeline.TimelineEvent
+
+/**
+ * This class contains a thread TimelineEvent along with a boolean that
+ * determines if the current user has participated in that event
+ */
+data class ThreadTimelineEvent(
```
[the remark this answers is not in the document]
ariskotsomitopoulos: Maybe, Thread API is not done yet, so we do not know yet the final model

vector/src/main/java/im/vector/app/features/command/Command.kt
```diff
@@ -24,41 +24,41 @@ import im.vector.app.R
...
+    RAINBOW("/rainbow", "<message>", R.string.command_description_rainbow, false, true),
+    RAINBOW_EMOTE("/rainbowme", "<message>", R.string.command_description_rainbow_emote, false, true),
+    CLEAR_SCALAR_TOKEN("/clear_scalar_token", "", R.string.command_description_clear_scalar_token, false, false),
+    SPOILER("/spoiler", "<message>", R.string.command_description_spoiler, false, true),
+    SHRUG("/shrug", "<message>", R.string.command_description_shrug, false, true),
+    LENNY("/lenny", "<message>", R.string.command_description_lenny, false, true),
+    PLAIN("/plain", "<message>", R.string.command_description_plain, false, true),
+    WHOIS("/whois", "<user-id>", R.string.command_description_whois, false, true),
+    DISCARD_SESSION("/discardsession", "", R.string.command_description_discard_session, false, false),
+    CONFETTI("/confetti", "<message>", R.string.command_confetti, false, false),
```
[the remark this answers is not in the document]
ariskotsomitopoulos: We have a specific [doc](https://docs.google.com/spreadsheets/d/16BTyouFCqqBYhA1rnDPOvRb7wlJACcmeqaq0td8Mehc/edit#gid=0) with POs for that, contains the commands that should be used within threads and why. Maybe you communicate with Danielle and request a change?

vector/src/main/java/im/vector/app/features/command/Command.kt
```diff
@@ -24,41 +24,41 @@ import im.vector.app.R
...
+    SNOWFALL("/snowfall", "<message>", R.string.command_snow, false, false),
```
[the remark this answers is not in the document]
ariskotsomitopoulos: Same with the above comment

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
ariskotsomitopoulos: keep in mind `CommandParser` is a **static** class (object). Can you explain a bit more what you mean?

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
ariskotsomitopoulos: Actually, it's not exactly the same. Current code with `?:` will also send message in current timeline as fallback (if null). Changing with if/else will not have the same behaviour

vector/src/main/java/im/vector/app/features/home/room/threads/arguments/ThreadListArgs.kt
```diff
@@ -0,0 +1,27 @@
...
+package im.vector.app.features.home.room.threads.arguments
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+data class ThreadListArgs(
+        val roomId: String,
+        val displayName: String?,
+        val avatarUrl: String?,
```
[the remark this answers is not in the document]
ariskotsomitopoulos: Currently it is, I wanted to make it able to change easily if business decide to user a thread specific icon but yes I can check that

vector/src/main/java/im/vector/app/features/home/room/threads/list/model/ThreadListModel.kt
ariskotsomitopoulos: Ok, I thought doc suggested to end with `model` also it automatically removes the model when you building epoxy components, so `somethingModel` can be written as` something{ }`. However, yes I will change it to be aligned with the whole application naming convention

vector/src/main/res/layout/fragment_thread_list.xml
ariskotsomitopoulos: Good to know, I might use this when I will ad API support calls to that screen

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
```diff
@@ -96,6 +96,11 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
 
     companion object {
 
+        /**
+         * Determines whether or not thread messages are enabled
+         */
+        var areThreadMessagesEnabled: Boolean = false
+
```
ariskotsomitopoulos, line 103: I do not really like my solution here, I tried to find a better way using DI but didn't work as expected. What I already tried:
- Interface in **matrix-sdk**
- Implementation in **vector**
- Bind Implementation to Interface in vector module

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/timeline/TimelineService.kt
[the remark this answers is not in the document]
bmarty: Ok, so splitting into 2 different interfaces will definitely make it clearer!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/timeline/TimelineSettings.kt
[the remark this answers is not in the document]
bmarty: Adding a `val` in the body of a `data class` is not the same thing than adding a new property (i.e. a `val` in the `constructor`) of a data class.
Replacing this line by `val isThreadTimeline = rootThreadEventId != null` will just compute this val once in the constructor.
Anyway this is a minor remark!

matrix-sdk-android-flow/src/main/java/org/matrix/android/sdk/flow/FlowRoom.kt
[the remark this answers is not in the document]
bmarty: Maybe create a `typealias`? (I am thinking of a WTF moment when developers will see this API)

```kotlin
typealias ThreadRootEvent = TimelineEvent
```

maybe?

vector/src/main/java/im/vector/app/features/command/Command.kt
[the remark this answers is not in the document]
bmarty: I did not know that doc, thanks.
OK, I will follow the PO decision, but from a user POV I do not see why I could not send confetti and snow in a thread :)

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewEvents.kt
bmarty: Have you seen this comment?

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
bmarty: I see, `replyInThread()` can return `null`. From a SDK user POV this is a bit weird, I would rather have an error in this case and not fallback to send the event in the regular timeline.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
```diff
@@ -429,6 +440,61 @@ class MessageActionsViewModel @AssistedInject constructor(@Assisted
...
+    /**
+     * Determine whether or no the selected event is a root thread event from within
+     * a thread timeline
+     */
+    private fun canViewInRoom(event: TimelineEvent,
+                              messageContent: MessageContent?,
+                              actionPermissions: ActionPermissions): Boolean {
+        // Only event of type EventType.MESSAGE are supported for the moment
+        if (!BuildConfig.THREADING_ENABLED) return false
+        if (!initialState.isFromThreadTimeline) return false
```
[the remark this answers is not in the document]
bmarty: I see, thanks. But the other test below are a bit weird to me. We should test if this is the first event of the RecyclerView, no?

vector/src/main/java/im/vector/app/features/home/room/threads/list/views/ThreadListFragment.kt
```diff
@@ -90,6 +91,12 @@ class ThreadListFragment @Inject constructor(
         renderToolbar()
     }
 
+    private fun initTextConstants() {
+        views.threadListEmptyNoticeTextView.text = String.format(
+                resources.getString(R.string.thread_list_empty_notice),
+                resources.getString(R.string.reply_in_thread))
```
bmarty, line 97: Why not using `getString(R.string.thread_list_empty_notice, getString(R.string.reply_in_thread))` ?

vector/src/main/java/im/vector/app/features/command/Command.kt
```diff
@@ -45,8 +45,8 @@ enum class Command(val command: String,
     REMOVE_USER("/remove", arrayOf("/kick"), "<user-id> [reason]", R.string.command_description_kick_user, false, false),
     CHANGE_DISPLAY_NAME("/nick", null, "<display-name>", R.string.command_description_nick, false, false),
     CHANGE_DISPLAY_NAME_FOR_ROOM("/myroomnick", arrayOf("/roomnick"), "<display-name>", R.string.command_description_nick_for_room, false, false),
-    ROOM_AVATAR("/roomavatar", null, "<mxc_url>", R.string.command_description_room_avatar, true /* Since user has to know the mxc url */, false),
-    CHANGE_AVATAR_FOR_ROOM("/myroomavatar", null, "<mxc_url>", R.string.command_description_avatar_for_room, true /* Since user has to know the mxc url */, false),
+    ROOM_AVATAR("/roomavatar", null, "<mxc_url>", R.string.command_description_room_avatar, true /* User has to know the mxc url */, false),
+    CHANGE_AVATAR_FOR_ROOM("/myroomavatar", null, "<mxc_url>", R.string.command_description_avatar_for_room, true /* User has to know the mxc url */, false),
```
bmarty, line 49: `Since` was used to justify the fact that those commands are for developer.
Anyway I am happy with this change.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/relation/RelationService.kt
```diff
@@ -153,5 +153,5 @@ interface RelationService {
      * from the backend
      * @param rootThreadEventId the root thread eventId
      */
-    suspend fun fetchThreadTimeline(rootThreadEventId: String): List<Event>
+    suspend fun fetchThreadTimeline(rootThreadEventId: String): Boolean
```
bmarty, line 156: Would be nice to document what is the meaning of a returned value `false` or `true`.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
bmarty: For global configuration of the SDK by the application, we use the class `MatrixConfiguration` if it helps.

ariskotsomitopoulos: It does not because `MatrixConfiguration` is initiated too early on Application level before @Injections are available like userPreferences etc.  I think the best solution is to store it in sharedPref within the matrix SDK

matrix-sdk-android-flow/src/main/java/org/matrix/android/sdk/flow/FlowRoom.kt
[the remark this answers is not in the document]
ariskotsomitopoulos: Nice idea +1

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/timeline/TimelineSettings.kt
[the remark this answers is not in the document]
ariskotsomitopoulos: Actually it's interesting to see another perspective! So this is a battle between `CPU` vs `RAM`
- `fun isThreadTimeline() = rootThreadEventId != null` will compute it every time but will never store anything else in the ram
-  `val isThreadTimeline = rootThreadEventId != null` will compute it only once and use more ram to store it

Both approaches work for me, yes its minor but interesting :)

vector/src/main/java/im/vector/app/features/command/Command.kt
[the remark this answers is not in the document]
ariskotsomitopoulos: Yes, I agree I will ask if we have any strong point for that 👍

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
ariskotsomitopoulos: Now it's clear! Totally will do that +1

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewEvents.kt
ariskotsomitopoulos: Yes, will update accordingly

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
ariskotsomitopoulos: Correct, this is supposed to never happen so I agree, will refactor it a bit

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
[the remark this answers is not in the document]
ariskotsomitopoulos: Thanks I agree, we can remove some of the tests, I will try to minimise them. However, checking only if it's the first cell is not enough while this might change in a future feature. Maybe insert a header or something prior to that so might be better to be decoupled.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/relation/RelationService.kt
ariskotsomitopoulos: Thanks, this code is not yet in use.  I will refactor it completely so I will update the documented accordingly

vector/src/main/java/im/vector/app/features/command/Command.kt
ariskotsomitopoulos: Sorry for that, I tried to remove some letters while the line was over 160 letters !!

bmarty:
- [ ] I have a crash when clicking on "View in room" from a thread (will attach a RS)

ariskotsomitopoulos:
> 
Thank you, no need for a RS, it's the `IndexOutOfBoundsException` due to the new timeline, I will try to fix that.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/TimelineControllerInterceptorHelper.kt
```diff
@@ -115,7 +115,10 @@ class TimelineControllerInterceptorHelper(private val positionOfReadMarker: KMut
     private fun MutableList<EpoxyModel<*>>.addForwardPrefetchIfNeeded(timeline: Timeline?, callback: TimelineEventController.Callback?) {
         val shouldAddForwardPrefetch = timeline?.hasMoreToLoad(Timeline.Direction.FORWARDS) ?: false
         if (shouldAddForwardPrefetch) {
-            val indexOfPrefetchForward = DEFAULT_PREFETCH_THRESHOLD.coerceAtMost(size - 1)
+            val indexOfPrefetchForward = DEFAULT_PREFETCH_THRESHOLD
+                    .coerceAtMost(size - 1)
+                    .coerceAtLeast(0)
```
bmarty, line 120: Just on the form: `coerceIn()` exists.

ariskotsomitopoulos: Nice, let's use that, thanks!

ariskotsomitopoulos: @bmarty  Actually, `coerceIn` is not exactly the same. So I will revert using `coerceIn`. When the range is empty or like [0 to -5] `coerceIn` will not work

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/timeline/TimelineService.kt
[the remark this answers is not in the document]
ariskotsomitopoulos: Done

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
ariskotsomitopoulos: Done!

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewEvents.kt
ariskotsomitopoulos: Done!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/threads/ThreadsService.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Returns whether or not the current user is participating in the thread
+     * @param rootThreadEventId the eventId of the current thread
+     */
+    fun isUserParticipatingInThread(rootThreadEventId: String): Boolean
+
+    /**
+     * Enhance the thread list with the edited events if needed
+     * @return the [LiveData] of [TimelineEvent]
```
bmarty, line 59: this comment is not accurate

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/threads/ThreadsService.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Get a live list of all the local unread threads for the specified roomId
+     * @return the [LiveData] of [TimelineEvent]
+     */
+    fun getNumberOfLocalThreadNotificationsLive(): LiveData<List<TimelineEvent>>
+
+    /**
+     * Get a list of all the local unread threads for the specified roomId
+     * @return the [LiveData] of [TimelineEvent]
```
bmarty, line 47: this comment is not accurate

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/threads/ThreadsService.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Get a list of all the TimelineEvents which have thread replies for the specified roomId
+     * @return the [LiveData] of [TimelineEvent]
+     */
+    fun getAllThreads(): List<TimelineEvent>
+
+    /**
+     * Get a live list of all the local unread threads for the specified roomId
+     * @return the [LiveData] of [TimelineEvent]
```
bmarty, line 41: this comment is not accurate

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/threads/ThreadsService.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Get a live list of all the TimelineEvents which have thread replies for the specified roomId
+     * @return the [LiveData] of [TimelineEvent]
+     */
+    fun getAllThreadsLive(): LiveData<List<TimelineEvent>>
+
+    /**
+     * Get a list of all the TimelineEvents which have thread replies for the specified roomId
+     * @return the [LiveData] of [TimelineEvent]
```
bmarty, line 35: this comment is not accurate

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/threads/ThreadsService.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.matrix.android.sdk.api.session.room.timeline.TimelineEvent
+
+/**
+ * This interface defines methods to interact with threads related features. It's implemented at the room level within the main timeline.
+ */
+interface ThreadsService {
+
+    /**
+     * Get a live list of all the TimelineEvents which have thread replies for the specified roomId
+     * @return the [LiveData] of [TimelineEvent]
```
bmarty, line 29: this comment is not accurate

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/threads/ThreadsService.kt
```diff
@@ -0,0 +1,68 @@
...
+    fun isUserParticipatingInThread(rootThreadEventId: String): Boolean
+
+    /**
+     * Enhance the thread list with the edited events if needed
+     * @return the [LiveData] of [TimelineEvent]
+     */
+    fun mapEventsWithEdition(threads: List<TimelineEvent>): List<TimelineEvent>
+
+    /**
+     * Marks the current thread as read. This is a local implementation
```
bmarty, line 64: Why do you mean by `This is a local implementation`? I think I understand, but for a SDK user who may read this documentation, it will be maybe not clear enough.

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
```diff
@@ -411,6 +404,29 @@ class CommandParser @Inject constructor() {
...
+    private fun getNotSupportedByThreads(isInThreadTimeline: Boolean, slashCommand: String): Command? {
+        if (isInThreadTimeline) {
+            notSupportedThreadsCommands.firstOrNull {
+                it.command == slashCommand
+            }?.let {
+                return it
+            }
+        }
+        return null
+    }
```
bmarty, line 428: Nit: could be re-write with
```kotlin
    private fun getNotSupportedByThreads(isInThreadTimeline: Boolean, slashCommand: String): Command? {
        return if (isInThreadTimeline) {
            notSupportedThreadsCommands.firstOrNull { it.command == slashCommand }
        } else {
            null
        }
    }
```

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -1947,14 +1930,11 @@ class TimelineFragment @Inject constructor(
...
             }
-            else                                 -> {
-                onThreadSummaryClicked(informationData.eventId, isRootThreadEvent)
-            }
             is MessageLocationContent            -> {
                 handleShowLocationPreview(messageContent, informationData.senderId)
             }
             else                                 -> {
-                Timber.d("No click action defined for this message content")
+                onThreadSummaryClicked(informationData.eventId, isRootThreadEvent)
```
bmarty, line 1937: This change is not ideal (but I know it's due to a recent merge), I would prefer to read:
```kotlin
if (isRootThreadEvent) {
    onThreadSummaryClicked(informationData.eventId)
} else {
    Timber.d("No click action defined for this message content")
}
```

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -1789,87 +1904,81 @@ class RoomDetailFragment @Inject constructor(
...
                 val action = RoomDetailAction.DownloadOrOpen(informationData.eventId, informationData.senderId, messageContent)
-                roomDetailViewModel.handle(action)
+                timelineViewModel.handle(action)
             }
             is EncryptedEventContent             -> {
-                roomDetailViewModel.handle(RoomDetailAction.TapOnFailedToDecrypt(informationData.eventId))
+                if (isRootThreadEvent) {
+                    onThreadSummaryClicked(informationData.eventId, isRootThreadEvent)
+                } else {
+                    timelineViewModel.handle(RoomDetailAction.TapOnFailedToDecrypt(informationData.eventId))
```
bmarty, line 1930: With this change, there is no way to get UISI details if the event is a root thread event. But when we have a UISI, can the app knows that this is a root thread event? If yes, I think we must change that.
CC @BillCarsonFr

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -1789,87 +1904,81 @@ class RoomDetailFragment @Inject constructor(
...
                     .setNegativeButton(R.string.no, null)
                     .setPositiveButton(R.string.yes) { _, _ ->
-                        roomDetailViewModel.handle(RoomDetailAction.CancelSend(action.eventId, false))
+                        timelineViewModel.handle(RoomDetailAction.CancelSend(action.eventId, false))
                     }
                     .show()
         }
     }
 
+    override fun onThreadSummaryClicked(eventId: String, isRootThreadEvent: Boolean) {
```
bmarty, line 1969: I would remove the second param, since it will always be true at the call sites

bmarty (commented):
Thanks for the update!
I added more comments

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ariskotsomitopoulos: Actually when called from within the TimelineFragment it is not always true

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ariskotsomitopoulos: No need to investigate that I can remove that to be sure, it will still be clickable the thread summary so we will have both actions :)

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/TimelineControllerInterceptorHelper.kt
bmarty: OK

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
bmarty: OK

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ariskotsomitopoulos: [line](line) will be true if this is a root thread event and false if this is a normal message, and will act accordignly

ariskotsomitopoulos: If the link does not work here is the part of code:
```
    override fun onEventCellClicked(informationData: MessageInformationData, messageContent: Any?, view: View, isRootThreadEvent: Boolean) {
        when (messageContent) {
            is MessageVerificationRequestContent -> {
                timelineViewModel.handle(RoomDetailAction.ResumeVerification(informationData.eventId, null))
            }
            is MessageWithAttachmentContent      -> {
                val action = RoomDetailAction.DownloadOrOpen(informationData.eventId, informationData.senderId, messageContent)
                timelineViewModel.handle(action)
            }
            is EncryptedEventContent             -> {
                    timelineViewModel.handle(RoomDetailAction.TapOnFailedToDecrypt(informationData.eventId))
            }
            is MessageLocationContent            -> {
                handleShowLocationPreview(messageContent, informationData.senderId)
            }
            else                                 -> {
                val handled = onThreadSummaryClicked(informationData.eventId, isRootThreadEvent)
                Timber.i(" ---->isRootThreadEvent = $isRootThreadEvent")
                if (!handled) {
                    Timber.d("No click action defined for this message content")
                }
            }
        }
    }
```

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -168,6 +171,11 @@ class MessageItemFactory @Inject constructor(
             return noticeItemFactory.create(params)
         }
 
+        // This is a thread event and we will [debug] display it when we are in the main timeline
```
bmarty, line 174: I would have move the comment inside the `if` block.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/TimelineEventVisibilityHelper.kt
```diff
@@ -117,7 +117,7 @@ class TimelineEventVisibilityHelper @Inject constructor(private val userPreferen
             rootThreadEventId: String?
     ): Boolean {
         // If show hidden events is true we should always display something
-        if (userPreferencesProvider.shouldShowHiddenEvents()) {
+        if (userPreferencesProvider.shouldShowHiddenEvents() && !isFromThreadTimeline) {
```
bmarty, line 120: Does it means that a message edition in a thread will not be displayed as a debug event in the thread? This is fine, just want to confirm. I will do some test when the APK will be ready

bmarty (commented):
Thanks for the update. I will do some test today

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/TimelineEventVisibilityHelper.kt
ariskotsomitopoulos: Exactly! It will be displayed in the main timeline. We can of course further improve all those thing when threads are more stable

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
ariskotsomitopoulos: True, updated

bmarty (approved):
Quick smoke test and it seems to work fine! Thank!
We can merge if CI is happy.
I will create an issue to test the new screens in the sanity test.

PR #4747, opened 2021-12-17
cleaning rendering of state event in timeline

BillCarsonFr:
Quick cleaning of state event rendering in timeline.
Added some validation around state_key, also state events are not encrypted


bmarty (approved):
LGTM, thanks! I will add a changelog after merging to develop.

EDIT: 9b94f1e370f0162ad5ff115d34a977e87990b0f5

PR #4758, opened 2021-12-17
Wrong emojis in composer

### LINKED ISSUE #4756: Emoji inconsistencies in composer
### Steps to reproduce

1. Add a new emoji such as :face_exhaling: via the keyboard
2. Notice the unicode components are displayed
3. Submitting the message then shows the correct emoji again 

#### What did you expect?
The see the correct emoji :face_exhaling:

#### What happened instead?
Got :open_mouth: :dash: 

Screenshot_20211217_094434

### Your phone model

Pocofone f1

### Operating system version

android 10 (MIUI 12.0.3.0)

1.3.10

### Homeserver

matrix.org

### Will you send logs?

No

ouchadam:
Fixes #4756 Inconsistent emoji input in the composer view

- Switches the `ComposeEditText` base view to the `AppCompatEditText` which now supports emojis out of the box

| BEFORE | AFTER |
| --- | --- |
|before-new-emoji|after-new-emoji

### DIFF
--- a/changelog.d/4756.bugfix
+++ b/changelog.d/4756.bugfix
@@ -1,0 +1,1 @@
+Fixes newer emojis rendering strangely when inserting from the system keyboard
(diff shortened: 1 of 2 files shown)


ganfra (commented):
Should we not remove completely vanniktech.emoji lib?

ouchadam:
> Should we not remove completely vanniktech.emoji lib?

it's being used for the dedicated emoji picker button in the composer (unless there's a separate artifact)
I wonder if we could invoke the system keyboard on the emoji page instead...

ganfra:
> 

I guess we should still support keyboard without emojis...

ganfra (approved):

bmarty (approved):

PR #4783, opened 2021-12-21
Unable to change avatar due to `NetworkOnMainThread`

### LINKED ISSUE #4767: It is not possiable to changing user  avatar
### Steps to reproduce

It is not possiable to changing user  avatar
image
image
cut and uploading photo
image
image

### Outcome

final Exception：android.os.NetworkOnMainThreadException
image
image

### Your phone model

18208727060

### Operating system version

Android 31

_No response_

### Homeserver

https://im.ynzzpt.cn/

### Will you send logs?

Yes

ouchadam:
Fixes #4767 Unable to change avatar

We had a report of `NetworkOnMainThread` whilst changing avatar, I've been unable to reproduce the issue but did notice we have an unnecessary main thread dispatcher being used (looks like it was introduced as part of the suspend migration).

- Switches the `main` dispatcher to `io`, the user has also confirmed this fixed their issue
- Bundles the file reading + temp file creation into the same io context 

| |
| --- |
|changing-avatar

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/profile/DefaultProfileService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/profile/DefaultProfileService.kt
@@ -68,7 +68,7 @@ internal class DefaultProfileService @Inject constructor(private val taskExecuto
     }
 
     override suspend fun updateAvatar(userId: String, newAvatarUri: Uri, fileName: String) {
-        withContext(coroutineDispatchers.main) {
+        withContext(coroutineDispatchers.io) {
             val response = fileUploader.uploadFromUri(newAvatarUri, fileName, MimeTypes.Jpeg)
             setAvatarUrlTask.execute(SetAvatarUrlTask.Params(userId = userId, newAvatarUrl = response.contentUri))
             userStore.updateAvatar(userId, response.contentUri)
--- a/changelog.d/4767.bugfix
+++ b/changelog.d/4767.bugfix
@@ -1,0 +1,1 @@
+Fixing unable to change change avatar in some scenarios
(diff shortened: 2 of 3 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/profile/DefaultProfileService.kt
```diff
@@ -68,7 +68,7 @@ internal class DefaultProfileService @Inject constructor(private val taskExecuto
     }
 
     override suspend fun updateAvatar(userId: String, newAvatarUri: Uri, fileName: String) {
-        withContext(coroutineDispatchers.main) {
+        withContext(coroutineDispatchers.io) {
```
ouchadam, line 71: this is the fix

ganfra: Its something we already had fixed, not sure why its back. But fileUploader should already dispatch on a background thread, its weird

ouchadam: yeah it's very strange, I checked the okhttp thread pool/interceptors as well, couldn't find any cases where the main thread was being used, unless on some devices the okhttp thread uses the main thread/looper for some reason :thinking:

ganfra (approved):
Thanks!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/profile/DefaultProfileService.kt
bmarty: This is also something which has been fixed by some fork on their side FWIW.

bmarty (approved):

PR #4790, opened 2021-12-21
Timeline recycling crash

ouchadam:
Tentatively fixes #4789 Timeline crashes when quickly scrolling or restoring state.

I was unable to reproduce this crash locally, the stack trace looks to be the same as when a `TextView` is modified after setting a `TextFuture`

```
Thread: main, Exception: java.lang.IllegalArgumentException: PrecomputedText's Parameters don't match the parameters of this TextView.
    at android.widget.TextView.setText(TextView.java:6206)
    at android.widget.TextView.setText(TextView.java:6124)
    at android.widget.TextView.setText(TextView.java:6076)
    at androidx.core.widget.TextViewCompat.setPrecomputedText(TextViewCompat.java:907)
    at androidx.appcompat.widget.AppCompatTextView.consumeTextFutureAndSetBlocking(AppCompatTextView.java:542)
    at androidx.appcompat.widget.AppCompatTextView.getText(AppCompatTextView.java:551)
    at androidx.emoji2.viewsintegration.EmojiInputFilter.filter(EmojiInputFilter.java:64)
    at android.widget.TextView.setText(TextView.java:6160)
    at android.widget.TextView.setText(TextView.java:6124)
    at android.widget.TextView.setText(TextView.java:6076)
```

My theory is that `RecyclerView` is recycling `TextViews` with non consumed futures which causes them to throw when the `TextView` options are rebound. This PR aims to solve the issue by resetting any present futures before calling `setText` which will allow `AppCompatTextView.consumeTextFutureAndSetBlocking` to be skipped if the view is rebound with `setText` directly (such is the case when emojis are present)

PR #4795, opened 2021-12-22
Emojis in code blocks crash

### LINKED ISSUE #4794: Crash when codeblocks contain emojis
### Steps to reproduce

1. Enable markdown
2. Open a room or send a message to a room with a code block containing :face_exhaling: 

#### What did you expect?
No crash

#### What happened instead?
We're seeing a variant of https://github.com/vector-im/element-android/issues/4691 but in codeblocks

```
Thread: main, Exception: com.airbnb.epoxy.ImmutableModelException: The model was changed during the bind call. Position: 2 Model: MessageBlockCodeItem_{message=VectorCharSequence(value=😮‍💨

Epoxy attribute fields on a model cannot be changed once the model is added to a controller. Check that these fields are not updated, or that the assigned objects are not mutated, outside of the buildModels method. The only exception is if the change is made inside an Interceptor callback. Consider using an interceptor if you need to change a model after it is added to the controller and before it is set on the adapter. If the model is already set on the adapter then you must call `requestModelBuild` instead to recreate all models.
     at com.airbnb.epoxy.EpoxyModel.validateStateHasNotChangedSinceAdded(EpoxyModel.java:466)
     at im.vector.app.features.home.room.detail.timeline.item.MessageBlockCodeItem_.handlePostBind(MessageBlockCodeItem_.java:56)
     at im.vector.app.features.home.room.detail.timeline.item.MessageBlockCodeItem_.handlePostBind(MessageBlockCodeItem_.java:26)
     at com.airbnb.epoxy.EpoxyViewHolder.bind(EpoxyViewHolder.java:77)
     at com.airbnb.epoxy.BaseEpoxyAdapter.onBindViewHolder(BaseEpoxyAdapter.java:118)
     at com.airbnb.epoxy.BaseEpoxyAdapter.onBindViewHolder(BaseEpoxyAdapter.java:20)
```

### Your phone model

_No response_

### Operating system version

android 10 (MIUI 12.0.3.0)

1.3.12

### Homeserver

matrix.org

### Will you send logs?

No

ouchadam:
Depends on #4790, Fixes #4794 Room crash when viewing/submitting emojis in code blocks (requires markdown to be enabled)

- Extracts a dedicated model `VectorCharSequence` to contain the binding options and remove duplication when applying the `setText` logic 

| BEFORE | AFTER | 
| --- | --- |
|Screenshot_20211222_123031|Screenshot_20211222_101004

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageBlockCodeItem.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageBlockCodeItem.kt
@@ -23,24 +23,22 @@ import com.airbnb.epoxy.EpoxyModelClass
 import im.vector.app.R
 import im.vector.app.core.epoxy.onClick
 import im.vector.app.core.extensions.setTextOrHide
-import im.vector.app.core.extensions.setTextWithEmojiSupport
+import im.vector.app.core.extensions.setVectorText
+import im.vector.app.features.html.VectorCharSequence
 import me.saket.bettermovementmethod.BetterLinkMovementMethod
 
 @EpoxyModelClass(layout = R.layout.item_timeline_event_base)
 abstract class MessageBlockCodeItem : AbsMessageItem<MessageBlockCodeItem.Holder>() {
 
     @EpoxyAttribute
-    var message: CharSequence? = null
-
-    @EpoxyAttribute
-    var bindingOptions: BindingOptions? = null
+    var message: VectorCharSequence? = null
 
     @EpoxyAttribute
     var editedSpan: CharSequence? = null
 
     override fun bind(holder: Holder) {
         super.bind(holder)
-        holder.messageView.setTextWithEmojiSupport(message, bindingOptions)
+        holder.messageView.setVectorText(message)
         renderSendState(holder.messageView, holder.messageView)
         holder.messageView.onClick(attributes.itemClickListener)
         holder.messageView.setOnLongClickListener(attributes.itemLongClickListener)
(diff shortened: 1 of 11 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageBlockCodeItem.kt
```diff
@@ -17,25 +17,28 @@
...
-    var message: CharSequence? = null
+    var message: VectorCharSequence? = null
 
     @EpoxyAttribute
     var editedSpan: CharSequence? = null
 
     override fun bind(holder: Holder) {
         super.bind(holder)
-        holder.messageView.text = message
+        holder.messageView.setVectorText(message)
```
ouchadam, line 41: this is the fix for the crash in the code blocks (applying the same emoji fix we're using for the other message types)

vector/src/main/java/im/vector/app/features/html/VectorCharSequenceFactory.kt
```diff
@@ -24,20 +24,22 @@ import im.vector.app.EmojiSpanify
 import im.vector.app.features.home.room.detail.timeline.item.BindingOptions
 import javax.inject.Inject
 
-class SpanUtils @Inject constructor(
+data class VectorCharSequence(val value: CharSequence, val bindingOptions: BindingOptions)
```
ouchadam, line 27: we could potentially override the hashcode/equals here to avoid epoxy throwing when spannables mutate but I chose to leave out the change from this PR

bmarty:
Probably have to be updated now that #4837 has been merged

ouchadam:
going to close this PR, we're only using text futures in the message item and #4837 fixes the crash

PR #4799, opened 2021-12-22
Outgoing call ringtone using wrong audio device

### LINKED ISSUE #4781: The ringtone plays through the "speaker mode" system instead of the "default calling" system when calling someone.
### Steps to reproduce

1. Call an individual
2. Put the phone up to your head, assuming it will behave like normal.
3. Experience hearing loss from the phone blaring at an unadjustable volume on speaker mode.
4. The call itself starts and it is no longer speaker mode, without any user interaction.

#### What did you expect?
I expected that the phone would behave like normal and not play at maximum "speaker mode" volume without me specifying for that to happen
#### What happened instead?
The app likely caused me hearing damage, this has happened multiple times because it is the practical behavior of every calling app I've ever used. Kind of annoyed because both the notification noise on the desktop app & this have been physically damaging to me & others and they don't seem to be getting worked on. (see https://github.com/vector-im/element-web/issues/5891#issuecomment-965898674)

### Your phone model

Pixel 4a 5G

### Operating system version

Android 11

Element version 1.3.9 Olm version 3.2.4

### Homeserver

yiff.social

### Will you send logs?

No

ouchadam:
Tentatively fixes #4781 wrong audio device being used for outgoing call ringtone

- In #4781 it's reported that the correct audio device is used when the call starts, the attempted fix is to use the same audio device configuration for the outgoing ringtone as we do for the call itself. 

I haven't been able to reproduce this locally, I'm assuming it's something related to LineageOS

### DIFF
--- a/vector/src/main/java/im/vector/app/core/services/CallRingPlayer.kt
+++ b/vector/src/main/java/im/vector/app/core/services/CallRingPlayer.kt
@@ -28,6 +28,8 @@ import android.os.VibrationEffect
 import android.os.Vibrator
 import androidx.core.content.getSystemService
 import im.vector.app.R
+import im.vector.app.features.call.audio.CallAudioManager
+import im.vector.app.features.call.webrtc.WebRtcCallManager
 import im.vector.app.features.notifications.NotificationUtils
 import org.matrix.android.sdk.api.extensions.orFalse
 import timber.log.Timber
@@ -94,7 +96,8 @@ class CallRingPlayerIncoming(
 }
 
 class CallRingPlayerOutgoing(
-        context: Context
+        context: Context,
+        private val callManager: WebRtcCallManager
 ) {
 
     private val applicationContext = context.applicationContext
@@ -102,7 +105,7 @@ class CallRingPlayerOutgoing(
     private var player: MediaPlayer? = null
 
     fun start() {
-        applicationContext.getSystemService<AudioManager>()?.mode = AudioManager.MODE_IN_COMMUNICATION
+        callManager.setAudioModeToCallType()
         player?.release()
         player = createPlayer()
         if (player != null) {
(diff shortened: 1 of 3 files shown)


vector/src/main/java/im/vector/app/core/services/CallRingPlayer.kt
```diff
@@ -94,15 +96,16 @@ class CallRingPlayerIncoming(
...
+        private val callManager: WebRtcCallManager
 ) {
 
     private val applicationContext = context.applicationContext
 
     private var player: MediaPlayer? = null
 
     fun start() {
-        applicationContext.getSystemService<AudioManager>()?.mode = AudioManager.MODE_IN_COMMUNICATION
+        callManager.setAudioModeToCallType()
```
ouchadam, line 108: another option could be to use the `RingtoneManager` but given that we don't already use it I'm assuming we'd prefer not to!

vector/src/main/java/im/vector/app/core/services/CallRingPlayer.kt
```diff
@@ -120,6 +123,12 @@ class CallRingPlayerOutgoing(
         }
     }
 
+    private fun WebRtcCallManager.setAudioModeToCallType() {
+        currentCall.get()?.let {
+            audioManager.setMode(if (it.mxCall.isVideoCall) CallAudioManager.Mode.VIDEO_CALL else CallAudioManager.Mode.AUDIO_CALL)
+        }
+    }
```
bmarty, line 130: Do we need a fallback if `currentCall.get()` is null?

Maybe better to do:
```kotlin
audioManager.setMode(if (currentCall.get()?.mxCall?.isVideoCall.orFalse()) CallAudioManager.Mode.VIDEO_CALL else CallAudioManager.Mode.AUDIO_CALL)
```

bmarty (commented):
LGTM, just a small remark

vector/src/main/java/im/vector/app/core/services/CallRingPlayer.kt
ouchadam: using a fallback sounds good :+1: I do wonder if it's possible to have a `null` call at this point in the flow (and if so should the entire flow throw?)

f89a2b4e56a3474c94610c35bbbf9e6092e47a40

bmarty: I do not know... I would say out of scope of this PR :/

bmarty (approved):
Thanks for the update

PR #4810, opened 2021-12-24
Fixing duplicated/empty notifications when joining/muting/leaving a VOIP call

### LINKED ISSUE #4804: In a 1:1 desktop voice or video call, muting the audio triggers an Android notification.
### Steps to reproduce

1. Have both the desktop and Android Element clients signed in to the same account. 
2. Start a 1:1 (voice or video) call on the desktop client.
3. Mute the audio in the desktop client.
4. Check the Android phone for notifications. 

When I asked the #element-android:matrix.org community about this, the response was, "it's always been like that," so it seems like I'm not the only affected user.

#### What did you expect?
I expect no notifications for this event.
Repeatedly muting to silence background/keyboard noise is common in an audio/video desktop call.

#### What happened instead?
Both parties in the call get an Android notification every time someone mutes a 1:1 desktop call.

### Your phone model

Pixel 4a 5G

### Operating system version

Andriod 12

1.3.12

### Homeserver

matrix.org

### Will you send logs?

No

ouchadam:
Fixes #4804 mute events displaying blank notifications

- It turns out the app was assuming all encrypted events were messages, which meant we were showing invalid notifications
- The fix is to only create `NotifiableMessageEvent` when the decrypted or clear content type is a `message`
- This also ends up fixing the debug and duplicated VOIP notifications appearing as w were using the same filtering logic for the timeline as the notifications

| BEFORE | AFTER |
| --- | --- |
|before-muting|Screenshot_20211224_111636

### DIFF
--- a/changelog.d/4804.bugfix
+++ b/changelog.d/4804.bugfix
@@ -1,0 +1,1 @@
+Fixing encrypted non message events showing up as notification messages (eg when a participant joins, mutes or leaves a voice call)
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM!

PR #4826, opened 2021-12-31
Nick color

bmarty:
Replacement for #2614.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/MatrixItemColorProvider.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/MatrixItemColorProvider.kt
@@ -22,6 +22,7 @@ import androidx.annotation.VisibleForTesting
 import im.vector.app.R
 import im.vector.app.core.resources.ColorProvider
 import org.matrix.android.sdk.api.util.MatrixItem
+import timber.log.Timber
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.math.abs
@@ -45,35 +46,38 @@ class MatrixItemColorProvider @Inject constructor(
     }
 
     fun setOverrideColors(overrideColors: Map<String, String>?) {
-        overrideColors?.forEach() {
+        cache.clear()
+        overrideColors?.forEach {
             setOverrideColor(it.key, it.value)
         }
     }
 
-    fun setOverrideColor(id: String, colorSpec: String?) : Boolean {
+    fun setOverrideColor(id: String, colorSpec: String?): Boolean {
         val color = parseUserColorSpec(colorSpec)
-        if (color == null) {
+        return if (color == null) {
             cache.remove(id)
-            return false
+            false
         } else {
-            cache.put(id, color)
-            return true
+            cache[id] = color
+            true
         }
     }
 
     @ColorInt
     private fun parseUserColorSpec(colorText: String?): Int? {
-        if (colorText.isNullOrBlank()) {
-            return null
-        }
-        try {
-            if (colorText.first() == '#') {
-                return (colorText.substring(1).toLong(radix = 16) or 0xff000000L).toInt()
-            } else {
-                return colorProvider.getColor(getUserColorByIndex(colorText.toInt()))
+        return if (colorText.isNullOrBlank()) {
+            null
+        } else {
+            try {
+                if (colorText.first() == '#') {
+                    (colorText.substring(1).toLong(radix = 16) or 0xff000000L).toInt()
+                } else {
+                    colorProvider.getColor(getUserColorByIndex(colorText.toInt()))
+                }
+            } catch (e: Throwable) {
+                Timber.e(e, "Unable to parse color $colorText")
+                null
             }
-        } catch (e: Throwable) {
-            return null
         }
     }
 
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/MatrixItemColorProvider.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/MatrixItemColorProvider.kt
@@ -16,6 +16,7 @@
 
 package im.vector.app.features.home.room.detail.timeline.helper
 
+import android.graphics.Color
 import androidx.annotation.ColorInt
 import androidx.annotation.ColorRes
 import androidx.annotation.VisibleForTesting
@@ -69,10 +70,10 @@ class MatrixItemColorProvider @Inject constructor(
             null
         } else {
             try {
-                if (colorText.first() == '#') {
-                    (colorText.substring(1).toLong(radix = 16) or 0xff000000L).toInt()
-                } else {
+                if (colorText.length == 1) {
                     colorProvider.getColor(getUserColorByIndex(colorText.toInt()))
+                } else {
+                    Color.parseColor(colorText)
                 }
             } catch (e: Throwable) {
                 Timber.e(e, "Unable to parse color $colorText")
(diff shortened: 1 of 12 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/MatrixItemColorProvider.kt
```diff
@@ -44,6 +46,42 @@ class MatrixItemColorProvider @Inject constructor(
...
+        }
+    }
+
+    @ColorInt
+    private fun parseUserColorSpec(colorText: String?): Int? {
+        return if (colorText.isNullOrBlank()) {
+            null
+        } else {
+            try {
+                if (colorText.length == 1) {
```
mitchnull, line 73: This will break if/when the palette is extended to 16 or more colors.  Maybe it would be better to try just try to parse it as `int`, and if that works then it's a color index (from the palette) otherwise parse with `Color.parseColor`.  Another option would be to check if the text starts with a digit to select the palette version

bmarty: I agree, but the plan is to change this dialog to do something more user friendly, so it's acceptable like that to me.

bmarty:
@ouchadam can you have a quick look on this PR please? Thanks.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/MatrixItemColorProvider.kt
mitchnull: OK, cool, thanks.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/accountdata/UserAccountDataTypes.kt
```diff
@@ -27,4 +27,5 @@ object UserAccountDataTypes {
     const val TYPE_ALLOWED_WIDGETS = "im.vector.setting.allowed_widgets"
     const val TYPE_IDENTITY_SERVER = "m.identity_server"
     const val TYPE_ACCEPTED_TERMS = "m.accepted_terms"
+    const val TYPE_OVERRIDE_COLORS = "im.vector.setting.override_colors"
```
ouchadam, line 30: does this type need to be added to the matrix spec somewhere?

vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileFragment.kt
```diff
@@ -321,6 +325,26 @@ class RoomMemberProfileFragment @Inject constructor(
...
     }
 
+    override fun onOverrideColorClicked(): Unit = withState(viewModel) { state ->
+        val inflater = requireActivity().layoutInflater
+        val layout = inflater.inflate(R.layout.dialog_base_edit_text, null)
+        val views = DialogBaseEditTextBinding.bind(layout)
+        views.editText.setText(state.userColorOverride)
+        views.editText.hint = "#000000"
+
+        MaterialAlertDialogBuilder(requireContext())
```
ouchadam, line 335: allowing the user to provide a hex string is quite verbose, perhaps another iteration of this could use a colour wheel UI

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/MatrixItemColorProvider.kt
ouchadam: being able to set the colour by index is quite cryptic, I wouldn't have expected an input with a hex string hint to also allow a 0-9 index 

something a UI iteration could help to improve

vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileViewModel.kt
```diff
@@ -121,6 +128,24 @@ class RoomMemberProfileViewModel @AssistedInject constructor(@Assisted private v
                 }
     }
 
+    private fun observeAccountData() {
+        session.flow()
+                .liveUserAccountData(UserAccountDataTypes.TYPE_OVERRIDE_COLORS)
+                .unwrap()
+                .map { it.content.toModel<Map<String, String>>() }
+                .map { userColorAccountDataContent ->
```
ouchadam, line 136: nit: could merge these maps 

```kotlin
.map { 
  val userColorAccountDataContent = it.content.toModel<Map<String, String>>()
  userColorAccountDataContent?.get(initialState.userId)
}
```

vector/src/main/res/values/strings.xml
```diff
@@ -2812,6 +2812,8 @@
     <string name="direct_room_profile_section_more_leave">Leave</string>
     <string name="room_profile_leaving_room">"Leaving the room…"</string>
 
+    <string name="room_member_override_nick_color">Override nick color</string>
```
ouchadam, line 2815: is the `nick` terminology used in other places, or should this be `display name`?

vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileFragment.kt
mitchnull: I didn't include a full-fledged UI for this in my original PR because I didn't have the necessary expertise to do it quickly, and it wasn't clear that the PR / idea would be accepted at all (this is mostly a POC and a working solution for us).  I do agree that a proper color selector should be created for this (that can select a color from the current palette or select and arbitrary RGB color)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/accountdata/UserAccountDataTypes.kt
mitchnull: I have another PR that implements the same functionality in element-desktop (https://github.com/matrix-org/matrix-react-sdk/pull/5626).  As far as I understand keys under `im.vector.setting` are free-to-use by clients, so I don't think it's absolutely necessary to include it in the matrix spec (apart from maybe documenting its existence so that other clients can pick this up if they wish)

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/MatrixItemColorProvider.kt
bmarty: Yes, this is really a hidden feature. Sort of easter egg.
I see in my dream a nice dialog with all the 8 predefined colors, a color picker and a field to manually enter a html color. Even a reset button, instead of validating a empty dialog to reset the color.
Also note that now using the word "black" or any other color that `Color.parse()` understand will also work, not sure for the other non Android-Element Matrix clients compatibility though.

vector/src/main/res/values/strings.xml
bmarty: Nick is used in the slash command `/myroomnick` but nowhere else AFAIK. The color will be used to tint the display name but also for the background of the default avatar. So not only the display name.
Maybe using "Override user color" is fine. Please vote with 👍 and 👎 !

vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileViewModel.kt
bmarty: Due to rework, thanks, will update

bmarty: 51c9c2f032a7709d18b56bbfb35def25e3d08a24

vector/src/main/res/values/strings.xml
ouchadam: no strong opinions from me on this :smile: maybe one for product cc @daniellekirkwood

ouchadam (approved):
LGTM :100:  small comment about the field UI label but we can always update it later

PR #4837, opened 2022-01-03
Safe epoxy char sequence

bmarty:
Using CharSequence in Epoxy item is not safe, since CharSequence are no immutable.
This PR introduce a wrapper for CharSequences used in Epoxy. If CharSequence is not mandatory, use immutable String.

- will fix issue handled in #4795 which contain other interesting stuff.
- tested with #2133 (which I am working on) using the latest markwon version with success \o/


vector/src/main/java/im/vector/app/core/epoxy/charsequence/EpoxyCharSequence.kt
```diff
@@ -0,0 +1,27 @@
...
+package im.vector.app.core.epoxy.charsequence
+
+/**
+ * Wrapper for a CharSequence, which support mutation of the CharSequence, which can happen during rendering
+ */
+class EpoxyCharSequence(val charSequence: CharSequence) {
+    private val hash = charSequence.toString().hashCode()
+
+    override fun hashCode() = hash
+    override fun equals(other: Any?) = other is EpoxyCharSequence && other.hash == hash
```
bmarty, line 26: We do not compare the String content, since the risk of collision is low enough, and Epoxy is mainly using hashCode() to compare items

vector/src/main/java/im/vector/app/core/epoxy/charsequence/Extensions.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 New Vector Ltd
+ * Copyright (c) 2022 New Vector Ltd
```
bmarty, line 2: Happy new year!

vector/src/main/java/im/vector/app/core/epoxy/profiles/notifications/RadioButtonItem.kt
```diff
@@ -33,7 +33,7 @@ import im.vector.app.core.extensions.setAttributeTintedImageResource
 abstract class RadioButtonItem : VectorEpoxyModel<RadioButtonItem.Holder>() {
 
     @EpoxyAttribute
-    var title: CharSequence? = null
+    var title: String? = null
```
bmarty, line 36: Whenever a CharSequence is not required, it simpler to use an immutable String

vector/src/main/java/im/vector/app/core/epoxy/charsequence/EpoxyCharSequence.kt
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.core.epoxy.charsequence
+
+/**
+ * Wrapper for a CharSequence, which support mutation of the CharSequence, which can happen during rendering
+ */
+class EpoxyCharSequence(val charSequence: CharSequence) {
```
ganfra, line 22: Maybe let the EpoxyCharSequence inherits CharSequence like this, to avoid accessing charSequence value everywhere 

```class EpoxyCharSequence(val charSequence: CharSequence) : CharSequence by charSequence```

**Review by ganfra — APPROVED** [review]
Just one comment, otherwise LGTM

File: vector/src/main/java/im/vector/app/core/epoxy/charsequence/EpoxyCharSequence.kt (hunk 4 above)
Comment (bmarty) [reply vector/src/main/java/im/vector/app/core/epoxy/charsequence/EpoxyCharSequence.kt:22]: Good idea, let me try it.

Comment (bmarty) [reply vector/src/main/java/im/vector/app/core/epoxy/charsequence/EpoxyCharSequence.kt:22]: It does not work, the TextView displays the Object and not the CharSequence... Weird

image

Comment (bmarty) [reply vector/src/main/java/im/vector/app/core/epoxy/charsequence/EpoxyCharSequence.kt:22]: According to @ganfra investigation, it can work using
```kotlin
class EpoxyCharSequence(val charSequence: CharSequence) : Spannable by SpannableString(charSequence) {...}
```
but I prefer not to create a new SpannableString for each charSequence.
Using `.charSequence` whenever it is necesssary is probably better, because it forces the developer to do it (no "hidden magic")

PR #4838, opened 2022-01-03
Support LaTex Math

bmarty:
Successor of #2133.

**EDIT**: Below is done

Need to be rebased after #4837 is merged, and this line will have to be updated from:
```kotlin
markwonPlugins?.forEach { plugin -> plugin.beforeSetText(holder.messageView, message as Spanned) }
```

to 
```kotlin
markwonPlugins?.forEach { plugin -> plugin.beforeSetText(holder.messageView, message?.charSequence as Spanned) }
```


ariskotsomitopoulos (approved):
It seems fine.

Minor, I would maybe create global variables regarding:
```
object Symbols {
        const val DOUBLE_DOLLAR = "$$"
        const val DOLLAR = "$"
      ....
}
```
```
object Tags {
        const val DIV = "div"
       ....
}
```

bmarty:
Thanks for the review. I will merge and fix ktlint on develop

PR #4841, opened 2022-01-03
Small cleanup

bmarty:
Remove unused methods.

Following change from PR #4745

### DIFF
--- a/vector/src/main/java/im/vector/app/core/utils/SystemUtils.kt
+++ b/vector/src/main/java/im/vector/app/core/utils/SystemUtils.kt
@@ -124,9 +124,9 @@ fun startNotificationChannelSettingsIntent(fragment: Fragment, channelID: String
 }
 
 fun startAddGoogleAccountIntent(context: Context, activityResultLauncher: ActivityResultLauncher<Intent>) {
+    val intent = Intent(Settings.ACTION_ADD_ACCOUNT)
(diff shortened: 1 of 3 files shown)


vector/src/main/java/im/vector/app/core/utils/SystemUtils.kt
```diff
@@ -135,9 +135,9 @@ fun startAddGoogleAccountIntent(context: Context, activityResultLauncher: Activi
 
 @RequiresApi(Build.VERSION_CODES.O)
 fun startInstallFromSourceIntent(context: Context, activityResultLauncher: ActivityResultLauncher<Intent>) {
+    val intent = Intent(Settings.ACTION_MANAGE_UNKNOWN_APP_SOURCES)
+            .setData(Uri.parse(String.format("package:%s", context.packageName)))
```
ouchadam, line 139: super nit: this was already here but it could be nice to avoid `String.format` in favour of a more kotlin idiomatic `"package:${context.packageName}" `

For this usecase it's not very relevant but`String.format` is also quite heavy relative to a `StringBuilder` https://cowtowncoder.medium.com/measuring-performance-of-java-string-format-or-lack-thereof-2e1c6a13362c

ouchadam (approved):

vector/src/main/java/im/vector/app/core/utils/SystemUtils.kt
bmarty: https://github.com/vector-im/element-android/issues/4859

PR #4842, opened 2022-01-03
Trying to fix integration tests

ariskotsomitopoulos:
### Fixing Integration test

The main problem is that GitHub actions is that it has limited resources and is not meant for that kind of jobs. In order to properly do a robust 100% solution we should welcome another CI/CD tool like Jenkins and trigger the build from here

For further improvements we should change the actual implementation of the integration tests along with the shared resources

- Synapse server can now start, it was broken
- Fix/Improve running them on **GitHub**
- Mark the not working tests with **@ignore** so that they will not run neither local or on github actions
- Advanced parsing the result to add a user friendly comment on the PR with the results
- Increase adb ram & also tried caching the adb

_On mac slave it works much better and consistent_

As you can see here the tests can run and the results are published like below:
Screenshot 2022-01-07 at 17 05 02


ariskotsomitopoulos:
### Matrix SDK
## Integration Tests Results
- `[org.matrix.android.sdk.session]`<br>passed="21" failures="0" errors="0" skipped="2"
- `[org.matrix.android.sdk.session]`<br>passed="21" failures="0" errors="0" skipped="2"
- - `[org.matrix.android.sdk.session]`<br>passed="21" failures="0" errors="0" skipped="2"

bmarty:
We will have to update the GitHub setting to make the actions required:
- [ ] `Integration Tests / Matrix SDK - Build Android Tests (pull_request)` - now passing!
- [ ] `Integration Tests / Matrix SDK - Running Integration Tests (28) (pull_request)` - now passing!
- [ ] `Integration Tests / App - Build Android Tests` (pull_request) (not required anymore, maybe because it has been renamed? Anyway it was a temporary GitHub action, which can be removed now)

Also the action `Compile Android tests` seems stuck.

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,208 @@
...
+# Enrich gradle.properties for CI/CD
+env:
+  CI_GRADLE_ARG_PROPERTIES: >
+    -Porg.gradle.jvmargs=-Xmx2g
+    -Porg.gradle.parallel=false
+
+jobs:
+  # Build Android Tests [Matrix SDK]
+  build-android-test-matrix-sdk:
+    name: Matrix SDK - Build Android Tests
```
bmarty, line 17: Do we want to keep this action since it is now done when we run the integration test?
It has been added (see the comment [here](https://github.com/vector-im/element-android/pull/4842/files#diff-82454b2fc887e9985b9348b8f3bca03d2f839645a4c3ba1f3f850014d7da5c8bL16)) to ensure that the tests are compiling by making this action "Required" on every PRs.
But now I think we can put the test run "Required".

Thinking more about it, since this action is more reliable than the running of integration test, maybe we can keep it.

dependencies.gradle
```diff
@@ -29,6 +29,7 @@ def vanniktechEmoji = "0.8.0"
 def mockk = "1.12.1"
 def espresso = "3.4.0"
 def androidxTest = "1.4.0"
+def androidxOrchestrator = "1.4.1"
```
bmarty, line 32: Why not just update the value of `androidxTest` above to `1.4.1`?

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/RetryTestRule.kt
```diff
@@ -0,0 +1,58 @@
...
+                var caughtThrowable: Throwable? = null
+
+                // implement retry logic here
+                for (i in 0 until retryCount) {
+                    try {
+                        base.evaluate()
+                        return
+                    } catch (t: Throwable) {
+                        caughtThrowable = t
+                        Log.e(TAG, description.displayName + ": run " + (i + 1) + " failed")
```
bmarty, line 49: Maybe also log the exception `t`?

bmarty: Also can we use `Timber` here? (Maybe not)

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/TestConstants.kt
```diff
@@ -23,7 +23,7 @@ object TestConstants {
     const val TESTS_HOME_SERVER_URL = "http://10.0.2.2:8080"
 
     // Time out to use when waiting for server response. 20s
-    private const val AWAIT_TIME_OUT_MILLIS = 20_000
+    private const val AWAIT_TIME_OUT_MILLIS = 30_000
```
bmarty, line 26: Please update the value in the comment (or maybe just remove "20s").

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/gossiping/WithHeldTests.kt
```diff
@@ -39,12 +41,14 @@ import org.matrix.android.sdk.internal.crypto.model.event.WithHeldCode
 
 @RunWith(AndroidJUnit4::class)
 @FixMethodOrder(MethodSorters.JVM)
+@LargeTest
```
bmarty, line 44: Where are effectively use this annotation `LargeTest` (and the other ones)? It's maybe not in this PR?

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/verification/SASTest.kt
```diff
@@ -528,7 +547,6 @@ class SASTest : InstrumentedTest {
         val aliceDeviceInfoFromBobPOV: CryptoDeviceInfo? = bobSession.cryptoService().getDeviceInfo(aliceSession.myUserId, aliceSession.cryptoService().getMyDevice().deviceId)
 
         // latch wait a bit again
-        Thread.sleep(1000)
```
bmarty, line 531: Please also remove the comment above

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/session/room/timeline/TimelineWithManyMembersTest.kt
```diff
@@ -31,8 +33,13 @@ import org.matrix.android.sdk.common.CommonTestHelper
 import org.matrix.android.sdk.common.CryptoTestHelper
 import java.util.concurrent.CountDownLatch
 
+/** !! Not working with the new timeline
+ *  Disabling it until the fix is made
```
bmarty, line 37: CC @ganfra

bmarty (commented):
Nice work, thanks a lot for working on that.
Some comment/questions.

.github/workflows/integration_tests.yml
ariskotsomitopoulos: Exactly thats why I kept it. While its not 100% reliable I think we should keep it

dependencies.gradle
ariskotsomitopoulos: As you can see below only `androidxOrchestrator` is yet it `1.4.1`
Screenshot 2022-01-07 at 17 53 50

bmarty: You are right, I read a bit fast the title at https://developer.android.com/jetpack/androidx/releases/test#core_141_2

Also confirmed that this should not be a problem to use various version by the sample at https://developer.android.com/jetpack/androidx/releases/test#declaring_dependencies

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/gossiping/WithHeldTests.kt
ariskotsomitopoulos: Yes, we are not currently using those annotations. The plan is to start using this annotation to our tests when everything is annotated. I have annotated some of the tests but not all of them

bmarty (approved):
Thanks for the update

PR #4846, opened 2022-01-04
Fix unread messages indicator on room list if no messages

### LINKED ISSUE #4749: Rooms marked as semi unread when being invited to a room (and more cases)
### Steps to reproduce

1.  someone invites me to a room.
2. I accept the room invite on web
3. I see this weird semi unread mark on Android:
image

#### What did you expect?
Not to have that.

Here is a video explaining this and showing some other weird behavior:
https://aarenet-my.sharepoint.com/:v:/p/chagai_friedlander/ETzk6F5vYYVPlcjyT10219QB9KsJzRDO-dW148wkCGa0Qw?e=I3OGKq

### Your phone model

Xiaomi Mi 11 lite

### Operating system version

IMG_20211217_113352

1.3.9

### Homeserver

Matrix.org

### Will you send logs?

No

langleyd:
If there are no messages in a room lastPreviewableEvent will be null. We should not indicate there is an unread if we don't have a lastPreviewableEvent.

Fixes #4749

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryUpdater.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryUpdater.kt
@@ -136,7 +136,7 @@ internal class RoomSummaryUpdater @Inject constructor(
 
         roomSummaryEntity.hasUnreadMessages = roomSummaryEntity.notificationCount > 0 ||
                 // avoid this call if we are sure there are unread events
-                !isEventRead(realm.configuration, userId, roomId, latestPreviewableEvent?.eventId)
+                latestPreviewableEvent?.let { !isEventRead(realm.configuration, userId, roomId, it.eventId) } ?: false
 
         roomSummaryEntity.setDisplayName(roomDisplayNameResolver.resolve(realm, roomId))
         roomSummaryEntity.avatarUrl = roomAvatarResolver.resolve(realm, roomId)
--- a/changelog.d/4749.bugfix
+++ b/changelog.d/4749.bugfix
@@ -1,0 +1,1 @@
+Fix for broken unread message indicator on the room list when there are no messages in the room.


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryUpdater.kt
```diff
@@ -136,7 +136,7 @@ internal class RoomSummaryUpdater @Inject constructor(
 
         roomSummaryEntity.hasUnreadMessages = roomSummaryEntity.notificationCount > 0 ||
                 // avoid this call if we are sure there are unread events
-                !isEventRead(realm.configuration, userId, roomId, latestPreviewableEvent?.eventId)
+                latestPreviewableEvent?.let { !isEventRead(realm.configuration, userId, roomId, it.eventId) } ?: false
```
bmarty, line 139: This is fine, but I am wondering if you should rather return `true` instead of `false` at this place: https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/ReadQueries.kt#L31

bmarty (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryUpdater.kt
ganfra: Nope, by default we should consider event as unread, it's safer

ganfra (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryUpdater.kt
bmarty: as discussed with @ganfra let's keep it like that.

PR #4823, opened 2021-12-30
Attachment Picker UI Improvements

onurays:
[Figma](https://www.figma.com/file/X4XTH9iS2KGJ2wFKDqkyed/Compound?node-id=2039%3A26421)

@gaelledel: I had to reuse existing audio and contact icons. 

EDIT Benoit:

*Previous record removed (commented)*

Latest version:

attachment

PR #4852, opened 2022-01-04
Add tooltip to explain what are the actions behind the icons

bmarty:
picker_tooltip

### DIFF
--- a/changelog.d/3444.bugfix
+++ b/changelog.d/3444.bugfix
@@ -1,1 +1,0 @@
-Attachment picker UI improvements
--- a/changelog.d/3444.feature
+++ b/changelog.d/3444.feature
@@ -1,0 +1,1 @@
+New attachment picker UI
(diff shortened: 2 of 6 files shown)


gaelledel (commented):
Approved given we're changing the copy that appears to the following:

- image: Send images and videos
- sticker: Send sticker
- attachment: Upload file
- Polls: Create poll
- Camera: Open camera
- Contacts: Open contacts

gaelledel (approved):

bmarty:
@gaelledel what is the wording for send audio?
image

Maybe just remove this item has it does not have a big added value compare to "files". WDYT?

gaelledel:
Yes, I actually asked in here earlier https://github.com/vector-im/element-android/pull/4823 
"Can we remove the audio button (headphones) and simply get users to fetch audio via the attachment button? The bar is getting crowded with new functionalities and there is no immediate use case to resurface audio as a specific key action."

So if we could remove this audio item, that'd be grant!

bmarty:
After update:

picker_tooltip

AnisTaluqdar:
When tapping on the plus button there has an animation issue. @bmarty can you check it?

bmarty:
@AnisTaluqdar which issue? Can you explain?

AnisTaluqdar:
Outside the circle. Seems a box is rotating. I think it is not intentional.

bmarty:
@AnisTaluqdar I see, thanks. It's due to the fact that we are using a PopupWindow to display this View. I am not sure how we can quickly avoid this side effect. Happy to hear some ideas.
Merging this PR for now.

PR #4863, opened 2022-01-05
Prefer usage of `any` or `none` rather `indexOfFirst` wherever it is possible

bmarty:
I read the change twice but please review carefully, I did not do any test.


ouchadam (approved):
:100:  for the change! any/all/none are awesome

I've reread this a few times and everything looks correct, we could create an extension to also handle the nullable cases but it's probably overkill

PR #4864, opened 2022-01-05
Enhance Linter Github Actions Reporting

ariskotsomitopoulos:
Improve Github Actions Linter reports:

- Add **ktlint** Artifact report
Screenshot 2022-01-05 at 19 37 37

- **kotlint reports will be also added as comment in the PR**
  - It will print exactly the lines that have problem
  - It will trim the output so the developer will be able to immediately find the class with the error
  - Support for all the different modules along with tests 
  - The same comment will be reused
  Screenshot 2022-01-05 at 22 41 08
  Screenshot 2022-01-05 at 22 56 08
- It is also able to identify the success and print the result appropriately
  Screenshot 2022-01-05 at 22 58 50

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -14,6 +14,7 @@ jobs:
       - name: Run code quality check suite
         run: ./tools/check/check_code_quality.sh
 
+# ktlint for all the modules
   ktlint:
     name: Kotlin Linter
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/quality.yml
```diff
@@ -23,12 +24,56 @@ jobs:
...
+        if: always()
+        id: get-comment-body
+        run: |
+          results="$(cat */*/build/reports/ktlint/ktlint*/ktlint*.txt */build/reports/ktlint/ktlint*/ktlint*.txt | sed -r "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g")"
+          if [ -z  "$results" ]; then
+            body="👍 ✅ 👍"
+          else
+            body="\`👎 ❌ 👎 ${results}\`"
+          fi
+          body="${body//'%'/'%25'}"
```
bmarty, line 43: I would apply all the treatment you are doing on the `else` block above and on the `results` var, before affecting the value prefixed with emojis to `body`.

.github/workflows/quality.yml
```diff
@@ -23,12 +24,56 @@ jobs:
...
+            */build/reports/ktlint/ktlint*/ktlint*.txt
+      - name: Handle Results
+        if: always()
+        id: get-comment-body
+        run: |
+          results="$(cat */*/build/reports/ktlint/ktlint*/ktlint*.txt */build/reports/ktlint/ktlint*/ktlint*.txt | sed -r "s/\x1B\[([0-9]{1,3}(;[0-9]{1,2})?)?[mGK]//g")"
+          if [ -z  "$results" ]; then
+            body="👍 ✅ 👍"
+          else
+            body="\`👎 ❌ 👎 ${results}\`"
```
bmarty, line 41: Small remark, the Emojis could be outside of the "`", no?

.github/workflows/quality.yml
```diff
@@ -23,12 +24,56 @@ jobs:
...
+      - name: Find Comment
+        if: always()
+        uses: peter-evans/find-comment@v1
+        id: fc
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          comment-author: 'github-actions[bot]'
+          body-includes: Ktlint Results
+      - name: Publish ktlint results to PR
+        if: always()
```
bmarty, line 64: OOI, is there a necessity to add this line?

.github/workflows/quality.yml
```diff
@@ -23,12 +24,56 @@ jobs:
...
+        uses: peter-evans/create-or-update-comment@v1
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            ### Ktlint Results
 
-# Lint for main module and all the other modules
+            ${{ steps.get-comment-body.outputs.body }}
+          reactions: rocket
```
bmarty, line 73: I do not see the added value to automatically add a reaction to the generated comment (except that it is nice to know that it is possible to do so)

bmarty (commented):
Nice work, thanks! A few little remarks

.github/workflows/quality.yml
ariskotsomitopoulos: Sure it would, however if you observe commenting with that lib do not support `\n` so I had to manually create the new lines by replacing via bash the aggregated results with `<br>`. But then I realised that **code blocks** do not support <br> new lines! 😵‍💫 So to support **newlines** along with code blocks I added multiple `code blocks` around `<br>`. So there should be an init `left_code_block` in order to work.

It was the tricky to support newlines with **codeblocks** without the lib supporting it

The best I can do is (that seems pretty nice):
👎 ❌ 👎`Failed `

.github/workflows/quality.yml
ariskotsomitopoulos: Yes I agree, it's really sensitive, if it can support multiple commands inside the else block I will change it

.github/workflows/quality.yml
ariskotsomitopoulos: haha yes I just wanted to test it, I will remove it

.github/workflows/quality.yml
bmarty: Ah, thanks for the explaination.
Maybe do not add Emoji there but use the `reactions`, different depending on the result?
Anyway this is fine like this.

.github/workflows/quality.yml
ariskotsomitopoulos: Totally, while when `ktlint` fails thats when we should publish the results, also that was the fix for the artifacts. The problem was that the jobs was killed so nothing happened after that

bmarty: Ok, understood, thanks!

ouchadam:
awesome change! :100

would love to see a v2 that has url clickthroughs for the failing lines :star_struck:

ouchadam (approved):

.github/workflows/quality.yml
bmarty: I see, thanks

.github/workflows/quality.yml
```diff
@@ -74,7 +118,6 @@ jobs:
         run: ./gradlew clean lint${{ matrix.target }}Release --stacktrace
       - name: Upload ${{ matrix.target }} linting report
         uses: actions/upload-artifact@v2
-        if: always()
```
bmarty, line 77: Just a last remark, will the lint report be uplodoad if the command fails (so if there are lint issue) if you remove this line?

bmarty: (related to your comment https://github.com/vector-im/element-android/pull/4864/files#r779175980)

ariskotsomitopoulos: No, without ` if: always()` the artifact will not be uploaded. While if a step fails all the subsequent ones will not run.

ariskotsomitopoulos: Edit:
Here is a different case than before because the way command `./gradlew clean lint` run needs to be succeed in order to generate the report! So there is no need to upload an artifact without the report. In comparison to `ktlint`
that the report is generated when the build fails.

bmarty: Yes, thanks!

bmarty (approved):

PR #4871, opened 2022-01-06
Toolbar is added to a views with QR code scan

### LINKED ISSUE #4644: Cannot exit scan QR code without scanning a QR code
### Problem
When a user wants to add another user by scanning the QR code, the full screen camera takeover appears. This view cannot be closed - the user _must_ scan a QR code which can be annoying if they clicked there by accident. 
The only way I found to close this view was to force quit the app.

| Current | Screens |
| --- | --- |
| DM | QR code |

### Solution
Add a material.io top bar (contextual action bar) with a title and close button that takes the user back to the previous screen.

Something like this:
Screenshot 2021-12-07 at 11 19 59

---
*Edit bmarty* TODO:
For all the titles to add you can reuse an existing string
- [ ] Add a title to the screen `CreateDirectRoomByQrCodeFragment` (see above)
- [ ] Add a title "Scan a QR code" to the screen `ScanUserCodeFragment`
- [ ] Add a title "Scan their code" to the screen `QrCodeScannerFragment`

fedrunov:
Toolbar is added to screens with QR code scanner. 

| CreateDirectRoomByQrCodeFragment | ScanUserCodeFragment | QrCodeScannerFragment |
| ------------- | ------------- | ------------- |
| photo_2022-01-06_17-14-41 | photo_2022-01-06_17-15-07| photo_2022-01-06_17-15-11|

fixes #4644

### DIFF
--- a/changelog.d/4644.misc
+++ b/changelog.d/4644.misc
@@ -1,0 +1,1 @@
+Toolbar is added to a views with QR code scan
--- a/vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomByQrCodeFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomByQrCodeFragment.kt
@@ -16,7 +16,9 @@
 
 package im.vector.app.features.createdirect
 
+import android.os.Bundle
 import android.view.LayoutInflater
+import android.view.View
 import android.view.ViewGroup
 import android.widget.Toast
 import com.airbnb.mvrx.activityViewModel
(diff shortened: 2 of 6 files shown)


bmarty (approved):
Very good first PR, thanks!
As discussed we should probably merge those 3 fragments and 2 layouts into a single implementation since they are really close and it will avoid code duplication.
Also if we want to improve this screen in the future (see [this TODO](https://github.com/vector-im/element-android/blob/main/vector/src/main/res/layout/fragment_qr_code_scanner.xml#L16)) we will have to do it only at one place.
I will create an issue to track that work.

PR #4727, opened 2021-12-15
FTUE Splash carousel MVP

ouchadam:
#4668 Part of #4584

- Adds a MVP splash carousel behind a WIP flag
- Will add a changelog once all the TODOs are done and the feature is enabled

TODO 
- Style the title fullstops
- Automatic rotation of the carousel pages
- Dark mode gradient (+ gradient tweaks from design)
- Add phone landscape support
- Replace temporary images 

| GIF | 
| --- | 
|after-carousel

| 10" TABLET PORTRAIT | 10" TABLET LANDSCAPE |
| --- | --- |
Screenshot_20211215_144613|Screenshot_20211215_144605

| 3.4" PHONE SMALL | 6.5" PHONE MID/LARGE |
| --- | --- |
small-carousel|carousel_mid-large

Issue #4585, opened 2021-11-30
Build FTUE Use Case screen


daniellekirkwood:
### Your use case

Once the user has chosen to Create an Account from the Splash Screen, we want to ask them about their motivation for downloading the app (their use case). 

One day, this will help us to personalise their experience once they hit home, for now it's more about making them feel comfortable, teasing what the app is capable of, and providing a buffer before the next question (which is harder to answer).

- [ ] Discuss if we can keep this info stored for creating cohorts

- Insert Figma link

---
_This issue is considered complete when the tasks are complete, the screen matches the Figma design, it's been reviewed (by Product, Design & Engineering), and merged._

daniellekirkwood:
![Screenshot 2022-02-21 at 18 28 38](https://user-images.githubusercontent.com/89144281/155009790-f43f0e3f-727c-40eb-a850-0ad265da1eae.png)

PR #4872, opened 2022-01-06
Enabling FTUE_AUTH onboarding variant

ouchadam:
Switches the build config to point to the `FTUE_AUTH` onboarding variant, this variant is a copy of the current (legacy login) with separate feature flags for future changes, eg  #4727 and #4585 

- The `FTUE_AUTH` flow also includes #4668  

| BEFORE SPLASH | AFTER SPLASH | 
| --- | --- |
|Screenshot_20220106_172749|Screenshot_20220106_173520

| BEFORE FEATURES | AFTER FEATURES |
| --- | --- |
|Screenshot_20220106_172757|Screenshot_20220106_173528

### DIFF
--- a/changelog.d/4872.misc
+++ b/changelog.d/4872.misc
@@ -1,0 +1,1 @@
+Enabling new FTUE Auth onboarding base, includes the "I already have an account" button in the splash
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -140,7 +140,7 @@ android {
         buildConfigField "String", "BUILD_NUMBER", "\"${buildNumber}\""
         resValue "string", "build_number", "\"${buildNumber}\""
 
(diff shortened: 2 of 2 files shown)


changelog.d/4872.misc
```diff
@@ -0,0 +1 @@
+Enabling new FTUE Auth onboarding base, includes the "I already have an account" button in the splash
```
ouchadam, line 1: wasn't sure if this was a `.feature` or `.misc` as we're only switching the build config value, could even be a new separate type (if we're able to have custom types)

bmarty: OK for misc. I think we are limited by towncrier for possible extension but maybe not.

bmarty (approved):

PR #4880, opened 2022-01-07
Updated splash carousel images and copy

ouchadam:
Part of #4584

- Updates the carousel images and copy to match the latest from the designs, including new dark mode assets https://www.figma.com/file/xdV4PuI3DlzA1EiBvbrggz/Login-Flow-v2?node-id=2235%3A41941
- Adds dark mode item background
- Styles the terminating full stop in the titles 

| 1 | 2 | 3 | 4 |
| --- | --- | --- | --- |
light_1|light_2|light_3|light_4
dark_1|dark_2|dark_3|dark_4

| TABLET PORTRAIT | TABLET LANDSCAPE | SMALL PHONE | MULTI WINDOW | MULTI WINDOW |
| --- | --- | --- | --- | --- |
|Screenshot_20220114_160815|Screenshot_20220114_160800|Screenshot_20220114_160616|Screenshot_20220114_160730|Screenshot_20220114_160656


vector/src/main/java/im/vector/app/features/onboarding/ftueauth/SplashCarouselStateFactory.kt
```diff
@@ -0,0 +1,75 @@
...
+import im.vector.app.features.themes.ThemeUtils
+import me.gujun.android.span.span
+import javax.inject.Inject
+
+class SplashCarouselStateFactory @Inject constructor(
+        private val context: Context,
+        private val stringProvider: StringProvider
+) {
+
+    fun create() = SplashCarouselState(listOf(
```
ouchadam, line 34: the state creation has been extracted out to a dedicated factory as we need to apply some span logic to the terminating `.`

vector/src/main/res/layout/item_splash_carousel.xml
```diff
@@ -39,7 +39,7 @@
 
         <ImageView
             android:id="@+id/carousel_item_image"
-            android:layout_width="wrap_content"
+            android:layout_width="match_parent"
```
ouchadam, line 42: now that we have consistent bounding boxes we can matching the parent (which is constrained by gutters)

vector/src/main/java/im/vector/app/features/onboarding/ftueauth/SplashCarouselStateFactory.kt
```diff
@@ -0,0 +1,76 @@
...
+                    R.string.ftue_auth_carousel_4_title.colorTerminatingFullStop(R.attr.colorAccent),
+                    R.string.ftue_auth_carousel_4_body,
+                    R.drawable.ic_splash_collaboration,
+                    R.drawable.bg_carousel_page_4
+            )
+    ))
+
+    private fun Int.colorTerminatingFullStop(@AttrRes color: Int): EpoxyCharSequence {
+        val string = stringProvider.getString(this)
+        val fullStop = "."
```
ouchadam, line 63: this is incredibly decoupled from the input text to allow languages which don't use full stops to be skipped, maybe a good idea or a terrible one, I'm not sure yet.

bmarty: It looks acceptable to me like this. Not a big deal anyway.

vector/src/main/java/im/vector/app/features/onboarding/ftueauth/SplashCarouselStateFactory.kt
```diff
@@ -0,0 +1,76 @@
...
+        val charSequence = if (string.endsWith(fullStop)) {
```
bmarty, line 64: Wondering how it will work with RTL languages. Not critical anyway.

vector/src/main/res/drawable/ic_splash_collaboration.xml
```diff
@@ -0,0 +1,1368 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="300dp"
+    android:height="300dp"
+    android:viewportWidth="300"
+    android:viewportHeight="300">
+  <group>
+    <clip-path
```
bmarty: pretty sure lint is already complaining about the complexity of those vectors...
Not a strong opinion neither, but probably quite bad for the performance than classical png/webp images

vector/src/main/res/values/strings.xml
```diff
@@ -2521,8 +2521,8 @@
     <string name="ftue_auth_carousel_1_body">End-to-end encrypted messaging for secure and independent communication, connected via Matrix.</string>
     <string name="ftue_auth_carousel_2_title">You\'re in control.</string>
     <string name="ftue_auth_carousel_2_body">Element lets you choose where you messages are stored, keeping you in control of your data.</string>
-    <string name="ftue_auth_carousel_3_title">Connect with anyone.</string>
-    <string name="ftue_auth_carousel_3_body">Element works with all Matrix-based apps and can even bridge into proprietary messengers.</string>
+    <string name="ftue_auth_carousel_3_title">Secure Messaging.</string>
+    <string name="ftue_auth_carousel_3_body">No phone number required, so you don\'t have to share those details with the outside world. No ads, no tracking or datamining.</string>
```
bmarty, line 2525: Is "no tracking" true regarding analytics? CC @daniellekirkwood

bmarty (commented):
Thanks, just some thoughts.

vector/src/main/java/im/vector/app/features/onboarding/ftueauth/SplashCarouselStateFactory.kt
ouchadam: | PSEUDO RTL |
| --- |
Screenshot_20220110_145700

it doesn't work and the `String` behaviour is even stranger...


```
value: ‏‮Own‬‏ ‏‮your‬‏ ‏‮conversions.‬‏
first char: "‏" last char: "‏"
starts with . : false, ends with . : false 
last index of . : 30, first index of . : 30
length: 33
```

bmarty: What is "Pseudo RTL"? Even when forcing RTL from the developer option of my phone, is does not reverse the Strings like that.

ouchadam: they're fake languages for testing locales https://developer.android.com/guide/topics/resources/pseudolocales

bmarty: TIL. But seems quite not useful if it breaks fun like `endsWith`.
I prefer to force the app to "Arabic" for instance in the "Choose language" screen.

vector/src/main/res/drawable/ic_splash_collaboration.xml
ouchadam: as these are one time use assets and they're quite complex I'll covert to .png

vector/src/main/res/values/strings.xml
daniellekirkwood: 👀  Good catch, @bmarty 

Even though we don't track without explicit consent from the user we'd like to change the content to remove the reference to tracking, that way it's super clear. 

"No ads, no datamining."

@ouchadam Is that ok?

ouchadam: sounds good to me :+1:

vector/src/main/res/drawable/ic_splash_collaboration.xml
ouchadam: https://github.com/vector-im/element-android/pull/4880/commits/a35dc7f99906c3d76e75bdef796f93fbbd64cbc0

vector/src/main/res/values/strings_no_weblate.xml
```diff
@@ -10,4 +10,6 @@
     <!-- Value MUST have 4 letters and MUST be in this list: https://www.unicode.org/iso15924/iso15924-codes.html. Example:  "Arab", "Cyrl", "Latn", etc. -->
     <string name="resources_script">Latn</string>
 
+    <string name="cut_the_slack_from_teams">Cut the slack from teams.</string>
```
bmarty, line 13: Maybe better to put it in [this file](https://github.com/vector-im/element-android/blob/main/vector/src/main/res/values/donottranslate.xml) and add the `translatable="false"` attribute

ouchadam: will do :+1:

vector/src/main/res/values/strings.xml
bmarty: Thanks!

vector/src/main/res/values/strings_no_weblate.xml
ouchadam: updated https://github.com/vector-im/element-android/pull/4880/commits/2958644a46635287914105b4d4a988cbed6e4260

ouchadam:
keeping as a draft until we finalise the carousel copy

vector/src/main/java/im/vector/app/features/onboarding/ftueauth/SplashCarouselItem.kt
```diff
@@ -35,7 +35,7 @@ abstract class SplashCarouselItem : VectorEpoxyModel<SplashCarouselItem.Holder>(
 
         holder.view.setBackgroundResource(item.pageBackground)
         holder.image.setImageResource(item.image)
-        holder.title.setText(item.title)
+        holder.title.text = item.title.charSequence
```
bmarty, line 38: hopefully there is not Emoji there.

vector/src/main/res/values/strings.xml
```diff
@@ -2523,9 +2523,9 @@
     <string name="ftue_auth_carousel_1_title">Own your conversations.</string>
     <string name="ftue_auth_carousel_1_body">End-to-end encrypted messaging for secure and independent communication, connected via Matrix.</string>
     <string name="ftue_auth_carousel_2_title">You\'re in control.</string>
-    <string name="ftue_auth_carousel_2_body">Element lets you choose where you messages are stored, keeping you in control of your data.</string>
-    <string name="ftue_auth_carousel_3_title">Connect with anyone.</string>
-    <string name="ftue_auth_carousel_3_body">Element works with all Matrix-based apps and can even bridge into proprietary messengers.</string>
+    <string name="ftue_auth_carousel_2_body">Element lets you choose where your messages are stored, keeping you in control of your data.</string>
```
bmarty, line 2526: We should not use hard-coded `Element` in the string. Use `{app_name}` and add `template` to the string key. (see other example in this file.

bmarty (commented):
One last remark, sorry.

vector/src/main/java/im/vector/app/features/onboarding/ftueauth/SplashCarouselItem.kt
ouchadam: :sweat_smile:   we should be okay as we've using the `EpoxyCharSequence` and there's no `setTextFuture` :crossed_fingers:

vector/src/main/res/values/strings.xml
ouchadam: https://github.com/vector-im/element-android/pull/4880/commits/57ccf2d16a115e47809e87dc63aac682dd45fc1e

also marks the previous strings for removal

bmarty (approved):
Thanks for the update

PR #4882, opened 2022-01-07
Olm lib is now hosted in MavenCentral

bmarty:
And upgrade to Olm <del>3.2.9</del> **3.2.10**

Sanity test pass.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,14 +36,6 @@ allprojects {
     apply plugin: "org.jlleitschuh.gradle.ktlint"
 
     repositories {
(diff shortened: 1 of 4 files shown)


bmarty:
There is a problem with the lib, it is not full:

image

I was also wondering why the lib was so small (see https://repo1.maven.org/maven2/org/matrix/android/olm-sdk/3.2.9/): only 26036 bytes.

bmarty:
Will have to wait for release 3.2.10 of Olm.

Issue #4884, opened 2022-01-09
Align toolbar usage


fedrunov:
#### What would you like to do?
Align the way toolbar is managed across the app

#### Why would you like to do it?

Currently on some screens toolbar's title, subtitle and navigation button (back button) are defined separately as roolbar's children, while on the other screens built-in toolbar's methods and widgets are used. This leads to a different margins, size and text appeareance. 

#### How would you like to achieve it?

there are 4 main use cases of toolbar in the app:

1. default one with just title, subtitle, navigation button and action buttons
2. default one + tabs control
3. default one + search
4. compex cases with a lot of widgets and/or custom position and appearance of title and subtitle

for first 3 cases toolbar's internal widgets can be utilized and also toolbar can be extracted to a separate resource file and included in a view. This will ensure consistancy between the screens and also will allow to write helper methods/builder to manage this casese

for 4th case it's not possible to reuse layout, but extracting toolbar to a separate layout file for each screen could make laoyuts easier to read. Incorporating this cases to a builder from previous paragraph is tricky and could not give a lot of benefit I think, so it wil be better to keep accessing widgets inside toolbar via bindings as it is now. But at least for title and subtitle text appearances from styles_toolbar.xml could be used (if fit design) to make it consistent with others and also to ease possible changes 

### Additional context

Differences in layout

![image](https://user-images.githubusercontent.com/66663241/148704089-a9e1d229-d6e8-4bce-9974-5676afeb9be6.png)  Main Screen 
![image](https://user-images.githubusercontent.com/66663241/148704102-b85beeab-f1b3-470d-a1be-2dfae72d7c2a.png)  Create room

PR #4887, opened 2022-01-10
toolbar management 

fedrunov:
Toolbar management rework

All toolbars across app now uses general text appearance for title and subtitle from ui-styles/styles_toolbar
Most of the toolbars now use built-in widgets for title, subtitle and nav button instead of custom
Added methods in fragment and activity to setup toolbar, which returns wrapper class with set of helper methods to ease toolbar configuration

| Before  | Now |
| ------------- | ------------- |
| Room Detail | Room Detail  |
| RoomDetail_old | RoomDetail_new |
| image  | image |
| image | image |

#4884


vector/src/main/java/im/vector/app/core/platform/VectorBaseActivity.kt
```diff
@@ -500,7 +501,7 @@ abstract class VectorBaseActivity<VB : ViewBinding> : AppCompatActivity(), Maver
     /**
      * Configure the Toolbar, with default back button.
      */
-    protected fun configureToolbar(toolbar: MaterialToolbar, displayBack: Boolean = true) {
```
fedrunov, line 503: currently just removed default value to not cause compile error. Later should be replaced completely by builder method

vector/src/main/java/im/vector/app/core/platform/VectorBaseFragment.kt
```diff
@@ -222,6 +223,10 @@ abstract class VectorBaseFragment<VB : ViewBinding> : Fragment(), MavericksView
         }
     }
 
+    //we need to decide if we want to use vectorBaseActivity here
+    //the problem is that it casts getActivity() result to non nullable type, so in case if it will return null, NullPointerException will be thrown
+    protected fun configureToolbar(toolbar: MaterialToolbar) = (activity as? VectorBaseActivity<*>)?.configureToolbar(toolbar) ?: ToolbarConfig(null, toolbar)
```
fedrunov, line 228: it's possible to remove toolbar argument if we will use same id for toolbar on every screen. But in that case we'll need to look it up with findviewById, which is slower that just pass reference from bindings. On the other side passing toolbar view every time you want to update title on a screen could be annoying boilerplate.

vector/src/main/java/im/vector/app/core/utils/ToolbarConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+    fun allowBack(isBackAllowed: Boolean) = apply { this.isBackAllowed = isBackAllowed }
+    fun setBackAsClose(useCloseButton: Boolean) = apply { this.useCloseButton = useCloseButton }
+    fun withCustomBackIcon(@DrawableRes customBackResId: Int) = apply { this.customBackResId = customBackResId }
+
+    fun configure(){
+        if (activity == null) {
+            return
+        }
+        activity.setSupportActionBar(toolbar)
+        toolbar.title = titleRes?.let { activity.getString(it) }
```
fedrunov, line 48: current implementation will force "null" on a title and subtitle in case if it wasn't set before configure. This could be a problem for cases when you want to update any of them separetely on a runtime. I didn't solve this case for this draft, because I'm not sure yet if it exists

vector/src/main/java/im/vector/app/features/contactsbook/ContactsBookFragment.kt
```diff
@@ -67,13 +68,17 @@ class ContactsBookFragment @Inject constructor(
...
         contactsBookViewModel.observeViewEvents {
             when (it) {
                 is ContactsBookViewEvents.Failure             -> showFailure(it.throwable)
                 is ContactsBookViewEvents.OnPoliciesRetrieved -> showConsentDialog(it)
             }.exhaustive
         }
+        configureToolbar(views.phoneBookToolbar)
+                .withTitle(R.string.contacts_book_title)
+                .allowBack(true)
+                .withCustomBackIcon(R.drawable.ic_x_18dp)
```
fedrunov, line 80: I'm not sure if we use X button everyevere, if so, it could be better to make it default and call this method only for other cases

vector/src/main/java/im/vector/app/core/utils/ToolbarConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+
+package im.vector.app.core.utils
+
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+import androidx.appcompat.app.AppCompatActivity
+import androidx.appcompat.content.res.AppCompatResources
+import com.google.android.material.appbar.MaterialToolbar
+
+class ToolbarConfig(val activity: AppCompatActivity?, val toolbar: MaterialToolbar) {
```
bmarty, line 25: Maybe activity could be required.

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomByQrCodeFragment.kt
```diff
@@ -63,12 +63,11 @@ class CreateDirectRoomByQrCodeFragment @Inject constructor() : VectorBaseFragmen
...
-
-        views.qrScannerClose.debouncedClicks {
-            requireActivity().onBackPressed()
-        }
-        views.qrScannerTitle.text = getString(R.string.add_by_qr_code)
+        configureToolbar(views.qrScannerToolbar)
+                .withTitle(R.string.add_by_qr_code)
+                .allowBack(true)
+                .withCustomBackIcon(R.drawable.ic_x_18dp)
+                .configure()
```
bmarty, line 70: We will get the click on back for free? (jsut to confirm)

vector/src/main/java/im/vector/app/core/utils/ToolbarConfig.kt
```diff
@@ -0,0 +1,60 @@
...
+    fun withTitle(@StringRes titleRes: Int) = apply {this.titleRes = titleRes }
+    fun withTitle(title: String) = apply {this.title = title}
+    fun withSubtitle(@StringRes subtitleRes: Int) = apply {this.subtitleRes = subtitleRes }
+    fun withSubtitle(subtitle: String) = apply {this.subtitle = subtitle}
+
+    fun allowBack(isBackAllowed: Boolean) = apply { this.isBackAllowed = isBackAllowed }
+    fun setBackAsClose(useCloseButton: Boolean) = apply { this.useCloseButton = useCloseButton }
+    fun withCustomBackIcon(@DrawableRes customBackResId: Int) = apply { this.customBackResId = customBackResId }
+
+    fun configure(){
```
ouchadam, line 43: we could split this responsibilities of this class by moving the config to a data class with a kotlin dsl style builder and then move the toolbar configuration to the activity (or extracted to a delegate)

dsl builders have the benefit of reducing boilerplate 

```kotlin
data class ToolbarConfig(
  var titleRes: Int? = null,
  var allowBack: Boolean = true,
)

// usage
activity.configureToolbar {
    titleRes = R.string.contacts_book_title
    allowBack = true
    customBackIcon = R.drawable.ic_x_18dp
}

fun VectorBaseActivity<*>.configureToolbar(builder: ToolbarConfig.() -> Unit) {
    val config = ToolbarConfig().apply { builder(this) }

    // apply the config to the toolbar
    setSupportActionBar(toolbar)
    toolbar.title = config.titleRes?.let { getString(it) }
    ...
}

```

what do you think?

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomByQrCodeFragment.kt
fedrunov: Yes

vector/src/main/java/im/vector/app/core/utils/ToolbarConfig.kt
fedrunov: Check new implementation. It's not exactly same, but also reduce amount of boilerplate

library/ui-styles/src/main/res/values/styles_toolbar.xml
```diff
@@ -24,14 +26,16 @@
...
+        <item name="android:textSize">18sp</item>
     </style>
 
     <!-- sub text -->
     <style name="Widget.Vector.TextView.ActionBarSubTitle" parent="TextAppearance.AppCompat.Widget.ActionBar.Subtitle">
-        <item name="android:textColor">?vctr_content_primary</item>
-        <item name="android:fontFamily">"sans-serif-medium"</item>
+        <item name="android:textColor">?vctr_content_secondary</item>
+        <item name="android:fontFamily">sans-serif</item>
+        <item name="fontFamily">sans-serif</item>
```
bmarty, line 38: There is an existing naming problem with those 2 styles. As they are TextAppearance, their name should start with `TextAppearance.Vector` and not `Widget.Vector.TextView`. Can you take the opportunity to rename the 2 styles please?
The two new names would be:
- `TextAppearance.Vector.Widget.ActionBar.Title`
- `TextAppearance.Vector.Widget.ActionBar.Subtitle`

(I am not sure on the naming convention though. @ganfra WDYT?)

bmarty (approved):
Nice rework!
LGTM, one small remark (or extra work)

vector/src/main/java/im/vector/app/core/utils/ToolbarConfig.kt
```diff
@@ -0,0 +1,82 @@
...
+     * */
+    fun setSubtitle(@StringRes subtitleRes: Int) = apply { toolbar.subtitle = activity?.getString(subtitleRes) }
+
+    /**
+     * Enables/disables navigate back button
+     *
+     * @param isAllowed defines if back button is enabled. Default [true]
+     * @param useCross defines if cross icon should be used instead of arrow. Default [false]
+     * */
+    fun allowBack(isAllowed: Boolean = true, useCross: Boolean = false) = apply {
```
ouchadam, line 72: what happens if we don't call `allowBack`? previously we were always calling `setDisplayShowHomeEnabled/setDisplayHomeAsUpEnabled`

fedrunov: Oh, I forgot to mention this. Initially I did it in way, when back was enabled by default, so there will be less boilerplate. But then, while I was reworking screens, I faced a problem when you can't say if back button enabled for a specific fragment or not, by just looking to it's code, since it was controlled by activity. So I had a thought that even though it will increase boilerplate, it's still might be valuable to force to call .allowBack if you want to show back button, since in this case it's explicit and you don't need to know about any default value/hidden calls/etc. But this can easily be reverted back to having back enabled by default and calling it only for a case when you need to not show it. What do you think about it? @bmarty could you also share you thoughts?

bmarty: I have no strong opinion, as far as there is no regression :)

ouchadam: I'm also easy, explicitly adding the back option sounds like a good refactoring step

I do wonder if we should only expose a subset of options to fragments which modify toolbars, only let them update the titles/options rather than determine the back behaviour and then force (or default) all activities with a toolbar (minus the top level home) to have a back/cross action, as we'll always be going a level deeper when navigating to them

bmarty: On a previous project, I had some open funs like `isModal(): Boolean`, `getTitle(): String?`, `getSubtitle(): String?`, in the parent fragment and parent activity and all the logic was managed there. The developer only had to answer the questions when implementing a new screen, it was quite handy. Here I think it's next level, as long as it's easy to use for the developer and that default behavior is sort of the expected one.

fedrunov: modal screens are still in discussion, but when we will implement, we could ofcourse take there some part of the logic, like cross icon for sure. About title and subtitle - I'd rather prefer having them in xml, when it's possibe, so it will be easier to see  together with rest part of UI. We can change mind later, but for now, if i got it correct, I will leave requirement to enable back button explicitly. Let me know if I got it wrong :)

ouchadam (approved):

bmarty:
@fedrunov you can run `./gradlew ktlintFormat` to fix some issues reported by the CI. Other issues have to be fixed manually. Then it will be possible to merge the PR.

Thanks!

PR #4888, opened 2022-01-10
Add comment only on ktlint fail

ariskotsomitopoulos:
This PR will contain the following changes:

- Do not add GitHub comments on successful `ktlint` runs
- Remove already existing comments when `ktlint` succeed


bmarty (approved):
Nice change, thanks!

PR #4893, opened 2022-01-10
Workaround to not to reuse poll option cells

### LINKED ISSUE #4735: Polls: unable to create a poll with more than 10 answers
The poll creation screen is broken if there are too many options, and the user scrolls.

The recycling of view mess up the answer contents.

I am afraid the user input can be lost in this case.

onurays:
Fixes #4735


bmarty:
I confirm it fixes the issue

bmarty (approved):

PR #4895, opened 2022-01-10
Avoid allowing null String for state_key.

bmarty:
Should always be an empty String according to the Matrix specification.

There is no functional change, just a change in the SDK API for clarity regarding the Matrix specs.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/state/StateService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/state/StateService.kt
@@ -68,8 +68,11 @@ interface StateService {
 
     /**
      * Send a state event to the room
+     * @param eventType The type of event to send.
+     * @param stateKey The state_key for the state to send. Can be an empty string.
+     * @param body The content object of the event; the fields in this object will vary depending on the type of event
      */
-    suspend fun sendStateEvent(eventType: String, stateKey: String?, body: JsonDict)
+    suspend fun sendStateEvent(eventType: String, stateKey: String, body: JsonDict)
 
     /**
      * Get a state event of the room
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/state/SendStateTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/state/SendStateTask.kt
@@ -26,7 +26,7 @@ import javax.inject.Inject
 internal interface SendStateTask : Task<SendStateTask.Params, Unit> {
     data class Params(
             val roomId: String,
-            val stateKey: String?,
+            val stateKey: String,
             val eventType: String,
             val body: JsonDict
     )
@@ -39,7 +39,7 @@ internal class DefaultSendStateTask @Inject constructor(
 
     override suspend fun execute(params: SendStateTask.Params) {
         return executeRequest(globalErrorReceiver) {
-            if (params.stateKey == null) {
+            if (params.stateKey.isEmpty()) {
                 roomAPI.sendStateEvent(
                         roomId = params.roomId,
                         stateEventType = params.eventType,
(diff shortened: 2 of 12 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/state/SendStateTask.kt
```diff
@@ -39,7 +39,7 @@ internal class DefaultSendStateTask @Inject constructor(
 
     override suspend fun execute(params: SendStateTask.Params) {
         return executeRequest(globalErrorReceiver) {
-            if (params.stateKey == null) {
+            if (params.stateKey.isEmpty()) {
```
bmarty, line 42: I think it is an improvement, I am not sure what would happen if `stateKey` was an empty String on the build URL (it would end by `/`.

ganfra, line 42: Why do we even need to test this?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/state/StateService.kt
```diff
@@ -68,8 +68,11 @@ interface StateService {
 
     /**
      * Send a state event to the room
+     * @param eventType The type of event to send.
+     * @param stateKey The state_key for the state to send. Can be an empty string.
+     * @param body The content object of the event; the fields in this object will vary depending on the type of event
      */
-    suspend fun sendStateEvent(eventType: String, stateKey: String?, body: JsonDict)
+    suspend fun sendStateEvent(eventType: String, stateKey: String, body: JsonDict)
```
ouchadam, line 75: are there other endpoints which use empty strings to represent the absence of a value? 

thinking out loud, from a public api perspective I would prefer to pass a null than an empty string as it gives us some extra type safety  (and then let the internal implementation convert to an empty string)

bmarty: We want the SDK API to follow as much as possible the Matrix specification/philosophy. In the spec, the `stateKey` cannot be null.
Ref: https://matrix.org/docs/spec/client_server/r0.6.1#put-matrix-client-r0-rooms-roomid-state-eventtype-statekey

bmarty: > extra type safety

I do not understand why?

ouchadam: a string can contain content or be empty, whereas a null can only be null

the spec is a little bit misleading here when it mentions empty strings because the value of "" is an implementation detail of building a url path, we never send `""` to the api.

it actually means to skip applying the `stateKey` segment, which means the `stateKey required` is actually a lie...

```
has a state key
/_matrix/client/r0/rooms/{roomId}/state/{eventType}/{stateKey}/

doesn't have a state key
/_matrix/client/r0/rooms/{roomId}/state/{eventType}/
```

assuming I've understood correctly! not a blocker for the PR, something I found interesting.

bmarty: In API paths we do not send "" (or we alsways do it, since its nothing!), but I am wondering if 
`/_matrix/client/r0/rooms/{roomId}/state/{eventType}/` is equivalent to
`/_matrix/client/r0/rooms/{roomId}/state/{eventType}` (note the missing trailing `/`).
Anyway I think the spec make a diff mainly for the Json of the Event content itself:
`"state_key": ""` is not equivalent to `"state_key": null` (or missing `"state_key"`).
that's maybe why it is a bit confusing all along.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/state/SendStateTask.kt
bmarty: probably if 
`/_matrix/client/r0/rooms/{roomId}/state/{eventType}/` is not equivalent to
`/_matrix/client/r0/rooms/{roomId}/state/{eventType}`

giomfo (approved):
LGTM - thx

ouchadam (approved):

PR #4911, opened 2022-01-11
Remove "X kicked X" terminology when removing people from a room

### LINKED ISSUE #4865: Remove "X kicked X" terminology when removing people from a room
To remove the wording "X kicked X" when someone is removed from a room and replace with "X removed X"

Because it is more acceptable terminology

---
*EDIT bmarty*
Ok, confirmed internally:
- replace all instance of "kick" and their variations to "remove" and their variations in the English wording. Translation will updated later following the regular Weblate process. No need to create new strings.
- add the slash command `/remove` but keep the existing `/kick` for now. They will have the same effect.

fedrunov:
there are three separate commits:

1. replacing "kick" with "remove" in code and resource. String reource's keys are unaffected, @bmarty said that will handle it later
2. adding alias to commands. 
3. adding aliases which are already exist in web app, taking from  https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/SlashCommands.tsx

also there are some small fixes/changes:

- case is now ignored (before /Nick returned an error, now works)
- fixed typo ("sPlashCommand") in method name and logs (could change in logs somehow affect analytics?) 
- added strict check for excess params for DISCARD_SESSION command, same as it is for CLEAR_SCALAR_TOKEN
- some commands didn't require param, while it looks like they should, so check was addde:
  /rainbow - posts empty message
  /spoiler - posts empty message
  /me - posts message with just user name
  /rainbowme - posts message with just user name

fixes #4865


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/members/MembershipService.kt
```diff
@@ -77,7 +77,7 @@ interface MembershipService {
     /**
      * Kick a user from the room
      */
-    suspend fun kick(userId: String, reason: String? = null)
+    suspend fun remove(userId: String, reason: String? = null)
```
bmarty, line 80: Maybe also update the comment above

bmarty, line 80: As it is a change in the SDK api, would be better to keep `kick` as an alias to remove and mark it as deprecated.
You can write in the interface something like:
```
@Deprecated("Use remove now")
suspend fun kick(userId: String, reason: String? = null) = remove(userId, reason)
```

vector/src/main/java/im/vector/app/features/command/Command.kt
```diff
@@ -24,42 +24,46 @@ import im.vector.app.R
...
-    WHOIS("/whois", "<user-id>", R.string.command_description_whois, false),
-    DISCARD_SESSION("/discardsession", "", R.string.command_description_discard_session, false),
-    CONFETTI("/confetti", "<message>", R.string.command_confetti, false),
-    SNOWFALL("/snowfall", "<message>", R.string.command_snow, false),
-    CREATE_SPACE("/createspace", "<name> <invitee>*", R.string.command_description_create_space, true),
-    ADD_TO_SPACE("/addToSpace", "spaceId", R.string.command_description_add_to_space, true),
-    JOIN_SPACE("/joinSpace", "spaceId", R.string.command_description_join_space, true),
-    LEAVE_ROOM("/leave", "<roomId?>", R.string.command_description_leave_room, true),
-    UPGRADE_ROOM("/upgraderoom", "newVersion", R.string.command_description_upgrade_room, true);
+enum class Command(val command: String, val aliases: Array<CharSequence>?, val parameters: String, @StringRes val description: Int, val isDevCommand: Boolean) {
```
bmarty, line 27: not a strong opinion about it, but why using `Array` instead of `List`?

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
```diff
@@ -32,12 +32,12 @@ object CommandParser {
      * @param textMessage   the text message
      * @return a parsed slash command (ok or error)
      */
-    fun parseSplashCommand(textMessage: CharSequence): ParsedCommand {
+    fun parseSlashCommand(textMessage: CharSequence): ParsedCommand {
```
bmarty, line 35: Good catch, nice typo, thanks :)

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
```diff
@@ -32,12 +32,12 @@ object CommandParser {
...
         // check if it has the Slash marker
         if (!textMessage.startsWith("/")) {
             return ParsedCommand.ErrorNotACommand
         } else {
-            Timber.v("parseSplashCommand")
+            Timber.v("parseSlashCommand")
```
bmarty, line 40: Changing the logs does not impact analytics.

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
```diff
@@ -361,57 +353,49 @@ object CommandParser {
...
-                    ParsedCommand.LeaveRoom(
-                            spaceIdOrAlias
-                    )
-                }
-                Command.UPGRADE_ROOM.command                 -> {
-                    val newVersion = textMessage.substring(Command.UPGRADE_ROOM.command.length).trim()
-                    if (newVersion.isEmpty()) {
-                        ParsedCommand.ErrorSyntax(Command.UPGRADE_ROOM)
+                Command.ADD_TO_SPACE.matches(slashCommand)                 -> {
+                    ParsedCommand.AddToSpace(spaceId = message)
```
bmarty, line 373: Maybe we miss some validation of `message` for this command and the one below.

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
```diff
@@ -361,57 +353,49 @@ object CommandParser {
...
                     // Unknown command
                     ParsedCommand.ErrorUnknownSlashCommand(slashCommand)
                 }
             }
         }
     }
+
+    private fun trimParts(message: CharSequence, messageParts: List<String>): String? {
+        val partsSize = messageParts.sumOf { it.length }
+        val gapsNumber = messageParts.size - 1
```
bmarty, line 398: I am wondering if it still works well if the user adds more space chars between the params

vector/src/main/res/values/strings.xml
```diff
@@ -25,8 +25,8 @@
     <string name="notice_direct_room_leave_by_you">You left the room</string>
     <string name="notice_room_reject">%1$s rejected the invitation</string>
     <string name="notice_room_reject_by_you">You rejected the invitation</string>
-    <string name="notice_room_kick">%1$s kicked %2$s</string>
-    <string name="notice_room_kick_by_you">You kicked %1$s</string>
+    <string name="notice_room_kick">%1$s removed %2$s</string>
```
bmarty, line 28: Can you add a TODO in XML comment above so that I will not forget to rename during the next weblate sync. Thanks. (only one comment will be enough)

changelog.d/4865.misc
```diff
@@ -0,0 +1 @@
+"/kick" command is replaced with "/remove"
```
bmarty, line 1: Maybe also mention the replacement in the wordings.

bmarty (commented):
Not a lot to say, this is really nice work, thanks!

vector/src/main/java/im/vector/app/features/command/CommandParser.kt
bmarty: (my bad eae8f993e642226bc4f7c9c565d41c368e7d0751 :) )

vector/src/main/java/im/vector/app/features/command/Command.kt
fedrunov: I did it because of spread operation used later here

`    val allAliases = arrayOf(command, *aliases.orEmpty())`

bmarty (approved):
Thanks for the update!

PR #4914, opened 2022-01-11
Disabling automatic carousel transitions on user interaction

ouchadam:
Part of #4584
 
- Disables the automatic carousel transitions once the user interacts with the carousel

| BEFORE | AFTER |
| --- | --- |
before-carousel|after-carousel

### DIFF
--- a/vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthSplashCarouselFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthSplashCarouselFragment.kt
@@ -80,17 +80,27 @@ class FtueAuthSplashCarouselFragment @Inject constructor(
                     "Branch: ${BuildConfig.GIT_BRANCH_NAME}"
             views.loginSplashVersion.debouncedClicks { navigator.openDebug(requireContext()) }
         }
+        views.splashCarousel.registerAutomaticUntilInteractionTransitions()
+    }
 
-        views.splashCarousel.apply {
-            var scheduledTransition: Job? = null
-            registerOnPageChangeCallback(object : ViewPager2.OnPageChangeCallback() {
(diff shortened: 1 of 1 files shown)


vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthSplashCarouselFragment.kt
```diff
@@ -80,17 +80,27 @@ class FtueAuthSplashCarouselFragment @Inject constructor(
...
-                override fun onPageSelected(position: Int) {
-                    scheduledTransition?.cancel()
+            override fun onPageScrolled(position: Int, positionOffset: Float, positionOffsetPixels: Int) {
+                hasUserManuallyInteractedWithCarousel = !isFakeDragging
+            }
+
+            override fun onPageSelected(position: Int) {
+                scheduledTransition?.cancel()
+                if (hasUserManuallyInteractedWithCarousel) {
+                    // stop the automatic transitions
```
bmarty, line 98: This is fine but maybe avoid having empty block. I think lint/compiler will also complain. Better to write `if(!...)` and just add the comment above.

bmarty (approved):
Thanks! Just one small remark

vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthSplashCarouselFragment.kt
ouchadam: in this case the comment counts as a non empty block, will invert the condition and remove the branch

ouchadam: 01ef46517e0358d0446a9b13fd0612e6286c4883

bmarty (approved):

PR #4918, opened 2022-01-11
Locking phones to portrait during FTUE onboarding

ouchadam:
Part of #4584 

- Makes use of a resource bucket flag to determine if the device is big enough to be considered a tablet and in turn, enable a landscape experience in the FTUE auth onboarding flow

| BEFORE PHONE | AFTER PHONE | AFTER LOGIN | AFTER TABLET |
| --- | --- | --- | --- |
|before-rotation|after-rotation|after-login|tablet-rotation

### DIFF
--- a/library/ui-styles/src/main/res/values/tablet.xml
+++ b/library/ui-styles/src/main/res/values/tablet.xml
@@ -2,5 +2,6 @@
 <resources>
 
     <dimen name="width_percent">1</dimen>
+    <bool name="is_tablet">false</bool>
 
 </resources>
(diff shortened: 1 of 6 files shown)


library/ui-styles/src/main/res/values/tablet.xml
```diff
@@ -2,5 +2,6 @@
 <resources>
 
     <dimen name="width_percent">1</dimen>
+    <bool name="is_tablet">false</bool>
```
bmarty, line 5: weird to see that in a file (that I) named `tablet.xml` :)

bmarty (approved):
LGTM 👍

ouchadam:
> Can I ask you to add the same rotation lock to the Activity `AnalyticsOptInActivity` please? I do not see any other screen where it could be applied at the moment.
> 
> Thanks!

will do :+1:

bmarty (approved):
Thanks!

PR #4924, opened 2022-01-12
Ensure algorithm instance is created and stored

BillCarsonFr:
Ensure that algorithm instance is created and cached when setEncryptionInRoom is called even if no alg change

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
@@ -578,7 +578,7 @@ internal class DefaultCryptoService @Inject constructor(
         // (for now at least. Maybe we should alert the user somehow?)
         val existingAlgorithm = cryptoStore.getRoomAlgorithm(roomId)
 
-        if (existingAlgorithm == algorithm) {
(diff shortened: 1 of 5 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -578,7 +578,7 @@ internal class DefaultCryptoService @Inject constructor(
         // (for now at least. Maybe we should alert the user somehow?)
         val existingAlgorithm = cryptoStore.getRoomAlgorithm(roomId)
 
-        if (existingAlgorithm == algorithm) {
+        if (existingAlgorithm == algorithm && roomEncryptorsStore.get(roomId) != null) {
```
ouchadam, line 581: would it be helpful to split the conditions into separate ifs/whens in order to provide unique logs?

ouchadam (approved):

ouchadam:
can confirm this fixes the issue for me

| BEFORE | AFTER |
| --- | --- |
Screenshot_20220112_102015|Screenshot_20220112_102147

ganfra (approved):

bmarty:
Also confirm the bugs are fixed on my side. I do not wait for the CI, only the changelog added.

bmarty (approved):

PR #4926, opened 2022-01-12
Add signing config for the release. No secret added

bmarty:
To be able to test the release build sooner during the release process.


BillCarsonFr (approved):
LGTM

Bubu:
This should be fine for f-droid. signing configs are automatically removed from the gradle files prior to running any builds. (apks are signed afterwards on a separate buildserver)

PR #4950, opened 2022-01-13
Add lambda in executeRequest params to define a custom retry policy

Florian14:
Previously, failed requests were retried according to a `canRetry` boolean passed in `executeRequest` method params. There is an exception for the 429 error with the `M_LIMIT_EXCEEDED` error code which were automatically retried without taking account of the `canRetry` flag.

The goal of this PR is to have more flexibility on the retry policy to be able to have a custom policy for a specific request. For example, a 429 error with a very long retry delay should no be retried but notify the user about the delay.

To keep the previous behaviour, the current retry policy has been kept and passed as the default value of the new `canRetryOnFailure` method parameter.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
@@ -19,6 +19,7 @@ package org.matrix.android.sdk.internal.network
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.delay
 import org.matrix.android.sdk.api.failure.Failure
+import org.matrix.android.sdk.api.failure.GlobalError
 import org.matrix.android.sdk.api.failure.MatrixError
 import org.matrix.android.sdk.api.failure.getRetryDelay
 import org.matrix.android.sdk.api.failure.shouldBeRetried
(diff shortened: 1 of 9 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
```diff
@@ -74,23 +84,16 @@ internal suspend inline fun <DATA> executeRequest(globalErrorReceiver: GlobalErr
...
-                    exception.error.code == MatrixError.M_LIMIT_EXCEEDED &&
-                    currentRetryCount < maxRetriesCount) {
-                // 429, we can retry
-                delay(exception.getRetryDelay(1_000))
-            } else if (canRetry && currentRetryCount < maxRetriesCount && exception.shouldBeRetried()) {
-                delay(currentDelay)
+            if (canRetryOnFailure(exception) && currentRetryCount < maxRetriesCount) {
+                delay(exception.getRetryDelay(currentDelay))
                 currentDelay = currentDelay.times(2L).coerceAtMost(maxDelayBeforeRetry)
-                // Try again (loop)
```
bmarty, line 86: Maybe restore this comment, it is still valid.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
```diff
@@ -74,23 +84,16 @@ internal suspend inline fun <DATA> executeRequest(globalErrorReceiver: GlobalErr
...
-            if (exception is Failure.ServerError &&
-                    exception.httpCode == 429 &&
-                    exception.error.code == MatrixError.M_LIMIT_EXCEEDED &&
-                    currentRetryCount < maxRetriesCount) {
-                // 429, we can retry
-                delay(exception.getRetryDelay(1_000))
-            } else if (canRetry && currentRetryCount < maxRetriesCount && exception.shouldBeRetried()) {
-                delay(currentDelay)
+            if (canRetryOnFailure(exception) && currentRetryCount < maxRetriesCount) {
+                delay(exception.getRetryDelay(currentDelay))
```
bmarty, line 88: This is not strictly equivalent with the previous version, when we had a 429, the default delay was 1s. But I think it is still fine.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/RequestExecutor.kt
```diff
@@ -15,22 +15,24 @@
  */
 
 package org.matrix.android.sdk.internal.network
+
+import org.matrix.android.sdk.internal.network.defaultRequestRetryPolicy as internalDefaultRequestRetryPolicy
```
bmarty, line 19: I do not think it is necessary to create an alias in this case.

bmarty (commented):
At first sight it looks good to me. I guess the next step is to use the new mechanism to change the login request?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
Florian14: Yes indeed, if we want to keep previous behaviour for 429, we should also return a fixed delay in the lambda (or null to use the default delay behaviour which is multiplied by 2 after each failure).

With my current implementation, receiving 4 successive 429 errors will lead to a total delay of 15 seconds (1 + 2 + 4 +8) instead of 4 seconds with the fixed delay. We should probably have to think about it. Was the previous behaviour originally expected for this special case ?

Note that the delay is passed as a default value in this special case, the delay passed in the 429 error body should be used, if any (it is not a mandatory field).

WDYT @bmarty?

Florian14:
> At first sight it looks good to me. I guess the next step is to use the new mechanism to change the login request?

Not sure, this is the case of Tchap fork, but should we apply it for all login flows? I haven't tried to trigger the error using matrix.org yet.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
bmarty: I see, the point is that if you want to have an automatic retry in case of 429, except if the delay is too long (which could be configurable by the app), it will be more difficult with your change.
So maybe keep the two cases (429 and canRetry) clearly separated.

Florian14: Ack. I close this PR and will create a new one.

Issue #4669, opened 2021-12-09
Permanently unresponsive (no sync) after returning from background


ManDay:
### Illustration

https://user-images.githubusercontent.com/1059827/147871105-a504390d-1365-43a0-aa2b-6ddb0953959e.mp4

### Steps to reproduce

Open a room and let the phone go to standby/display off after delay. Then unlock the phone again, thus returning to the room.

### Outcome

The "sync" spinner at the top border will appear and start spinning indefinitely.

From that point on

* no synchronisation will ever succeed
* "Clear cache" hangs indefinitely on the initial splash screen, too
* Closing Element Android has no effect

The only solution that I've found so far is using "Force Stop" on Element Android in the device settings.

### Your phone model

Cubot J5

### Operating system version

Android 9

### Application version and app store

Element 1.3.7 F-8b98c8fe

### Homeserver

matrix.org

### Will you send logs?

Yes

ManDay:
About reproducability: It happens occasionally with a high probability. I have not been able to reproduce it in small rooms, although once the problem occurs, sync fails for all rooms. On larger rooms like #matrix I would say it happens in 8 out of 10 trials.

ManDay:
If it is of any significance: When this happens (I've joined more than 3 rooms as of recently and now it happens with a probability of >60%, even in the overview of rooms and when returning to the app - this has now forced me to use *Element Web* because *Element Android* is basically unusable), cutting the network -- which normally causes a warning about *Airplane Mode* -- does not have any effect, either: The loading indicator just keeps spinning.

ManDay:
@SpiritCroc from Schildi (which has the same problem from the Element Codebase) has told me that a video might clarify the problem, I've added it to the opening post. You can see how after several tabbing back-and-forth Element bricks in the indefinite sync. I suspect that the stochastic doesn't follow the number of switches, because it happens with a 80% probability when I come back to Element after it has been in the background for several minutes.

amas0:
Just wanted to jump in and say that I'm also experiencing this issue with high probability upon returning to Element from it being in the background. I've found I can fix it temporarily by closing the app completely (removing it from my background apps) and reopening it. 

I am running 1.3.12 (G-b5040) on a Galaxy S21.

ManDay:
[...]

PR #4952, opened 2022-01-14
Force HTTP/1.1

abaker:
Quite frequently when I am on mobile data I will experience #4669. The incremental sync state will be idle, but the progress bar will be visible for one minute before throwing a timeout exception and displaying a connectivity error bar

I found some open OkHttp issues that were possibly related (3146, 3278 and 4455). I didn't drop links because I don't know if they're the cause, but these led me to disabling HTTP/2 which seems to resolve the issue.

I have attached a recording with the following setup:
- Phone1 as a mobile hotspot
- Phone2 in airplane mode connected to hotspot, proxying connections through a laptop to record traffic

I'm running two builds of element in split screen on phone2. The top app (dark mode) is this PR, the bottom app (light mode) is v1.3.14 from the Play Store. The apps show a blue airplane mode bar instead of a red connectivity error bar because I'm connecting through the hotspot.

Based on the proxy traffic it appears that phone1 briefly loses internet connectivity and the HTTP/2 sync request gets stuck for some reason.

It may be possible to fix this without disabling HTTP/2, e.g. listening for connectivity changes and evicting the OkHttp connection pool.

https://user-images.githubusercontent.com/220908/149430467-5a1c40ee-660f-4269-aad1-28b835c5b1e9.mp4

---

### DIFF
--- a/changelog.d/4669.bugfix
+++ b/changelog.d/4669.bugfix
@@ -1,0 +1,1 @@
+Fix sync timeout after returning from background
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/di/NetworkModule.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/di/NetworkModule.kt
@@ -22,6 +22,7 @@ import dagger.Module
 import dagger.Provides
 import okhttp3.ConnectionSpec
 import okhttp3.OkHttpClient
+import okhttp3.Protocol
 import okhttp3.logging.HttpLoggingInterceptor
 import org.matrix.android.sdk.BuildConfig
 import org.matrix.android.sdk.api.MatrixConfiguration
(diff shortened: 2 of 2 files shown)


bmarty:
@T-bond you down voted the PR what is your concern?

T-bond:
@bmarty ups. Sorry, I accidentally misclicked the reaction, and did not notice. I don't think we will loose anything forcing HTTP 1.1

BillCarsonFr:
FTR probably related to https://github.com/square/okhttp/issues/3278 and https://github.com/square/okhttp/issues/4455

bmarty:
FTR some internal feedback:

> based on my understanding of how okhttp manages connections, it's probably fine to use HTTP/1.1 instead of HTTP/2, in that I don't expect this to result in masses of TCP connections being created as I suspect okhttp does make use of persistent connections with existing connections in the connection pool. If I were you I would mock up a basic test jig to confirm this though: with a very simple HTTP server which holds open connections for 30s or so then just make a bunch of requests to it using your okhttp configuration and make sure you only see a max of N TCP connections in netstat or equiv, where N = the connection pool limit

> longer term, I would be concerned that okhttp has issues like this open for 3+ years. Your choice of HTTP library is utterly critical to the functioning of the app, so if there's big failure modes like "you can't send HTTP requests for N minutes whilst things timeout" then this is probably an indication that maybe you should be looking for an alternative

> https://github.com/google/volley looks to be google maintained

abaker:
There is some documentation for okhttp connections [here](https://square.github.io/okhttp/connections/).

Charles proxy lists a client connection number and server connection number for each request, and I can see multiple http/1.1 calls with matching connection numbers. I assume this means connections are being reused.

bmarty:
Thanks for your test @abaker 👍

PR #4960, opened 2022-01-17
Avoid deleting sent messages on non room events

ariskotsomitopoulos:
Following fix #4928 this is an improvement that prevents SENT messages to be deleted when there are **no** room events received

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/RoomSyncHandler.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/RoomSyncHandler.kt
@@ -514,6 +514,9 @@ internal class RoomSyncHandler @Inject constructor(private val readReceiptHandle
      * we clear all SENT events, and we are sure that we will receive it from /sync or pagination
      */
     private fun fixStuckLocalEcho(rooms: List<RoomEntity>) {
+        // when there are not room events, there is no need to delete SENT messages
+        // this might be useful for events like typing etc
+        if(rooms.isNullOrEmpty()) return
         rooms.forEach { roomEntity ->
             roomEntity.sendingTimelineEvents.filter {  timelineEvent ->
                 timelineEvent.root?.sendState ==  SendState.SENT


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/RoomSyncHandler.kt
```diff
@@ -514,6 +514,9 @@ internal class RoomSyncHandler @Inject constructor(private val readReceiptHandle
      * we clear all SENT events, and we are sure that we will receive it from /sync or pagination
      */
     private fun fixStuckLocalEcho(rooms: List<RoomEntity>) {
+        // when there are not room events, there is no need to delete SENT messages
+        // this might be useful for events like typing etc
+        if (rooms.isNullOrEmpty()) return
```
bmarty, line 519: If there are only typing events, I am not sure that `rooms` will be empty.
Also we can have the case where we have new Events from other users, so we can still have a blink. @ganfra WDYT?

ariskotsomitopoulos: I agree for the first part, I will update it a bit to check timeline events and not ephemeral etc. Now regarding the other we can fix this by removing them when we found an event from the same user! Do you think that will work?

bmarty: > removing them when we found an event from the same user

Yes, maybe. Would be nice to have @ganfra feedback as weel

ariskotsomitopoulos: Also keep in mind, that on iOS they delete it much earlier (on SENT), so I think we improved the experience by delaying it until those constraints

ganfra: I had flickering issues on friday sending reply and quote events (not sure the type is relevant).
Also, maybe just cleaning SENT events after a paused sync (look at SyncThread) would be enough?

ariskotsomitopoulos: I think, the flickering will mainly stop with that fix, while now we only delete local echos on specific cases. Maybe, pausing on syncThread will work, but not sure what will happened with that approach when there is no internet connection and the paused sync is called.

ariskotsomitopoulos: I would suggest to merge it while it will for sure eliminate a lot of blinking, and if we observe more, we follow a different approach

ganfra: OK

bmarty (approved):

PR #4962, opened 2022-01-17
Fix crash on API <24 and make sure this error will not occur again.

bmarty:
Fix crash on API <24 and make sure this error will not occur again.

### DIFF
--- a/library/core-utils/src/main/java/im/vector/lib/core/utils/compat/MutableCollectionCompat.kt
+++ b/library/core-utils/src/main/java/im/vector/lib/core/utils/compat/MutableCollectionCompat.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2022 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.lib.core.utils.compat
+
+import android.os.Build
+
+fun <E> MutableCollection<E>.removeIfCompat(predicate: (E) -> Boolean) {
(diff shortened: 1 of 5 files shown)


library/core-utils/src/main/java/im/vector/lib/core/utils/compat/MutableCollectionCompat.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package im.vector.lib.core.utils.compat
+
+import android.os.Build
+
+fun <E> MutableCollection<E>.removeIfCompat(predicate: (E) -> Boolean) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+        removeIf(predicate)
+    } else {
+        removeAll(filter(predicate).toSet())
```
bmarty, line 25: I have been asked to convert to Set for performance reason.

ouchadam (approved):

library/core-utils/src/main/java/im/vector/lib/core/utils/compat/MutableCollectionCompat.kt
ouchadam: as this is only for the _keys to remove_ there shouldn't be any changes for lists with duplicated items?

```kotlin
listOf(1, 1, 2).removeIfCompat { it == 2 }  // would still expect [1, 1] 
```

vector/lint.xml
```diff
@@ -40,6 +40,7 @@
     <issue id="RtlSymmetry" severity="error" />
 
     <!-- Code -->
+    <issue id="NewApi" severity="error" />
```
ouchadam, line 43: :100:

library/core-utils/src/main/java/im/vector/lib/core/utils/compat/MutableCollectionCompat.kt
bmarty: Now that you say it, I am not sure what will happen with
`listOf(1, 1, 2).removeIfCompat { it == 1 }  // would still expect [2] but I think we will get [1, 2]`
Let me play a bit with it.

bmarty: Seems to work as expected:
image
(https://pl.kotl.in/vIVLDMXe-)

ouchadam: thanks for checking!

bmarty:
Now included in #4971

PR #4963, opened 2022-01-17
Add case for local or non public homeserver.

bmarty:
Quick update of the issue template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -57,8 +57,9 @@ body:
     id: homeserver
     attributes:
       label: Homeserver
-      description: Which server is your account registered on?
-      placeholder: e.g. matrix.org
+      description: |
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -57,8 +57,9 @@ body:
     id: homeserver
     attributes:
       label: Homeserver
-      description: Which server is your account registered on?
-      placeholder: e.g. matrix.org
+      description: |
+        Which server is your account registered on? If it is a local or non public homeserver, please tell us what is the homeserver implementation (ex: Synapse/Dendrite/etc.) and its version.
```
kittykat, line 61: local or non-public homeserver… and the version

kittykat (approved):

.github/ISSUE_TEMPLATE/bug.yml
bmarty, line 61:
```suggestion
        Which server is your account registered on? If it is a local or non-public homeserver, please tell us what is the homeserver implementation (ex: Synapse/Dendrite/etc.) and the version.
```

bmarty: done

PR #4973, opened 2022-01-18
Fixing onboarding crash when signing in/up on non `matrix.org` server with SSO

### LINKED ISSUE #4969: App crashes on attempt to log in to SAML-only homeserver
### Steps to reproduce

1. Log out
2. On the start screen, tap "I already have an account"
3. Tap "Other"
4. In the "Address" field, type the address of a homeserver that supports only SAML/SSO logins, such as https://chat.mozilla.org

#### What did you expect?
The app should prompt to continue with a SAML/SSO login.

#### What happened instead?
The app crashes.

### Your phone model

Samsung Galaxy S8

### Operating system version

Android 9

1.3.16-dev

### Homeserver

chat.mozilla.org

### Will you send logs?

Yes

ouchadam:
Fixes #4969 

The crash occurs because the updated login flow attempts to forward the user to the sign in or sign up pages after fetching the homeserver login configuration, this is only correct for the `matrix.org` server.

- Only applies the sign in/up forwarding if the current server selection is `matrix.org`, for other server types we instead allow the `OnLoginFlowRetrieved` event to continue the flow (this is what the `LoginViewModel` [originally did](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/login/LoginViewModel.kt#L819) 

| OTHER BEFORE | OTHER AFTER |
| --- | --- |
|before-sso-crash|after-sso-fix

| EMS BEFORE | EMS AFTER |
| --- | --- |
|before-ems|after-ems

*matrix flow - no changes 

| SIGN UP MATRIX | SIGN IN MATRIX |
| --- | --- |
|get-started-matrix|i-already-have-account-matrix

### DIFF
--- a/changelog.d/4969.bugfix
+++ b/changelog.d/4969.bugfix
@@ -1,0 +1,1 @@
+Fixes sign in/up crash when selecting ems and other server types which use SSO
--- a/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
@@ -826,13 +826,17 @@ class OnboardingViewModel @AssistedInject constructor(
             }
 
             withState {
-                when (it.onboardingFlow) {
-                    OnboardingFlow.SignIn -> handleUpdateSignMode(OnboardingAction.UpdateSignMode(SignMode.SignIn))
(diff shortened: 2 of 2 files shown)


bmarty (approved):
LGTM, thanks. Will do some test directly from the hotfix branch.

bmarty:
fix tested OK from the branch hotfix 👍

PR #4975, opened 2022-01-18
Including onboarding auth options in sanity test 

ouchadam:
Adds crawling the `get started` and `I already have an account` onboarding flows to the sanity test

- Fixes the _other_ servers using the sign in + sign up screen instead of the forwarded sign up or sign in #4973
- Checks _other_ servers with live `chat.mozilla.org` and `matrix.org` in order to verify SSO only and multi option flows 

| GIF |
| --- |
after-sanity


bmarty (approved):
Perfect, thanks!

Issue #1497, opened 2020-06-17
Sending Stickers from Dimension doesn't work


Bubu:
Unfortunatelty riotX cannot send stickers when dimension is configured as an integration manager.
I just get this screen:
![image](https://user-images.githubusercontent.com/105185/84943153-3c5d4900-b0e4-11ea-9f06-da217d209cef.png)

Sending stickers from riot-web works fine.

jeffverk:
I notice that Riot web/desktop asks permission to identify me (with the option to remember yes/no) before proceeding to the stickers. RiotX does not do this and just gives the 403.

ganfra:
Any chance you could create a test accout on your homeserver so I can try with Dimension?

Bubu:
Sent privately

atomauro:
facing the same problem, anybody with a new idea? i already checked the credentials, i might that there is some flag that enables auth for all users from dimension server

Bubu:
For anyone wondering about this, I've since switched to https://github.com/maunium/stickerpicker for sending stickers. It works across all element version.

acheng-floyd:
When i debug element android , i found it's the same problem with "[Fix room_id check when adding user widgets](https://github.com/matrix-org/matrix-react-sdk/pull/7448)" post by bink on matrix-react-sdk issues. I'm trying to find whre the problem is 
![image](https://user-images.githubusercontent.com/62228888/148696020-9bc57b07-25f6-4186-a89c-c8f70c855352.png)

PR #4984, opened 2022-01-19
Fix Sending Stickers from Dimension doesn't work

acheng-floyd:
Fixes: Sending Stickers from Dimension doesn't work #1497

A saclar token is needed by self deployed dimension when using sticker picker.
Fix roomid check is no needed when using sticker picker

Noted: there still problem to enable sticker picker for the first time, if we want to use sticker picker, we should enable it on element-web then we can use sticker picker in android with self deployed dimension

Here's what your changelog entry will look like:

🐛 Bug Fixes
Fixes: Sending Stickers from Dimension doesn't work #1497. Contributed by @achengfloyd

Before fix:
image

After fix:
image


vector/src/main/java/im/vector/app/features/widgets/WidgetPostAPIHandler.kt
```diff
@@ -211,12 +212,23 @@ class WidgetPostAPIHandler @AssistedInject constructor(@Assisted private val roo
      * @param eventData the modular data
      */
     private fun getWidgets(widgetPostAPIMediator: WidgetPostAPIMediator, eventData: JsonDict) {
-        if (checkRoomId(widgetPostAPIMediator, eventData)) {
-            return
+        //if (checkRoomId(widgetPostAPIMediator, eventData)) {
```
ouchadam, line 215: if it's no longer needed lets delete the code rather than commenting out

vector/src/main/java/im/vector/app/features/widgets/WidgetPostAPIHandler.kt
```diff
@@ -211,12 +212,23 @@ class WidgetPostAPIHandler @AssistedInject constructor(@Assisted private val roo
...
+        //    return
+        //}
+        val roomIdInEvent = eventData["room_id"] as String?
```
ouchadam, line 218: I'm wondering if this could be simplified to 

```kotlin
val allWidgets = when {
    eventData.isMissingRoomId() -> session.widgetService().getUserWidgets()
    roomId == eventData.roomId  -> session.widgetService().getRoomWidgets(roomId) + session.widgetService().getUserWidgets()
    else                        -> {
        widgetPostAPIMediator.sendError(stringProvider.getString(R.string.widget_integration_room_not_visible), eventData)
        return
    }
}
```

acheng-floyd: That's looks nice, but i try your code in my ide, it looks like this
image

I'm sorry i'm really not familiar with kotlin, i just try to use the easy way i could understand~~

ouchadam: ah my bad! this was a mix of pseudo code, these could be implemented as kotlin extension functions on the `JsonDict`, we could also extract another for matching against the room id  

```kotlin
private fun getWidgets(widgetPostAPIMediator: WidgetPostAPIMediator, eventData: JsonDict) {
  val allWidgets = when {
      eventData.isMissingRoomId()        -> session.widgetService().getUserWidgets()
      eventData.hasMatchingRoom(roomId)  -> session.widgetService().getRoomWidgets(roomId) + session.widgetService().getUserWidgets()
      else                               -> {
          widgetPostAPIMediator.sendError(stringProvider.getString(R.string.widget_integration_room_not_visible), eventData)
          return
      }
  }
  ... 
}

private fun JsonDict.isMissingRoomId() = (this["room_id"] as? String).isNullOrEmpty()
private fun JsonDict.hasMatchingRoom(roomId: String) = (this["room_id"] as? String) == roomId
```

bmarty:
OK, I have made a small mistake when fixing the conflict, the code will not compile. Can you add a sign-off to the PR, so that I can merge it to one of my branch to be able to fix the issue? Thanks @acheng-floyd

bmarty:
Stale PR, feel free to update and re-open.

PR #4987, opened 2022-01-19
Try to reduce APK size for API29+ - WIP

bmarty:
Exclude ffmpeg lib for device  API 29+. The lib is is quite huge, and was introduced in August 2021, in version 1.1.15 with voice message support. See the impact on APK size on the graph from the PlayStore below:

image

With this change, we can see 2 different APKs generated depending on the new variant dimension `api`:

image

On the left, APK for API 21 to 28 and on the right, the APK for API 29+
We can see that the APK download size estimation is reduced from 57.2 to 44.9 for device running API 29+.

### Remaining work

Still have to check the versionCode. I think they MUST be all different when submitting to the PlayStore

Let's see what the CI is doing with this new flavor dimension.

Also I open the PR for comment, this is maybe not the best way to build various APK depending on API level. It can become a bit complicated if there is another feature which require the same type of work, but with a different API level split.


vector/src/api29/java/im/vector/app/features/voice/VoicePlayerHelper.kt
```diff
@@ -16,18 +16,12 @@
...
 
-class VoiceRecorderProvider @Inject constructor(
-        private val context: Context
-) {
-    fun provideVoiceRecorder(): VoiceRecorder {
-        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
-            VoiceRecorderQ(context)
-        } else {
-            VoiceRecorderL(context)
-        }
```
bmarty, line 31: git diff is wrong, this file is actually deleted.

ouchadam (approved):
this is a great action! :100: if we eventually move to bundles this will still be relevant as we can supply multiple `.aabs` to target different api levels

bmarty:
I close the PR, since we can live with that. Also it will require to update the release flow, we could do that later

PR #4995, opened 2022-01-19
Do not auto-retry on 429 in case of too long retry delay

Florian14:
Replace #4950 

Propagate the M_LIMIT_EXCEEDED error when the provided delay is too long (exceeds 32 seconds, eg. after three failed login attempts).

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
@@ -33,7 +33,8 @@ import java.io.IOException
  *
  * @param globalErrorReceiver will be use to notify error such as invalid token error. See [GlobalError]
  * @param canRetry if set to true, the request will be executed again in case of error, after a delay
- * @param maxDelayBeforeRetry the max delay to wait before a retry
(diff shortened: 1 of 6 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/Request.kt
```diff
@@ -74,23 +74,26 @@ internal suspend inline fun <DATA> executeRequest(globalErrorReceiver: GlobalErr
...
 
-            if (exception is Failure.ServerError &&
-                    exception.httpCode == 429 &&
-                    exception.error.code == MatrixError.M_LIMIT_EXCEEDED &&
-                    currentRetryCount < maxRetriesCount) {
+            if (exception.isLimitExceededError() && currentRetryCount < maxRetriesCount) {
                 // 429, we can retry
-                delay(exception.getRetryDelay(1_000))
+                val retryDelay = exception.getRetryDelay(1_000)
+                if (retryDelay <= maxDelayBeforeRetry) {
```
bmarty, line 80: Maybe update the documentation of `maxDelayBeforeRetry` to inform that it also has an effect on 429 management.

bmarty (approved):
LGTM, thanks for this new version. I like it!

bmarty (approved):
Thanks for the update!

PR #4999, opened 2022-01-19
Fix an error in string resource (#4997)

### LINKED ISSUE #4997: Missing plurals in localization
### Steps to reproduce

[poll_total_vote_count_before_ended_and_not_voted](https://translate.element.io/translate/element-android/element-app/de/?checksum=42583658a10ba344) has 3 plural forms (0, 1, >1) in the source string, but the translated strings have not. This is not exclusive to german.

grafik

### Outcome

-

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

bmarty:
Fixes #4997

### DIFF
--- a/changelog.d/4997.bugfix
+++ b/changelog.d/4997.bugfix
@@ -1,0 +1,1 @@
+Fix an error in string resource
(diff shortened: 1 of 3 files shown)


ouchadam:
I'm a little confused by the issue of using `zero`, does weblate not support unique plurals per locale?

onurays (approved):
Surprised and lgtm!

bmarty:
> I'm a little confused by the issue of using `zero`, does weblate not support unique plurals per locale?

Using `zero` for English language is not correct, see https://developer.android.com/guide/topics/resources/string-resource#Plurals:

>In English, a string for zero is ignored even if the quantity is 0, because 0 isn't grammatically different from 2, or any other number except 1 ("zero books", "one book", "two books", and so on)

On some other language, `zero` can have specific plural form. So If we want to special case `zero` in English (and so in all the languages), we have to do it manually like on this PR.

Other languages may need to define `zero` for the plural `poll_total_vote_count_before_ended_and_not_voted` to be correct regarding the plural form, but in the fact, this translation will never be used in that case.

ouchadam:
:100:  ahhh it's a an empty case rather than 0 specific plural

PR #5009, opened 2022-01-20
Storing and tracking the onboarding messaging use case 

ouchadam:
Part of #4585 

- Adds storing and tracking of the selected messaging use case as part of the onboarding journey (this will eventually be used to better tailor the first time experience)

- Adds `VectorAnalytics.updateUserProperties` to allow for updating the identity/user properties outside of the analytics id/consent flow, eg locally capturing the onboarding selection and only sending after the user has consented.`Posthog.Identify()` accumulates a local cache which is only sent once a user has consented `Posthog.OptOut(false)`


ouchadam:
marked as draft whilst confirming if the identity payload is coming through

bmarty (approved):
LGTM, some remarks

vector/src/main/java/im/vector/app/features/onboarding/FtueUseCase.kt
```diff
@@ -16,9 +16,24 @@
...
+    TEAMS("teams"),
+    COMMUNITIES("communities"),
+    SKIP("skip");
+
+    companion object {
+        fun from(persistedValue: String) = values().first { it.persistableValue == persistedValue }
+    }
+}
+
+fun FtueUseCase.toTrackingValue(): Identity.FtueUseCaseSelection {
```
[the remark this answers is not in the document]
ouchadam: will do the same (in a separate `FtueUseCaseExt.kt` file) :+1:

vector/src/main/java/im/vector/app/features/analytics/VectorAnalytics.kt
```diff
@@ -70,4 +71,9 @@ interface VectorAnalytics {
      * Track a displayed screen
      */
     fun screen(screen: VectorAnalyticsScreen)
+
+    /**
+     * Update user specific properties
+     */
+    fun updateUserProperties(identity: Identity)
```
ouchadam, line 78: @bmarty I'll also move this to the `AnalyticsTracker`

vector/src/main/java/im/vector/app/features/onboarding/store/OnboardingStore.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.core.stringPreferencesKey
+import androidx.datastore.preferences.preferencesDataStore
+import im.vector.app.features.onboarding.FtueUseCase
+import kotlinx.coroutines.flow.first
+import javax.inject.Inject
+
+private val Context.dataStore: DataStore<Preferences> by preferencesDataStore(name = "vector_onboarding")
```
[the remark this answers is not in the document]
ouchadam: assuming we're happy to avoid using a database for the time being~ I've updated the onboarding store to be _by user id_ https://github.com/vector-im/element-android/pull/5009/commits/efbdd7048715a3a2a943e4f99157ba46dc886fe8

vector/src/main/java/im/vector/app/features/onboarding/FtueUseCase.kt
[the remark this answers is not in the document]
ouchadam: https://github.com/vector-im/element-android/pull/5009/commits/7f751680fcbf921c218a032189b679a8df584882

vector/src/main/java/im/vector/app/features/analytics/VectorAnalytics.kt
ouchadam: https://github.com/vector-im/element-android/pull/5009/commits/928354312a69662963889a08f53eff893304a524

vector/src/main/java/im/vector/app/features/onboarding/store/OnboardingStore.kt
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Local storage for:
+ * - messaging use case (Enum/String)
+ */
+class OnboardingStore @Inject constructor(
+        private val context: Context
+) {
+    suspend fun readUseCase(userId: String) = context.dataStore.data.first().let { preferences ->
```
bmarty, line 39: Q: is it possible to have a OnBoardingStore scoped to the session to avoid passing the `userId` to each of its fun?

ouchadam: we could do it for the `MainActivity` injection site (for cleaning up the local storage) but not the `OnboardingViewModel` as the session doesn't exist yet

alternatively we could have a provider/factory as a session extension instead of injecting the dependency, what do you think?

```kotlin
Session.onboardingStore(context: Context) = OnboardingStore(context, myUserId)
```

bmarty: this looks neat!

vector/src/main/java/im/vector/app/features/onboarding/store/OnboardingStore.kt
```diff
@@ -0,0 +1,61 @@
...
+            settings.remove(myUserId.toUseCaseKey())
+        }
+    }
+
+    suspend fun clear() {
+        context.dataStore.removeKeysWithPrefix(myUserId)
+    }
+
+    private fun String.toUseCaseKey() = stringPreferencesKey("$this-use_case")
+}
```
bmarty, line 61: This class could be simplified with:
```kotlin
class OnboardingStore(
        private val context: Context,
        myUserId: String
) {
    private val useCasePreferences = stringPreferencesKey("$myUserId-use_case")

    suspend fun readUseCase() = context.dataStore.data.first().let { preferences ->
        preferences[useCasePreferences]?.let { FtueUseCase.from(it) }
    }

    suspend fun setUseCase(useCase: FtueUseCase) {
        context.dataStore.edit { settings ->
            settings[useCasePreferences] = useCase.persistableValue
        }
    }

    suspend fun resetUseCase() {
        context.dataStore.edit { settings ->
            settings.remove(useCasePreferences)
        }
    }

    suspend fun clear() {
        resetUseCase()
    }
}
```

and `removeKeysWithPrefix` is not necessary since we know have a new per userId store object.

I am also wondering if we could make it more generic, so its goal will be to store any data app side related to the session. I have in mind the storage of a client secret for https://github.com/matrix-org/sygnal/issues/70.

The class name could be something like `SessionDataStore` (open for better ideas!)

ouchadam: sounds good to me :+1:

ouchadam: went with `VectorSessionStore`
https://github.com/vector-im/element-android/pull/5009/commits/42f5ba8746c231f68ce7806bb7812a317edf133b

vector/src/main/java/im/vector/app/features/session/VectorSessionStore.kt
```diff
@@ -14,48 +14,46 @@
...
  * - messaging use case (Enum/String)
  */
-class OnboardingStore constructor(
+class VectorSessionStore constructor(
         private val context: Context,
-        private val myUserId: String
+        myUserId: String
 ) {
 
+    private val Context.dataStore: DataStore<Preferences> by preferencesDataStore(name = "vector_session_store_$myUserId")
```
bmarty, line 37: I am wondering if directly using the `userId` may lead to some trouble: special chars - low risk, length - higher risk. Maybe using a hash of it would be better.

ouchadam: very good point, I keep forgetting that user-id is actually a matrix id rather than a uuid, will change

bmarty: Yes. "Matrix Id" is not really a specific concept, but it is sometimes used for user ids. https://spec.matrix.org/v1.1/appendices/#identifier-grammar is an interesting reading on the subject

ouchadam: https://github.com/vector-im/element-android/pull/5009/commits/cbdeb54fdbd687561f043ad807d0620c8a722523

bmarty (approved):
LGTM, thanks for all the updates!

PR #5012, opened 2022-01-20
Missing fdroid notifications

### LINKED ISSUE #5003: Notifications are broken on 1.3.14 F-Droid
### Steps to reproduce

1. Enable Caffeine for testing purposes this prevents your phone sleeping.
2. Observe that **no notifications** come through when Element is closed and not running as a background app, even though the "Listening for notifications" persistent notification is present.
3. Launch Element and then press the home button to move it to the background (running as a background app), note that **no notifications** come through.
4. Finally bring Element to the foreground but do not have it focused on a chat that would generate a notification and you will finally have **notifications displayed**.

Summing up, currently for the F-Droid build notifications are broken in every aspect unless Element is in the foreground.

The problem was not occurring on the previous Element build from F-Droid (1.3.9).

#### What did you expect?
Notifications should be received no matter what state Element is in.

#### What happened instead?
No notifications are received unless the app is open and in the foreground.

### Your phone model

Asus Zenfone Selfie ZD551KL

### Operating system version

Android 7.1.2 (LineageOS 14.1)

Element version 1.3.14 olm version 3.2.10 from F-Droid

### Homeserver

matrix.org

### Your phone model

Motorola Moto E Dual SIM (2nd gen) XT1521

### Operating system version

Android 10 (LineageOS 17.1)

Element version 1.3.14 olm version 3.2.10 from F-Droid

### Homeserver

matrix.org

### Will you send logs?

Yes

ouchadam:
Fixes #5003 Fdroid notifications no longer working.

The workaround is to switch the background mode to `optimised for realtime` 

- We've had a small regression in the automatic background sync worker where the `periodic` flag was no longer being set and instead using the default value of `false`
- The named arguments should help combat this in the future!

https://github.com/vector-im/element-android/pull/4614/files#diff-0b0656b4f8d1f73f2a4d5844acec8bf35b732e315e5355800f74145c7fc82fa1L117

### DIFF
--- a/changelog.d/5003.bugfix
+++ b/changelog.d/5003.bugfix
@@ -1,0 +1,1 @@
+Fixing missing notifications in FDroid variants using `optimised for battery` background sync mode
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/job/SyncWorker.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/job/SyncWorker.kt
@@ -151,6 +151,7 @@ internal class SyncWorker(context: Context, workerParameters: WorkerParameters, 
                             sessionId = sessionId,
                             timeout = serverTimeoutInSeconds,
                             delay = delayInSeconds,
+                            periodic = true,
                             forceImmediate = forceImmediate
                     )
             )


bmarty (approved):
Good catch, thanks!

bmarty (commented):
I am not that sure that they can be side effect. I will read it again.

bmarty (changes requested):
(Not sure how to revert approval)

bmarty (approved):
Ok, I have done more check, I think this is fine!

PR #5019, opened 2022-01-21
Quick fixes on auto uisi RS

BillCarsonFr:
Quick fixes on auto rageshake lab feature:

-> Align name and format with web (snake case, use json)
-> Fix an issue with auto rageshake disabling itself on restart
-> Update the report server app name (so that it goes to a dedicated private repo)

Depends on https://github.com/matrix-org/matrix-ansible-private/pull/4247 (do not merge before)

### DIFF
--- a/vector/src/main/java/im/vector/app/AutoRageShaker.kt
+++ b/vector/src/main/java/im/vector/app/AutoRageShaker.kt
@@ -67,8 +67,8 @@ class AutoRageShaker @Inject constructor(
 
     fun initialize() {
         observeActiveSession()
-        // It's a singleton...
         enable(vectorPreferences.labsAutoReportUISI())
+        // It's a singleton...
         vectorPreferences.subscribeToChanges(this)
 
         // Simple rate limit, notice that order is not
@@ -142,7 +142,7 @@ class AutoRageShaker @Inject constructor(
                 withCrashLogs = true,
                 withKeyRequestHistory = true,
                 withScreenshot = false,
-                theBugDescription = "Auto-reporting decryption error (recipient)",
+                theBugDescription = "Auto-reporting decryption error",
                 serverVersion = "",
                 canContact = false,
                 customFields = mapOf("auto_uisi" to buildString {
--- a/vector/src/main/java/im/vector/app/AutoRageShaker.kt
+++ b/vector/src/main/java/im/vector/app/AutoRageShaker.kt
@@ -68,6 +68,7 @@ class AutoRageShaker @Inject constructor(
     fun initialize() {
         observeActiveSession()
         // It's a singleton...
+        enable(vectorPreferences.labsAutoReportUISI())
         vectorPreferences.subscribeToChanges(this)
 
         // Simple rate limit, notice that order is not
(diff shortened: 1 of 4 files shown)


vector/src/main/java/im/vector/app/AutoRageShaker.kt
```diff
@@ -68,6 +68,7 @@ class AutoRageShaker @Inject constructor(
     fun initialize() {
         observeActiveSession()
         // It's a singleton...
+        enable(vectorPreferences.labsAutoReportUISI())
```
bmarty, line 71: Please do not insert new code between a line of code and its comment :)

vector/src/main/java/im/vector/app/features/rageshake/BugReporter.kt
```diff
@@ -259,13 +259,23 @@ class BugReporter @Inject constructor(
...
+                                when (reportType) {
+                                    ReportType.AUTO_UISI_SENDER,
+                                    ReportType.AUTO_UISI           -> {
+                                        addFormDataPart("app", "element-auto-uisi")
+                                    }
+                                    else -> {
+                                        addFormDataPart("app", "riot-android")
+                                    }
+                                }
+                            }
```
bmarty, line 278: OK... You already expect me to say something about that, no? :)

bmarty (commented):
Nice, thanks.
We wait to merge this PR.

vector/src/main/java/im/vector/app/AutoRageShaker.kt
```diff
@@ -221,17 +224,19 @@ class AutoRageShaker @Inject constructor(
                 withCrashLogs = true,
                 withKeyRequestHistory = true,
                 withScreenshot = false,
-                theBugDescription = "UISI detected $matchingIssue",
+                theBugDescription = "Auto-reporting decryption error \nRecipient rageshake: $matchingIssue",
```
duxovni, line 227: If we want this to be totally consistent with element-web, it'd be
```suggestion
                theBugDescription = "Auto-reporting decryption error (sender)\nRecipient rageshake: $matchingIssue",
```
but that's not a huge deal, and we have the labels now in any case so maybe element-web doesn't need this either anymore.

duxovni (changes requested):
Otherwise LGTM in terms of consistency with element-web.

vector/src/main/java/im/vector/app/features/rageshake/BugReporter.kt
bmarty: I mean having something like`.addFormDataPart("app", getRageshakeApp(reportType))` would be nicer

BillCarsonFr: you read my mind, or did I read yours?

vector/src/main/java/im/vector/app/AutoRageShaker.kt
BillCarsonFr: updated

vector/src/main/java/im/vector/app/features/rageshake/BugReporter.kt
BillCarsonFr: done

vector/src/main/java/im/vector/app/AutoRageShaker.kt
BillCarsonFr: done

bmarty (approved):
Thanks for the update!

PR #5023, opened 2022-01-21
Refresh Tokens

### LINKED ISSUE #4943: Implement refresh tokens
Add support for refresh token:
https://github.com/matrix-org/matrix-doc/blob/main/proposals/2918-refreshtokens.md
https://github.com/matrix-org/synapse/pull/11427

Implementation Highlights:
- We add the `refresh_token` parameter to login/register requests to receive a `refresh_token`(used to request new access tokens) and an `access_token` in addition to a `expires_in_ms` interval (to estimate client side when the token expires).
- We should handle the cases where `refresh_token` is empty (server may not support refresh tokens or not have it enabled),  expires_in_ms is empty ( the access token does not expire).
- There is a `/refresh` endpoint where the `refresh_token` can be provided to gain a new access token.
- Refresh tokens rotate, each time we call the `refresh` endpoint we also get a new `refresh_token`. Therefore we must be careful around the time of each refresh that if we have multiple concurrent requests we don't make multiple refresh requests or mishandle the new `refresh_token`.
- We can preemptively refresh the token shortly before it expires to minimise any disruption to the user(waiting on the refresh request before a user request can be made). On iOS the server timeout we set on `sync` requests is 30s, so sometime like 60s seems reasonable for preemptive expiry.

Rollout plan:
1. Distribute an alpha build or maybe add to dev feature flags.
2. Enable for all beta users and monitor for a period of time.
3. Ship to prod and monitor.

langleyd:
Resolves #4943 

### What's in the PR
- Hardcodes refresh_tokens on for the moment
- Refresh tokens requested during login
- Refresh tokens and access token expiry parsed in login response
- Access Token provider synchronises refresh so that we have a leader refresh requests while other requests are blocked and received the new access token when the leader succeeds.
- Adds refresh task for unauthenticated refresh api.
- Added interceptor logic may be trigger a refresh before a request if it is deemed the toke is no longer valid(e.g. expired). But also triggers a refresh on the a 401 `UNKOWN_TOKEN` from the server(the source of truth on token validity).
- Handles cases of both access token auth/refresh token auth becoming unauthenticated and logs to analytics for monitoring.

#TODO
- [x] Verify we don't need to handle multiprocess
- [x] Add config to enable/disable refresh tokens and test both enabled/disabled
- [x] Test sequential login/logouts
- [x] That soft/hard logouts work and that clients gracefully logs out when either refresh token or access token based auth become unauthenticated.
- [x] Add analytics event for [UnauthenticatedError](https://github.com/matrix-org/matrix-analytics-events/blob/main/types/kotlin/UnauthenticatedError.kt)
- [x] Open issue for separate bug found: https://github.com/vector-im/element-android/issues/5192


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/AuthenticationService.kt
```diff
@@ -63,6 +64,13 @@ interface AuthenticationService {
      */
     fun getRegistrationWizard(): RegistrationWizard
 
+    /**
+     * Return a RefreshWizard, refresh our authentication with the homeserver.
+     *
+     * See [RefreshWizard] for more details.
+     */
+    fun getRefreshWizard(sessionId: String): RefreshWizard
```
bmarty, line 72: I do not think that we need to introduce a `RefreshWizard` in the API that the SDK exposes to the application.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/data/Credentials.kt
```diff
@@ -36,6 +36,14 @@ data class Credentials(
          * An access token for the account. This access token can then be used to authorize other requests.
          */
         @Json(name = "access_token") val accessToken: String,
+        /**
+         * The interval in milliseconds that the access token will expire in.
+         */
+        @Json(name = "expires_in_ms") val expiresInMs: Long?,
+        /**
+         * The interval in milliseconds that the access token will expire in.
```
bmarty, line 44: Update the comment?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/login/LoginWizard.kt
```diff
@@ -54,7 +54,9 @@ interface LoginWizard {
 
     /**
      * Login to the homeserver by sending a custom JsonDict.
-     * The data should contain at least one entry "type" with a String value.
+     * The data should contain:
+     * - at least one entry "type" with a String value.
+     * - a "refresh_token" param to indicate if refresh tokens should be used.
```
bmarty, line 59: I think the param (boolean) `refresh_token`, should not be provided here, but added later by the SDK since this is a feature of the SDK to support token refreshing.
So I would not change this doc.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/refresh/RefreshWizard.kt
```diff
@@ -0,0 +1,34 @@
...
+ */
+interface RefreshWizard {
+    /**
+     * Refresh out access token.
+     *
+     * @param refreshToken the refresh token to be used to get a new access token. They rotate so should be used only once, the result contains the next one.
+     * @return a [RefreshResult] if the refresh is successful. Contains the new access token, the refresh token and any expiry on the access token.
+     */
+    suspend fun refresh(refreshToken: String): RefreshResult
+}
```
bmarty, line 34: Not needed I think.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/AuthAPI.kt
```diff
@@ -125,6 +127,14 @@ internal interface AuthAPI {
     @POST(NetworkConstants.URI_API_PREFIX_PATH_R0 + "login")
     suspend fun login(@Body loginParams: JsonDict): Credentials
 
+    /**
+     * Refresh the access token given a refresh token.
+     * @param refreshParams the refresh token used to
+     */
+    @Headers("CONNECT_TIMEOUT:60000", "READ_TIMEOUT:60000", "WRITE_TIMEOUT:60000")
+    @POST(NetworkConstants.URI_API_PREFIX_PATH_V1 + "refresh")
+    suspend fun refresh(@Body refreshParams: RefreshParams): RefreshResult
```
bmarty, line 136: I would call this method `refreshToken` for clarity.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/data/PasswordLoginParams.kt
```diff
@@ -30,6 +30,7 @@ internal data class PasswordLoginParams(
         @Json(name = "identifier") val identifier: Map<String, String>,
         @Json(name = "password") val password: String,
         @Json(name = "type") override val type: String,
+        @Json(name = "refresh_token") override val refreshToken: Boolean,
```
bmarty, line 33: I would rename the `val` `withRefreshTokenSupport` for clarity. We may find a better name.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/RetrofitExtensions.kt
```diff
@@ -91,10 +89,11 @@ private fun toFailure(errorBody: ResponseBody?, httpCode: Int, globalErrorReceiv
...
                 }
-                httpCode == HttpURLConnection.HTTP_UNAUTHORIZED && /* 401 */
-                        matrixError.code == MatrixError.M_UNKNOWN_TOKEN                                        -> {
-                    globalErrorReceiver?.handleGlobalError(GlobalError.InvalidToken(matrixError.isSoftLogout.orFalse()))
-                }
+// If using refresh tokens we won't want this as 401 will just mean refresh your token.
+//                httpCode == HttpURLConnection.HTTP_UNAUTHORIZED && /* 401 */
+//                        matrixError.code == MatrixError.M_UNKNOWN_TOKEN                                        -> {
+//                    globalErrorReceiver?.handleGlobalError(GlobalError.InvalidToken(matrixError.isSoftLogout.orFalse()))
+//                }
```
bmarty, line 96: How softLogout is managed now?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/parsing/CredentialsResponseAdapter.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+class CredentialsResponseAdapter {
+    @FromJson
+    fun credentialsFromResponse(responseCredentials: Credentials): Credentials {
+        if (responseCredentials.expiresInMs == null || responseCredentials.expiryTs != null) {
+            // expiryTs already estimated or doesn't apply return existing credential
+            return responseCredentials
+        }
+        // We have received a credential response from the server estimate the expiry datetime.
+        return responseCredentials.copy(expiryTs = System.currentTimeMillis() + responseCredentials.expiresInMs)
```
bmarty, line 34: `System.currentTimeMillis()` is not ideal to manage datetime for this use case I think, since the phone date can be changed by the user. I have no better idea though, will think about it.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
```diff
@@ -16,13 +16,99 @@
...
+        private val mutex = Mutex()
+    }
+
+    object Constants {
+        /**
+        The time interval before the access token expires that we will start trying to refresh the token.
+        This avoids us having to block other users requests while the token refreshes.
+        Choosing a value larger than DEFAULT_DELAY_MILLIS + DEFAULT_LONG_POOL_TIMEOUT_SECONDS guarantees we will at least have attempted it before expiry.
+         */
+        const val PREEMPT_REFRESH_EXPIRATION_INTERVAL = 60000
```
bmarty, line 52: could be declared in the `companion object` above (no need for a new `Constants` object.)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
```diff
@@ -16,13 +16,99 @@
...
+    }
+
+    override fun getToken(): String? {
+        var accessToken: String?
+        // We synchronise in a blocking fashion here so that when refresh is required, a single request becomes the leader.
+        // On successful refresh the leader saves the credential and the new access token then becomes available to other requests when they are unblocked.
+        // We should never send multiple refresh requests as refresh tokens are single-use(they rotate with a new one returned in the response).
+        // Mishandled via race conditions and we could become unauthenticated.
+        runBlocking {
```
bmarty, line 61: instead of using `runBlocking`, the fun `getToken` should be declared with `suspend`.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
```diff
@@ -16,13 +16,99 @@
...
+import org.matrix.android.sdk.internal.network.toFailure
+import retrofit2.HttpException
+import timber.log.Timber
 import javax.inject.Inject
 
 internal class HomeserverAccessTokenProvider @Inject constructor(
         @SessionId private val sessionId: String,
-        private val sessionParamsStore: SessionParamsStore
+        private val sessionParamsStore: SessionParamsStore,
+        private val authenticationService: AuthenticationService,
```
bmarty, line 38: Do not inject services here. Instead you should create a `RefreshTokenTask`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
```diff
@@ -16,13 +16,99 @@
...
+                    .getRefreshWizard(sessionId)
+                    .refresh(credentials.refreshToken)
+        } catch (throwable: Throwable) {
+            val serverError = throwable
+                    .let { it as? HttpException }
+                    ?.let { it.toFailure(globalErrorReceiver) as? Failure.ServerError }
+            if (serverError != null) {
+                Timber.d("## Token Refresh: Failed to refresh access token. error: $serverError")
+                if (serverError.error.code == M_UNKNOWN_TOKEN || serverError.error.code == M_FORBIDDEN) {
+                    Timber.d("## Token Refresh: Requires logout.")
```
bmarty, line 92: throw the `twowable` again to propagate?

bmarty (changes requested):
Nice work. I've made a few remarks.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/login/LoginWizard.kt
langleyd: makes sense

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/data/PasswordLoginParams.kt
langleyd: `enableRefreshTokenAuth`?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/RetrofitExtensions.kt
langleyd: If refreshTokenAuth is disable this would continue to function as it was before. If it is enable this error not just indicates we need to refresh. If this error gets returned from the refresh endpoint *then* respect it as before and look for the `soft_logout` as before. It's details in the [spec](https://github.com/matrix-org/matrix-doc/blob/main/proposals/2918-refreshtokens.md) linked in the issue.


> If a client uses an access token that has expired, the server will respond with an M_UNKNOWN_TOKEN error, preferably with the soft_logout parameter set to true to improve the user experience in legacy clients. Thus, if a client receives an M_UNKNOWN_TOKEN error, and it has a refresh token available, it should no longer assume that it has been logged out, and instead attempt to refresh the token. If the client was in fact logged out, then the server will respond with an M_UNKNOWN_TOKEN error to the token refresh request, possibly with the soft_logout parameter set.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/parsing/CredentialsResponseAdapter.kt
langleyd: Yea, the accuracy of the date isn't too important as we are just adding the expiry to whatever the user has set. So it will still expire at the time + `expiresInMs`. If the user does change the time forward in to the future and return to the app, it will likely just cause a refresh. Which is probably what happens most times you use the app anyway. So not sure there it's all that bad?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
langleyd: Would we then just have to do a `runBlocking` in AccessTokenIntercept?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
langleyd: 👍 Yea haven't really worked out exactly how I'm gonna handle some of the error cases here. Will be working more on this.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/parsing/CredentialsResponseAdapter.kt
langleyd: The implementation is now complete and matches what we do on iOS. `expiresInMs` is used to preemptively estimate when the token needs to be updated, but ultimately we reply on the server returning `M_UNKNOWN_TOKEN` as the source of truth as to when we need to refresh. This translates to a [check before each request](https://github.com/vector-im/element-android/pull/5023/files#diff-47d9f959e736383577e4c16f04856294d5888805287086720b6d3dfbf7087261R37) to see if we need to preemptively refresh the token based on our estimate. But if the server returns a `M_UNKNOWN_TOKEN` for a request we also refresh the token and retry the original request [here](https://github.com/vector-im/element-android/pull/5023/files#diff-47d9f959e736383577e4c16f04856294d5888805287086720b6d3dfbf7087261R46).


So if the user moves:
 - moves the time into the future, it will cause a token refresh when we open the app, after which we will create a new estimate based on the response and the new time.
 - moves the time into the past, the expiry estimate will be incorrect, when the token expires the server will respond with an `M_UNKNOWN_TOKEN`, we will refresh and create a new estimate based on the response and the new time.

Let me know what you think.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
langleyd: Moved it to the `AccessTokenInterceptor`.  Let me know what you think?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
langleyd: Error handling added.

vector/src/main/java/im/vector/app/features/signout/soft/SoftLogoutActivity.kt
```diff
@@ -73,7 +73,10 @@ class SoftLogoutActivity : LoginActivity() {
                 )
             }
             is SoftLogoutViewEvents.ClearData        -> {
-                MainActivity.restartApp(this, MainActivityArgs(clearCredentials = true))
+                MainActivity.restartApp(this, MainActivityArgs(
+                        clearCredentials = true,
+                        isUserLoggedOut = true
```
langleyd, line 78: I think from SoftLogout if we are clearing the data we should specify that the user is now logged out, as the text in softlogout view suggests will happen:
> Clear it if you’re finished using this device, or want to sign in to another account.

 Otherwise we try and send the logout request which will always fail(as we no longer have a valid access token.). I have seen the app get into a weird state where an alert keeps prompting us if we would like to retry the logout request and it becomes difficult to get out of the app.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/parsing/CredentialsResponseAdapter.kt
bmarty: I think it is OK, thanks!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DefaultLoginWizard.kt
```diff
@@ -62,14 +64,16 @@ internal class DefaultLoginWizard(
                     address = login,
                     password = password,
                     deviceDisplayName = initialDeviceName,
-                    deviceId = deviceId
+                    deviceId = deviceId,
+                    enableRefreshTokenAuth = matrixConfiguration.enableRefreshTokenAuth
```
bmarty, line 68: I would not introduce `MatrixConfiguration.enableRefreshTokenAuth`, which is used as a global configuration of the SDK, but rather add a boolean to the API `login` (and probably some other API) so that the caller can choose to enable refresh token facility or not per session.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/failure/Extensions.kt
```diff
@@ -24,6 +24,10 @@ import org.matrix.android.sdk.internal.di.MoshiProvider
 import java.io.IOException
 import javax.net.ssl.HttpsURLConnection
 
+fun Failure.ServerError.isTokenUnknownError() =
```
bmarty, line 27: Can we keep the format `fun Throwable.isTokenUnknownError()` and check the type as per the other extensions in this file?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/data/RefreshResult.kt
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.api.auth.data
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class RefreshResult(
```
bmarty, line 23: Could be internal and moved to the internal package I think.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/parsing/CredentialsResponseAdapter.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+class CredentialsResponseAdapter {
+    @FromJson
+    fun credentialsFromResponse(responseCredentials: Credentials): Credentials {
+        if (responseCredentials.expiresInMs == null || responseCredentials.expiryTs != null) {
+            // expiryTs already estimated or doesn't apply, return existing credential.
+            return responseCredentials
+        }
+        // We have received a credential response from the server, estimate the expiry datetime.
+        return responseCredentials.copy(expiryTs = System.currentTimeMillis() + responseCredentials.expiresInMs)
```
bmarty, line 34: I understand reading this code again that `expiryTs` does not come from the server, do it?
This a bit weird to create a customized Json adapter to handle this.
Not sure how to implement it in another way, I will think more about it.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionModule.kt
```diff
@@ -285,6 +289,19 @@ internal abstract class SessionModule {
...
+        @SessionScope
+        fun providesAuthApi(
+                @Unauthenticated okHttpClient: Lazy<OkHttpClient>,
+                retrofitFactory: RetrofitFactory,
+                homeServerConnectionConfig: HomeServerConnectionConfig
+        ): AuthAPI {
+            val homeServerUrl = homeServerConnectionConfig.homeServerUriBase.toString()
+            return retrofitFactory.create(okHttpClient, homeServerUrl)
+                    .create(AuthAPI::class.java)
+        }
```
bmarty, line 304: I do not really like to add a provider for `AuthAPI`. It breaks a bit the arch of login/registration flow.

I would create a new `RefreshTokenAPI` interface, that can be created using an injected Retrofit instance (like [here](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/RoomModule.kt#L133) for instance)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/identity/IdentityAccessTokenProvider.kt
```diff
@@ -16,12 +16,14 @@
...
+import org.matrix.android.sdk.api.failure.Failure
 import org.matrix.android.sdk.internal.network.token.AccessTokenProvider
 import org.matrix.android.sdk.internal.session.identity.data.IdentityStore
 import javax.inject.Inject
 
 internal class IdentityAccessTokenProvider @Inject constructor(
         private val identityStore: IdentityStore
 ) : AccessTokenProvider {
-    override fun getToken() = identityStore.getIdentityData()?.token
+    override var supportsRefreshTokens = false
```
bmarty, line 27:
```suggestion
    override val supportsRefreshTokens = false
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/identity/IdentityModule.kt
```diff
@@ -48,10 +49,11 @@ internal abstract class IdentityModule {
         @SessionScope
         @AuthenticatedIdentity
         fun providesOkHttpClient(@UnauthenticatedWithCertificate okHttpClient: OkHttpClient,
-                                 @AuthenticatedIdentity accessTokenProvider: AccessTokenProvider): OkHttpClient {
+                                 @AuthenticatedIdentity accessTokenProvider: AccessTokenProvider,
+                                 globalErrorReceiver: GlobalErrorReceiver): OkHttpClient {
```
bmarty, line 53: Since this is for the identity server, I do not think we can use the GlobalErrorReceiver here. I would make it nullable in `addAccessTokenInterceptor`

vector/src/main/java/im/vector/app/features/signout/soft/SoftLogoutActivity.kt
bmarty: OK

bmarty (commented):
Some more comments, please let me know what you think.
Sorry this is a very sensitive subject, I want to be sure that there is no issue and the implementation is crystal clear.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DefaultLoginWizard.kt
langleyd: This also reminds me I never added `refresh_token` param to the registration request.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DefaultLoginWizard.kt
```diff
@@ -100,9 +100,7 @@ internal class DefaultLoginWizard(
 
     override suspend fun loginCustom(data: JsonDict): Session {
         val loginParams = data.toMutableMap()
-        if (matrixConfiguration.enableRefreshTokenAuth) {
-            loginParams["refresh_token"] = true
-        }
+        loginParams["refresh_token"] = enableRefreshTokenAuth
```
bmarty, line 103: This is fine, but not strictly equivalent to the previous version where the value was not set to the Json body when equal to `false`. Do you confirm?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionModule.kt
```diff
@@ -290,17 +290,17 @@ internal abstract class SessionModule {
...
         @Provides
+        @JvmStatic
         @SessionScope
-        fun providesAuthApi(
+        fun providesRefreshTokenAPI(
                 @Unauthenticated okHttpClient: Lazy<OkHttpClient>,
                 retrofitFactory: RetrofitFactory,
                 homeServerConnectionConfig: HomeServerConnectionConfig
-        ): AuthAPI {
+        ): RefreshTokenAPI {
```
bmarty, line 300: Can you have `retrofit: Retrofit` as the only parameter?

bmarty (changes requested):
Some more remarks.

@ganfra can you have a quick look on this PR please? especially the refresh token management itself.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DefaultLoginWizard.kt
langleyd: 👍 Yea I changed it to have that behaviour to be consistent with the other login api's where the parameter is always present.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionModule.kt
langleyd: The `okHttpClient` here is unauthenticated. Would I have to provide another unauthenticated version of `Retrofit`? is there any example of where we do that?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/data/RefreshResult.kt
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.internal.auth.data
+
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+data class RefreshResult(
```
bmarty, line 23: Could be `internal`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionModule.kt
bmarty: I see your (good) point about potential endless loop

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/AuthenticationService.kt
```diff
@@ -53,15 +53,19 @@ interface AuthenticationService {
      * Return a LoginWizard, to login to the homeserver. The login flow has to be retrieved first.
      *
      * See [LoginWizard] for more details
+     * @param enableRefreshTokenAuth whether to enable refresh token based auth (in beta, hence default is false for now)
+     * instead of long lived access tokens.
      */
-    fun getLoginWizard(): LoginWizard
+    fun getLoginWizard(enableRefreshTokenAuth: Boolean = false): LoginWizard
```
ganfra, line 59: Should this flag enabled on MatrixConfiguration instead? It would avoid passing this everywhere and directly be injected in the few places where it makes sense.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionModule.kt
ganfra: Yeah you can keep it like that it's ok

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/token/HomeserverAccessTokenProvider.kt
```diff
@@ -16,13 +16,107 @@
...
+import timber.log.Timber
 import javax.inject.Inject
 
 internal class HomeserverAccessTokenProvider @Inject constructor(
         @SessionId private val sessionId: String,
-        private val sessionParamsStore: SessionParamsStore
+        private val sessionParamsStore: SessionParamsStore,
+        private val refreshTokenTask: RefreshTokenTask,
+        private val globalErrorReceiver: GlobalErrorReceiver
 ) : AccessTokenProvider {
```
ganfra, line 40: Okhttp has an API to handle accessToken and refreshToken when needed, might be the opportunity to use it instead of a custom interceptor/provider ? 
See this article: https://www.lordcodes.com/articles/authorization-of-web-requests-for-okhttp-and-retrofit for example

ganfra (commented):
Some thoughts

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/AuthenticationService.kt
langleyd: I actually had it that way initially. @bmarty felt it would be better to be a more explicit part of the api on the `AuthenticationService` vs a config people might not know about. I don't really have a strong opinion on it. :)

bmarty:
Will have to be done again, if we want to support this using https://www.lordcodes.com/articles/authorization-of-web-requests-for-okhttp-and-retrofit.

I guess this is probably better to create another PR.

PR #5043, opened 2022-01-24
Track performance

bmarty:
Quick PR to add 3 events in the next release about sync perf.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/statistics/StatisticEvent.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/statistics/StatisticEvent.kt
@@ -23,15 +23,15 @@ sealed interface StatisticEvent {
     /**
      * Initial sync request and response downloading, not including parsing and storage of response
      */
-    data class InitialSyncRequest(val durationMs: Int) : StatisticEvent
+    data class InitialSyncRequest(val durationMs: Int, val nbOfRooms: Int) : StatisticEvent
 
     /**
      * Initial sync treatment: parsing and storage of response
      */
-    data class InitialSyncTreatment(val durationMs: Int) : StatisticEvent
+    data class InitialSyncTreatment(val durationMs: Int, val nbOfRooms: Int) : StatisticEvent
 
     /**
      * Incremental sync event
      */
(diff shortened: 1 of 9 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/statistics/StatisticEvent.kt
```diff
@@ -0,0 +1,37 @@
...
+sealed interface StatisticEvent {
+    /**
+     * Initial sync request and response downloading, not including parsing and storage of response
+     */
+    data class InitialSyncRequest(val durationMs: Int, val nbOfRooms: Int) : StatisticEvent
+
+    /**
+     * Initial sync treatment: parsing and storage of response
+     */
+    data class InitialSyncTreatment(val durationMs: Int, val nbOfRooms: Int) : StatisticEvent
```
bmarty, line 31: Reading it again, I think we could merge this data class with the previous one.

bmarty: https://github.com/vector-im/element-android/pull/5043/commits/cf86fe34d5abd46221e35ac663dbfcc73fea196c

pixlwave (approved):
Yep, this aligns with the tracking on iOS. Implementation looks neat from what I can understand too! 🙂

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncTask.kt
```diff
@@ -239,6 +255,45 @@ internal class DefaultSyncTask @Inject constructor(
...
 
+    /**
+     * Aggregator to send stat event.
+     */
+    class SyncStatisticsData(
+            val isInitSync: Boolean,
+            val isAfterPause: Boolean
+    ) {
+        val startTime = SystemClock.elapsedRealtime()
+        var requestInitSyncTime = startTime
```
ouchadam, line 266: not a blocker, could be nice to avoid mutating the `SyncStatisticsData` and instead create a concrete `StatisticEvent.*` for the different types of sync (with immutable timestamps), I can understand why it's done this way as we're accumulating the data

ouchadam (approved):

PR #5051, opened 2022-01-25
Remove some usage of MatrixCallback

bmarty:
I have reverted the change on the crypto part as we are not sure about the side effect.

The rest seems OK, after a quick smoke test.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
@@ -740,14 +740,22 @@ class TimelineViewModel @AssistedInject constructor(
     }
 
     private fun handleUndoReact(action: RoomDetailAction.UndoReaction) {
-        room.undoReaction(action.targetEventId, action.reaction)
+        viewModelScope.launch {
+            tryOrNull {
+                room.undoReaction(action.targetEventId, action.reaction)
+            }
+        }
     }
 
     private fun handleUpdateQuickReaction(action: RoomDetailAction.UpdateQuickReactAction) {
         if (action.add) {
             room.sendReaction(action.targetEventId, action.selectedReaction)
         } else {
-            room.undoReaction(action.targetEventId, action.selectedReaction)
+            viewModelScope.launch {
+                tryOrNull {
+                    room.undoReaction(action.targetEventId, action.selectedReaction)
+                }
+            }
         }
     }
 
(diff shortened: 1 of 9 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -1320,16 +1320,15 @@ internal class DefaultCryptoService @Inject constructor(
...
         cryptoCoroutineScope.launch(coroutineDispatchers.crypto) {
             Timber.tag(loggerTag.value).d("prepareToEncrypt() roomId:$roomId Check room members up to date")
             // Ensure to load all room members
             try {
                 loadRoomMembersTask.execute(LoadRoomMembersTask.Params(roomId))
             } catch (failure: Throwable) {
                 Timber.tag(loggerTag.value).e("prepareToEncrypt() : Failed to load room members")
-                callback.onFailure(failure)
-                return@launch
+                throw failure
```
ouchadam, line 1331: If I've understood correctly, throwing here shouldn't have any behaviour changes as it's the same as `awaitCallback`?

ouchadam (approved):
assuming there aren't more changes on the way!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
bmarty: I am not sure :/
I will do more test, or I should say I will do test 🙈

bmarty: The app crashes :/

bmarty: @ouchadam I have reverted this change as we (V and G) are not sure about the impact on the crypto code here.

ouchadam: thanks for checking! :+1:

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
```diff
@@ -740,14 +740,22 @@ class TimelineViewModel @AssistedInject constructor(
     }
 
     private fun handleUndoReact(action: RoomDetailAction.UndoReaction) {
-        room.undoReaction(action.targetEventId, action.reaction)
+        viewModelScope.launch {
+            tryOrNull {
+                room.undoReaction(action.targetEventId, action.reaction)
```
ouchadam, line 745: nit: could be handy to log the error case

ouchadam (approved):
no blockers from me, looks good! :+1:

PR #5062, opened 2022-01-25
Ensure mapbox legal mention are displayed on the map

bmarty:
We should not hide the legal mention at bottom left of the map.

|Before|After|
|-|-|
|image|image|

Note: The legal mention is correctly displayed in the timeline, in the message bottom sheet and in the location message detail.

### DIFF
--- a/vector/src/main/res/layout/fragment_location_sharing.xml
+++ b/vector/src/main/res/layout/fragment_location_sharing.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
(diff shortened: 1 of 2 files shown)


ouchadam (approved):

ouchadam:
great catch!

vector/src/main/res/layout/fragment_location_sharing.xml
```diff
@@ -7,7 +7,9 @@
     <im.vector.app.features.location.MapTilerMapView
```
ouchadam, line 7: not a blocker, something I noticed in the file layout - we could flatten the nested constraint layouts by using a `Group` with a background for the bottom bar

bmarty: Yes, Using a `Flow` is better. https://github.com/vector-im/element-android/pull/5062/commits/6107fb644d7e7810ff83da543700260c81fd5fce

onurays (approved):
Nice catch, thank you!

PR #5069, opened 2022-01-26
Allow ending polls if user created it

### LINKED ISSUE #5067: Poll cannot end in some unencrypted rooms
### Steps to reproduce

1. Create a poll in an unencrypted room (the problem do not exist in all the unencrypted rooms) 
2. Vote
3. End the poll

I observed this issue in two rooms, but in some others works. 

#### What did you expect?
The poll to be end
#### What happened instead?
The poll did not End

### Your phone model

_No response_

### Operating system version

_No response_

1.3.16

### Homeserver

_No response_

### Will you send logs?

No

onurays:
Fixes #5067

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
@@ -446,6 +446,17 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
                               isLocalEcho: Boolean) {
         val pollEventId = content.relatesTo?.eventId ?: return
 
+        val pollOwnerId = getPollEvent(roomId, pollEventId)?.root?.senderId
+        val isPollOwner = pollOwnerId == event.senderId
+
+        val powerLevelsHelper = stateEventDataSource.getStateEvent(roomId, EventType.STATE_ROOM_POWER_LEVELS, QueryStringValue.NoCondition)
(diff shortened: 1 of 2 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -470,10 +463,13 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
             return
         }
 
+        val pollOwnerId = getPollEvent(roomId, pollEventId)?.root?.senderId
+        val isPollOwner = pollOwnerId == sessionManager.getSessionComponent(sessionId)?.session()?.myUserId
```
bmarty, line 467: - You should not compare `pollOwnerId` with `myUserId` but rather with `event.senderId`.
- Also I would move those check at the beginning of the fun to avoid querying the database (`EventAnnotationsSummaryEntity` and `PollResponseAggregatedSummaryEntity`) for nothing.

bmarty (changes requested):
There is a bug I think, see my comment

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
onurays: Omg, done. Thanks for catching.

bmarty (approved):
Thanks for the fix!

bmarty:
Quick smoke test OK!

PR #5070, opened 2022-01-26
Test screens in the settings

### LINKED ISSUE #4709: Update sanity test
Sanity test has to be updated:
- Test the new legal screen, and update the help and about test (I forgot to do it in the related PR)
- Test analytics opt-in screen (from the setting) and ensure when auto-display is enabled it will test it correctly.

bmarty:
Fixes #4709.

No need for a changelog here I think.


ouchadam (approved):
:100:

PR #5071, opened 2022-01-26
Instruction for the next weblate sync

bmarty:
I miss that during the previous string cleanup in #5059 

@ouchadam do you agree with this plan?

### DIFF
--- a/vector/src/main/res/values/strings.xml
+++ b/vector/src/main/res/values/strings.xml
@@ -2516,11 +2516,13 @@
     <string name="ftue_auth_carousel_1_title">Own your conversations.</string>
     <string name="ftue_auth_carousel_2_title">You\'re in control.</string>
     <string name="ftue_auth_carousel_3_title">Secure messaging.</string>
+    <!-- TODO Remove -->
     <string name="ftue_auth_carousel_4_title">Cut the slack from teams.</string>
 
     <string name="ftue_auth_carousel_body_secure">Secure and independent communication that gives you the same level of privacy as a face-to-face conversation in your own home.</string>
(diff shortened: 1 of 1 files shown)


vector/src/main/res/values/strings.xml
```diff
@@ -2516,11 +2516,13 @@
     <string name="ftue_auth_carousel_1_title">Own your conversations.</string>
     <string name="ftue_auth_carousel_2_title">You\'re in control.</string>
     <string name="ftue_auth_carousel_3_title">Secure messaging.</string>
+    <!-- TODO Remove -->
     <string name="ftue_auth_carousel_4_title">Cut the slack from teams.</string>
 
     <string name="ftue_auth_carousel_body_secure">Secure and independent communication that gives you the same level of privacy as a face-to-face conversation in your own home.</string>
     <string name="ftue_auth_carousel_body_control">Choose where your conversations are kept, giving you control and independence. Connected via Matrix.</string>
     <string name="ftue_auth_carousel_body_encrypted">End-to-end encrypted and no phone number required. No ads or datamining.</string>
+    <!-- TODO Rename to ftue_auth_carousel_4_title and move the string with the 3 other titles -->
```
ouchadam, line 2525: sounds good to me :+1:  I originally didn't reuse the numbered prefix as I realised (a bit too late...) that it was unnecessarily coupling us to the positions of the list but I'm happy for it to come back for consistency

ouchadam (approved):

vector/src/main/res/values/strings.xml
bmarty: The coupling with the actual position in the carroussel does not really matter from a translator POV.
I keep that in mind and will probably change also the body to have couples like
- `ftue_auth_carousel_secure_title`
- `ftue_auth_carousel_secure_body`

PR #5073, opened 2022-01-26
Fixing concurrent modification espresso errors 

ouchadam:
We're seeing concurrent modification exceptions within the sanity test suite when using the `waitForActivity` idling resource.

This is caused by `waitForActivity` not taking into account instantly idle activities eg already in a resumed state, combined with our activity matching being `Class` type based rather than instance it means we end up registering callbacks which are leaked and unexpectedly trigger throughout the rest of the test run (ultimately causing concurrent modification exceptions)

We haven't been seeing this with the current test suite due to few usages of `waitForActivity` and rarely revisiting screens after destroying them


bmarty (approved):
LGTM, thanks!

PR #5079, opened 2022-01-27
WIP and low priority: upgrade konfetti to 2.0.2

bmarty:
Update Konfetti to the latest version

Compilation warning is now fixed, probably due to the update of library Jitsi.

The rendering is not strictly equivalent, but still acceptable to me.

|Before|After|
|-|-|
|confetti_before|confetti_after|

We are using the "rain" party, but we could choose a different configuration, from the sample app:

image

Also reporting that on the celebration screen (after account creation), the confetti are displayed below the other views, which is a bit strange to me, and not equivalent to what we have in the timeline. This PR does not change that.

### DIFF
--- a/vector/build.gradle
+++ b/vector/build.gradle
@@ -446,7 +446,8 @@ dependencies {
     implementation 'me.leolin:ShortcutBadger:1.1.22@aar'
 
     // Chat effects
-    implementation 'nl.dionsegijn:konfetti:1.3.2'
+    implementation 'nl.dionsegijn:konfetti-xml:2.0.1'
+
     implementation 'com.github.jetradarmobile:android-snowfall:1.2.1'
     // DI
     implementation libs.dagger.hilt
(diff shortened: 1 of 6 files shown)


vector/build.gradle
```diff
@@ -451,7 +451,8 @@ dependencies {
     implementation 'me.leolin:ShortcutBadger:1.1.22@aar'
 
     // Chat effects
-    implementation 'nl.dionsegijn:konfetti:1.3.2'
+    implementation 'nl.dionsegijn:konfetti-xml:2.0.1'
```
bmarty, line 454:
```suggestion
    implementation 'nl.dionsegijn:konfetti-xml:2.0.2'
```

ouchadam:
onboarding confetti still works and looks better IMO! :+1: 

| |
| --- |
after-onboarding-confetti

ouchadam (approved):

bmarty:
(I have added a changelog, no need for a full CI build)

PR #5080, opened 2022-01-27
Updating encrypted room list message previews on key updates

### LINKED ISSUE #4867: [Rooms list] Last messages are stuck encrypted whereas the related keys have been received
### Steps to reproduce

1. Log in with an account that is logged in on another session. The last message of the most recent encrypted rooms is decrypted on this existing session
2. Verify the new session with the existing one
3. The rooms list is not updated whereas the keys are received, all the last messages are displayed encrypted. If we open one of the room the messages are decrypted and when we go back to the rooms list the last message is displayed correctly

### Outcome

This sounds like the rooms list is not updated whereas the room summaries have been updated

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

ouchadam:
Fixes #4867 - Message previews not decrypting upon receiving crypto keys

Follows the example of the timeline and adds a `session.cryptoService().addNewSessionListener` which refreshes the `RoomSummary.LatestPreviewContent` whenever we get a new key for the given room, the decryption will only be attempted if the current preview content doesn't contain an already decrypted json

<details>
<summary>
Timeline example
</summary>

https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/home/room/detail/RoomDetailViewModel.kt#L159
https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt#L121
https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/DefaultTimeline.kt#L235
https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/LoadTimelineStrategy.kt#L143
https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineEventDecryptor.kt#L67
</details>

- Also removes the `onNewSession` function from the base `IMXDecrypting` interface as it's only used for Megolm sessions and allows us to better ensure a room id is available 

*before shows a new message coming in at the end, not the original message being decrypted

| BEFORE | AFTER |
| --- | --- |
|before-verify|after-verify

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
@@ -242,4 +242,12 @@ interface RoomService {
      */
     fun getFlattenRoomSummaryChildrenOfLive(spaceId: String?,
                                             memberships: List<Membership> = Membership.activeMemberships()): LiveData<List<RoomSummary>>
+
+    /**
+     * Refreshes the RoomSummary LatestPreviewContent for the given @param roomId
+     * If the roomId is null, all rooms are updated
+     *
+     * This is useful for refreshing summary content with encrypted messages after receiving new room keys
+     */
+    fun refreshJoinedRoomSummaryPreviews(roomId: String?)
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/MegolmSessionDataImporter.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/MegolmSessionDataImporter.kt
@@ -22,6 +22,8 @@ import org.matrix.android.sdk.internal.crypto.MXOlmDevice
 import org.matrix.android.sdk.internal.crypto.MegolmSessionData
 import org.matrix.android.sdk.internal.crypto.OutgoingGossipingRequestManager
 import org.matrix.android.sdk.internal.crypto.RoomDecryptorProvider
+import org.matrix.android.sdk.internal.crypto.algorithms.megolm.MXMegolmDecryption
+import org.matrix.android.sdk.internal.crypto.algorithms.olm.MXOlmDecryption
 import org.matrix.android.sdk.internal.crypto.model.ImportRoomKeysResult
 import org.matrix.android.sdk.internal.crypto.model.rest.RoomKeyRequestBody
 import org.matrix.android.sdk.internal.crypto.store.IMXCryptoStore
@@ -76,7 +78,11 @@ internal class MegolmSessionDataImporter @Inject constructor(private val olmDevi
                     outgoingGossipingRequestManager.cancelRoomKeyRequest(roomKeyRequestBody)
 
                     // Have another go at decrypting events sent with this session
-                    decrypting.onNewSession(megolmSessionData.senderKey!!, sessionId!!)
+                    when (decrypting) {
+                        is MXMegolmDecryption -> {
+                            decrypting.onNewSession(megolmSessionData.roomId, megolmSessionData.senderKey!!, sessionId!!)
+                        }
+                    }
                 } catch (e: Exception) {
                     Timber.e(e, "## importRoomKeys() : onNewSession failed")
                 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/MegolmSessionDataImporter.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/MegolmSessionDataImporter.kt
@@ -23,7 +23,6 @@ import org.matrix.android.sdk.internal.crypto.MegolmSessionData
 import org.matrix.android.sdk.internal.crypto.OutgoingGossipingRequestManager
 import org.matrix.android.sdk.internal.crypto.RoomDecryptorProvider
 import org.matrix.android.sdk.internal.crypto.algorithms.megolm.MXMegolmDecryption
-import org.matrix.android.sdk.internal.crypto.algorithms.olm.MXOlmDecryption
 import org.matrix.android.sdk.internal.crypto.model.ImportRoomKeysResult
 import org.matrix.android.sdk.internal.crypto.model.rest.RoomKeyRequestBody
 import org.matrix.android.sdk.internal.crypto.store.IMXCryptoStore
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
@@ -318,7 +318,7 @@ internal class MXMegolmDecryption(private val userId: String,
 
             outgoingGossipingRequestManager.cancelRoomKeyRequest(content)
 
-            onNewSession(senderKey, roomKeyContent.sessionId)
+            onNewSession(roomKeyContent.roomId, senderKey, roomKeyContent.sessionId)
         }
     }
 
@@ -328,9 +328,9 @@ internal class MXMegolmDecryption(private val userId: String,
      * @param senderKey the session sender key
      * @param sessionId the session id
      */
-    override fun onNewSession(senderKey: String, sessionId: String) {
+    fun onNewSession(roomId: String?, senderKey: String, sessionId: String) {
         Timber.tag(loggerTag.value).v("ON NEW SESSION $sessionId - $senderKey")
-        newSessionListener?.onNewSession(null, senderKey, sessionId)
+        newSessionListener?.onNewSession(roomId, senderKey, sessionId)
     }
 
     override fun hasKeysForKeyRequest(request: IncomingRoomKeyRequest): Boolean {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
@@ -325,6 +325,7 @@ internal class MXMegolmDecryption(private val userId: String,
     /**
      * Check if the some messages can be decrypted with a new session
      *
+     * @param roomId the room id where the new Megolm session has been created for, may be null when importing from external sessions
      * @param senderKey the session sender key
      * @param sessionId the session id
      */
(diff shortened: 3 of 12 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/DefaultRoomService.kt
```diff
@@ -90,6 +94,23 @@ internal class DefaultRoomService @Inject constructor(
         return roomSummaryDataSource.getRoomSummaries(queryParams, sortOrder)
     }
 
+    override fun refreshJoinedRoomSummaryPreviews(roomId: String?) {
+        val roomSummaries = getRoomSummaries(roomSummaryQueryParams {
```
ouchadam, line 98: due to nullability in the deeper crypto layers the room id isn't always guaranteed, I wasn't sure if we should skip those cases or attempt to refresh the entire room list

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
```diff
@@ -229,4 +229,12 @@ interface RoomService {
...
     fun getFlattenRoomSummaryChildrenOfLive(spaceId: String?,
                                             memberships: List<Membership> = Membership.activeMemberships()): LiveData<List<RoomSummary>>
+
+    /**
+     * Refreshes the RoomSummary LatestPreviewContent for the given @param roomId
+     * If the roomId is null, all rooms are updated
+     *
+     * This is useful for refreshing summary content with encrypted messages after receiving new room keys
+     */
+    fun refreshJoinedRoomSummaryPreviews(roomId: String?)
```
bmarty, line 239: According to your video and following the implementation, the changes LGTM.
But I am wondering why we have to expose such API to the app. Can't this new decryption attempt be managed only SDK side? Since the SDK maintain the latestPreviewableEvent in the RoomSummary to me it would make more sense.

vector/src/main/java/im/vector/app/features/home/HomeDetailViewModel.kt
```diff
@@ -149,6 +157,7 @@ class HomeDetailViewModel @AssistedInject constructor(
     override fun onCleared() {
         super.onCleared()
         callManager.removeProtocolsCheckerListener(this)
+        session.cryptoService().removeSessionListener(refreshRoomSummariesOnCryptoSessionChange)
```
bmarty, line 160: This is the only change app side and it's clear that we listen for callback from the SDK to trigger a request to the SDK. So I think it confirms that it could be managed only by the SDK.

bmarty (commented):
Nice fix!
Some though about it.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
ouchadam: moving to the sdk also makes sense to me, the tricky part is where~

in the timeline it's managed by the SDK `DefaultTimeline` abstraction which acts as the datasource and holder for the listeners to allow the SDK to manage the crypto/pagination callbacks, from the client `ViewModel` we simply call `timeline.start()`
 
currently the room summaries is just a data model which clients observe using `RoomService.getRoomSummariesLive`, we could add a `start/stop` to the `RoomService` to internally register for crypto callbacks or create another abstraction on top of the summaries similar to `Timeline` :thinking:  what do you think?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/MegolmSessionDataImporter.kt
```diff
@@ -76,7 +77,11 @@ internal class MegolmSessionDataImporter @Inject constructor(private val olmDevi
                     outgoingGossipingRequestManager.cancelRoomKeyRequest(roomKeyRequestBody)
 
                     // Have another go at decrypting events sent with this session
-                    decrypting.onNewSession(megolmSessionData.senderKey!!, sessionId!!)
+                    when (decrypting) {
```
BillCarsonFr, line 80: For things like that I used the extension pattern. For example there is the IMXWithHeldExtension that is only supported by MXMegolmDecryption and not by MXOlmDecryption. So in code you can do 
```
if (alg is IMXWithHeldExtension) {
            alg.onRoomKeyWithHeldEvent(withHeldContent)
        }
```

Maybe we can use also this principles for new session? Or let me know if you think it's too overengineered

vector/src/main/java/im/vector/app/features/home/HomeDetailViewModel.kt
BillCarsonFr: Yeah I think it makes sense to keep it in the SDK too (my initial thought was to do it in app too, but it probably should be in the SDK).

I was thinking of letting the RoomSummary having an instance of TimelineEventDecryptor.
Then for every undecrypted lastMessage, we just request the TimelineEventDecryptor to decrypt it, it will then by itself listen to session and retry to decrypt.

Two points to check:
- When the TimelineEventDecryptor succesfully decrypt an event (and will update EventEntity), would this trigger an update to the LiveRoomSummaries?
- We have to encure that it's the same TimelineEventDecryptor that is used during the session lifetime (if not it won't remember all the event to decrypt). It should belong some how to the session

Does it make sense?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
```diff
@@ -327,9 +327,9 @@ internal class MXMegolmDecryption(private val userId: String,
      * @param senderKey the session sender key
      * @param sessionId the session id
      */
-    override fun onNewSession(senderKey: String, sessionId: String) {
+    fun onNewSession(roomId: String?, senderKey: String, sessionId: String) {
```
Florian14, line 330: Please, can you update the documentation about this new param? Also, it would be great if we could have some documentation about the NewSessionListener interface (which would probably be the same as this method) :)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/actions/MegolmSessionDataImporter.kt
ouchadam: I think that could be quite nice as we're not using any of the other functions from the `MXMegolmDecryption`

maybe a bit of a _hot take_ but I'm also wondering if we need the base interface? it seems that only `fun decryptEvent(event: Event, timeline: String): MXEventDecryptionResult` is used by both olm and megolm

when talking to the `roomDecryptorProvider.getOrCreateRoomDecryptor` we tend to already know which type of decrypter we need, in the case of `MegolmSessionDataImporter` I wouldn't expect to see any olm decrypters popping up

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
Florian14: Thanks for the explanation, the link between the megolm session and the room was unclear to me!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
ouchadam: using the `FlowSession` abstraction we can wrap and combine the room and crypto services 

```kotlin
fun liveRoomSummaries(queryParams: RoomSummaryQueryParams, sortOrder: RoomSortOrder = RoomSortOrder.NONE): Flow<List<RoomSummary>> {
    val refreshRoomSummariesOnCryptoSessionChange = object : NewSessionListener {
        override fun onNewSession(roomId: String?, senderKey: String, sessionId: String) {
            session.refreshJoinedRoomSummaryPreviews(roomId)
        }
    }

    return session.getRoomSummariesLive(queryParams, sortOrder).asFlow()
            .startWith(session.coroutineDispatchers.io) { session.getRoomSummaries(queryParams, sortOrder) }
            .onStart { session.cryptoService().addNewSessionListener(refreshRoomSummariesOnCryptoSessionChange) }
            .onCompletion { session.cryptoService().removeSessionListener(refreshRoomSummariesOnCryptoSessionChange) }
}
```

with the current keys being updated in background workers it makes it very difficult (and potentially easier to introduce memory leaks) to attach a key listener in the lower levels of the sync handler or dynamically spawning them when calling the room summary update

conceptually I like the idea of an abstraction on top of the Session api (similar to the `SessionFlow`) which would interact and coordinate with the separate services to provide a source of truth

BillCarsonFr (approved):
After rebase I am ok to merge

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
Florian14: Thanks for the docs update! It would be perfect if you could also add the new param in the doc of that function to be up to date with your changes

Florian14 (approved):
Just a small remark, SGTM otherwise 💯

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
ouchadam: ah my bad, I thought you meant the interface, will update this method as well :+1:

Florian14: My comment was for both, the interface and the current method which already had documentation :)

bmarty (approved):
LGTM, but still a bit disappointed that we have to do changes app side. It will not be easy for other application using the SDK to do the same thing.
Since it fixes the issue, I do not want to block merging this PR.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
ouchadam: updated both and provided context on why the value may be null https://github.com/vector-im/element-android/commit/57bf044df0092e9bd9306d17264cfe49e16e4495

ouchadam:
@bmarty created an SDK issue to track moving the logic https://github.com/vector-im/element-android/issues/5678

PR #5084, opened 2022-01-27
Fix location crash

bmarty:
When the user want to share their location and then disable the location on their device, there was a crash on Android < R.

Now the user get a dialog and can safely close the screen.

**Update**

I use the same PR to fix more implementation issue:
- Some lifecycle method was not called.
- Add a ProgressBar when the user wait to see his location. It can take several seconds on my device
- Move some logic to the ViewModel.
- pin was not rendered if the user has no avatar
- fix race condition between map ready / location update
- **NEW** Give priority to GPS provider over the other ones

There is still some problems:
- [x] a regression from this PR: In a timeline, it there are several location events, the pin is not displayed on each item. I am pretty sure it's a stupid bug, I will investigate later.
- [x] In the timeline we do not cleanup properly the maps, I think it can lead to bug due to bad releasing of resources.
- [x] Same problem in the bottomsheet (message detail)


onurays (approved):
Surprised, approved, thanks a lot for testing.

ouchadam (approved):

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/LocationPinProvider.kt
```diff
@@ -54,22 +55,36 @@ class LocationPinProvider @Inject constructor(
...
+                    // FIXME The doc says it has to be implemented and should free resources
+                    Timber.d("## Location: onLoadCleared")
+                }
+
+                override fun onLoadFailed(errorDrawable: Drawable?) {
+                    Timber.w("## Location: onLoadFailed")
+                    errorDrawable ?: return
+                    val pinDrawable = createPinDrawable(errorDrawable)
+                    cache[userId] = pinDrawable
+                    callback(pinDrawable)
```
bmarty, line 75: Fixes the issue when the user does not have an avatar

vector/src/main/java/im/vector/app/features/location/MapState.kt
```diff
@@ -18,15 +18,9 @@ package im.vector.app.features.location
 
 import android.graphics.drawable.Drawable
 
-interface VectorMapView {
```
bmarty, line 21: This interface was implemented but not really used as a type

vector/src/main/java/im/vector/app/features/location/MapTilerMapView.kt
```diff
@@ -28,63 +27,85 @@ import com.mapbox.mapboxsdk.plugins.annotation.SymbolManager
...
-                this.map = map
-                this.style = style
-                onMapReady()
+                mapRefs = MapRefs(
+                        map,
+                        SymbolManager(this, map, style),
+                        style
+                )
+                pendingState?.let { render(it) }
+                pendingState = null
```
bmarty, line 68: Trying to fix race condition here

vector/src/main/java/im/vector/app/features/location/MapTilerMapView.kt
```diff
@@ -28,63 +27,85 @@ import com.mapbox.mapboxsdk.plugins.annotation.SymbolManager
...
-                .zoom(zoom)
+                .zoom(INITIAL_MAP_ZOOM)
                 .build()
     }
 
-    override fun getCurrentZoom(): Double? {
-        return map?.cameraPosition?.zoom
-    }
-
-    override fun onClick(callback: () -> Unit) {
```
bmarty, line 81: this was not used

vector/src/main/res/layout/fragment_location_sharing.xml
```diff
@@ -48,4 +48,13 @@
...
         android:textStyle="bold" />
 
+    <ProgressBar
+        android:id="@+id/shareLocationGpsLoading"
+        android:layout_width="32dp"
+        android:layout_height="32dp"
+        android:layout_marginEnd="@dimen/layout_horizontal_margin"
+        app:layout_constraintBottom_toBottomOf="@id/shareLocationContainer"
+        app:layout_constraintEnd_toEndOf="@id/shareLocationContainer"
+        app:layout_constraintTop_toTopOf="@id/shareLocationContainer" />
```
bmarty, line 58: image

vector/src/main/java/im/vector/app/features/location/LocationTracker.kt
```diff
@@ -50,34 +50,28 @@ class LocationTracker @Inject constructor(
...
+                ?.forEach { provider ->
+                    Timber.d("## LocationTracker. track location using $provider")
+
+                    // Send last known location without waiting location updates
+                    locationManager.getLastKnownLocation(provider)?.let { lastKnownLocation ->
+                        Timber.d("## LocationTracker. lastKnownLocation")
+                        callback?.onLocationUpdate(lastKnownLocation.toLocationData())
+                    }
+
+                    locationManager.requestLocationUpdates(
```
bmarty, line 64: Is it OK to call this multiple times this `fun` and call only once `locationManager?.removeUpdates(this)` in `onStop`? Maybe yes but I want to be sure.

vector/src/main/java/im/vector/app/features/location/MapState.kt
onurays: Yes, I tried to be flexible in case of changing the map library but has problems with epoxy.

vector/src/main/java/im/vector/app/features/location/LocationTracker.kt
onurays: Yes, you can see the location icon on Status Bar is disappearing when fragment is destroyed.

bmarty: Ah didn't notice that. It does not disappear on my side

bmarty: (I have to put the entire app into bg to make it disappear to be more precise.

ouchadam (approved):
working well in my local test

DoM1niC:
for me the Marker is there but no Map Render in Timeline but the rest to seems work now :)

bmarty:
> for me the Marker is there but no Map Render in Timeline but the rest to seems work now :)

Thanks for your feedback!

Loading images in the timeline can take a few seconds but is working here.

vector/src/main/java/im/vector/app/features/location/UrlMapProvider.kt
```diff
@@ -0,0 +1,58 @@
...
+            append(zoom)
+            append("/")
+            append(width)
+            append("x")
+            append(height)
+            append(".png")
+            append(keyParam)
+            if (!resources.getBoolean(R.bool.is_rtl)) {
+                // On LTR languages we want the legal mentions to be displayed on the bottom left of the image
+                append("&attribution=bottomleft")
```
onurays, line 54: Nice catch.

onurays (approved):
Thank you for the great collaboration!

DoM1niC:
> can

Mhhh... I wait more then 30 Sec without success here 🥺https://user-images.githubusercontent.com/1436958/151638693-96a96509-34dd-4470-b334-5a014820ff02.mp4

bmarty:
@DoM1niC any chance that you put a breakpoint [here](https://github.com/vector-im/element-android/blob/b14e557c36deef390fa3973f34c8d3cf568ba75b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageLocationItem.kt#L47) and get the URL `location`, then try to open it from the browser of your device? The image should be displayed.

vector/src/test/java/im/vector/app/features/location/LocationDataTest.kt
```diff
@@ -0,0 +1,60 @@
...
+                LocationData(latitude = 12.34, longitude = 56.78, uncertainty = null)
+        // Space are ignored (trim)
+        parseGeo("geo: 12.34,56.78;13.56") shouldBeEqualTo
+                LocationData(latitude = 12.34, longitude = 56.78, uncertainty = 13.56)
+        parseGeo("geo:12.34,56.78; 13.56") shouldBeEqualTo
+                LocationData(latitude = 12.34, longitude = 56.78, uncertainty = 13.56)
+    }
+
+    @Test
+    fun invalidCases() {
```
onurays, line 43: Perfect, thanks for the test cases!

onurays:
> @DoM1niC any chance that you put a breakpoint [here](https://github.com/vector-im/element-android/blob/b14e557c36deef390fa3973f34c8d3cf568ba75b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageLocationItem.kt#L47) and get the URL `location`, then try to open it from the browser of your device? The image should be displayed.

[I can reproduce it](https://api.maptiler.com/maps/basic/static/29.0902101,40.9564908,17.0/915x550.png?key=fU3vlMsMn4Jb6dnEIFsx&attribution=bottomleft)

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsEpoxyController.kt
```diff
@@ -74,9 +77,11 @@ class MessageActionsEpoxyController @Inject constructor(
         val formattedDate = dateFormatter.format(date, DateFormatKind.MESSAGE_DETAIL)
         val body = state.messageBody.linkify(host.listener)
         val bindingOptions = spanUtils.getBindingOptions(body)
-        val locationData = state.timelineEvent()?.root?.getClearContent()?.toModel<MessageLocationContent>(catchError = true)?.let {
-            LocationData.create(it.getUri())
-        }
+        val locationUrl = state.timelineEvent()?.root?.getClearContent()
+                ?.toModel<MessageLocationContent>(catchError = true)
+                ?.toLocationData()
+                ?.let { urlMapProvider.buildStaticMapUrl(it, INITIAL_MAP_ZOOM_IN_TIMELINE, 1200, 800) }
```
ouchadam, line 83: nit: is it worth extracting these to width/height constants or using named params?

ouchadam (approved):

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsEpoxyController.kt
bmarty: Yes, and I think the values are not correct (since not related to the actual ImageView size). Anyway, could be improved later, this is not a big issue, Glide will crop the image anyway (I think)
Thanks for the review!

PR #5087, opened 2022-01-28
Add all screens about Spaces in allScreensTest

### LINKED ISSUE #4890: Add all screens about Spaces in allScreensTest
When Spaces support has been added to the project, lots of new screens have been added, but they are not yet covered by the  test https://github.com/vector-im/element-android/blob/main/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt#L55, which is run before each release.

The goal for this test is to cover every single screen of the app.

Ideally PR adding new screens to the app MUST also edit this test for the new screens to be covered, but it is often forgotten.

fedrunov:
Following screens were added to UiAllScreensSanityTest:

SpaceDirectoryFragment
SpaceListFragment
SpaceSettingsFragment
SpaceAddRoomFragment
SpaceLeaveAdvancedFragment
SpaceManageRoomsFragment
ChoosePrivateSpaceTypeFragment
ChooseSpaceTypeFragment
CreateSpaceDetailsFragment
CreateSpaceDefaultRoomsFragment
CreateSpaceAdd3pidInvitesFragment

fixes #4890


vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
```diff
@@ -95,6 +96,23 @@ class UiAllScreensSanityTest {
             }
         }
 
+        elementRobot.space {
```
ouchadam, line 99: for some more context about the robot pattern, it aims to create page models for each of the pages/screens/interactable components and return to a consistent state after usage

it allows for building and combining flows together like 

```
elementRobot.settings {
    advancedSettings { crawl() }
}

elementRobot.withDeveloperMode {
    settings {
        advancedSettings { crawlDeveloperOptions() }
    }
}
```
 
with that in mind there's some minor changes we could make to the spaces to make them easier to reuse

- parameterised space creation
- robot for the space menu (with ability to reopen itself)

```kotlin
elementRobot.space {
    val spaceName = UUID.randomUUID().toString()
    createSpace {
        crawl()
        createPublicSpace(spaceName)
    }
    spaceMenu(spaceName) {
        listMembers()
        settings()
        exploreRooms()

        // could be internal to the interactions that close the menu
        keepMenuOpen {
            invitePeople()
            addRoom()
            addSpace()
            leaveSpace()
        }
    }
}
```

what do you think?

bmarty: (maybe those useful information could be paste in the documentation of the project somewhere)

ouchadam: very good point :+1:  I'll update the UI tests doc

vector/src/androidTest/java/im/vector/app/ui/robot/SpaceRobot.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    fun invitePeople() {
+        clickOn(R.id.invitePeople)
+        waitUntilDialogVisible(withId(R.id.inviteByMxidButton))
+        clickOn(R.id.inviteByMxidButton)
+        waitUntilActivityVisible<InviteUsersToRoomActivity> {
+            waitUntilViewVisible(withId(R.id.userListRecyclerView))
+        }
+        EspressoHelper.getCurrentActivity()!!.finish()
```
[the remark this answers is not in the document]
fedrunov: in this case it will require to call it two times to close activity and also there will be the third one few lines below and it looks a little bit weird, isn't it better to just close activity explicitly, since what's what we want to achieve?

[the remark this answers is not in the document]
ouchadam: you're correct and both methods achieve the same result, the main difference is that calling finish directly is not possible for real users so there's the potential that our tests end up with different behaviour

I agree that adding all the _backs_ is a bit tedious and verbose but it does provide interesting UX commentary 

```
// close keyboard
pressBack()

// close popup
pressBack()

// exit screen
pressBack()
```

vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt
ouchadam: added https://github.com/vector-im/element-android/pull/5228

ouchadam (approved):
thanks for updating :100:

bmarty (approved):

Issue #5088, opened 2022-01-28
Fix call status timeline events mixed up(ended, missed and declined)


langleyd:
In some cases missed call/declined status messages and call ended(+ duration) status are displayed in place of each other.

Behaviour is slightly different with virtual rooms

bmarty:
In element 1.3.18

PR #5089, opened 2022-01-28
CallTileTimelineItem.CallStatus.MISSED renders both missed and reject…

langleyd:
Fixes https://github.com/vector-im/element-android/issues/5088

The if statement switches between `CallStatus.MISSED` and `CallStatus.ENDED`. `CallStatus.MISSED` renders both missed calls and not answered calls. So i think the logic makes more sense to check that the call was not answered vs the old code which checks all events are from one side.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/CallItemFactory.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/CallItemFactory.kt
@@ -101,7 +101,11 @@ class CallItemFactory @Inject constructor(
                 createCallTileTimelineItem(
                         roomSummary = roomSummary,
                         callId = callEventGrouper.callId,
-                        callStatus = if (!callEventGrouper.callWasAnswered()) CallTileTimelineItem.CallStatus.MISSED else CallTileTimelineItem.CallStatus.ENDED,
+                        callStatus = if (callEventGrouper.callWasAnswered()) {
+                            CallTileTimelineItem.CallStatus.ENDED
+                        } else {
+                            CallTileTimelineItem.CallStatus.MISSED
+                        },
                         callKind = callKind,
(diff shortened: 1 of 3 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/CallItemFactory.kt
```diff
@@ -101,7 +101,7 @@ class CallItemFactory @Inject constructor(
                 createCallTileTimelineItem(
                         roomSummary = roomSummary,
                         callId = callEventGrouper.callId,
-                        callStatus = if (callEventGrouper.callWasMissed()) CallTileTimelineItem.CallStatus.MISSED else CallTileTimelineItem.CallStatus.ENDED,
+                        callStatus = if (!callEventGrouper.callWasAnswered()) CallTileTimelineItem.CallStatus.MISSED else CallTileTimelineItem.CallStatus.ENDED,
```
bmarty, line 104: Can I ask you to remove `!` and swap if and else blocks to simplify reading?

bmarty (approved):
LGTM, once again, could be nice to have @ganfra's blessing

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/CallItemFactory.kt
ouchadam: was about to say the same thing :smile:

bmarty: :). also the CI complains that the line is too long

bmarty: Done

bmarty (approved):

PR #5091, opened 2022-01-28
Timeline recycling crash

ouchadam:
It looks like the originally intended fix from #4790 was accidentally merged out

The theory is `setText / text = message / setting text with emojis` is attempting to use a previously set `text future` from a recycled text view. This doesn't usually cause a problem but the emoji layer is particularly strict with futures.


bmarty (approved):
Thanks and sorry for the code vanishing :/

ouchadam:
it was definitely my bad :joy:

PR #5093, opened 2022-01-28
UI test clean ups

ouchadam:
Ui test cleanups 

- Fixes the missing `when` cases when compiling the UI tests with warnings enabled
- Replaces hardcoded timeout for the reaction picker with `waitUntil*`, some times the CI tests fail because the emojis aren't loaded in time
- Replaces `Barista.longClickOn` with the stock espresso long click which we're also using to open the reaction menu, the UI tests have been failing on the CI at the point of long clicking a previously added reaction
- Adds a wait for ensure the message reaction is present before attempting to long click it

### DIFF
--- a/vector/src/androidTest/java/im/vector/app/ui/robot/MessageMenuRobot.kt
+++ b/vector/src/androidTest/java/im/vector/app/ui/robot/MessageMenuRobot.kt
@@ -17,12 +17,16 @@
 package im.vector.app.ui.robot
 
 import androidx.test.espresso.Espresso.pressBack
+import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.espresso.matcher.ViewMatchers.withText
 import com.adevinta.android.barista.interaction.BaristaClickInteractions.clickOn
 import com.adevinta.android.barista.interaction.BaristaListInteractions.clickListItem
 import com.google.android.material.bottomsheet.BottomSheetBehavior
 import im.vector.app.R
+import im.vector.app.espresso.tools.waitUntilActivityVisible
+import im.vector.app.espresso.tools.waitUntilViewVisible
 import im.vector.app.features.home.room.detail.timeline.edithistory.ViewEditHistoryBottomSheet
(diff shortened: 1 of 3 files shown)


vector/src/androidTest/java/im/vector/app/ui/robot/MessageMenuRobot.kt
```diff
@@ -54,7 +58,10 @@ class MessageMenuRobot(
     fun addReactionFromEmojiPicker() {
         clickOn(R.string.message_add_reaction)
         // Wait for emoji to load, it's async now
-        sleep(2000)
+        waitUntilActivityVisible<EmojiReactionPickerActivity> {
+            waitUntilViewVisible(withId(R.id.emojiRecyclerView))
+            waitUntilViewVisible(withText("😀"))
+        }
```
bmarty, line 64: That's so much better, thanks!

vector/src/androidTest/java/im/vector/app/ui/robot/RoomDetailRobot.kt
```diff
@@ -103,7 +104,7 @@ class RoomDetailRobot {
 
     private fun longClickReaction(quickReaction: String) {
         withRetry {
-            longClickOn(quickReaction)
+            onView(withText(quickReaction)).perform(longClick())
```
bmarty, line 107: weird that it fixes something here... Does it mean that we should not use `longClickOn` anymore?

bmarty (approved):
Thanks for reducing the test flakiness!

vector/src/androidTest/java/im/vector/app/ui/robot/RoomDetailRobot.kt
ouchadam: I don't know if this will fix the CI flakyness~ but we weren't using `longClickOn` anywhere else :smile: and the `longClick()` was working well for the other interactions (open the message menu and room details)

Issue #5038, opened 2022-01-24
Missing push notifications (Google Play variant)


ouchadam:
Centralising #5028 #5016 #4862 #4393 

We're getting reports that the google play variant of the app is missing/skipping notifications, this also manifests as notifications not being marked as read when reading from other sessions

To help troubleshoot this

- What date and time did the sender send the missed message (with timezone information) - We'll cross reference with push server outages
- Are you using multiple clients, if so, have any of the notification settings been modified?
- Does the issue happen for all rooms/dms?
- Are you running a private homeserver, if so do you know the version? `{homeserver_url}/_synapse/admin/v1/server_version` eg https://matrix-client.matrix.org/_synapse/admin/v1/server_version

---
 
- Whilst having the issue, screenshot the list in Settings -> Advanced settings -> Push rules. We're looking to see if a rule has incorrectly started filtering out messages
- Does Element receive raw pushes? A raw `Push` counter can be displayed with  Advanced settings -> Enable `Developer mode` -> Enable `Show debug info on screen`.  

---

- Does switching room notifications to `none` and back to `all messages` fix the issue?

| PUSH RULES - 1| PUSH RULES - 2 |
| --- | --- |
![Screenshot_20220124_171938](https://user-images.githubusercontent.com/1848238/150832377-d8460824-3b17-4bee-8368-ab7fe72187e4.png)|![Screenshot_20220124_171945](https://user-images.githubusercontent.com/1848238/150832373-24de7d21-c586-4268-bf10-c6f77f413dfb.png)

| RAW PUSH COUNTER | ELEMENT WEB NOTIFICATION SETTINGS | ELEMENT WEB ROOM SETTINGS | 
| --- | --- | --- |
![2022-01-24T12:01:50,573509360+00:00](https://user-images.githubusercontent.com/1848238/150779630-df3bbdee-b428-4fff-9d4b-d9735a433de2.png)|![2022-01-24T16:21:50,592516708+00:00](https://user-images.githubusercontent.com/1848238/150822183-06a70acc-0d81-40a1-ae56-e39eaea42efd.png)|![2022-01-24T16:23:21,958254940+00:00](https://user-images.githubusercontent.com/1848238/150822512-06e1d039-89c1-4e79-860a-8ccb8d86ec8e.png)

Karamelmar:
Got the notifications running again on my F2 Pro fdroid version after switching "Background Sync" in Settings > Notifications from "Optimized for battery" to "Optimized for real time".
[...]

PR #5102, opened 2022-01-31
Increased push rules logging

ouchadam:
An attempt to help confirm the cause of #5038 

- Adds extra logging around the push rules and promotes some existing logs to `debug` from `verbose` to allow them to be included in rage shakes without having to enable verbose developer logs

output
```
[PushRules] Found 3 out of 3 to check for push rules with 17 rules
[PushRules] matched 3 out of 3
```


bmarty (approved):
LGTM, thanks!

Issue #3907, opened 2021-08-26
Textviews with a medium style don't apply sans-serif-medium font


yostyle:
### Steps to reproduce

1. In debug enable developer mode
2. Restart the app
3. Go to Debug view from home drawer
4. Click on "TEXT VIEWS LIGHT" button 

### What did you expect?

All texts in textviews with a medium style should have the sans-serif-medium font.

### What happened?

We have a text with sans-serif font if the style is medium type.

### Your phone model

Samsung Galaxy A50

### Operating system version

Android 10

### Have you submitted a rageshake?

No

bmarty:
Fixed in 1.3.18

PR #5106, opened 2022-01-31
Enabling FTUE Use Case

ouchadam:
#4585 Enables the FTUE Use case screen by default

| LIGHT | DARK | 
| --- | --- |
use-case-light|use-case-dark

| SMALL DEVICE TOP | SMALL DEVICE BOTTOM | TABLET PORTRAIT | TABLET LANDSCAPE |
| --- | --- | --- | --- |
Screenshot_20220202_184858|Screenshot_20220202_184903|Screenshot_20220202_184834|Screenshot_20220202_184847

### DIFF
--- a/vector/src/main/res/layout/fragment_ftue_auth_use_case.xml
+++ b/vector/src/main/res/layout/fragment_ftue_auth_use_case.xml
@@ -79,7 +79,7 @@
 
         <TextView
             android:id="@+id/useCaseOptionOne"
-            style="@style/Widget.Vector.TextView.Subtitle"
+            style="@style/Widget.Vector.TextView.Subtitle.Medium"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
             android:layout_marginBottom="16dp"
@@ -97,7 +97,7 @@
 
         <TextView
             android:id="@+id/useCaseOptionTwo"
-            style="@style/Widget.Vector.TextView.Subtitle"
+            style="@style/Widget.Vector.TextView.Subtitle.Medium"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
             android:layout_marginBottom="16dp"
@@ -115,7 +115,7 @@
 
         <TextView
             android:id="@+id/useCaseOptionThree"
-            style="@style/Widget.Vector.TextView.Subtitle"
+            style="@style/Widget.Vector.TextView.Subtitle.Medium"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
             android:layout_marginBottom="16dp"
@@ -133,10 +133,12 @@
 
         <TextView
             android:id="@+id/useCaseSkip"
+            style="@style/Widget.Vector.TextView.Body"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
             android:clickable="true"
             android:gravity="center"
+            android:textColor="?vctr_content_secondary"
             app:layout_constraintBottom_toTopOf="@id/contentFooterSpacing"
             app:layout_constraintEnd_toEndOf="@id/useCaseGutterEnd"
             app:layout_constraintStart_toStartOf="@id/useCaseGutterStart"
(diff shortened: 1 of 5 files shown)


ouchadam:
@bmarty updating the test caught https://github.com/matrix-org/matrix-analytics-events/pull/20

bmarty:
> @bmarty updating the test caught [matrix-org/matrix-analytics-events#20](https://github.com/matrix-org/matrix-analytics-events/pull/20)

I do not see the link between our sanity test and the analytics. Can you enlighten me please?

ouchadam:
@bmarty the tests were using the back flow and resetting the identity tracking with "ftueUseCase" = null which turned out to remove the property from the map which then resulted in a crash as the `identify` must be called with either a valid analytics id or non empty properties

amshakal:
Hello! It's looking good. Just a few small comments:

1. For dark mode, can we change the colour of 'Not sure yet? You can' and 'looking to join an exitsing server?' to A9B2BC? since it's not the most important information on the screen, id reserve our secondary grey colour for this.
2. Similarly for light mode, can we change the colour of the items mentioned above to 737D8C?
3. Can we semi bold the button options - Friends and family, team and communities?
4. For the outline of these three buttons, Friends and family, team and communities, can we change the outline stroke to 6F7882?

TY!

ouchadam:
@amshakal with fixes 

| BEFORE | AFTER |
| --- | --- |
|Screenshot_20220131_142327|after-light
Screenshot_20220131_142312|after-dark

We spoke internally and will keep the same option border colours for the time being as they match the existing server selection screen

kojid0:
@ouchadam 
I think you missed the third bullet point of her (semi bold the button options).
Also, I'd prefer to have "You can" not be clickable, only "skip this question".

vector/src/main/res/layout/fragment_ftue_auth_use_case.xml
```diff
@@ -79,7 +79,7 @@
 
         <TextView
             android:id="@+id/useCaseOptionOne"
-            style="@style/Widget.Vector.TextView.Subtitle"
+            style="@style/Widget.Vector.TextView.Subtitle.Medium"
```
bmarty, line 82: Note that #3907 is a thing and I have still not found a good solution for it. So basically this change has no effect (I think), which does not mean that we should not do it.

ouchadam:
@kojid0 create catch! I'll update and sync with @amshakal about the font weight as it's set to medium figma  

:+1: for separating the `you can` and `skip this question`, the designs have also been updated (or I imagined them the wrong way originally :thinking: )

vector/src/main/res/layout/fragment_ftue_auth_use_case.xml
ouchadam: I see... the figma typography is a little misleading here then :sweat_smile: Seems like semi bold is the correct weighting

vector/src/main/res/values/strings.xml
```diff
@@ -2533,8 +2533,8 @@
     <string name="ftue_auth_use_case_option_two">Teams</string>
     <string name="ftue_auth_use_case_option_three">Communities</string>
     <!-- Note to translators: the %s is replaced by the content of ftue_auth_use_case_skip_partial -->
-    <string name="ftue_auth_use_case_skip">Not sure yet? %s</string>
-    <string name="ftue_auth_use_case_skip_partial">You can skip this question</string>
+    <string name="ftue_auth_use_case_skip">Not sure yet? You can %s</string>
+    <string name="ftue_auth_use_case_skip_partial">skip this question</string>
```
bmarty, line 2537: Happy to see that your plan to put the new strings at a temporary location to avoid making Weblate handle them is useful here!

bmarty (approved):
LGTM, thanks!

vector/src/main/res/values/strings.xml
ouchadam: same! it does mean there's a slight delay with when the translations can be done but it does save us doing multiple redundant translations

ouchadam:
PR rebased on develop with the font weight fix #5135 
- screenshots updates

bmarty (approved):
LGTM, thanks for the update!

amshakal (approved):

daniellekirkwood (approved):

PR #5107, opened 2022-01-31
Give more visibility to the sanity test nighly by adding a badge to the README

bmarty:
Rendered

image


ouchadam (approved):
:100: 

unfortunately the reactions long press is still failing, I'm wondering if the touch zone is too small on the emulator :thinking:

PR #5117, opened 2022-02-01
Handling use case screen in sanity tests 

ouchadam:
Whilst fixing the sanity tests after enabling the use case screen, 2 crashes were found

- Multiple `VectorSessionStore` instances were attempting to create separate dataStore instances which then throw because the same backing file was being accessed, fixed by introducing a thread safe lazy provider `(String) -> DataStore<Preferences>` to allow for dynamic created stores to be cached and reused   

- Posthog doesn't support sending empty properties and a blank analytics id when identifying a user/session. This was caused by the analytics models filtering out null properties https://github.com/matrix-org/matrix-analytics-events/pull/20, I've manually merged in the changes for the identity model.

*note 
- I haven't created a changelog entry as the feature hasn't been enabled yet, are the crash fixes worth their own separate entries?

### DIFF
--- a/vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
+++ b/vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
@@ -28,6 +28,13 @@ import kotlin.reflect.KProperty
 /**
  * Provides a singleton datastore cache
  * allows for lazily fetching a datastore instance by key to avoid creating multiple stores for the same file
+ * Based on https://androidx.tech/artifacts/datastore/datastore-preferences/1.0.0-source/androidx/datastore/preferences/PreferenceDataStoreDelegate.kt.html
+ *
+ * Makes use of a ReadOnlyProperty in order to provide a simplified api on top of a Context
+ * ReadOnlyProperty allows us to lazily access the backing property instead of requiring it upfront as a dependency
+ * <pre>
+ * val Context.dataStoreProvider by dataStoreProvider()
+ * </pre>
  */
 fun dataStoreProvider(): ReadOnlyProperty<Context, (String) -> DataStore<Preferences>> {
     return MappedPreferenceDataStoreSingletonDelegate()
--- a/vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
+++ b/vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
@@ -40,7 +40,7 @@ private class MappedPreferenceDataStoreSingletonDelegate : ReadOnlyProperty<Cont
         { key ->
             dataStoreCache.getOrPut(key) {
                 PreferenceDataStoreFactory.create {
-                    context.preferencesDataStoreFile(key)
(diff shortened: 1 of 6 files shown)


vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Provides a singleton datastore cache
+ * allows for lazily fetching a datastore instance by key to avoid creating multiple stores for the same file
+ */
+fun dataStoreProvider(): ReadOnlyProperty<Context, (String) -> DataStore<Preferences>> {
+    return MappedPreferenceDataStoreSingletonDelegate()
+}
+
+private class MappedPreferenceDataStoreSingletonDelegate : ReadOnlyProperty<Context, (String) -> DataStore<Preferences>> {
```
ouchadam, line 36: this is a simplified version of `by preferencesDataStore(name = "")`

vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+        { key ->
+            dataStoreCache.getOrPut(key) {
+                PreferenceDataStoreFactory.create {
+                    context.preferencesDataStoreFile(key)
+                }
+            }
+        }
+    }
+
+    override fun getValue(thisRef: Context, property: KProperty<*>) = provider.invoke(thisRef)
```
ouchadam, line 49: `ReadOnlyProperty` allows us to lazily read a value at the point of read access, which in our case we use to lazily create a context based dependency

vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
bmarty: Nit: I would prefer to see your comment in the file itself, instead of in this PR. I think it is better and more visible for future reference.

vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
bmarty: Same remark :)

bmarty (approved):
Thanks.
About adding entry in the changelog, this is necessary if it fixes a crash that could happen in the current app, and not only during test. Else I think it's fine to omit it.

vector/src/main/java/im/vector/app/core/datastore/DataStoreProvider.kt
ouchadam: will add :+1:

ouchadam: combined both comments into the public api doc https://github.com/vector-im/element-android/pull/5117/commits/4b11ac6e90362ee66024c3a4d1fe1763bce00c03

bmarty (approved):
Thanks for the update!

PR #5127, opened 2022-02-02
Support generic location pin

onurays:
I am not satisfied with code duplications, @bmarty please advise if you see some quick wins.

Screenshot_20220202_142853


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLocationContent.kt
```diff
@@ -64,4 +64,12 @@ data class MessageLocationContent(
...
 
     fun getBestGeoUri() = locationInfo?.geoUri ?: geoUri
+
+    /**
+     * @return true if location asset is different than LocationAssetType.SELF
+     */
+    fun isGenericLocation(): Boolean {
+        // Should behave like m.self if locationAsset is null
+        return locationAsset != null && locationAsset.type != LocationAssetType.SELF
+    }
```
bmarty, line 74: I think it will be simpler to test the opposite, so something like:
```kotlin
    fun isSelfLocation(): Boolean {
        // Should behave like m.self if locationAsset is null
        return locationAsset?.type == null || locationAsset.type == LocationAssetType.SELF
    }
```

vector/src/main/java/im/vector/app/features/location/Config.kt
```diff
@@ -18,6 +18,7 @@ package im.vector.app.features.location
 
 const val MAP_BASE_URL = "https://api.maptiler.com/maps/streets/style.json"
 const val STATIC_MAP_BASE_URL = "https://api.maptiler.com/maps/basic/static/"
+const val USER_PIN_NAME = "USER_PIN_NAME"
```
bmarty, line 21: It should be more something like `const val GENERIC_PIN_NAME = "GENERIC_PIN_NAME"`, no?

vector/src/test/java/im/vector/app/features/location/LocationDataTest.kt
```diff
@@ -57,4 +60,16 @@ class LocationDataTest {
         parseGeo("ge o:12.34,56.78;13.56").shouldBeNull()
         parseGeo("geo :12.34,56.78;13.56").shouldBeNull()
     }
+
+    @Test
+    fun genericLocationTest() {
+        val contentWithNullAsset = MessageLocationContent(body = "", geoUri = "", locationAsset = null)
+        contentWithNullAsset.isGenericLocation() shouldBeEqualTo(false)
```
bmarty, line 67: Nit: better to use `.shouldBeFalse()` and `.shouldBeTrue()`

bmarty (commented):
I do not see too many code duplication, that's fine.
Some small remarks after a static review

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLocationContent.kt
onurays: Done.

vector/src/test/java/im/vector/app/features/location/LocationDataTest.kt
onurays: Done. I didn't know these exist :O

vector/src/main/java/im/vector/app/features/location/Config.kt
onurays: Not exactly, but I changed it to `DEFAULT_PIN_ID` for now. Refactoring will be needed during live location implementation.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsEpoxyController.kt
```diff
@@ -82,7 +82,7 @@ class MessageActionsEpoxyController @Inject constructor(
                 ?.toModel<MessageLocationContent>(catchError = true)
         val locationUrl = locationContent?.toLocationData()
                 ?.let { urlMapProvider.buildStaticMapUrl(it, INITIAL_MAP_ZOOM_IN_TIMELINE, 1200, 800) }
-        val locationOwnerId = if (locationContent?.isGenericLocation().orFalse()) null else state.informationData.matrixItem.id
+        val locationOwnerId = if (locationContent?.isSelfLocation().orFalse()) state.informationData.matrixItem.id else null
```
bmarty, line 85: I think now you have to use `orTrue()` but I may be wrong 🤯

vector/src/test/java/im/vector/app/features/location/LocationDataTest.kt
```diff
@@ -62,14 +64,14 @@ class LocationDataTest {
...
     @Test
-    fun genericLocationTest() {
+    fun selfLocationTest() {
         val contentWithNullAsset = MessageLocationContent(body = "", geoUri = "", locationAsset = null)
-        contentWithNullAsset.isGenericLocation() shouldBeEqualTo(false)
+        contentWithNullAsset.isSelfLocation().shouldBeTrue()
 
         val contentWithNullAssetType = MessageLocationContent(body = "", geoUri = "", locationAsset = LocationAsset(type = null))
-        contentWithNullAssetType.isGenericLocation() shouldBeEqualTo(true)
+        contentWithNullAssetType.isSelfLocation().shouldBeTrue()
```
bmarty, line 72: Weird, there is a logic change here. But the newer version seems fine. So the previous test was wrong?

bmarty (commented):
Some late remarks

vector/src/test/java/im/vector/app/features/location/LocationDataTest.kt
onurays: > If m.asset is missing from the location's content the client should render it as m.self as that will be the most common use case. Otherwise, if it's not missing but the type is invalid or unknown the client should attempt to render it as a generic location.

Yes, MSC confused me. But last one seems ok. I will also check with other clients.

bmarty: OK, so the correct test should be `contentWithNullAssetType.isSelfLocation().shouldBeFalse()`, in this case no?

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsEpoxyController.kt
onurays: If the content is accidentally null, then we have a bigger problem :)

vector/src/test/java/im/vector/app/features/location/LocationDataTest.kt
onurays: We just discussed internally and the missing `type` is `m.asset`.

bmarty (approved):
Thanks for the update.
Please add a file for the change log and this PR can be merged!

PR #5129, opened 2022-02-02
Fix fallback permalink when threads are disabled

### LINKED ISSUE #5128: Fix permalinks when threads are disabled
Fix permalinks navigation when threads are disabled along with the fallback `in reply to`

ariskotsomitopoulos:
Fixing permalink navigation when threads are disabled. Closes #5128


bmarty (approved):
LGTM

PR #5136, opened 2022-02-02
Defensive coding to ensure encryption when room was once e2e

BillCarsonFr:
Some defensive coding on the crypto side to ensure that a room that was once known locally has encrypted (with a valid algorithm) will still stays encrypted.
This is very defensive and resist to `state resets` variations

In the CryptoRoom entity we add a new boolean to remember if the room was once encrypted with a valid algorithm.
We still store the alg got from the state, in order to properly give feedback of misconfigured room

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
@@ -113,6 +113,14 @@ interface CryptoService {
 
     fun isRoomEncrypted(roomId: String): Boolean
 
+    /**
+     * This is a bit different than isRoomEncrypted
+     * A room is encrypted when there is a m.room.encryption state event in the room (malformed/invalid or not)
+     * But the crypto layer has additional guaranty to ensure that encryption would never been reverted
+     * It's defensive coding out of precaution (if ever state is reset)
+     */
+    fun shouldEncryptInRoom(roomId: String?): Boolean
+
     fun encryptEventContent(eventContent: Content,
                             eventType: String,
                             roomId: String,
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
@@ -113,14 +113,6 @@ interface CryptoService {
 
     fun isRoomEncrypted(roomId: String): Boolean
 
-    /**
-     * This is a bit different than isRoomEncrypted
-     * A room is encrypted when there is a m.room.encryption state event in the room (malformed/invalid or not)
-     * But the crypto layer has additional guaranty to ensure that encryption would never been reverted
-     * It's defensive coding out of precaution (if ever state is reset)
-     */
-    fun shouldEncryptInRoom(roomId: String?): Boolean
-
     fun encryptEventContent(eventContent: Content,
                             eventType: String,
                             roomId: String,
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/migration/MigrateCryptoTo015.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/migration/MigrateCryptoTo015.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2022 The Matrix.org Foundation C.I.C.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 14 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
```diff
@@ -113,6 +113,14 @@ interface CryptoService {
 
     fun isRoomEncrypted(roomId: String): Boolean
 
+    /**
+     * This is a bit different than isRoomEncrypted
+     * A room is encrypted when there is a m.room.encryption state event in the room (malformed/invalid or not)
+     * But the crypto layer has additional guaranty to ensure that encryption would never been reverted
+     * It's defensive coding out of precaution (if ever state is reset)
+     */
+    fun shouldEncryptInRoom(roomId: String?): Boolean
```
ganfra, line 122: If we have roomId as nullable here we should do the same for isRoomEncrypted I guess

ganfra (approved):
One remark otherwise LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
BillCarsonFr: roomId in event are optional (but almost always there), all codes ends up having optional, and function call wrapped in a let with an elvis.
It's a bit annoying, I let the caller pass an optional and return false if null in the code. I find the ending code more readable

ganfra: Yeah I know...the fact is I think we shouldn't have nullable roomId in Event :/

bmarty, line 122: Do we really need to expose this to the application?
Using the CryptoStore may be enough internally

Also we can have `isRoomEncrypted` which return `false` and `shouldEncryptInRoom` which return `true`. How the UI should react in this case? Updating the UI will be maybe for another PR so this is the rational to expose that API?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/send/DefaultSendService.kt
```diff
@@ -303,7 +303,7 @@ internal class DefaultSendService @AssistedInject constructor(
     private fun internalSendMedia(allLocalEchoes: List<Event>, attachment: ContentAttachmentData, compressBeforeSending: Boolean): Cancelable {
         val cancelableBag = CancelableBag()
 
-        allLocalEchoes.groupBy { cryptoSessionInfoProvider.isRoomEncrypted(it.roomId!!) }
+        allLocalEchoes.groupBy { cryptoStore.roomWasOnceEncrypted(it.roomId!!) }
```
bmarty, line 306: We still have `!!` here, this is not ideal :/ but this is not Event coming from the wild so I think it's fine...

bmarty (commented):
Some thoughts

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
BillCarsonFr: The source of trust is still the presence or not of a state event of type m.room.encryption, so the UI will always reflect that. For example UI will detect a missconfigured encryption.
The shouldEncrypt is more some internal protection, and should not be reflected on UI.

Also I don't expose it anymore

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/send/DefaultSendService.kt
BillCarsonFr: Yes it's a private function, so acceptable :/

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/migration/MigrateCryptoTo015.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package org.matrix.android.sdk.internal.crypto.store.db.migration
+
+import io.realm.DynamicRealm
+import org.matrix.android.sdk.internal.crypto.MXCRYPTO_ALGORITHM_MEGOLM
+import org.matrix.android.sdk.internal.crypto.store.db.model.CryptoRoomEntityFields
+import org.matrix.android.sdk.internal.util.database.RealmMigrator
+
+// Version 14L Update the way we remember key sharing
```
bmarty, line 24: This comment need to be updated (or removed). I can delete it on develop.

bmarty (approved):
Thanks for the update!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/store/db/migration/MigrateCryptoTo015.kt
bmarty: (95c00a1cce66946ce1c62d6f143891db575d0287)

PR #5137, opened 2022-02-02
Sync: avoid deleting root event of CurrentState on gappy sync

bmarty:


ouchadam (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/RealmSessionStoreMigration.kt
```diff
@@ -445,4 +446,13 @@ internal class RealmSessionStoreMigration @Inject constructor(
                     }
                 }
     }
+
+    private fun migrateTo22(realm: DynamicRealm) {
+        Timber.d("Step 21 -> 22")
+        val hasMissingStateEvent = realm.where("CurrentStateEventEntity").isNull(CurrentStateEventEntityFields.ROOT.`$`).findFirst() != null
```
BillCarsonFr, line 452: I you left some room, wouldn't this lead to null root and trigger an unwanted initial sync?

bmarty: Yes, we have updated the query together. Thanks!

PR #5142, opened 2022-02-03
Aligning use case tracking with iOS

ouchadam:
Aligning use case user property tracking with iOS

- Filters out null user properties as we want to avoid 1 device resetting values for others, we allow new non absent values to be set
- Delays applying the use case property until the point of account creation to avoid needing to reset the value.

### DIFF
--- a/changelog.d/5142.bugfix
+++ b/changelog.d/5142.bugfix
@@ -1,1 +1,1 @@
-Aligns use case identifying with iOS implementation
+Analytics: aligns use case identifying with iOS implementation
--- a/changelog.d/5142.bugfix
+++ b/changelog.d/5142.bugfix
@@ -1,0 +1,1 @@
+Aligns use case identifying with iOS implementation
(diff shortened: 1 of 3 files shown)


changelog.d/5142.bugfix
```diff
@@ -0,0 +1 @@
+Aligns use case identifying with iOS implementation
```
bmarty, line 1: Maybe prefix the content with `Analytics: ` else it can be a bit confusing.
Or create a new extension for towncrier `.analytics` which will end to a dedicated section in the Changelog?

bmarty (approved):
LGTM, just a small remark about the changelog.

changelog.d/5142.bugfix
ouchadam: a new section could be handy :thinking: I think for now the prefix might be enough~

ouchadam: 3f3cac0489342ea0fdcd90fe80237bd711bfe77c

PR #5145, opened 2022-02-03
Fix location rendering in timeline if map cannot be loaded

### LINKED ISSUE #5143: Fix location rendering in timeline if map cannot be loaded
### Steps to reproduce

- Turn on airplane mode
- Share location

### Outcome

- An indication of failed map should be rendered.

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

onurays:
Fixes #5143 

[Figma](https://www.figma.com/file/uthEK6xuo9hSQRdWvPAre5/Location-Sharing?node-id=427%3A48024)

Screenshot_20220203_161053


gaelledel (approved):

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageLocationItem.kt
```diff
@@ -46,20 +53,32 @@ abstract class MessageLocationItem : AbsMessageItem<MessageLocationItem.Holder>(
         GlideApp.with(holder.staticMapImageView)
                 .load(location)
                 .apply(RequestOptions.centerCropTransform())
-                .into(holder.staticMapImageView)
+                .listener(object : RequestListener<Drawable> {
+                    override fun onLoadFailed(e: GlideException?, model: Any?, target: Target<Drawable>?, isFirstResource: Boolean): Boolean {
+                        holder.staticMapPinImageView.setImageResource(R.drawable.ic_location_pin_failed)
```
bmarty, line 58: This is maybe a design decision, but when the map fail to load, we could still draw the user avatar if available, no?

bmarty (approved):
LGTM, just one question (not blocking).

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageLocationItem.kt
bmarty: CC @gaelledel

onurays: Yes it is a design decision and it seems like a bug/ugly to use user avatar.

bmarty: I agree, thanks!

PR #5168, opened 2022-02-05
Fix back pagination conduit

### LINKED ISSUE #5166: Repeated timeline, many read receipts
### Steps to reproduce

Log in to a conduit account.
Accept a federated invite into a new room.

The problem seems to be related to the limited flag for syncs or the /messages endpoint.

### Your phone model

-

### Operating system version

-

-

### Homeserver

conduit.rs

### Will you send logs?

No

BillCarsonFr:
Fixes #5166
@ganfra Could you have a look at it? and check for any side effects?
Basically EA is relying on the`end` field of /messages call, but as per spec it should be omitted if no events .
Conduit is doing as per spec, and it's breaking EA back pagination.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TokenChunkEventPersistor.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TokenChunkEventPersistor.kt
@@ -67,14 +67,25 @@ internal class TokenChunkEventPersistor @Inject constructor(
                 .awaitTransaction { realm ->
                     Timber.v("Start persisting ${receivedChunk.events.size} events in $roomId towards $direction")
 
+                    /**
+                     * As per spec
+                     * if no further events are available (either because we have reached the start of the timeline,
+                     * or because the user does not have permission to see any more events), the end property is omitted from the response.
+                     * Synapse is not omitting it and EA was relying on it
+                     */
+                    val chunkEnd = if (receivedChunk.events.isEmpty()) {
+                        receivedChunk.start
+                    } else {
+                        receivedChunk.end
+                    }
                     val nextToken: String?
                     val prevToken: String?
                     if (direction == PaginationDirection.FORWARDS) {
-                        nextToken = receivedChunk.end
+                        nextToken = chunkEnd
                         prevToken = receivedChunk.start
                     } else {
                         nextToken = receivedChunk.start
-                        prevToken = receivedChunk.end
+                        prevToken = chunkEnd
                     }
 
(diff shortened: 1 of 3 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TokenChunkEventPersistor.kt
```diff
@@ -67,14 +67,25 @@ internal class TokenChunkEventPersistor @Inject constructor(
...
                     Timber.v("Start persisting ${receivedChunk.events.size} events in $roomId towards $direction")
 
+                    /**
+                     * As per spec
+                     * if no further events are available (either because we have reached the start of the timeline,
+                     * or because the user does not have permission to see any more events), the end property is omitted from the response.
+                     * Synapse is not omitting it and EA was relying on it
+                     */
+                    val chunkEnd = if (receivedChunk.events.isEmpty()) {
+                        receivedChunk.start
```
ganfra, line 77: ReceivedChunk.start should be the same as the from param, so we will be blocked too, no?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TokenChunkEventPersistor.kt
```diff
@@ -67,14 +67,25 @@ internal class TokenChunkEventPersistor @Inject constructor(
                 .awaitTransaction { realm ->
                     Timber.v("Start persisting ${receivedChunk.events.size} events in $roomId towards $direction")
 
+                    /**
+                     * As per spec
+                     * if no further events are available (either because we have reached the start of the timeline,
+                     * or because the user does not have permission to see any more events), the end property is omitted from the response.
+                     * Synapse is not omitting it and EA was relying on it
+                     */
+                    val chunkEnd = if (receivedChunk.events.isEmpty()) {
```
timokoesters, line 76: Checking events.is_empty is wrong here. There may be events, but if those are the 'last' events in the room, 'end' will not be set.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TokenChunkEventPersistor.kt
timokoesters: I think the 'start' is set but 'end' is unset (=null?). Probably chunkEnd=null causes these problems in Element Android

timokoesters (changes requested):

bmarty:
#5166 has been fixed by #6318, we think we do not need this change anymore. Closing.

PR #5176, opened 2022-02-07
Retrieve map style url from .well-known

### LINKED ISSUE #5175: Retrieve map style url from .well-known
### Your use case

Check if `org.matrix.msc3488.tile_server. map_style_url ` exists in `.well-known` and use it instead of hardcoded one.

_No response_

### Additional context

_No response_

onurays:
Fixes #5175.

[MSC](https://github.com/matrix-org/matrix-doc/blob/matthew/location/proposals/3488-location.md#well-known-configuration)

### DIFF
--- a/vector/src/main/java/im/vector/app/features/location/UrlMapProvider.kt
+++ b/vector/src/main/java/im/vector/app/features/location/UrlMapProvider.kt
@@ -32,14 +32,16 @@ class UrlMapProvider @Inject constructor(
 ) {
     private val keyParam = "?key=${BuildConfig.mapTilerKey}"
 
+    private val fallbackMapUrl = buildString {
+        append(MAP_BASE_URL)
+        append(keyParam)
+    }
+
     suspend fun getMapUrl(): String {
-        return tryOrNull { rawService.getElementWellknown(session.sessionParams) }
+        val upstreamMapUrl = tryOrNull { rawService.getElementWellknown(session.sessionParams) }
                 ?.mapTileServerConfig
                 ?.mapStyleUrl
-                ?: buildString {
-                    append(MAP_BASE_URL)
-                    append(keyParam)
-                }
+        return upstreamMapUrl ?: fallbackMapUrl
     }
 
     fun buildStaticMapUrl(locationData: LocationData,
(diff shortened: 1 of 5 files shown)


vector/src/main/java/im/vector/app/features/location/UrlMapProvider.kt
```diff
@@ -19,17 +19,27 @@ package im.vector.app.features.location
...
+        private val session: Session,
+        private val rawService: RawService
 ) {
     private val keyParam = "?key=${BuildConfig.mapTilerKey}"
 
-    // This is static so no need for a fun
-    val mapUrl = buildString {
-        append(MAP_BASE_URL)
-        append(keyParam)
+    suspend fun getMapUrl(): String {
```
ouchadam, line 35: small nit: this might be easier to parse if we separate the fallback with its own variable


```kotlin
private val fallbackMapUrl = buildString {
    append(MAP_BASE_URL)
    append(keyParam)
}

suspend fun getMapUrl(): String {
    val upstreamMapUrl = tryOrNull { rawService.getElementWellknown(session.sessionParams) }
            ?.mapTileServerConfig
            ?.mapStyleUrl
    return upstreamMapUrl ?: fallbackMapUrl
}
```

what do you think?

onurays: Much easier to read! I will commit, thanks.

onurays: Done.

vector/src/main/java/im/vector/app/features/location/UrlMapProvider.kt
```diff
@@ -19,19 +19,31 @@ package im.vector.app.features.location
...
     }
 
+    suspend fun getMapUrl(): String {
+        val upstreamMapUrl = tryOrNull { rawService.getElementWellknown(session.sessionParams) }
+                ?.mapTileServerConfig
+                ?.mapStyleUrl
+        return upstreamMapUrl ?: fallbackMapUrl
+    }
+
     fun buildStaticMapUrl(locationData: LocationData,
```
bmarty, line 47: I think this fun should also be impacted by the url set up in the .well-known file, since AFAIU, it can contain another API key.

PR #5179, opened 2022-02-07
Remove redundant highlight on add poll option button

### LINKED ISSUE #5178: Remove redundant highlight on add poll option button
### Steps to reproduce

Highlighting "add option" button seems weird.

### Outcome

Remove highlight of the button

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

onurays:
Fixes #5178.

### DIFF
--- a/vector/src/main/java/im/vector/app/core/ui/list/GenericButtonItem.kt
+++ b/vector/src/main/java/im/vector/app/core/ui/list/GenericButtonItem.kt
@@ -75,8 +75,11 @@ abstract class GenericButtonItem : VectorEpoxyModel<GenericButtonItem.Holder>() 
         val textStyle = if (bold) Typeface.BOLD else Typeface.NORMAL
         holder.button.setTypeface(null, textStyle)
 
-        holder.button.rippleColor = if (highlight) ColorStateList.valueOf(ThemeUtils.getColor(holder.view.context, R.attr.colorSecondary)) else
+        holder.button.rippleColor = if (highlight) {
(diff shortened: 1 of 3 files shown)


vector/src/main/java/im/vector/app/core/ui/list/GenericButtonItem.kt
```diff
@@ -70,6 +75,12 @@ abstract class GenericButtonItem : VectorEpoxyModel<GenericButtonItem.Holder>()
         val textStyle = if (bold) Typeface.BOLD else Typeface.NORMAL
         holder.button.setTypeface(null, textStyle)
 
+        holder.button.rippleColor = if (highlight) {
```
ouchadam, line 78: was `R.attr.colorSecondary` also the previous highlight colour? just in case we've changed the default value for a generic component which might be used in other places

onurays: Good point, button has `style="@style/Widget.Vector.Button.Text"` and this style has `<item name="colorControlHighlight">?colorSecondary</item>`. But I will check if `rippleCount` has the same effect with `colorControlHighlight`.

ouchadam: thanks!

onurays: Actually was not the exact color. Sorry, now fixed.

ouchadam (approved):

PR #5181, opened 2022-02-08
#4533:  improve headers UI in room/messages lists

### LINKED ISSUE #4533: Update `Direct Messages` and `Rooms` headers from UI

Remove the `Direct Messages` and `Rooms` expandable headers from the DMs and Rooms tabs respectively.

It is annoyingly frequent that I accidentally tap the header and hide the chat list when I'm intending to open the top chat.

Moreover, I don't see the point of these headers. Not only are they buggy (switching the tab and switching back expands them again), but seemingly redundant. The title bar of the app already tells you which tab you have opened. The headers seem to have been copied from the Element Web UI directly, without an actual use case. 

The only expandable header that makes sense is `Invites`, because you might get a lot at once, and want to hide them so you can focus on your existing chat list.

I would like to see the 2 headers removed completely. The `Invites` header can have a distinctly visible UI boundary so that it's clear when the invites end and the existing DMs/Rooms start.

## Solution

Based on our chat, here is the proposed solution. The changes I have made are:
- Created a band under collapsible /expandable headers
- Added no. of items in the list (similar to ios)
- Changed type style of header to make it clearer

Screenshot 2021-12-14 at 19 55 01

Link to Figma file: https://www.figma.com/file/QugKzfgL4gISlsmluw9EVb/Change-expandable-headers?node-id=4%3A2593

Also, sections should only appear here if they have items in them.

mnaturel:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : Enhancement

## Content

Improvement for UI of headers sections in the rooms and messages lists. A counter of the list items has been added and the notification badge has been moved to the right.

## Motivation and context

Closes #4533 

## Screenshots / GIFs

New header style : 

## Tests

- Go to the home screen
- Select the rooms tab
- Check headers are correctly displayed with the correct number of items
- Check headers are collapsable/expandable
- Check notification counter is correctly displayed
- Select the messages tab
- Check headers are correctly displayed with the correct number of items
- Check headers are collapsable/expandable
- Check notification counter is correctly displayed
- Open the drawer menu
- Switch space
- Check the lists are correctly updated
- Go to search screen
- Check number of items is correct when changing text search 

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 11

## Checklist


mnaturel:
I have achieved most of the required changes. However, without making to many changes, I have some troubles to understand how to listen for query parameters changes to update the item list count to cover missing some use cases: 

- Search by text
- Space switching

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryDataSource.kt
```diff
@@ -230,6 +239,16 @@ internal class RoomSummaryDataSource @Inject constructor(
         }
     }
 
+    fun getCountFlow(queryParams: RoomSummaryQueryParams): Flow<Int> =
+            realmSessionProvider
+                    .withRealm { realm -> roomSummariesQuery(realm, queryParams).findAllAsync() }
+                    .toFlow()
+                    // need to create the flow on a context dispatcher with a thread with attached Looper
+                    .flowOn(coroutineDispatchers.main)
```
ouchadam, line 247: this is a little bit strange, we need to switch to the main thread in order to query the count?  

we have some other flow based pipelines we don't need to do the switch https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android-flow/src/main/java/org/matrix/android/sdk/flow/FlowSession.kt :thinking:

mnaturel: Yes, I see why it could be strange but this is the only solution I found to make it work. I am using the [toFlow() extension of RealmResult](https://github.com/realm/realm-java/blob/973d43725d3bc2a8bd053892ddab0ff2856daa12/realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmResultsExtensions.kt) and I didn't find any existing usage in the project. I have seen there is an [issue about the documentation of this extension](https://github.com/realm/realm-java/issues/7168) meaning the thread we collect the flow should be the same as the one used for creating the result...

amshakal (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/summary/RoomSummaryDataSource.kt
ouchadam: having read a little more into this, it's not as troublesome as I originally thought, realm's async is keeping us off of the main thread but it needs to be triggered from a thread with a looper

initially I was thought this might have pushed us into doing the database query on the main thread

vector/src/main/java/im/vector/app/features/home/room/list/RoomCategoryItem.kt
```diff
@@ -46,14 +46,18 @@ abstract class RoomCategoryItem : VectorEpoxyModel<RoomCategoryItem.Holder>() {
             DrawableCompat.setTint(it, tintColor)
         }
         holder.unreadCounterBadgeView.render(UnreadCounterBadgeView.State(unreadNotificationCount, showHighlighted))
-        holder.titleView.setCompoundDrawablesWithIntrinsicBounds(null, null, expandedArrowDrawable, null)
         holder.titleView.text = title
+        with(holder.counterView) {
+            text = ""
```
ouchadam, line 51: does this reset the counter amount?

mnaturel: Here it does not display the counter at all. For what I understood of the code, this item may be used in a screen to add a room but I didn't succeed to find where in the app. I didn't reach creation of this item in debug mode. So, I just supposed we don't need a counter in this use case. Do you know how to open the screen which uses this item ?

ouchadam: from what I can tell `RoomCategoryItem/RoomCategoryItem_` is only used by the `AddRoomListController` which is used by the `SapceAddRoomFragment`

which I believe is this screen...

Screenshot_20220210_102639

to get here I did, `menu drawer` -> `add space` -> `private, just me`, this screen should also be accessible by editing an existing space to add/remove rooms

ouchadam: if we don't want to have a counter here, it could be nice to extract a constant for the empty copy with a name like 

```kotlin
const val NO_COUNTER = ""
```

mnaturel: I will try to add the counter in this screen as well to have consistent UI.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
```diff
@@ -216,6 +217,11 @@ interface RoomService {
                                           pagedListConfig: PagedList.Config = defaultPagedListConfig,
                                           sortOrder: RoomSortOrder = RoomSortOrder.ACTIVITY): UpdatableLivePageResult
 
+    /**
+     * Retrieve a flow on the the number of rooms.
```
bmarty, line 221:
```suggestion
     * Retrieve a flow on the number of rooms.
```

vector/src/main/java/im/vector/app/features/home/room/list/RoomListFragment.kt
```diff
@@ -373,6 +380,18 @@ class RoomListFragment @Inject constructor(
         }
     }
 
+    private fun listenItemCount(section: RoomsSection, sectionAdapter: SectionHeaderAdapter) {
```
bmarty, line 383: Nit: I would rename to `listenToItemCount`, or `observeItemCount` (Rx style). Not sure what is the Flow style naming convention. `collectItemCount`?

vector/src/main/java/im/vector/app/features/home/room/list/SectionHeaderAdapter.kt
```diff
@@ -79,14 +80,17 @@ class SectionHeaderAdapter constructor(
         }
 
         fun bind(roomsSectionData: RoomsSectionData) {
-            binding.roomCategoryTitleView.text = roomsSectionData.name
-            val tintColor = ThemeUtils.getColor(binding.root.context, R.attr.vctr_content_secondary)
-            val expandedArrowDrawableRes = if (roomsSectionData.isExpanded) R.drawable.ic_expand_more else R.drawable.ic_expand_less
-            val expandedArrowDrawable = ContextCompat.getDrawable(binding.root.context, expandedArrowDrawableRes)?.also {
-                DrawableCompat.setTint(it, tintColor)
+            with(binding) {
```
bmarty, line 83: I am not a big fan of `with` (and `apply` for the same reason). It ends up with several `this` in the block and this is a source of bugs, and not good for the readability, especially on GitHub). We can discuss this later.

bmarty (commented):
Some few remarks, but globally looking good to me!

vector/src/main/java/im/vector/app/features/home/room/list/RoomListFragment.kt
mnaturel: Okay, I will go for `observeItemCount` to be agnostic of the used observation framework (could be LiveData, Flow, Rx, etc.).

vector/src/main/java/im/vector/app/features/home/room/list/SectionHeaderAdapter.kt
mnaturel: On my side, I use a lot operators like `with`, `apply`, `run` or `let` when there are at least 2 operators applied to a variable. But I reverted back the usage here. I understand your point of view since it changes the scope, you must pay attention what you do inside.

mnaturel:
@ouchadam I have fixed all comments and rebase the branch on develop. Can you take a look again if everything is okay to be merged?

ouchadam (approved):
thanks for updating, LGTM :100:

opusforlife2:
> Also, sections should only appear here if they have items in them.

I can't tell from the PR description. Was this addressed?

mnaturel:
> > Also, sections should only appear here if they have items in them.
> 
> I can't tell from the PR description. Was this addressed?

This part was in fact already implemented before this PR. There is a field named `hidden` in the [`SectionHeaderAdapter` class](https://github.com/vector-im/element-android/blob/4939a986895e0c597234a9b8e82ab7b687492141/vector/src/main/java/im/vector/app/features/home/room/list/SectionHeaderAdapter.kt#L40). This field is set to true if there are no items corresponding to the header.

opusforlife2:
Oh wait. That was the wrong question, sorry. What I want to know is, if `Direct Messages` or `Rooms` are the only sections visible, does the section header disappear too? Since there is only one section, there is no need for the header, after all.

mnaturel:
> Oh wait. That was the wrong question, sorry. What I want to know is, if `Direct Messages` or `Rooms` are the only sections visible, does the section header disappear too? Since there is only one section, there is no need for the header, after all.

Okay I see. When there is only one visible section, we decided to show the header. It is necessary to keep it since it includes the notifications counter and the current number of items as well.

mnaturel:
> I see. This was the main reason I opened #4533:
> 
> > It is annoyingly frequent that I accidentally tap the header and hide the chat list when I'm intending to open the top chat.
> 
> So, my issue isn't actually solved by this PR. 😅

Hello, sorry for the late reply. I see your point. The height of the headers has been increased. Maybe now it is better to avoid this scenario? If it does not fix this problem, I do not have a solution right now for that. It will need to rethink the UI and for all platforms to be consistent. CC: @daniellekirkwood

opusforlife2:
Actually, the description of the merged PR #5580 seems to imply that the change I'm seeking is already implemented. There are no screenshots, though, so I'm not sure. Could you please check?

mnaturel:
The actual changes is in fact in #5616. When there is a single section, it is not collapsable. Is was released in [the v1.4.7](https://github.com/vector-im/element-android/releases/tag/v1.4.7). It looks like the capture below:

opusforlife2:
Alright! That does indeed solve my problem. Thanks for confirming!

Maybe #4533 should be linked to that PR instead, for posterity?

PR #5183, opened 2022-02-08
join and leave methods moved from MembershipService to RoomService an…

fedrunov:
There are methods to join and leave room/space in `MembershipService` but in context of this service (as part of `Room`) it's not always possible to validate if current instance is `Room` or `Space` and `JoinRoomTask` is always used, even though for spaces we should use `JoinSpaceTask`. This PR removes `join` and `leave` methods from `MembershipService` and replaces it with a separate implementation in `RoomService` and `SpaceService` according to context. This will allow us to properly track events for analytics and also will prevent possible bugs when Space is treated as Room

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
@@ -184,8 +184,8 @@ internal class DefaultSpaceService @Inject constructor(
         return joinSpaceTask.execute(JoinSpaceTask.Params(spaceIdOrAlias, reason, viaServers))
     }
 
-    override suspend fun leaveSpace(roomId: String, reason: String?) {
-        leaveRoomTask.execute(LeaveRoomTask.Params(roomId, reason))
+    override suspend fun leaveSpace(spaceId: String, reason: String?) {
+        leaveRoomTask.execute(LeaveRoomTask.Params(spaceId, reason))
     }
 
     override suspend fun rejectInvite(spaceId: String, reason: String?) {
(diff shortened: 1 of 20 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
```diff
@@ -184,6 +184,10 @@ internal class DefaultSpaceService @Inject constructor(
         return joinSpaceTask.execute(JoinSpaceTask.Params(spaceIdOrAlias, reason, viaServers))
     }
 
+    override suspend fun leaveSpace(roomId: String, reason: String?) {
```
ganfra, line 187: spaceId instead of roomId?

ganfra (changes requested):
Small remarks otherwise ok for me!

ouchadam (approved):
also looks good to me! just missing a changelog entry,  as it's touching the sdk api I guess we'll need to use .sdk for the file type

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
fedrunov: initially changed to spaceId, but since we use Room for Space sometimes wasn't sure if it won't be confusing. Like
leaveSpace(spaceId = room.roomId). What do you think?

ganfra (approved):
Thanks for the changes!

vector/src/main/java/im/vector/app/features/spaces/leave/SpaceLeaveAdvancedViewModel.kt
```diff
@@ -72,14 +72,14 @@ class SpaceLeaveAdvancedViewModel @AssistedInject constructor(
                     try {
                         state.selectedRooms.forEach {
                             try {
-                                session.getRoom(it)?.leave(null)
+                                session.leaveRoom(it)
```
bmarty, line 75: Behind the scene I think this is equivalent to leave a room or a space, but can `it` be a `spaceId` sometimes here?

fedrunov: As far i understand we leave all selected rooms in a space here. Maybe there could be also subspaces, but I think rooms is more often and more relevant here.

PR #5185, opened 2022-02-08
Decoupling Matrix creation from Application

ouchadam:
Decouples the creation of our `Matrix` instance from the application by using the DI factory instead.
This has the benefit of allowing the `MatrixConfiguration` to use injectable dependencies.

However, it means the `Matrix.getInstance` will throw if we attempt to access a non injected instance before an injected instance as it'll attempt to read the configuration from the application which no longer implements the provider interface  (maybe a good thing for consistency?)

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
@@ -99,12 +99,28 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
         private lateinit var instance: Matrix
         private val isInit = AtomicBoolean(false)
 
+        /**
+         * Creates a new instance of Matrix, it's recommend to manage this instance as a singleton.
+         * To make use of the built in singleton use Matrix.initialise() and/or Matrix.getInstance(context) instead
+         **/
+        fun createInstance(context: Context, matrixConfiguration: MatrixConfiguration): Matrix {
+            return Matrix(context.applicationContext, matrixConfiguration)
+        }
+
+        /**
+         * Initializes a singleton instance of Matrix for the given MatrixConfiguration
+         * This instance will be returned by Matrix.getInstance(context)
+         */
         fun initialize(context: Context, matrixConfiguration: MatrixConfiguration) {
             if (isInit.compareAndSet(false, true)) {
                 instance = Matrix(context.applicationContext, matrixConfiguration)
             }
         }
 
+        /**
+         * Either provides an already initialized singleton Matrix instance or queries the application context for a MatrixConfiguration.Provider
+         * to lazily create and store the instance.
+         */
         fun getInstance(context: Context): Matrix {
             if (isInit.compareAndSet(false, true)) {
                 val appContext = context.applicationContext
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
@@ -129,7 +129,8 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
                     instance = Matrix(appContext, matrixConfiguration)
                 } else {
                     throw IllegalStateException("Matrix is not initialized properly." +
-                            " You should call Matrix.initialize or let your application implements MatrixConfiguration.Provider.")
+                            " If you want to manage your own Matrix instance use Matrix.createInstance" +
+                            " otherwise you should call Matrix.initialize or let your application implement MatrixConfiguration.Provider.")
                 }
             }
             return instance
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
@@ -122,6 +122,7 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
          * Either provides an already initialized singleton Matrix instance or queries the application context for a MatrixConfiguration.Provider
          * to lazily create and store the instance.
          */
+        @Suppress("deprecation") // suppressing warning as this method is unused but is still provided for SDK clients
         @Deprecated("Use Matrix.createInstance and manage the instance manually")
         fun getInstance(context: Context): Matrix {
             if (isInit.compareAndSet(false, true)) {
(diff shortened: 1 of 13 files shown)


vector/src/main/java/im/vector/app/VectorApplication.kt
```diff
@@ -220,16 +218,9 @@ class VectorApplication :
         }
     }
 
-    override fun providesMatrixConfiguration(): MatrixConfiguration {
```
ouchadam, line 223: the main downside to using `MatrixConfiguration.Provider` is that it occurs during `super.onCreate` which means the dependencies declared in the application haven't been injected yet, making the object creation lifecycle a bit misleading

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
```diff
@@ -107,7 +110,17 @@ object VectorStaticModule {
     }
 
     @Provides
-    fun providesMatrix(context: Context): Matrix {
+    fun providesMatrixConfiguration(context: Context): MatrixConfiguration {
+        return MatrixConfiguration(
+                applicationFlavor = BuildConfig.FLAVOR_DESCRIPTION,
+                roomDisplayNameFallbackProvider = VectorRoomDisplayNameFallbackProvider(context)
```
bmarty, line 116: We could probably also provide `RoomDisplayNameFallbackProvider` as a param.

vector/src/main/java/im/vector/app/VectorApplication.kt
```diff
@@ -77,7 +75,6 @@ import androidx.work.Configuration as WorkConfiguration
 @HiltAndroidApp
 class VectorApplication :
         Application(),
-        MatrixConfiguration.Provider,
```
bmarty, line 80: Maybe delete this interface? Or can it still be used by other application?

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
```diff
@@ -107,7 +110,17 @@ object VectorStaticModule {
...
+        )
+    }
+
+    @Provides
+    @Singleton
+    fun providesMatrix(context: Context, configuration: MatrixConfiguration): Matrix {
+        Matrix.initialize(context, configuration)
         return Matrix.getInstance(context)
```
bmarty, line 124: Will this work since the context is not implementing `MatrixConfiguration.Provider` anymore?

https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt#L108

bmarty (commented):
Some quick feedback, would be nice to have @ganfra POV too.

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
bmarty: Ah, just read again the description, sorry

vector/src/main/java/im/vector/app/VectorApplication.kt
ouchadam: it could be used by other clients of the SDK, if we agree that the Matrix singleton API should avoid querying the application then we could mark the interface a deprecate and schedule removing it? potentially with some migration steps~ 

what do you think?

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
ouchadam: will do! :+1:

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
ouchadam: I did it :smile: having the separate method really helps clarify the ambiguity fd2d9287e7b536ac259f950ad66ecfb11ef029f8

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
ouchadam: 674aea97a8db26f63bf2c72ab68412368e1d6d2e

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
```diff
@@ -99,12 +99,28 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
         private lateinit var instance: Matrix
         private val isInit = AtomicBoolean(false)
 
+        /**
+         * Creates a new instance of Matrix, it's recommend to manage this instance as a singleton.
+         * To make use of the built in singleton use Matrix.initialise() and/or Matrix.getInstance(context) instead
```
bmarty, line 104:
```suggestion
         * To make use of the built in singleton use Matrix.initialize() and/or Matrix.getInstance(context) instead
```

US English :)

vector/src/main/java/im/vector/app/features/rageshake/BugReporter.kt
```diff
@@ -73,7 +73,8 @@ class BugReporter @Inject constructor(
         private val versionProvider: VersionProvider,
         private val vectorPreferences: VectorPreferences,
         private val vectorFileLogger: VectorFileLogger,
-        private val systemLocaleProvider: SystemLocaleProvider
+        private val systemLocaleProvider: SystemLocaleProvider,
+        private val matrix: Matrix
```
bmarty, line 77: I like that!

vector/src/main/java/im/vector/app/VectorApplication.kt
ganfra: Yes I think we should remove this and the inner singleton to avoid mistakes! 
Regarding the process, we could indeed mark them as deprecated in a first place.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
```diff
@@ -99,12 +99,28 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
...
         fun initialize(context: Context, matrixConfiguration: MatrixConfiguration) {
             if (isInit.compareAndSet(false, true)) {
                 instance = Matrix(context.applicationContext, matrixConfiguration)
             }
         }
 
+        /**
+         * Either provides an already initialized singleton Matrix instance or queries the application context for a MatrixConfiguration.Provider
+         * to lazily create and store the instance.
+         */
```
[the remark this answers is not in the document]
ouchadam: https://github.com/vector-im/element-android/pull/5185/commits/d05af1cb047f9c91e563960af269ca8382cbcc66

bmarty (approved):
Thanks for the update and thanks @ganfra for your review!

Issue #5198, opened 2022-02-10
Rooms with virtual rooms do not render the merged timeline with call events.


langleyd:
On environments with virtual rooms for voip signalling only 1 room shows call events.

### Steps
1. Make calls in multiple rooms on environment§ with virtual rooms for voip signalling.
2. Go in to each room.
### Observed
 Only one room shows the call status events in the timeline.
### Expected
All rooms show the call status events in the timeline.

PR #5199, opened 2022-02-10
Rooms with virtual rooms do not render the merged timeline with call events.

langleyd:
Updates getAccountDataEvents function to match its description.

Fixes https://github.com/vector-im/element-android/issues/5198


ouchadam (approved):
the fix makes sense to me, will defer to @ganfra

ganfra (approved):
LGTM, thanks!

Issue #5201, opened 2022-02-10
Call transfer with consult fails to make outgoing consultation call.


langleyd:
### Steps
1. Alice on Android rings bob
2. Alice makes a call transfer with consult checked to Charlie.
### Expected
Outgoing Consultation call from Alice to Charlie is made.
### Observed
Charlie never receives the call invite on their side.

PR #5083, opened 2022-01-27
Hold call for transfer selection

langleyd:
Fixes https://github.com/vector-im/element-android/issues/5081

PR #5203, opened 2022-02-10
Call transfer with consult fails to make outgoing consultation call.

langleyd:
Fixes https://github.com/vector-im/element-android/issues/5201

### First part of fix

Intent information(`EXTRA_MODE`) that is pass to attachViewRenderers  [here](https://github.com/vector-im/element-android/blob/11986ec9ef508b93be68860e5eb28460535592ce/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt#L505) is not present on intent when `onNewIntent` is called, so the outgoing call is never placed. Fix is to set the intent [here](https://github.com/vector-im/element-android/pull/5203/files#diff-7ccb9f5a9496cb142b284f942f64646e7e3580890854d161ab4e7bdaeca74ec2R170).

### Second part of fix

A follow on issue was related to a recent [PR](https://github.com/vector-im/element-android/pull/5083) that puts other caller on hold when going to the transfer selection and off hold [when they cancel](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt#L532).

With consult transfers, As the consult call is started in a new task, the call transfer selection activity immediately gets returned with `RESULT_CANCELLED` (rather than `RESULT_OK` from the `Complete` view action in turn calling `setResult`). This means we do an additional unintended unhold and this places the new call in the incorrect hold state. 

### 2 options for a solution
- Start changing how the call activity stack works(changing launch options and flags etc)
- Use the transfer screen just for transfer selection, Return the selection as an activity result and do the actual transfer back on the actual call activity/view model. 

The second option seemed pretty reasonable to me and we actually already do [the second part of the consult transfer from the call view model](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/call/VectorCallViewModel.kt#L331) already. So went with that.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
@@ -16,7 +16,6 @@
 
 package im.vector.app.features.call.transfer
 
-import android.app.Activity
 import android.content.Context
 import android.content.Intent
 import android.os.Bundle
--- a/vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
@@ -104,7 +104,7 @@ class CallTransferActivity : VectorBaseActivity<ActivityCallTransferBinding>() {
     }
 
     companion object {
-            const val EXTRA_TRANSFER_RESULT = "EXTRA_TRANSFER_RESULT"
+        const val EXTRA_TRANSFER_RESULT = "EXTRA_TRANSFER_RESULT"
         fun newIntent(context: Context, callId: String): Intent {
             return Intent(context, CallTransferActivity::class.java).also {
                 it.putExtra(Mavericks.KEY_ARG, CallTransferArgs(callId))
--- a/vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
@@ -26,6 +26,7 @@ import com.google.android.material.tabs.TabLayoutMediator
 import dagger.hilt.android.AndroidEntryPoint
 import im.vector.app.R
 import im.vector.app.core.error.ErrorFormatter
+import im.vector.app.core.extensions.exhaustive
 import im.vector.app.core.platform.VectorBaseActivity
 import im.vector.app.databinding.ActivityCallTransferBinding
 import kotlinx.parcelize.Parcelize
@@ -55,8 +56,8 @@ class CallTransferActivity : VectorBaseActivity<ActivityCallTransferBinding>() {
 
         callTransferViewModel.observeViewEvents {
             when (it) {
-                is CallTransferViewEvents.Complete        -> handleComplete()
-            }
+                is CallTransferViewEvents.Complete -> handleComplete()
+            }.exhaustive
         }
 
         sectionsPagerAdapter = CallTransferPagerAdapter(this)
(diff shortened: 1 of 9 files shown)


vector/src/main/java/im/vector/app/features/call/transfer/CallTransferViewModel.kt
```diff
@@ -22,22 +22,16 @@ import dagger.assisted.AssistedFactory
...
-import im.vector.app.features.call.dialpad.DialPadLookup
 import im.vector.app.features.call.webrtc.WebRtcCall
 import im.vector.app.features.call.webrtc.WebRtcCallManager
-import im.vector.app.features.createdirect.DirectRoomHelper
-import kotlinx.coroutines.launch
-import org.matrix.android.sdk.api.extensions.orFalse
 import org.matrix.android.sdk.api.session.call.CallState
 import org.matrix.android.sdk.api.session.call.MxCall
 
 class CallTransferViewModel @AssistedInject constructor(@Assisted initialState: CallTransferViewState,
```
ouchadam, line 32: is this class still needed? without the handle method it looks like it doesn't do much!

vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
```diff
@@ -82,25 +79,32 @@ class CallTransferActivity : VectorBaseActivity<ActivityCallTransferBinding>() {
...
+            setResult(RESULT_OK, intent)
+        } else {
+            setResult(RESULT_OK)
+        }
         finish()
     }
 
     companion object {
-
+            const val EXTRA_TRANSFER_RESULT = "EXTRA_TRANSFER_RESULT"
```
ouchadam, line 107: nit: the formatting is a little funky here

vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
```diff
@@ -82,25 +79,32 @@ class CallTransferActivity : VectorBaseActivity<ActivityCallTransferBinding>() {
             when (views.callTransferTabLayout.selectedTabPosition) {
                 CallTransferPagerAdapter.USER_LIST_INDEX -> {
                     val selectedUser = sectionsPagerAdapter.userListFragment?.getCurrentState()?.getSelectedMatrixId()?.firstOrNull() ?: return@debouncedClicks
-                    val action = CallTransferAction.ConnectWithUserId(views.callTransferConsultCheckBox.isChecked, selectedUser)
-                    callTransferViewModel.handle(action)
+                    val result = CallTransferResult.ConnectWithUserId(views.callTransferConsultCheckBox.isChecked, selectedUser)
+                    handleComplete(result)
```
ouchadam, line 83: to double check my understanding, is this this line the fix? (by returning the action as a result to the starting activity so that we can handle with the original `VectorCallViewModel`)

langleyd: yup!

ouchadam (approved):
small non blocking question around whether we can remove a file, otherwise looking good! :100:

vector/src/main/java/im/vector/app/features/call/transfer/CallTransferViewModel.kt
langleyd: Yea doesn't do much but it still listens for the call completion and sends a `CallTransferViewEvents.Complete`, which is still required and feels like correct place for this. So I would tend to favour consistency and what somebody reading for the first time might find expectable vs optimising for less classes. But totally happy to put it somewhere else if it isn't shoehorned. :)

vector/src/main/java/im/vector/app/features/call/transfer/CallTransferActivity.kt
langleyd: Well actually there was 2 parts to the fix, this fixed the second part. I reformatted the description of the PR to make ti clearer what the 2 parts were.

vector/src/main/java/im/vector/app/features/call/transfer/CallTransferViewModel.kt
ouchadam: no strong opinions from my side, would agree that using a consistent pattern outweighs the benefit of removing a little bit of boilerplate

bmarty (approved):
LGTM, i have push a small cleanup commit.

Issue #5149, opened 2022-02-03
Reactions appear cut off


Throlo:
![Screenshot_20220203-161812_Element_1](https://user-images.githubusercontent.com/82319992/152414699-f968568d-270c-44a1-b2b0-da16a5b50673.jpg)

The reactions appear cut off regardless of whether it is a room or a direct chat.

kojid0:
Cannot reproduce. Your emoji looks way bigger than mine. What font size did you select in settings?

kojid0:
> "Largest"

Alright, can reproduce now. Looks like the button doesn't resize along with the other elements.

ganfra:
Closed by #5204

Issue #4674, opened 2021-12-09
Reactions on a message are limited to 8 distinct reactions.


bmarty:
We apply a maximum of 8 different displayed reactions on the timeline, with no clue that more reactions have been added to the message. This is here: https://github.com/vector-im/element-android/blob/main/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/AbsBaseMessageItem.kt#L75

All the reactions are visible in the bottom sheet after long click on the message and "View Reactions" though.

|Timeline|BottomSheet|
|-|-|
|<img width="292" alt="image" src="https://user-images.githubusercontent.com/3940906/145490684-de9cd808-a76d-49b6-b1c7-3cf35d204e3e.png">|<img width="292" alt="image" src="https://user-images.githubusercontent.com/3940906/145490715-d60e376b-113a-4538-a3c6-1f071c701ca9.png">|

Element Web has a "show all" action, which displays all the reactions on the timeline.

|Limited|Expanded|
|-|-|
|<img width="486" alt="image" src="https://user-images.githubusercontent.com/3940906/145490796-e9d2feb3-3ccc-41e4-bb10-95b6d2dd2ff9.png">|<img width="567" alt="image" src="https://user-images.githubusercontent.com/3940906/145490776-f81d331b-0fb8-4b01-b7a8-124c81141c58.png">|

On mobile we could add the "show all" action and maybe just open the bottom sheet with all the reactions.

ganfra:
We are also supposed to add a quick action to add more reactions : https://github.com/vector-im/element-android/issues/3344

daniellekirkwood:
Tagging Message Bubbles as we will implement this as a part of that initiative

ganfra:
Waiting for @amshakal inputs

daniellekirkwood:
Tagging as Needs-Design and awaiting for Figma link/direction from @amshakal  :)

amshakal:
Hi great suggestion. I think we should add the capability for a user to expand and see all reactions by both methods - show all as well as the bottom sheet on long press, giving people the capability to go deeper into 'who-sent-what-reaction' as they wish to. 

I also think we should also as a button to add reaction as suggested in https://github.com/vector-im/element-android/issues/3344.
<img width="309" alt="Screenshot 2022-02-08 at 7 22 18 pm" src="https://user-images.githubusercontent.com/85162587/153060459-1d9fd3e1-d637-44e1-990a-e23cb1abff29.png">

[...]

Issue #3344, opened 2021-05-14
'add reaction' interaction on the row of reactions to encourage more (web parity)


BillCarsonFr:
### First point

Actual 
<img width="158" alt="image" src="https://user-images.githubusercontent.com/9841565/118264605-36b8e080-b4b8-11eb-9ee8-f9fcb266a412.png">

After
<img width="274" alt="image" src="https://user-images.githubusercontent.com/9841565/118264673-53551880-b4b8-11eb-8b6d-ef7a419cc6b8.png">

### Second point

Take the opportunity to think again on the bottom sheet of a message?

Currently we have 

<img width="287" alt="image" src="https://user-images.githubusercontent.com/3940906/140287251-2109a6d1-112a-45e0-a34b-f3a8cfda4773.png">

"Add Reaction" action is far from the quick reactions row, especially when this is a message sent by the user (since they can also edit and remove)

daniellekirkwood:
Tagging Message Bubbles as we will implement this as a part of Bubbles project

ganfra:
Done on #5204

PR #5204, opened 2022-02-10
Feature/fga/reactions UI improvements

ganfra:
This PR improves the UI of reactions by handling #3344 #4674 and #5149
It also fixes reactions order on Bubble and UI echo.

| COLLAPSED |EXPANDED | 
| --- | --- |
Screenshot_20220210-190439|Screenshot_20220210-190444


amshakal:
Looks good! 

The designs were made with the assumption that the text in each reaction is black in colour. Can we match the colour and style of '3 more' and 'show less' to match the text style used for reactions? i.e. grey and bold text?

ganfra:
> will this also fix #5149 ?

Oh yeah probably!

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/AbsBaseMessageItem.kt
```diff
@@ -83,23 +113,36 @@ abstract class AbsBaseMessageItem<H : AbsBaseMessageItem.Holder> : BaseEventItem
                 reactionButton.isEnabled = reaction.synced
                 holder.reactionsContainer.addView(reactionButton)
             }
+            if (reactions.count() > MAX_REACTIONS_TO_SHOW) {
+                val showReactionsTextView = createReactionTextView(holder, 6)
```
ouchadam, line 117: could be handy to have some named arguments for the hardcoded values for some extra context

ouchadam (approved):
LGTM as well! :100:

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/AbsBaseMessageItem.kt
ganfra: Actually, I'm changing this to a Style as it's shared with ReactionButton

amshakal (approved):
TY!

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageInformationData.kt
```diff
@@ -55,6 +54,16 @@ data class ReferencesInfoData(
...
 ) : Parcelable
 
+@Parcelize
+data class ReactionsSummaryData(
+        /*List of reactions (emoji,count,isSelected)*/
+        val reactions: List<ReactionInfoData>? = null,
+        val showAll: Boolean = false,
+        val onShowMoreClicked: () -> Unit,
+        val onShowLessClicked: () -> Unit,
+        val onAddMoreClicked: () -> Unit
```
bmarty, line 64: Weird to see some lambda in data class

vector/src/main/res/values/strings.xml
```diff
@@ -3779,4 +3779,7 @@
     <string name="tooltip_attachment_poll">Create poll</string>
     <string name="tooltip_attachment_location">Share location</string>
 
+    <string name="message_reaction_show_less">Show less</string>
+    <string name="message_reaction_show_more">"%1$d more"</string>
```
bmarty, line 3783: You should have used a `plurals` :/

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageInformationData.kt
ouchadam: not sure what will happen when we attempt to parcelize :thinking:

PR #5229, opened 2022-02-14
Live Threads Timeline

ariskotsomitopoulos:
Integrate /relations API to create a live threads timeline.


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/Event.kt
```diff
@@ -201,7 +201,11 @@ data class Event(
      */
     fun getDecryptedTextSummary(): String? {
         if (isRedacted()) return "Message Deleted"
-        val text = getDecryptedValue() ?: return null
+        val text = getDecryptedValue() ?: run {
```
bmarty, line 204: The fun `getDecryptedTextSummary` is weird, there is no i18n, I think I want to delete it...

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/RealmSessionStoreMigration.kt
```diff
@@ -57,7 +58,7 @@ internal class RealmSessionStoreMigration @Inject constructor(
     override fun equals(other: Any?) = other is RealmSessionStoreMigration
     override fun hashCode() = 1000
 
-    val schemaVersion = 24L
+    val schemaVersion = 25L
```
bmarty, line 61: The value you want to set is 26, there was a mistake, the previous upgrade did not increment the version.

bmarty (commented):
Some remarks. Can you add a file for the changelog please?

Also @ganfra can you have a look on this PR too?

ariskotsomitopoulos:
You can add comments in this PR if you like but do not merge it yet, it is the part1 of a large PR that will also include live thread list. So whatever comments I see will be on the next PR.

I am doing that not have again one huge PR :)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/RealmSessionStoreMigration.kt
ariskotsomitopoulos: Thanks yes, I already fixed that in the` live thread list` that will also contain this PR

PR #5233, opened 2022-02-14
Enhance RealmMigrator to dynamically calculate schema version

ariskotsomitopoulos:
Removes some boilerplate code to avoid writing the number in every migration

**Before**
`class MigrateAuthTo003(realm: DynamicRealm) : RealmMigrator(realm) {
`

**After**
`class MigrateAuthTo003(realm: DynamicRealm) : RealmMigrator(realm,3) {
`

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/database/RealmMigrator.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/database/RealmMigrator.kt
@@ -18,10 +18,13 @@ package org.matrix.android.sdk.internal.util.database
 
 import io.realm.DynamicRealm
 import io.realm.RealmObjectSchema
+import org.matrix.android.sdk.internal.util.database.exceptions.RealmSchemaVersionException
 import timber.log.Timber
 
-abstract class RealmMigrator(private val realm: DynamicRealm,
-                             private val targetSchemaVersion: Int) {
+abstract class RealmMigrator(private val realm: DynamicRealm) {
+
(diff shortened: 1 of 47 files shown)


ariskotsomitopoulos:
I don't have a strong point to merge it.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/database/RealmMigrator.kt
```diff
@@ -18,10 +18,13 @@ package org.matrix.android.sdk.internal.util.database
...
 import io.realm.DynamicRealm
 import io.realm.RealmObjectSchema
+import org.matrix.android.sdk.internal.util.database.exceptions.RealmSchemaVersionException
 import timber.log.Timber
 
-abstract class RealmMigrator(private val realm: DynamicRealm,
-                             private val targetSchemaVersion: Int) {
+abstract class RealmMigrator(private val realm: DynamicRealm) {
+
+    private val targetSchemaVersion = calculateSchemaVersion(this::class.java.simpleName)
```
ouchadam, line 26: this will break if any forks have R8/proguard class name obfuscating enabled, might be worth adding a proguard rule to be 100% sure the names are kept (if we want to use reflection)

ariskotsomitopoulos: Yes, or we can skip it completely its fine by me.

ariskotsomitopoulos:
No point to merge it while this number is just for logging

Issue #5123, opened 2022-02-02
Add completion for @room as per element web


bmarty:
When typing "@r" to a room I should be proposed a completion for "@room", if my power level let me notify the room like this.

There are also sections in the completion popup that we could also introduce, see below:

<img width="1061" alt="image" src="https://user-images.githubusercontent.com/3940906/152116983-c4d9f4af-b349-496d-871b-371022c316db.png">

mnaturel:
Hi @amshakal , can you review the design part? I have added screenshots (both in light and dark theme) in #5237 . Thank you.

PR #5237, opened 2022-02-15
#5123: @room tag suggestion

mnaturel:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Adding possibility to notify everyone in a room using "@room" tag. The ability to notify everyone depends on the user permission level in the room.

## Motivation and context

See #5123 

## Screenshots / GIFs

TODO

## Tests

- Go to a room screen where you have enough permission to notify everyone (e.g. a room you created)
- Type "@" 
- Check that "@room" appears in the suggestion list in a dedicated section 
- Check the suggestion list corresponds to what is typed
- Select the "@room" tag
- Check the tag appears as a "chip"
- Write a message
- Send the message
- Check all users of the room are notified of the message
- Go to a room screen where you don't have enough permission to notify everyone
- Type "@" 
- Check that "@room" does not appear in the suggestion list

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 11

## Checklist


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/MatrixItem.kt
```diff
@@ -35,7 +35,20 @@ sealed class MatrixItem(
...
-        MatrixItem(id, displayName?.removeSuffix(ircPattern), avatarUrl) {
+            MatrixItem(id, displayName?.removeSuffix(IRC_PATTERN), avatarUrl) {
+        init {
+            if (BuildConfig.DEBUG) checkId()
+        }
+
+        override fun updateAvatar(newAvatar: String?) = copy(avatarUrl = newAvatar)
+    }
+
+    // TODO is it correct to represent it by a Matrix Item ? => to confirm
```
mnaturel, line 46: I was asking myself if the "@room" tag can be represented as a `MatrixItem` since in messages, it is just raw text and it is not converted to any matrix item link like others. I choose to represent it as `MatrixItem` since it was the easiest way to implement it.

amshakal:
Hello! Looks good to me. Just wondering if we need the title 'room notifications' here and if it's redundant.
154087389-8038cbd7-0356-4e53-bd90-f9113d4333ed

mnaturel:
> Hello! Looks good to me. Just wondering if we need the title 'room notifications' here and if it's redundant. 154087389-8038cbd7-0356-4e53-bd90-f9113d4333ed

Hi! I copied the same UX as in the web. When there is only "notify everyone" tag the header section is still displayed in the web. We have added headers since there may be some users as well in a dedicated section. However, we could hide the headers when there are only either users or room notification. In this case no need for headers. And when we have both types of tag, there will be headers. What do you think?

amshakal:
Exactly! Let's show it only when we have both. :) TY!

amshakal:
Okay, that makes sense, thanks for sharing the annimation, it helps to look at it change from users + room to only room. Let's keep the header then, approving the PR :)

bmarty:
OK, thanks, so the plan is to hide the headers when the user does not have the permission to send `@room`.

@mnaturel can you do the change please?

Also I want to highlight that the original issue was just to add completion for `@room` and you actually also add support for `@room` pills. This is great!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/MatrixItem.kt
bmarty: This approach is fine to me, thanks!

mnaturel:
> OK, thanks, so the plan is to hide the headers when the user does not have the permission to send `@room`.
> 
> @mnaturel can you do the change please?
> 
> Also I want to highlight that the original issue was just to add completion for `@room` and you actually also add support for `@room` pills. This is great!

Done, I have changed the header visibility rule.

PR #5239, opened 2022-02-15
Fixing missing identify properties 

### LINKED ISSUE #5234: Missing identity tracking during onboarding
when calling `Posthog.identify` during the onboarding, the `use case` property is cached within `posthog` but is not sent when the user consents to the tracking. 

As we have the use case value stored locally, we can include the use case selection as part of the initial identify when the app launches.

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Re-applies specified `UserProperties` each time the client does an initial `identify` (cold launches and on providing tracking consent)   
- Lifts the `GlobalScope` and `PostHog` creation to dagger providers for testability
- Adds unit tests around the `DefaultVectorAnalytics` and  `LateInitUserPropertiesFactory` which involves creating a bunch of fakes to enable testing

## Motivation and context

Fixes #5234 

Fixes the missing identity properties from within the onboarding flow (pre consent) specifically the onboarding use case. This was caused by `posthog.identify` being called whilst the client was opt'd out which unfortunately doesn't send the cached properties once consent is given, the fix is for us to manually send the stored properties as part of the initial identify which occurs after the user has consented for the first time or after during cold launch after consenting 

```json
"$set": {
  "ftueUseCaseSelection": "PersonalMessaging",
 ....
}
```

## Screenshots / GIFs

No UI changes

## Tests

Whilst observing https://posthog-poc.lab.element.dev/events, creating a new account and check the properties of the `identify` event

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Sv2 31

### DIFF
--- a/vector/src/main/java/im/vector/app/features/analytics/impl/DefaultVectorAnalytics.kt
+++ b/vector/src/main/java/im/vector/app/features/analytics/impl/DefaultVectorAnalytics.kt
@@ -16,13 +16,11 @@
 
 package im.vector.app.features.analytics.impl
 
-import android.content.Context
 import com.posthog.android.Options
 import com.posthog.android.PostHog
 import com.posthog.android.Properties
-import im.vector.app.BuildConfig
-import im.vector.app.config.analyticsConfig
 import im.vector.app.core.di.NamedGlobalScope
+import im.vector.app.features.analytics.AnalyticsConfig
 import im.vector.app.features.analytics.VectorAnalytics
 import im.vector.app.features.analytics.itf.VectorAnalyticsEvent
 import im.vector.app.features.analytics.itf.VectorAnalyticsScreen
@@ -42,8 +40,9 @@ private val IGNORED_OPTIONS: Options? = null
 
 @Singleton
 class DefaultVectorAnalytics @Inject constructor(
-        private val context: Context,
+        private val postHogFactory: PostHogFactory,
         private val analyticsStore: AnalyticsStore,
+        private val analyticsConfig: AnalyticsConfig,
         @NamedGlobalScope private val globalScope: CoroutineScope
 ) : VectorAnalytics {
     private var posthog: PostHog? = null
@@ -85,9 +84,9 @@ class DefaultVectorAnalytics @Inject constructor(
     }
 
     override fun init() {
+        createAnalyticsClient()
         observeUserConsent()
         observeAnalyticsId()
-        createAnalyticsClient()
     }
 
     private fun observeAnalyticsId() {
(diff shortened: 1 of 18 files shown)


vector/src/main/java/im/vector/app/features/analytics/impl/DefaultVectorAnalytics.kt
```diff
@@ -41,15 +40,30 @@ private val IGNORED_OPTIONS: Options? = null
...
+            null
+        }
+    }
 
     // Cache for the store values
     private var userConsent: Boolean? = null
     private var analyticsId: String? = null
 
+    override fun init() {
+        observeUserConsent()
```
ouchadam, line 63: `createAnalyticsClient` has been removed from the custom init and inlined straight into the class field to help simplify the flow

previously `createAnalyticsClient` was also calling 

```kotlin
optOutPostHog()
identifyPostHog()
```
 
however, these calls were being made before values were set for the `userConsent` and `analyticsIds`, making them noops

```kotlin
private var userConsent: Boolean? = null
private var analyticsId: String? = null
...

// optOutPostHog
userConsent?.let { posthog?.optOut(!it) }

// identifyPostHog
val id = analyticsId ?: return


```

vector/src/main/java/im/vector/app/features/analytics/impl/DefaultVectorAnalytics.kt
```diff
@@ -82,85 +96,41 @@ class DefaultVectorAnalytics @Inject constructor(
...
-    private fun identifyPostHog() {
+    private suspend fun identifyPostHog() {
         val id = analyticsId ?: return
         if (id.isEmpty()) {
             Timber.tag(analyticsTag.value).d("reset")
             posthog?.reset()
         } else {
             Timber.tag(analyticsTag.value).d("identify")
-            posthog?.identify(id)
+            posthog?.identify(id, lateInitUserPropertiesFactory.createUserProperties()?.getProperties()?.toPostHogUserProperties(), IGNORED_OPTIONS)
```
ouchadam, line 116: this line is the fix, specifically creating and including the late init user properties

vector/src/test/java/im/vector/app/test/fakes/FakePostHog.kt
```diff
@@ -0,0 +1,81 @@
...
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.mockkStatic
+import io.mockk.verify
+
+class FakePostHog {
+
+    init {
+        // workaround to avoid PostHog.HANDLER failing
+        mockkStatic(Looper::class)
```
ouchadam, line 32: this workaround is caused by `mockk` being unable to stop java static fields from being run and posthog has a

```java
  static final Handler HANDLER =
      new Handler(Looper.getMainLooper()) {
        @Override
        public void handleMessage(Message msg) {
          throw new AssertionError("Unknown handler message received: " + msg.what);
        }
      };
```

PR #5257, opened 2022-02-18
track screen event when user enters the screen 

### LINKED ISSUE #5256: Track Screen analytics events on screen enter
### Your use case

Currently we track screen event when user leaves the screen to measure duration. It causes some problems like sidebar only tracked when it's closed and we don't close it when user selects somethings, so userpaths in analytics will be wrong. Also it's much easier to debug and test analytics if screens are tracked immediately.

_No response_

### Additional context

_No response_

fedrunov:
closes #5256


vector/src/main/java/im/vector/app/features/analytics/screen/ScreenEvent.kt
```diff
@@ -25,8 +24,6 @@ import timber.log.Timber
  * Track a screen display. Unique usage.
  */
 class ScreenEvent(val screenName: MobileScreen.ScreenName) {
```
bmarty, line 26: The class `ScreenEvent`has been created to measure the time the screen is displayed. Now that we do not do that anymore, I would vote to just remove it.

bmarty (changes requested):
Con you remove `ScreenEvent` please?

Issue #5218, opened 2022-02-12
App crash on homeserver registration


yatadev:
### Steps to reproduce

1. Where are you starting? What can you see?
- Try to register on my homeserver with registration tokens enabled
- A modal pops up, asking if i want to open the web client for registration
- The app crashes if i click yes

#### What did you expect?
Browser should open the homeserver url
#### What happened instead?
App crash

### Your phone model

Samsung galaxy s8+

### Operating system version

Android 9

### Application version and app store

Element v1.3.18, App Store

### Homeserver

Synapse v1.52.0

### Will you send logs?

No

bmarty:
I can repro the crash by hacking the code (adding `!` at [this line](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthVariant.kt#L127)).

```
Caused by: java.lang.ClassCastException: im.vector.app.features.onboarding.OnboardingActivity cannot be cast to im.vector.app.features.signout.soft.SoftLogoutActivity
        at im.vector.app.features.signout.soft.SoftLogoutViewModel$Companion.initialState(SoftLogoutViewModel.kt:59)
```

Adding it to our milestone

mnaturel:
Hi @ouchadam @bmarty , in order to find the best fix for that crash, I am wondering the purpose of the`SoftLogoutViewModel`. I see it is used in [FtueAuthWebFragment](https://github.com/vector-im/element-android/blob/f7aedff418cbec3a3f66c71b561a2b6e36d37699/vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthWebFragment.kt) and [LoginWebFragment](https://github.com/vector-im/element-android/blob/f7aedff418cbec3a3f66c71b561a2b6e36d37699/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt) and [LoginWebFragment2](https://github.com/vector-im/element-android/blob/f7aedff418cbec3a3f66c71b561a2b6e36d37699/vector/src/main/java/im/vector/app/features/login2/LoginWebFragment2.kt).

I am asking the question because the `session` is injected into ViewModel meaning user should be authenticated, right? But for what I understand, it is used for sessionRecovery purpose which may indicating user is no more authenticated. And in the onboarding case I guess user will never be authenticated.

mnaturel:
[...]

PR #5258, opened 2022-02-18
#5218: Fix crash at registration when redirecting to Web View

mnaturel:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Removing usage of `SoftLogoutViewModel` in `FtueAuthWebFragment` since it seems the code which is using it cannot be reached.
- Changing the way we retrieve the current active `Session` to initialize the `ViewState` used in `SoftLogoutViewModel`: there is no more reference to the `Activity` using the `ViewModel` which caused the crash (ClassCastException).

## Motivation and context

See #5218.

## Tests

- Edit temporarily the code to reproduce the flow leading to web view redirection as explained in [this comment](https://github.com/vector-im/element-android/issues/5218#issuecomment-1042746067).
- Start onboarding
- Choose to create a new account on "Matrix.org" server for example
- Wait to see the redirection dialog
- Choose "yes" to be redirected to web view
- Check the app does not crash and displays the registration web view

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 11

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
@@ -51,8 +51,6 @@ class LoginWebFragment @Inject constructor(
         private val assetReader: AssetReader
 ) : AbstractLoginFragment<FragmentLoginWebBinding>() {
 
-    // TODO I noticed in the Git history this variable was a local variable inside notifyViewModel method
-    // TODO was there any reason ? To be able to create this ViewModel we must be sure we will have a valid session
     private val softLogoutViewModel: SoftLogoutViewModel by activityViewModel()
 
     override fun getBinding(inflater: LayoutInflater, container: ViewGroup?): FragmentLoginWebBinding {
--- a/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
@@ -51,6 +51,7 @@ class LoginWebFragment @Inject constructor(
         private val assetReader: AssetReader
 ) : AbstractLoginFragment<FragmentLoginWebBinding>() {
 
+    // TODO confirm the need of this viewModel
     val softLogoutViewModel: SoftLogoutViewModel by activityViewModel()
 
     override fun getBinding(inflater: LayoutInflater, container: ViewGroup?): FragmentLoginWebBinding {
--- a/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
@@ -72,7 +72,6 @@ class LoginWebFragment @Inject constructor(
     override fun updateWithState(state: LoginViewState) {
         setupTitle(state)
 
-        // TODO check how it is possible to arrive in this case
         isForSessionRecovery = state.deviceId?.isNotBlank() == true
 
         if (!isWebViewLoaded) {
--- a/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
@@ -51,8 +51,9 @@ class LoginWebFragment @Inject constructor(
         private val assetReader: AssetReader
 ) : AbstractLoginFragment<FragmentLoginWebBinding>() {
 
-    // TODO confirm the need of this viewModel
-    val softLogoutViewModel: SoftLogoutViewModel by activityViewModel()
+    // TODO I noticed in the Git history this variable was a local variable inside notifyViewModel method
+    // TODO was there any reason ? To be able to create this ViewModel we must be sure we will have a valid session
+    private val softLogoutViewModel: SoftLogoutViewModel by activityViewModel()
 
     override fun getBinding(inflater: LayoutInflater, container: ViewGroup?): FragmentLoginWebBinding {
         return FragmentLoginWebBinding.inflate(inflater, container, false)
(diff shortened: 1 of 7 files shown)


vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
```diff
@@ -51,7 +51,9 @@ class LoginWebFragment @Inject constructor(
         private val assetReader: AssetReader
 ) : AbstractLoginFragment<FragmentLoginWebBinding>() {
 
-    val softLogoutViewModel: SoftLogoutViewModel by activityViewModel()
+    // TODO I noticed in the Git history this variable was a local variable inside notifyViewModel method
+    // TODO was there any reason ? To be able to create this ViewModel we must be sure we will have a valid session
+    private val softLogoutViewModel: SoftLogoutViewModel by activityViewModel()
```
mnaturel, line 56: I am wondering if creating `SoftLogoutViewModel` is safe here since it needs a valid active `Session` to be created. Otherwise, it will crash. Inspecting the Git history, I see it was once created only when needed SoftLogout which may be safer. Should we do something?

vector/src/main/java/im/vector/app/features/signout/soft/SoftLogoutViewModel.kt
```diff
@@ -56,15 +57,23 @@ class SoftLogoutViewModel @AssistedInject constructor(
...
-                    homeServerUrl = activity.session.sessionParams.homeServerUrl,
-                    userId = userId,
-                    deviceId = activity.session.sessionParams.deviceId ?: "",
-                    userDisplayName = activity.session.getUser(userId)?.displayName ?: userId,
-                    hasUnsavedKeys = activity.session.hasUnsavedKeys()
-            )
+            val sessionHolder = EntryPoints.get(viewModelContext.app(), SingletonEntryPoint::class.java)
+                    .activeSessionHolder()
+
+            return if (sessionHolder.hasActiveSession()) {
```
ouchadam, line 63: :100: this fix makes sense to me, keeping the softlogout logic in the logout viewmodel

vector/src/main/java/im/vector/app/features/signout/soft/SoftLogoutViewModel.kt
```diff
@@ -56,15 +57,23 @@ class SoftLogoutViewModel @AssistedInject constructor(
...
-            val activity: SoftLogoutActivity = (viewModelContext as ActivityViewModelContext).activity()
-            val userId = activity.session.myUserId
-            return SoftLogoutViewState(
-                    homeServerUrl = activity.session.sessionParams.homeServerUrl,
-                    userId = userId,
-                    deviceId = activity.session.sessionParams.deviceId ?: "",
-                    userDisplayName = activity.session.getUser(userId)?.displayName ?: userId,
-                    hasUnsavedKeys = activity.session.hasUnsavedKeys()
-            )
+            val sessionHolder = EntryPoints.get(viewModelContext.app(), SingletonEntryPoint::class.java)
```
ouchadam, line 60: I'm not sure this is the correct way to inject into the initial state, we can provide other instances via https://github.com/vector-im/element-android/pull/5258/files#diff-a57f93c54f89c6899826efa9c728f09e6172d330131c1ae6a33e2ac1cfb26a5fR54

mnaturel: I copy/paste the same logic I found in other `initialState(viewModelContext: ViewModelContext)` methods. But I don't understand your suggestion. Can you explain it more, please?

ouchadam: ah no worries, I didn't realise `hiltMavericksViewModelFactory` was our own helper which also directly interacts with the `EntryPoints`

I was imagining the `@Assisted initialState: SoftLogoutViewState` would allow us to do... 

```kotlin
class Factory(activeSessionHolder) {
  fun create() = InitialState(activeSessionHolder.foo)
}
```

vector/src/main/java/im/vector/app/features/signout/soft/SoftLogoutViewModel.kt
```diff
@@ -56,15 +57,23 @@ class SoftLogoutViewModel @AssistedInject constructor(
...
+
+                SoftLogoutViewState(
+                        homeServerUrl = session.sessionParams.homeServerUrl,
+                        userId = userId,
+                        deviceId = session.sessionParams.deviceId.orEmpty(),
+                        userDisplayName = session.getUser(userId)?.displayName ?: userId,
+                        hasUnsavedKeys = session.hasUnsavedKeys()
+                )
+            } else {
+                null
```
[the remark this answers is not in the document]
mnaturel: In fact, Mavericks handles null viewState: the doc says: 
`Returns:
the initial state. If null, the state class constructors will be used for initial state creation.`
But I fixed it since there is no default constructor with empty arguments. With null it crashes. Thanks! 
However, here `SoftLogoutViewModel` cannot be instantiated if there is no active `Session` so this `else` branch should never be reached... And this is why, I removed it from `FtueAuthWebFragment`.

bmarty (approved):
I have done some test and the fallback to register is not crashing anymore.
Note that it does not work on matrix.org (see https://github.com/matrix-org/synapse/issues/7676). But maybe on other homeserver it can work, so Element Android should still support it (and not crash!)

Testing a bit the soft logout, there is no crash, but there is also (another) issue on matrix.org, I will rise a dedicated issue to handle it.

Anyway do you want to cleanup the PR before merging?

vector/src/main/java/im/vector/app/features/login/LoginWebFragment.kt
bmarty: There is no crash according to my test

ouchadam: it's worth mentioning this fragment is only used when the onboarding is set to `OnboardingVariant.LEGACY`

ouchadam (approved):
assuming the feature behaves as expected, all good from me as well! :+1:

Issue #4164, opened 2021-10-05
Display a notification if there is unsent message(s)


bmarty:
User may not be aware that a message they have sent is blocked on the device, because of an error, because of no network coverage, etc.

This is what WhatsApp is displaying in such circumstances:

<img width="323" alt="image" src="https://user-images.githubusercontent.com/3940906/136000343-7a0b61a6-a4a6-4cec-894f-4438fa4f889e.png">

daniellekirkwood:
Great, thanks @bmarty - do you need content written, or do we have something already ?

Flam3z:
> What about...
> 
> If cause is due to network connection:
> 
> > Some messages not sent
> > Check your connection. Messages will send automatically when reconnected.
> 
> If the user needs to select Retry:
> 
> > Message not sent
> > There was a problem sending your message. Please try again.
> 
> If cause unknown:
> 
> > You message was not sent
> > Please try again.

Context within the notifcations is very good for the user at understanding what exactly went wrong instead of a message not sending for whatever reason that the user may not know.

PR #5287, opened 2022-02-22
#4164 Display a notification if there is unsent message(s)

ahmed-radhouane:
Signed-off-by: Ahmed Radhouane Belkilani <arbelkilani@gmail.com>

 
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- update postTask in EventSenderProcessorCoroutine.kt in order to catch exception and return new type of Cancelable
- adding new state MessageUnsentException to handle the new exception state and display the notification.
- adding the new notification in NotificationUtils.kt in order to handle the different messages depending on exception type.
- throw exception from related tasks.
- adding new strings resources depending on exception type.
- increase max lines to 2800 lines.
- adding changelog file.

## Motivation and context

After sending message and in case of error, we should show a notification informing user of cause of error. 

## Screenshots / GIFs
Network Error | Server Error | Unknown Error
--- | --- | ---
Capture d’écran 2022-02-21 à 17 37 33 | Capture d’écran 2022-02-21 à 17 37 02 | Capture d’écran 2022-02-21 à 17 40 13

## Tests

- Access to Room or DM
- Send a message with bad internet connection
- A notification appear if message was not sent. 

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Android 8.1, Android 9

## Checklist

### DIFF
--- a/changelog.d/4164.bugfix
+++ b/changelog.d/4164.bugfix
@@ -1,0 +1,1 @@
+Display a notification if there is unsent message(s)
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/Cancelable.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/Cancelable.kt
@@ -28,6 +28,8 @@ interface Cancelable {
     fun cancel() {
         // no-op
     }
+
+    data class Exception(val exception: Throwable) : Cancelable
 }
 
 object NoOpCancellable : Cancelable
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendEventTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendEventTask.kt
@@ -65,7 +65,6 @@ internal class DefaultSendEventTask @Inject constructor(
                 Timber.d("Event: $it just sent in ${params.event.roomId}")
             }
         } catch (e: Throwable) {
-//            localEchoRepository.updateSendState(params.event.eventId!!, SendState.UNDELIVERED)
             throw e
         }
     }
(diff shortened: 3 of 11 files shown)


BillCarsonFr (changes requested):
Hello, I just had a quick look and I am a bit surprised on how it's done.
Given that there is already a method called when a task did fail to be sent in the eventSender, you could probably just add some listener mechanism and a the app layer you could then decide to show or not a notification (if the app is backgrounded or you are in a different room, smth like that).

But this still won't work if the app is killed or "dozed". For that to work you might want to have some sort of worker that check for unsent event and if it's there since too long you could then deschedule it and show a notification? This could avoid sending a message  sent long after you wanted (if you entered airplane mode for several hours?)

ahmed-radhouane:
> Hello, I just had a quick look and I am a bit surprised on how it's done. Given that there is already a method called when a task did fail to be sent in the eventSender, you could probably just add some listener mechanism and a the app layer you could then decide to show or not a notification (if the app is backgrounded or you are in a different room, smth like that).
> 
> But this still won't work if the app is killed or "dozed". For that to work you might want to have some sort of worker that check for unsent event and if it's there since too long you could then deschedule it and show a notification? This could avoid sending a message sent long after you wanted (if you entered airplane mode for several hours?)

> Hello thank you for the replay. 
the event sender tells me that there is a failed message but doesn't describe which kind of error was, a network , server or other which was asked to do to change the error message. There where I found a difficulties. If there is another method that could provide it. I will be grateful if you direct me toward. 

For the state killed app. I'm 100% percent ok with you. I guess a service is needed like the one used to intercept incoming calls. If you have other examples also that can help me as a reference for a solution. It's more than welcome. 

We can make a call or you can send me direct message, if you can that we discuss it in more details. 

Thank you

bmarty:
As discussed IRL, we shouls probably rely on an Android Service to display the notification, even if the app is killed.

I let you dig more in this lead and try to provide a better solution for this complexe use case. Thanks!

bmarty:
Closing as we will implement that maybe differently.

PR #5292, opened 2022-02-22
#4319: Fix DM navigation in member profile screen

### LINKED ISSUE #4319: Direct message not initiated after telling the app to do so via the space member list
### Steps to reproduce

1. open a space 
2. use context menu of different space (not sure if the Space must be different)
3. list members 
4. tap on a member (you don't have a chat with <- maybe not required either) 
5. tap direct message 
6. see Space menu

#### What did you expect?

See DM opening

#### What happened instead?

See Space menu

https://user-images.githubusercontent.com/20560137/138476277-40a1cd59-bf91-439d-86ea-a3c3816f414e.mp4

### Your phone model

Fairphone 3+

### Operating system version

Android 10

Element version 1.3.4 [40103042] (G-b4036) from GitHub

### Homeserver

matrix.org

### Will you send logs?

No

mnaturel:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Moving the navigation handling to open direct message from `TimelineFragment` to `RoomMemberProfileFragment`.

## Motivation and context

Closes #4319 

## Tests

- Open drawer menu
- Click on a space menu
- Click on "List members" button
- Click on a member
- Click on "Direct message"
- Check the screen with corresponding direct message room is displayed
- Go to a room where there are different members
- Click on a member in the timeline
- Click on "Direct message"
- Check the screen with corresponding direct message room is displayed 

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 11

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileFragment.kt
@@ -127,6 +127,7 @@ class RoomMemberProfileFragment @Inject constructor(
                 is RoomMemberProfileViewEvents.ShareRoomMemberProfile      -> handleShareRoomMemberProfile(it.permalink)
                 is RoomMemberProfileViewEvents.ShowPowerLevelValidation    -> handleShowPowerLevelAdminWarning(it)
                 is RoomMemberProfileViewEvents.ShowPowerLevelDemoteWarning -> handleShowPowerLevelDemoteWarning(it)
+                is RoomMemberProfileViewEvents.OpenRoom                    -> handleOpenRoom(it)
                 is RoomMemberProfileViewEvents.OnKickActionSuccess         -> Unit
                 is RoomMemberProfileViewEvents.OnSetPowerLevelSuccess      -> Unit
                 is RoomMemberProfileViewEvents.OnBanActionSuccess          -> Unit
@@ -140,6 +141,10 @@ class RoomMemberProfileFragment @Inject constructor(
     private fun setupLongClicks() {
         headerViews.memberProfileNameView.copyOnLongClick()
         headerViews.memberProfileIdView.copyOnLongClick()
+    }
+
+    private fun handleOpenRoom(event: RoomMemberProfileViewEvents.OpenRoom) {
+        navigator.openRoom(requireContext(), event.roomId, null)
     }
 
     private fun handleShowPowerLevelDemoteWarning(event: RoomMemberProfileViewEvents.ShowPowerLevelDemoteWarning) {
@@ -297,8 +302,7 @@ class RoomMemberProfileFragment @Inject constructor(
     }
 
     override fun onOpenDmClicked() {
-        roomDetailPendingActionStore.data = RoomDetailPendingAction.OpenOrCreateDm(fragmentArgs.userId)
-        vectorBaseActivity.finish()
+        viewModel.handle(RoomMemberProfileAction.OpenOrCreateDm(fragmentArgs.userId))
     }
 
     override fun onJumpToReadReceiptClicked() {
--- a/vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileViewModel.kt
@@ -32,6 +32,7 @@ import im.vector.app.core.extensions.exhaustive
 import im.vector.app.core.mvrx.runCatchingToAsync
 import im.vector.app.core.platform.VectorViewModel
 import im.vector.app.core.resources.StringProvider
+import im.vector.app.features.createdirect.DirectRoomHelper
 import im.vector.app.features.displayname.getBestName
 import im.vector.app.features.home.room.detail.timeline.helper.MatrixItemColorProvider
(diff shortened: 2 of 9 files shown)


vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileFragment.kt
```diff
@@ -297,8 +302,7 @@ class RoomMemberProfileFragment @Inject constructor(
     }
 
     override fun onOpenDmClicked() {
-        roomDetailPendingActionStore.data = RoomDetailPendingAction.OpenOrCreateDm(fragmentArgs.userId)
-        vectorBaseActivity.finish()
+        viewModel.handle(RoomMemberProfileAction.OpenOrCreateDm(fragmentArgs.userId))
```
bmarty, line 305: This has the effect to add a room to the Activity stack. This is fine, but this is a side effect of this change.
So when the user will close the RoomDetailActivity with the DM, they will see the RoomMemberProfile Fragment again.

vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileViewModel.kt
```diff
@@ -167,9 +169,25 @@ class RoomMemberProfileViewModel @AssistedInject constructor(
...
+                directRoomHelper.ensureDMExists(action.userId)
+            } catch (failure: Throwable) {
+                _viewEvents.post(RoomMemberProfileViewEvents.Failure(failure))
+                return@launch
+            }
+            if (roomId != initialState.roomId) {
+                _viewEvents.post(RoomMemberProfileViewEvents.OpenRoom(roomId = roomId))
+            }
+        }
+    }
```
bmarty, line 189: This code is duplicate of the code in `TimelineViewModel`.
We should find a way to avoid code duplication in ViewModels (maybe out of scope of this PR)

bmarty (commented):
The fix is fine, but this is a change in the navigation. Is it confirmed that we want to do this change?

Maybe a video before / after showing the diff can help @daniellekirkwood to understand the diff.

vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileFragment.kt
mnaturel: Yes It was intended. In my opinion it is consistent to go back to member profile screen in term of navigation. But I will try to make a video to explain the changes as you suggested.

mnaturel:
> CC: @kittykat

I have just seen your comment. I have just added GIFs in the associated issue.

kittykat:
@mnaturel Can we make the back button go to the timeline instead of the user profile? (Please wait for confirmation from @niquewoodhouse before changing behaviour.)

My thoughts are that normally a user would want to message only one user from a memberlist rather than continuing to take other actions on that user's profile or on the member list.

vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileViewModel.kt
bmarty: Ah, my bad, sorry

mnaturel:
> @mnaturel Can we make the back button go to the timeline instead of the user profile? (Please wait for confirmation from @niquewoodhouse before changing behaviour.)
> 
> My thoughts are that normally a user would want to message only one user from a memberlist rather than continuing to take other actions on that user's profile or on the member list.

Hello @kittykat , to answer your question, I can explain why I changed the back navigation. There are 2 navigation paths leading to a room member profile:

1. From room timeline, click on a profile picture
2. From home, click on drawer menu, click on space settings, click on list members entry

For the case 1, I see why we would like to go back to timeline and it may be easy to handle technically. But for the case 2, which is the original bug we fix here, I don't understand why we would go back to a room timeline and if so, which room we would choose since the user is coming back from home screen.
This is the reason, I choose to be consistent in navigation paths and I kept the back stack history to make it simple in both cases.

So I may find a solution to handle case 1, but for case 2, I am not sure we should do something else.

bmarty:
Is it possible to fix #4319, so case 1, without changing the other existing navigation of case 2?

kittykat:
[I talked to mnaturel and current status of PR is Nique and I are considering best course of action. I would be leaning towards `Space header menu -> List members -> User -> Direct message -back button-> Space` for the second case.]

mnaturel:
> [I talked to mnaturel and current status of PR is Nique and I are considering best course of action. I would be leaning towards `Space header menu -> List members -> User -> Direct message -back button-> Space` for the second case.]

@kittykat Okay. Just to be sure I understood well, to sum up:

- when coming from a room timeline we want to go back to that timeline instead of the member profile.
- when coming from "List members" entry of a space menu settings, we want to go back to the home screen (see picture below) instead of the member profile.



Is that correct?

kittykat:
@mnaturel Please stay with your original proposal for now (the back button goes back screen by screen along the whole path) and we can review the interaction when we're not in a rush as we would ideally do some user testing around it because it's a big change.

bmarty (approved):
Thanks for the discussion.
We may have complain about the change of navigation when opening the DM from the room member list, but we will see!
It's worth noting that we can also open a DM from the timeline by clicking on a user avatar and choosing "Direct message". So I guess the long path is less used.

PR #5300, opened 2022-02-22
tracking number of spaces user joined

### LINKED ISSUE #5299: Add a user property for how many spaces the user is in
Add a user property for how many spaces the user is in

separated from #5013 to make it easier to review and discuss

fedrunov:
closes #5299


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
```diff
@@ -165,12 +170,28 @@ internal class SyncResponseHandler @Inject constructor(
         Timber.v("On sync completed")
         cryptoSyncHandler.onSyncCompleted(syncResponse)
 
+        monarchy.awaitTransaction {
+            trackUserSpaceNumber(it)
+        }
```
bmarty, line 175: Is it worth doing that treatment on **every** sync response?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/Session.kt
```diff
@@ -294,6 +294,11 @@ interface Session :
          */
         fun onNewInvitedRoom(session: Session, roomId: String) = Unit
 
+        /**
+         * Called after sync and provide number of spaces user joined to track in analytics
+         */
+        fun onJoinedSpaceCounted(session: Session, numSpaces: Int) = Unit
```
bmarty, line 300: Have you considered implementing that in the interface `StatisticEvent` instead? I think it will have more sense.

bmarty (commented):
Some thoughts about the implementation

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/Session.kt
fedrunov: Yes, I was thinking about it, but usage of `StatisticsEvent` is

    ```
override fun onStatisticsEvent(session: Session, statisticEvent: StatisticEvent) {
        statisticEvent.toListOfPerformanceTimer().forEach {
            analyticsTracker.capture(it)
        }
    }
```
So I'll either need to add some ugly workaround for a new Event to not resolve one specific case to `PerfomanceTimer` Event, or rework it completely. Wasn't sure that this change worth it. What do you think?

File: matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt (hunk 1 above)
Comment (fedrunov) [reply matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt:175]: I've tried to do that, but I haven't found a way to implement it in a better way. I thought to only do that when there are some rooms to join in Sync, but in that case we won't send initial count before user will join any room. And we can't access here any type of app-level cache, like SharedPreferences to validate if we've send it already, since this code is in sdk module. So to do that we will require to implement some interface and then provide implementation from app module, which looks overcomplicated for this. I thought that number of rooms and spaces is relatively small and so this counting should not take a long tome.

File: matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/Session.kt (hunk 2 above)
Comment (bmarty) [reply matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/Session.kt:300]: Yes, I have seen that, I guess we will have to update the code on the app side (change `statisticEvent.toListOfPerformanceTimer()`). But I think we will end up with a better SDK API, which is a higher priority. Can you do it please?

File: matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt (hunk 1 above)
Comment (bmarty) [reply matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt:175]: OK
@ganfra what do you think?

Comment (ganfra) [reply matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt:175]: Yes I don't think this should be done in the SDK at all if it's only for Analytics...
Maybe you can listen in the App a combination of SyncState and RoomSummary to know when it has changed?

Comment (ouchadam) [reply matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt:175]: +1 for avoiding the concept of tracking within the SDK, it feels like we should be able to register event listeners and track from the element side

File: vector/src/main/java/im/vector/app/AppStateHandler.kt [hunk 3]
```diff
@@ -125,11 +133,28 @@ class AppStateHandler @Inject constructor(
...
+                        observeSyncStatus(session)
                     }
                 }
                 .launchIn(coroutineScope)
     }
 
+    private fun observeSyncStatus(session: Session) {
+        session.getSyncStatusLive()
+                .asFlow()
+                .filterIsInstance<SyncStatusService.Status.IncrementalSyncDone>()
```
ouchadam, line 145: we can use `distinct` to avoid keeping track of the previous state https://github.com/vector-im/element-android/pull/5300/files#diff-d5584663de96dcd881e4441839a2a6e44b50a58268b0b9af4d17be21fba9897fR152

vector/src/main/java/im/vector/app/AppStateHandler.kt
```diff
@@ -125,11 +133,28 @@ class AppStateHandler @Inject constructor(
...
+                .onEach {
+                    handleJoinedSpaceStatistics(session.spaceService().getRootSpaceSummaries().size)
+                }.launchIn(session.coroutineScope)
```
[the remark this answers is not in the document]
fedrunov: Why do you think so? `observeSyncStatus` called when session is changed and it subscribes to a sync status.

vector/src/main/java/im/vector/app/AppStateHandler.kt
fedrunov: 👍  added one line below

vector/src/main/java/im/vector/app/AppStateHandler.kt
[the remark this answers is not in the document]
fedrunov: check latest changes, it's now a little bit different

ouchadam (approved):
LGTM! will leave to @ganfra  (and confirming the space switching behaviour)

ganfra (approved):
Ok, LGTM now, thanks

PR #5312, opened 2022-02-23
Log `since` token when doing an incremental sync

erikjohnston:
This aids debugging when `/sync` does something weird server side. Since we already log a line each time we do an incremental sync this shouldn't cause a problem.

---

I haven't tested this in any shape or form, nor do I know if this is *actually* where we want to add this log line. Sorry for the drive by PR, but found myself wanting this when investigating https://github.com/matrix-org/synapse/issues/11916.

### DIFF
--- a/changelog.d/5312.misc
+++ b/changelog.d/5312.misc
@@ -1,0 +1,1 @@
+Log the `since` token used when doing an incremental sync.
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, thanks!

PR #5318, opened 2022-02-23
Also record `next_batch` on `/sync` response

erikjohnston:
In the same vein as #5312.


changelog.d/5318.misc
```diff
@@ -0,0 +1 @@
+Log the `since` token used and `next_batch` token returned when doing an incremental sync.
```
bmarty, line 1: Only updating the other file is enough, but do not worry I will delete this one later.

bmarty (approved):
LGTM

changelog.d/5318.misc
erikjohnston: Ah, I figured I may as well combine the two PRs into a changelog entry (as they really should have been one PR), but YMMV :)

erikjohnston: (FTR towncrier merges multiple changelog files with the same content into a single line in the final changelog)

Issue #5200, opened 2022-02-10
Mobile FTUE: Profile Personalisation screens (Android)


ouchadam:
Android specific https://github.com/vector-im/element-meta/issues/132

- [x] Account created screen #5158 
- [x] Add support for additional homeserver capabilities (display name/avatar) #5208 
- [x] Choose a display name screen #5211
- [x] Add a profile picture screen #5323
- [x] Personalisation complete screen #5389
- [x] Avoid showing personalisation screens if the homeserver has disabled them #5375
- [x] Confirm error messaging and handling (local rules extracted to https://github.com/vector-im/element-meta/issues/204) #5517  
- [x] Enable feature - product/design sign off #5519 

Also:
- [ ] https://github.com/vector-im/element-meta/issues/132
- [x] https://github.com/vector-im/element-android/issues/5099

Issue #3633, opened 2021-07-05
[SDK] Creates inline classes to wrap UserId, RoomId...


ganfra:
We should wrap some primitive classes with inline classes (https://kotlinlang.org/docs/inline-classes.html#representation).
It'll allow us to have some extension methods just for these classes and not available on the primitive class.

PR #5323, opened 2022-02-23
FTUE - Choose a display picture

ouchadam:
## Type of change

- [x] WIP Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Introduces a profile picture personalisation screen to the onboarding flow
- Also extracts and tests the uri filename resolver since I was in the area 

## Motivation and context

To allow users to change their profile picture during the account creation 
Part of #5200 

## Screenshots / GIFs

| TINY | TALL | TABLET | GIF
| --- | --- | --- | --- |
Screenshot_20220223_154259|profile-tall|profile-tablet|after-profile-picture

## Tests

- With the personalisation feature flag enabled
- Create a new account

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s):

### DIFF
--- a/library/ui-styles/src/main/res/values-h480dp/dimens.xml
+++ b/library/ui-styles/src/main/res/values-h480dp/dimens.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item name="ftue_auth_profile_picture_height" format="float" type="dimen">0.15</item>
+    <item name="ftue_auth_profile_picture_icon_height" format="float" type="dimen">0.05</item>
+</resources>
--- a/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
@@ -47,6 +47,7 @@ import im.vector.app.features.login.ReAuthHelper
 import im.vector.app.features.login.ServerType
 import im.vector.app.features.login.SignMode
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import org.matrix.android.sdk.api.MatrixPatterns.getDomain
 import org.matrix.android.sdk.api.auth.AuthenticationService
--- a/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
@@ -149,6 +149,7 @@ class OnboardingViewModel @AssistedInject constructor(
             is OnboardingAction.PostViewEvent              -> _viewEvents.post(action.viewEvent)
             is OnboardingAction.UpdateDisplayName          -> updateDisplayName(action.displayName)
             OnboardingAction.UpdateDisplayNameSkipped      -> _viewEvents.post(OnboardingViewEvents.OnDisplayNameSkipped)
+            OnboardingAction.UpdateProfilePictureSkipped   -> TODO()
         }.exhaustive
     }
 
--- a/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
@@ -64,6 +64,7 @@ import org.matrix.android.sdk.api.failure.Failure
 import org.matrix.android.sdk.api.failure.MatrixIdFailure
 import org.matrix.android.sdk.api.session.Session
 import timber.log.Timber
+import java.util.UUID
 import java.util.concurrent.CancellationException
 
 /**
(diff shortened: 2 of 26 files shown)


library/ui-styles/src/main/res/values-h480dp/dimens.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item name="ftue_auth_profile_picture_height" format="float" type="dimen">0.15</item>
+    <item name="ftue_auth_profile_picture_icon_height" format="float" type="dimen">0.05</item>
```
bmarty, line 4: Those 2 values are the same than in default resource, is it a mistake?

vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
```diff
@@ -899,14 +904,59 @@ class OnboardingViewModel @AssistedInject constructor(
             val activeSession = activeSessionHolder.getActiveSession()
             try {
                 activeSession.setDisplayName(activeSession.myUserId, displayName)
-                setState { copy(asyncDisplayName = Success(Unit)) }
+                setState {
+                    copy(
+                            asyncDisplayName = Success(Unit),
```
bmarty, line 909: Can't it be `asyncDisplayName = Success(displayName)` instead of storing the value in `personalizationState`. The remark is also applicable to `selectedPictureUri` I guess.

vector/src/main/res/layout/fragment_ftue_profile_picture.xml
```diff
@@ -0,0 +1,166 @@
...
+                android:contentDescription="@string/ftue_profile_picture_title"
+                android:src="@drawable/ic_camera_plain"
+                app:layout_constraintBottom_toBottomOf="@id/pos"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintHeight_percent="0.55"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toTopOf="@id/pos"
+                app:tint="?vctr_content_secondary">
+
+            </ImageView>
```
bmarty, line 91: Could use `/>` for empty tag block :)

vector/src/test/java/im/vector/app/test/fakes/FakeSession.kt
```diff
@@ -34,10 +35,13 @@ class FakeSession(
         mockkStatic("im.vector.app.core.extensions.SessionKt")
     }
 
+    override val myUserId: String = "a-user-id"
```
bmarty, line 38: I am wondering if we should have a valid userId here like `@fake:server.fake`. WDYT?

bmarty (approved):
Thanks for adding all those tests!
A few remarks, but that's great work! Thanks!

library/ui-styles/src/main/res/values-h480dp/dimens.xml
ouchadam: ah great catch :eyes: ! they can be deleted, they were separated whilst checking with design about the different screen sizes, we chose to maintain the same _% of screen_  for small screens and allow the entire avatar to be clickable instead of making the image bigger

vector/src/main/res/layout/fragment_ftue_profile_picture.xml
ouchadam: will do :+1:

vector/src/test/java/im/vector/app/test/fakes/FakeSession.kt
ouchadam: sounds great to me! makes me wonder if a dedicated type for `UserId` or `MatrixId` could be helpful (might raise a ticket for a [value class](https://kotlinlang.org/docs/inline-classes.html) PoC)

vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
ouchadam: this is a good point, at the moment those asyncs are only being used for showing/hiding the loading view, in `Login2` they're replaced with a single `isLoading: Boolean`, I was thinking of doing the same but wanted to do all the asyncs at the same time  

WDYT?

library/ui-styles/src/main/res/values-h480dp/dimens.xml
ouchadam: https://github.com/vector-im/element-android/pull/5323/commits/1fc1935374c214c1e4fe7a9205c2cb248e7740db

vector/src/main/res/layout/fragment_ftue_profile_picture.xml
ouchadam: https://github.com/vector-im/element-android/pull/5323/commits/3a2dea2c3b9ffae4224715fa28acea406e90976b

vector/src/test/java/im/vector/app/test/fakes/FakeSession.kt
ouchadam: ah it already exists https://github.com/vector-im/element-android/issues/3633

https://github.com/vector-im/element-android/pull/5323/commits/deec923026c8a6d158e9396f961b8c1a8284e69f for the value update

vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
bmarty: That sounds like a plan! OK for me. What I want to avoid is having multiple sources of truth.

bmarty (approved):
Thanks for the update. I let you merge it!

vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
ouchadam: will raise a PR for it :100:

Issue #5327, opened 2022-02-23
Add concurrency to the GitHub workflows to auto-cancel older runs of some actions in PRs.


langleyd:
Within a particular PR we currently run actions for every commit. Some of these are very resource heavy like integration tests.
While we might want to run these for every commit on `develop` and `main` branches, within a PR we can probably afford to cancel jobs for all but the most recent commit.

For `main` and `develop` `github.sha` can be used in the group id to ensure every commit starts a job.
For feature branches `github.ref` in the group id to ensure all but the most recent job are cancelled.

Candidates for concurrency in PR branches are:
- Integration test
- Unit tests
- GPlay/FDriod Builds
- Quality checks

Example on element-ios -> https://github.com/vector-im/element-ios/pull/5101

Documentation on concurrency -> https://docs.github.com/en/actions/using-jobs/using-concurrency

PR #5328, opened 2022-02-23
Add concurrency to the GitHub workflows to auto-cancel older runs of actions in PRs.

langleyd:
Resolves https://github.com/vector-im/element-android/issues/5327


.github/workflows/quality.yml
```diff
@@ -18,6 +18,10 @@ jobs:
   ktlint:
     name: Kotlin Linter
     runs-on: ubuntu-latest
+    # Allow all jobs on main and develop. Just one per PR.
+    concurrency:
+      group: ${{ github.ref == 'refs/heads/main' && format('klint-main-{0}', github.sha) || github.ref == 'refs/heads/develop' && format('klint-develop-{0}', github.sha) || format('klint-{0}', github.ref) }}
```
michaelkaye, line 23:
```suggestion
      group: ${{ github.ref == 'refs/heads/main' && format('ktlint-main-{0}', github.sha) || github.ref == 'refs/heads/develop' && format('ktlint-develop-{0}', github.sha) || format('ktlint-{0}', github.ref) }}
```
s/klint/ktlint/ to match job name

.github/workflows/quality.yml
```diff
@@ -116,6 +124,10 @@ jobs:
       fail-fast: false
       matrix:
         target: [ Gplay, Fdroid ]
+    # Allow all jobs on develop. Just one per PR.
+    concurrency:
+      group: ${{ github.ref == 'refs/heads/develop' && format('apk-lint-develop-{0}', github.sha) || format('apk-lint-{0}-{1}', matrix.target, github.ref) }}
```
michaelkaye, line 129:
```suggestion
      group: ${{ github.ref == 'refs/heads/develop' && format('apk-lint-develop-{0}-{1}', matrix.target, github.sha) || format('apk-lint-{0}-{1}', matrix.target, github.ref) }}
```

We want to concurrently run Gplay and Fdroid in develop - they'll both have the same commit hash.

.github/workflows/tests.yml
```diff
@@ -15,6 +15,10 @@ jobs:
   unit-tests:
     name: Run Unit Tests
     runs-on: ubuntu-latest
+    # Allow all jobs on main and develop. Just one per PR.
+    concurrency:
+      group: ${{ github.ref == 'refs/heads/main' && format('unit-test-main-{0}', github.sha) || github.ref == 'refs/heads/develop' && format('unit-test-develop-{0}', github.sha) || format('unit-test-{0}', github.ref) }}
```
michaelkaye, line 20:
```suggestion
      group: ${{ github.ref == 'refs/heads/main' && format('unit-tests-main-{0}', github.sha) || github.ref == 'refs/heads/develop' && format('unit-tests-develop-{0}', github.sha) || format('unit-tests-{0}', github.ref) }}
```

s/unit-test/unit-tests/ to match job name.

michaelkaye (changes requested):
Intent of PR is great, looks good.

A few nits around naming (to keep consistency), but I think we have a correctness issue with the missing `matrix.target`

Also wondering if we should add comments for the other workflows as well (Even if it just says "sanity_test.yml: run every time") so we are reminded to consider it in future.

.github/workflows/quality.yml
langleyd: Ah you're dead right, good catch.

langleyd:
Nice suggestions @michaelkaye , thanks for the review!

michaelkaye (approved):
This looks good to me.

ouchadam:
out of interest, what's the reasoning for needing more than the head ref? `group: ${{ github.head_ref }}`

michaelkaye:
> out of interest, what's the reasoning for needing more than the head ref? `group: ${{ github.head_ref }}`

Groups cross workflows, so the per-workflow name is required to avoid concurrency blocking across workflows ; the instances where we use the SHA hash is to ensure multiple runs on the same workflow can happen; the instances where the matrix target is also included is when we want both sides of the target to run.

(never used this before, but that was my understanding when i reviewed)

ouchadam:
thanks for the explanation!

ouchadam (approved):

PR #5339, opened 2022-02-24
Render integration test results inline.

michaelkaye:
Render integration test results inline (including timing information for each test), while preserving the formatting for the PR comment integration.

### DIFF
--- a/tools/ci/render_test_output.py
+++ b/tools/ci/render_test_output.py
@@ -38,6 +38,6 @@ for xmlfile in xmlfiles:
             print(f"::error file={testname}::{message} in {time}s")
             print(child.text)
     body = f"passed={passed} failures={failures} errors={errors} skipped={skipped}"
-    print(f"::set-output name={suitename}::passed={body}"
+    print(f"::set-output name={suitename}::passed={body}")
     print("::endgroup::")
 
--- a/tools/ci/render_test_output.py
+++ b/tools/ci/render_test_output.py
@@ -10,7 +10,10 @@ import xml.etree.ElementTree as ET
 suitename = sys.argv[1]
 xmlfiles = sys.argv[2:]
 
+print(f"Arguments: {sys.argv}") 
+
 for xmlfile in xmlfiles:
+    print(f"Handling: {xmlfile}") 
     tree = ET.parse(xmlfile)
     
     root = tree.getroot()
--- a/tools/ci/render_test_output.py
+++ b/tools/ci/render_test_output.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python3
+#
+# Renders a list of xml files taken as arguments into GHA-styled messages in groups.
+# Explicitly aims not to have any dependencies, to reduce installation load.
+# Should just be able to run in the context of your standard github runner.
+
+import sys
+import xml.etree.ElementTree as ET
+
+xmlfiles= sys.argv[1:]
+
+for xmlfile in xmlfiles:
+    tree = ET.parse(xmlfile)
+    
+    root = tree.getroot()
+    name = root.attrib['name']
+    name = root.attrib['time']
+    success = int(root.attrib['tests']) - int(root.attrib['failures']) - int(root.attrib['errors'])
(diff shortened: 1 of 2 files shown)


bmarty (commented):
Could be nice to have a review from @ariskotsomitopoulos :)

ariskotsomitopoulos:
@michaelkaye  Nice work, it's more clear and scalable to have the logic in a separate python file, so even if we have one large run it will still work. I can see the error `FileNotFoundError` but this is expected due to the current error/fail of the integration tests

ariskotsomitopoulos (approved):

michaelkaye:
Rebased against develop so we can see the actual test run in this PR before merging.

michaelkaye:
```
  File "./tools/ci/render_test_output.py", line 20, in <module>
    passed = int(root.attrib['passed'])
KeyError: 'passed'
```

Glad I ran this again. Now i'm confused why this is doing this.

A xml file i have locally shows:
```
<testsuite name="im.vector.app.ui.UiAllScreensSanityTest" tests="1" failures="1" errors="0" skipped="0" time="66.418" timestamp="2022-02-14T11:31:50" hostname="localhost">
```

as the line that is being parsed by the sed, and the sed definitely produces a result that has `passed=` in it, so there must be a way for the testsuite XML to contain the attribute `passed=`, but when we test in our python there isn't. WTF?

I'm going to fall back to calculating passed based on the other fields, and return that.

michaelkaye:
> ```
>   File "./tools/ci/render_test_output.py", line 20, in <module>
>     passed = int(root.attrib['passed'])
> KeyError: 'passed'
> ```
> 
> Glad I ran this again. Now i'm confused why this is doing this.
> 
> A xml file i have locally shows:
> 
> ```
> <testsuite name="im.vector.app.ui.UiAllScreensSanityTest" tests="1" failures="1" errors="0" skipped="0" time="66.418" timestamp="2022-02-14T11:31:50" hostname="localhost">
> ```
> 
> as the line that is being parsed by the sed, and the sed definitely produces a result that has `passed=` in it, so there must be a way for the testsuite XML to contain the attribute `passed=`, but when we test in our python there isn't. WTF?
> 
> I'm going to fall back to calculating passed based on the other fields, and return that.

I've just worked out what's actually going on here, I was just not understanding how it was meant to work --

the passed= field was coming from the echo of the original script, not from the XML - so one part of me was reading it as "the thing you pass to the output, so it's passed=<body>", the other part of me was reading the end result and going "ah, that's a value called passed".

All worked out how, hopefully the next run will fix it.

michaelkaye:
"Code Quality Checks / Kotlin Linter (pull_request) Failing after 1m — Kotlin Linter" is failing but i don't think we touched it in this change. It seems generally stable, so unsure why it's unhappy now.

https://buildkite.com/matrix-dot-org/element-android/builds/6383#2b3b62e4-a499-4cc4-867e-9e41bd29bfee also failing on buildkite for similar reason. Perhaps upstream are having issues or the package has been deleted?

michaelkaye:
I was trying to find out where test runs had gone; but the previous version was showing "passed=<total tests of all results>" and is now showing "passed=<successful tests>" , so previously 
```
[org.matrix.android.sdk.session]
passed="25" failures="16" errors="0" skipped="2"
```
Is replaced by:
```
[org.matrix.android.sdk.session]
passed=6 failures=17 errors=0 skipped=2
```

`25 = 17 + 6 + 2` which is where the original passed=25 came from, I believe.

tools/ci/render_test_output.py
```diff
@@ -0,0 +1,45 @@
...
+
+# Potentially rewrite as an independent action, use handlebars to template result
+import sys
+import xml.etree.ElementTree as ET
+suitename = sys.argv[1]
+xmlfiles = sys.argv[2:]
+
+print(f"Arguments: {sys.argv}") 
+
+for xmlfile in xmlfiles:
```
ouchadam, line 15: for the future, this could make a fantastic independent action

ouchadam (approved):
great catch with the test output counts! :100:

PR #5345, opened 2022-02-24
Support both unstable and stable prefixes

### LINKED ISSUE #5340: Support both stable and unstable prefixes
### Steps to reproduce

Currently we only support, parse unstable prefixes which will cause problems when they become stable.

### Outcome

We need to continue sending unstable prefixes but be able to read stable prefixes too.

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

onurays:
Fixes #5340

### DIFF
--- a/changelog.d/5340.bugfix
+++ b/changelog.d/5340.bugfix
@@ -1,0 +1,1 @@
+Support both stable and unstable prefixes
--- a/changelog.d/5340.bugfix
+++ b/changelog.d/5340.bugfix
@@ -1,1 +1,1 @@
-Support both stable and unstable prefixes
+Support both stable and unstable prefixes for Events about Polls and Location
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLocationContent.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLocationContent.kt
@@ -39,37 +39,46 @@ data class MessageLocationContent(
          */
         @Json(name = "geo_uri") val geoUri: String,
 
+        @Json(name = "m.relates_to") override val relatesTo: RelationDefaultContent? = null,
+        @Json(name = "m.new_content") override val newContent: Content? = null,
         /**
          * See https://github.com/matrix-org/matrix-doc/blob/matthew/location/proposals/3488-location.md
          */
-        @Json(name = "org.matrix.msc3488.location") val locationInfo: LocationInfo? = null,
-
-        @Json(name = "m.relates_to") override val relatesTo: RelationDefaultContent? = null,
-        @Json(name = "m.new_content") override val newContent: Content? = null,
-
+        @Json(name = "org.matrix.msc3488.location") val unstableLocationInfo: LocationInfo? = null,
+        @Json(name = "m.location") val locationInfo: LocationInfo? = null,
+        /**
+         * Exact time that the data in the event refers to (milliseconds since the UNIX epoch)
+         */
+        @Json(name = "org.matrix.msc3488.ts") val unstableTs: Long? = null,
+        @Json(name = "m.ts") val ts: Long? = null,
+        @Json(name = "org.matrix.msc1767.text") val unstableText: String? = null,
+        @Json(name = "m.text") val text: String? = null,
         /**
          * m.asset defines a generic asset that can be used for location tracking but also in other places like
          * inventories, geofencing, checkins/checkouts etc.
          * It should contain a mandatory namespaced type key defining what particular asset is being referred to.
          * For the purposes of user location tracking m.self should be used in order to avoid duplicating the mxid.
          */
-        @Json(name = "m.asset") val locationAsset: LocationAsset? = null,
-
-        /**
-         * Exact time that the data in the event refers to (milliseconds since the UNIX epoch)
-         */
-        @Json(name = "org.matrix.msc3488.ts") val ts: Long? = null,
-
-        @Json(name = "org.matrix.msc1767.text") val text: String? = null
+        @Json(name = "org.matrix.msc3488.asset") val unstableLocationAsset: LocationAsset? = null,
+        @Json(name = "m.asset") val locationAsset: LocationAsset? = null
 ) : MessageContent {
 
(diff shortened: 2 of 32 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLocationContent.kt
```diff
@@ -39,29 +39,65 @@ data class MessageLocationContent(
...
-         * Exact time that the data in the event refers to (milliseconds since the UNIX epoch)
-         */
-        @Json(name = "org.matrix.msc3488.ts") val ts: Long? = null,
+    @Transient
+    var locationInfo: LocationInfo? = null
+    get() = stableLocationInfo ?: unstableLocationInfo
+    set(value) {
+        field = value
+        unstableLocationInfo = value
+    }
```
bmarty, line 58: This is too complex.

```kotlin
    @Json(name = "org.matrix.msc3488.location") val unstableLocationInfo: LocationInfo? = null
    @Json(name = "m.location") val stableLocationInfo: LocationInfo? = null
```

must be val and in the constructor.

then you can have a helper

```kotlin
fun getLocationInfo() = stableLocationInfo ?: unstableLocationInfo
```

Nothing more is needed.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/PollType.kt
```diff
@@ -27,9 +27,15 @@ enum class PollType {
...
+
     /**
      * Results should be only revealed when the poll is ended.
      */
     @Json(name = "org.matrix.msc3381.poll.undisclosed")
-    UNDISCLOSED
+    UNDISCLOSED,
+
+    @Json(name = "m.poll.undisclosed")
+    UNDISCLOSED_STABLE
```
bmarty, line 40: I would rather rename the other to `UNDISCLOSED_UNSTABLE` and name this one `UNDISCLOSED`, since at the end we will have only this one.

bmarty (changes requested):
Some remark.

On a more general thought, I think we should just use the stable values and do not bother with unstable ones, even if it will break history. Anyway unstable prefix are not supposed to be used long term, but this is a real mess I agree.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/PollType.kt
onurays: Done.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLocationContent.kt
onurays: Thank you, much more cleaner. We should always use unstable prefixes while MSC is not finalized.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLocationContent.kt
```diff
@@ -40,72 +40,45 @@ data class MessageLocationContent(
         @Json(name = "geo_uri") val geoUri: String,
 
         @Json(name = "m.relates_to") override val relatesTo: RelationDefaultContent? = null,
-        @Json(name = "m.new_content") override val newContent: Content? = null
+        @Json(name = "m.new_content") override val newContent: Content? = null,
+        /**
+         * See https://github.com/matrix-org/matrix-doc/blob/matthew/location/proposals/3488-location.md
+         */
+        @Json(name = "org.matrix.msc3488.location") val unstableLocationInfo: LocationInfo? = null,
+        @Json(name = "m.location") val stableLocationInfo: LocationInfo? = null,
```
bmarty, line 48: Please only use `unstable` prefix, and never use `stable` prefix.
Applicable to all the fileds in this PR.

bmarty (changes requested):
One more remark on the latest commit.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLocationContent.kt
onurays: We should be able to read the stable one too. When MSC is finalized and we start supporting stable prefixes then older versions should be able to read these fields. Do I miss something here?

bmarty: Yes, sorry, I was not clear.
What I want to see here is:
```kotlin
        @Json(name = "org.matrix.msc3488.location") val unstableLocationInfo: LocationInfo? = null,
        @Json(name = "m.location") val locationInfo: LocationInfo? = null,
```

`stableLocationInfo` -> `locationInfo`

onurays: Then we need an alternative naming for `fun getLocationInfo()`

onurays: What do you think about `getBestLocationInfo()`

bmarty: That's OK.
Ideally those management should be quite hidden in the SDK. But it's fine like this I guess.

bmarty (approved):
Thanks for the update!
Can be squashed and merged on develop

bmarty:
OK, I have force pushed the branch with your change, and added an extra commit.

changelog.d/5340.bugfix
```diff
@@ -0,0 +1 @@
+Support both stable and unstable prefixes
```
bmarty, line 1:
```suggestion
Support both stable and unstable prefixes for Events about polls
```

bmarty (approved):
LGTM! (there is still a code quality error to fix)

PR #5348, opened 2022-02-24
Upgrade to Android stem 2.0.0

### LINKED ISSUE #5341: Upgrade to LikeTheSalad/android-stem 2.0.0
Further instructions can be found here: https://github.com/LikeTheSalad/android-stem/issues/10#issuecomment-1046044368 : 

> The new version 2.0.0 is now available, it is no longer needed to define your templates by adding the "template_" to your string names anymore, which means that Lint shouldn't treat them as separate strings, thus avoiding the "translation" issue. Although, please bear in mind that a lot of things have changed in this new version 2.0.0, there's a rebranding along with new dependencies parameters and plugin ID, so in order to add the new version into your project, I recommend you to check again the README's instructions: https://github.com/LikeTheSalad/android-stem#21--changes-to-your-roots-buildgradle-file

- [x] Upgrade gradle file as per instruction in https://github.com/LikeTheSalad/android-stem#21--changes-to-your-roots-buildgradle-file
- [x] Strings key have to be renamed to remove the `template_` prefix. It has to be done during a Webate sync.
- [x] The code in lint.xml with `<ignore path="**/generated/resolved/**/resolved.xml" />` may be removed during the migration.

bmarty:
Fixes #5341 

Upgrade the dependency to generate strings using template.

It requires changing lots of strings key. But Weblate is currently locked, so this is fine.

1st and 3rd commit are interesting, the 2nd one is basically replacing `<string name="template_` by `<string name="` everywhere.

Tested OK on this screen for instance, where title and subtitle contains a template:

image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ buildscript {
         classpath 'com.google.gms:google-services:4.3.10'
         classpath 'org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:3.3'
         classpath 'com.google.android.gms:oss-licenses-plugin:0.10.4'
-        classpath "com.likethesalad.android:string-reference:1.2.2"
+        classpath "com.likethesalad.android:stem-plugin:2.0.0"
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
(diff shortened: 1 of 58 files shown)


ariskotsomitopoulos (approved):
Its seems fine to me

PR #5352, opened 2022-02-24
Delete bunch of unused strings - WIP

bmarty:
Remove all unused strings (about 700 strings removed). See the last file on this diff (**value/strings.xml**). Lots of strings were coming from Riot or has been replaced by other strings. I prefer to remove them all and add again later if necessary. The goal is to avoid translators (= real human!) to translate unused strings.

- 2 layouts were unused too, so deleted.
- some string resource from build.gradle has been deleted too, they were not used and can be replaced by BuildConfig field.
- `lint.xml` file has been updated to ignore other sources of unused resources. Weblate is locked, so I want to focus on String here. Other unused resources can be handled later in a separate PR.

Now unused strings will be considered as lint error, we may change this later, since PR cannot delete unused string, but mark them as to be deleted. We may just ignore those errors until the next Weblate sync, but this is not ideal.

**WIP**

All translations has been temporarily removed to work only on the original strings.

If the CI is happy, next step is to restore all the translations (revert first commit) and delete all the translations of deleted strings.

Once PR is merge, a sync with Weblate will be done and Weblate will be unlocked.


BillCarsonFr (approved):

bmarty:
Force pushed and git history rewritten.

If CI is happy, this could be merged and Weblate synced and unlocked.

PR #5361, opened 2022-02-25
Lifting debug overrides to their own abstraction

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content
Lifts the debug only options to their own `VectorOverrides` class, this class is explicitly for overriding behaviour for testing and is not meant to be user facing. The `release` build type will use the interface defaults, effectively the same as `VectorFeatures`.

## Motivation and context

To avoid leaking debug only content within the `VectorDataStore` and provide a single point of reference for adding feature overrides for testing

## Screenshots / GIFs

No UI changes

## Tests

- Open the debug menu -> private settings 
- enable force dial pad tab
- see the dial pad tab in the home activity

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Sv2 (31)

### DIFF
--- a/vector/src/debug/java/im/vector/app/features/debug/di/FeaturesModule.kt
+++ b/vector/src/debug/java/im/vector/app/features/debug/di/FeaturesModule.kt
@@ -23,8 +23,11 @@ import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
 import im.vector.app.features.DefaultVectorFeatures
+import im.vector.app.features.DefaultVectorOverrides
 import im.vector.app.features.VectorFeatures
+import im.vector.app.features.VectorOverrides
 import im.vector.app.features.debug.features.DebugVectorFeatures
+import im.vector.app.features.debug.features.DebugVectorOverrides
 
 @InstallIn(SingletonComponent::class)
 @Module
@@ -32,6 +35,9 @@ interface FeaturesModule {
 
     @Binds
     fun bindFeatures(debugFeatures: DebugVectorFeatures): VectorFeatures
+
+    @Binds
+    fun bindOverrides(debugOverrides: DebugVectorOverrides): VectorOverrides
 
     companion object {
 
@@ -43,6 +49,16 @@ interface FeaturesModule {
         @Provides
         fun providesDebugVectorFeatures(context: Context, defaultVectorFeatures: DefaultVectorFeatures): DebugVectorFeatures {
             return DebugVectorFeatures(context, defaultVectorFeatures)
+        }
+
+        @Provides
+        fun providesDefaultVectorOverrides(): DefaultVectorOverrides {
+            return DefaultVectorOverrides()
+        }
+
+        @Provides
+        fun providesDebugVectorOverrides(context: Context): DebugVectorOverrides {
+            return DebugVectorOverrides(context)
         }
     }
 }
--- a/vector/src/debug/java/im/vector/app/features/debug/features/DebugVectorOverrides.kt
+++ b/vector/src/debug/java/im/vector/app/features/debug/features/DebugVectorOverrides.kt
@@ -23,36 +23,32 @@ import androidx.datastore.preferences.core.booleanPreferencesKey
 import androidx.datastore.preferences.core.edit
 import androidx.datastore.preferences.preferencesDataStore
 import im.vector.app.features.VectorOverrides
-import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 import org.matrix.android.sdk.api.extensions.orFalse
 
 private val Context.dataStore: DataStore<Preferences> by preferencesDataStore(name = "vector_overrides")
-private val forceDialPadDisplay = booleanPreferencesKey("force_dial_pad_display")
-private val forceLoginFallback = booleanPreferencesKey("force_login_fallback")
+private val keyForceDialPadDisplay = booleanPreferencesKey("force_dial_pad_display")
+private val keyForceLoginFallback = booleanPreferencesKey("force_login_fallback")
 
 class DebugVectorOverrides(private val context: Context) : VectorOverrides {
 
-    override fun forceDialPad() = forceDialPadDisplayFlow
-    override fun forceLoginFallback() = forceLoginFallbackFlow
+    override val forceDialPad = context.dataStore.data.map { preferences ->
+        preferences[keyForceDialPadDisplay].orFalse()
+    }
+
+    override val forceLoginFallback = context.dataStore.data.map { preferences ->
+        preferences[keyForceLoginFallback].orFalse()
+    }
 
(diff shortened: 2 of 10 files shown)


vector/src/debug/java/im/vector/app/features/debug/features/DebugVectorOverrides.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    suspend fun setForceDialPadDisplay(force: Boolean) {
+        context.dataStore.edit { settings ->
+            settings[forceDialPadDisplay] = force
+        }
+    }
+
+    private val forceDialPadDisplayFlow: Flow<Boolean> = context.dataStore.data.map { preferences ->
+        preferences[forceDialPadDisplay].orFalse()
+    }
```
ericdecanini, line 47: Is there any reason we have these separate from the two override functions above? I think it would be much simpler to remove these vals and move their bodies to those functions

vector/src/debug/java/im/vector/app/features/debug/settings/DebugPrivateSettingsViewModel.kt
```diff
@@ -44,11 +44,12 @@ class DebugPrivateSettingsViewModel @AssistedInject constructor(
     }
 
     private fun observeVectorDataStore() {
-        vectorDataStore.forceDialPadDisplayFlow.setOnEach {
-            copy(dialPadVisible = it)
+        debugVectorOverrides.forceDialPad().setOnEach {
+            copy(
+                    dialPadVisible = it
+            )
```
ericdecanini, line 50: Not something to action as this is more of a personal preference thing but I never understood opting for this format rather than a single line `copy(dialPadVisible = it)`. It just personally looks cleaner to me

vector/src/release/java/im/vector/app/core/di/FeaturesModule.kt
```diff
@@ -31,4 +33,9 @@ object FeaturesModule {
     fun providesFeatures(): VectorFeatures {
         return DefaultVectorFeatures()
     }
+
+    @Provides
+    fun providesOverrides(): VectorOverrides {
+        return DefaultVectorOverrides()
+    }
```
ericdecanini, line 40: Same with the above DI comment. Unfortunately I can't run builds on my machine so I can't confirm this myself but it's worth taking a look imo

vector/src/debug/java/im/vector/app/features/debug/settings/DebugPrivateSettingsViewModel.kt
bmarty: Yes, when there is only one line it's OK. I think, it's for future hypothetical code to split into 3 lines, so if for instance in the future we have:
```kotlin
copy(
    dialPadVisible = it,
    other = blah
)
```

But this is definitely not a strong constraint.

vector/src/debug/java/im/vector/app/features/debug/di/FeaturesModule.kt
```diff
@@ -44,5 +50,15 @@ interface FeaturesModule {
         fun providesDebugVectorFeatures(context: Context, defaultVectorFeatures: DefaultVectorFeatures): DebugVectorFeatures {
             return DebugVectorFeatures(context, defaultVectorFeatures)
         }
+
+        @Provides
+        fun providesDefaultVectorOverrides(): DefaultVectorOverrides {
+            return DefaultVectorOverrides()
+        }
```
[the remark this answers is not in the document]
bmarty: It's maybe because of `@InstallIn(SingletonComponent::class)` anotation of the interface `FeaturesModule` but I am not a big specialist of the DI.

vector/src/debug/java/im/vector/app/features/debug/features/DebugVectorOverrides.kt
bmarty: the 2 members of the interface could also be `val`s I think.

bmarty (commented):
Thanks @ouchadam for the PR, look neat.
Thanks @ericdecanini for the review. I have nothing more to add.
I let you 2 manage this PR until merging on develop!

vector/src/debug/java/im/vector/app/features/debug/di/FeaturesModule.kt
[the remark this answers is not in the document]
ouchadam: there's two parts to this

- the default implementations avoid being directly injectable as we want to avoid leaking their implementations and ensure the consumption is only done via `VectorFeatures`
- we have separate debug and feature DI modules, debug configurations need to know the implementation details in order to provide fallbacks

```kotlin
// Both debug and release need to provide the interface VectorFeatures

// Release only provides the down scoped interface 
@Provides
fun providesDefaultVectorFeatures(): VectorFeatures {
    return DefaultVectorFeatures()
}

// Debug 
// Defaults are provided but not accessible outside debug variant, provides binding source for @Binds
@Provides
fun providesDefaultVectorFeatures(): DefaultVectorFeatures {
    return DefaultVectorFeatures()
}

// debug features which make use of the of the default features
@Provides
fun providesDebugVectorFeatures(context: Context, defaultVectorFeatures: DefaultVectorFeatures): DebugVectorFeatures {
    return DebugVectorFeatures(context, defaultVectorFeatures)
}

// Provides the debug instance as the VectorFeatures for the rest of the app
@Binds
fun bindFeatures(debugFeatures: DebugVectorFeatures): VectorFeatures
```

vector/src/debug/java/im/vector/app/features/debug/settings/DebugPrivateSettingsViewModel.kt
ouchadam: I don't have any strong feelings for either, I chose to stick with the consistency of the file, happy to change!

vector/src/debug/java/im/vector/app/features/debug/features/DebugVectorOverrides.kt
ouchadam: will inline into vals :+1:

vector/src/debug/java/im/vector/app/features/debug/di/FeaturesModule.kt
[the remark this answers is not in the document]
ericdecanini: I see, this does shine some light on it, thanks!

vector/src/debug/java/im/vector/app/features/debug/features/DebugVectorOverrides.kt
ouchadam: updated https://github.com/vector-im/element-android/pull/5361/commits/7637ee45447f052cc1675dd7bb1c589c165a2659

bmarty (approved):

ericdecanini (approved):
Thanks for making the changes! LGTM

PR #5380, opened 2022-02-28
Refactoring for safer olm and megolm session usage

BillCarsonFr:
Rework following investigations on strange rolling UISIs errors.
Logs contained a lot of BAD_MAC errors on both olm and megolm.

Olm bad macs can pretty much be anything, like using wrong/desync session or concurrency, or just false positive logs when iterating other known sessions (though proliferation of session might be suspicious)

Megolm could be more probably concurrent decryption, as well as desync with cache/store, as the current code was not very safe.

- The code has been refactored to create spefic classes to handle cache of sessions, and move this responsability out of the realm crypto store. Access are guarded by synchronized directive.
- Concurrent decryption attempt are guarded with mutex (per olm/megolm session)
- Access to olm account has been protected against concurrent access.
- Refactor the addInboundroupSession logic (there was some crash report of NPE there) and code could cause cache beeing out of sync
- Better logging
- *EDIT* Cleaner unwedging with better logs. Unwedging done at the end of the sync now,  some synchronisation added. 

Added a new E2E sanity test that the following is working:
- Proper key sharing to a set of user in e2e room
- Proper key backup restore from a new session
- Proper key forwarding
- Accept and replace existing session with a better forwarded one
 
 *EDIT* Resurrected unwedging test


matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/E2eeSanityTests.kt
```diff
@@ -0,0 +1,648 @@
...
+    @Test
+    fun testSendingE2EEMessages() {
+        val cryptoTestData = cryptoTestHelper.doE2ETestWithAliceAndBobInARoom(true)
+        val aliceSession = cryptoTestData.firstSession
+        val e2eRoomID = cryptoTestData.roomId
+
+        val aliceRoomPOV = aliceSession.getRoom(e2eRoomID)!!
+
+        // add some more users and invite them
+        val otherAccounts = listOf("benoit", "valere", "ganfra") // , "adam", "manu")
```
bmarty, line 78: 👋

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/E2eeSanityTests.kt
```diff
@@ -0,0 +1,648 @@
...
+        sentEventIds.forEach { sentEventId ->
+            val event = newBobSession.getRoom(e2eRoomID)!!.getTimeLineEvent(sentEventId)!!.root
+            testHelper.runBlockingTest {
+                try {
+                    newBobSession.cryptoService().decryptEvent(event, "")
+                    fail("Should not be able to decrypt event")
+                } catch (error: MXCryptoError) {
+                    val errorType = (error as? MXCryptoError.Base)?.errorType
+                    if (expectedError == null) {
+                        Assert.assertTrue(errorType != null)
```
bmarty, line 640: Nit: Could use `assertNotNull`

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/E2eeSanityTests.kt
```diff
@@ -0,0 +1,648 @@
...
+                    } else {
+                        assertEquals(expectedError, errorType, "Message expected to be UISI")
```
bmarty, line 642: I though there was an error, because the `message` should be the first parameter, but there are some sort of method aliases indeed:

image

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/PreShareKeysTest.kt
```diff
@@ -92,7 +90,7 @@ class PreShareKeysTest : InstrumentedTest {
         // Just send a real message as test
         val sentEvent = testHelper.sendTextMessage(aliceSession.getRoom(e2eRoomID)!!, "Allo", 1).first()
 
-        assertEquals(megolmSessionId, sentEvent.root.content.toModel<EncryptedEventContent>()?.sessionId, "Unexpected megolm session")
+        assertEquals("Unexpected megolm session", megolmSessionId, sentEvent.root.content.toModel<EncryptedEventContent>()?.sessionId,)
```
bmarty, line 93: If it was the source of the failure, that's a good catch, even if just running the test should show the error rather clearly :)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -133,7 +143,18 @@ internal class MXOlmDevice @Inject constructor(
...
     fun getMaxNumberOfOneTimeKeys(): Long {
-        return store.getOlmAccount().maxOneTimeKeys()
+        return doWithOlmAccount { it.maxOneTimeKeys() }
+    }
+
+    /**
+     * Olm account access should be synchronized
+     */
+    private fun <T> doWithOlmAccount(block: (OlmAccount) -> T): T {
+        return store.getOlmAccount().let { olmAccount ->
```
bmarty, line 153: To avoid future mistake I would move the fun `doWithOlmAccount` to the store and remove getOlmAccount(), WDYT?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmSessionWrapper.kt
```diff
@@ -25,7 +26,9 @@ data class OlmSessionWrapper(
         // The associated olm session.
         val olmSession: OlmSession,
         // Timestamp at which the session last received a message.
-        var lastReceivedMessageTs: Long = 0) {
+        var lastReceivedMessageTs: Long = 0,
+
+        var mutex: Mutex = Mutex()) {
```
bmarty, line 31: Could be a `val`

bmarty (commented):
That's amazing work. I am not sure I understand all the code related to the crypto part. I am glad that we have some unit test. And you added lots of unit test which is also very nice. Thanks!
I would be more confident if we can have a review from someone in the crypto team. WDYT? @poljar maybe?

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/PreShareKeysTest.kt
BillCarsonFr: I didn't check others, but there are prob more

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/E2eeSanityTests.kt
BillCarsonFr: Done

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/E2eeSanityTests.kt
BillCarsonFr: :/

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
BillCarsonFr: yeah good idea

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmSessionWrapper.kt
BillCarsonFr: done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmEncryption.kt
```diff
@@ -456,42 +464,48 @@ internal class MXMegolmEncryption(
...
-
-        Timber.tag(loggerTag.value).i(" reshareKey: sharing keys for session $senderKey|$sessionId:$chainIndex with device $userId:$deviceId")
+        val usersDeviceMap = try {
+            ensureOlmSessionsForDevicesAction.handle(devicesByUser)
+        } catch (failure: Throwable) {
+            null
+        }
+        val olmSessionResult = usersDeviceMap?.getObject(userId, deviceId)
+        if (olmSessionResult?.sessionId == null) {
+            return false.also {
```
poljar, line 474: Is there a point to this `false.also`? Why not put the log line above the return and just `return false`? The same `log -> return` pattern is happening a couple lines bellow but `also {}` wasn't used there.

Perhaps we should either use `also {}` in such scenarios or not, but not mix/match.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -819,7 +865,8 @@ internal class MXOlmDevice @Inject constructor(
     private fun getSessionForDevice(theirDeviceIdentityKey: String, sessionId: String): OlmSessionWrapper? {
         // sanity check
         return if (theirDeviceIdentityKey.isEmpty() || sessionId.isEmpty()) null else {
-            store.getDeviceSession(sessionId, theirDeviceIdentityKey)
+            olmSessionStore.getDeviceSession(sessionId, theirDeviceIdentityKey)
+//            store.getDeviceSession(sessionId, theirDeviceIdentityKey)
```
poljar, line 869: Another one, let's not keep cryptically commented code in the tree.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -257,14 +279,15 @@ internal class MXOlmDevice @Inject constructor(
             // this session
             olmSessionWrapper.onMessageReceived()
 
-            store.storeSession(olmSessionWrapper, theirIdentityKey)
+            olmSessionStore.storeSession(olmSessionWrapper, theirIdentityKey)
+//            store.storeSession(olmSessionWrapper, theirIdentityKey)
```
poljar, line 283: Let's remove this or document why it's still here.

poljar (approved):
Left some small nits, logic wise nothing seems to be wrong.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OlmSessionStore.kt
```diff
@@ -0,0 +1,164 @@
...
+     *
+     * @param deviceKey the public key of the other device.
+     * @return A set of sessionId, or empty if device is not known
+     */
+    @Synchronized
+    fun getDeviceSessionIds(deviceKey: String): List<String> {
+        return internalGetAllSessions(deviceKey)
+    }
+
+    private fun internalGetAllSessions(deviceKey: String): MutableList<String> {
```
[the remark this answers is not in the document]
BillCarsonFr: mmm, initialy it was called also in another method (that was already synchronized), so I created this. But looks like there is no point now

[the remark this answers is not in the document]
BillCarsonFr: Done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OlmSessionStore.kt
```diff
@@ -0,0 +1,164 @@
...
+                if (!persistedKnownSessions.contains(cachedSessionId)) {
+                    persistedKnownSessions.add(cachedSessionId)
+                }
+            }
+        }
+        return persistedKnownSessions
+    }
+
+    /**
+     * Retrieve an end-to-end session between the logged-in user and another
```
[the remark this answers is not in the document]
BillCarsonFr: fixed (and some more in crypto store interface)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/OlmSessionStore.kt
```diff
@@ -0,0 +1,164 @@
...
+     * device.
+     *
+     * @param sessionId the session Id.
+     * @param deviceKey the public key of the other device.
+     * @return The Base64 end-to-end session, or null if not found
```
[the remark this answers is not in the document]
BillCarsonFr: Fixed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmEncryption.kt
BillCarsonFr: yes fixed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
BillCarsonFr: done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
BillCarsonFr: done

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/TestConstants.kt
```diff
@@ -23,7 +23,7 @@ object TestConstants {
     const val TESTS_HOME_SERVER_URL = "http://10.0.2.2:8080"
 
     // Time out to use when waiting for server response.
-    private const val AWAIT_TIME_OUT_MILLIS = 30_000
```
michaelkaye, line 26: I found myself doing this in response to slow integration tests in #5459 - do you have any more information on why 60s is better?

michaelkaye (commented):
I've recently commented out some flaky integration tests that are in this area: https://github.com/vector-im/element-android/issues/5449 - they seemed to be some sort of issue between two things opening a realm db at the same time. Do you think these concurrent access changes you  mention might have helped there - is it worth un-ignoring those tests?

Also, it might be good to run the nightly integration tests against a PR like this that introduces more integration tests - otherwise we won't compile and run them until the night after merge.

You can do that in Actions -> "Nightly build" (in left sidebar) -> then in the blue bar, select this branch to run the workflow against.

It doesn't run them automatically any more as it takes a good long time and we don't want to slow down the PR merge time too much.

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/common/TestConstants.kt
BillCarsonFr: When doing test with several sessions everything is getting a lot slower (like E2E sanity test), each session is slower to sync, etc...
Everytime we test something with await/waitWithLatch it's using that as default timeout, and test are then failing when waiting from something coming back from the sync.
It also happens when running all tests, I wonder if we don't clean properly between test?
Previously I used the `ANDROIDX_TEST_ORCHESTRATOR` option to ensure that all test start with a clear state, but looks like it's not working anymore.

michaelkaye: We definitely don't clean up the server between tests (and if you just run a demo server locally, then that won't be cleaned up unless you explicitly do so)

bmarty (approved):
Let's merge this PR, 1.4.4 has been released.

Issue #5311, opened 2022-02-22
SoftLogout UX is broken if the homeserver supports both password and SSO


bmarty:
SoftLogout UX is broken if the homeserver supports both password and SSO like for instance matrix.org.

Mainly because this block is empty:
```kotlin
is LoginMode.SsoAndPassword -> {
}
```
https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/signout/soft/SoftLogoutController.kt#L125

In this case, the app should display the password form, if the user has created the session using a password. If SSO was used, a message like "Please clear data and log in again to your account" can be displayed.

The fact that the user has created the session using a password is not stored. This info can be stored in the `SessionParamsEntity` as a nullable boolean. A database migration will be required.

- boolean is null (we do not have the info) -> Display the password form and a message like "If you have signed in using an account from another provider, please clear data and log in again to your account"
- boolean is true or false: see above

Simulating a soft logout can be done by adding this line: 

```kotlin
globalErrorReceiver.handleGlobalError(GlobalError.InvalidToken(softLogout = true))
```

[here](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncTask.kt#L153) (for instance)

Critical because the user cannot recover their session in case of soft logout.

langleyd:
Duplicate of https://github.com/vector-im/element-android/issues/5192, whichever one you want to keep open.

PR #5398, opened 2022-03-02
Fixes broken SoftLogout UX for homeservers that support both Password and SSO

ericdecanini:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes an error in SoftLogout where if your LoginMode was SsoAndPassword, no form would show up on the soft logout screen.

This was done by adding a new field LoginType in SessionParamsEntity which remembers the method the user used to log in.

## Motivation and context

Merging this PR closes https://github.com/vector-im/element-android/issues/5311

I also took this as an opportunity to refactor some classes around the session creation in order to make them easier to read and more testable & maintainable.

I did try to add tests for this too, but due to hardware constraints I couldn't test them on my local machine and using the CI didn't give me the info I need to properly write them to pass (failing because of MockkException, NoClassDefFoundException, etc) so I removed them in [this commit](https://github.com/vector-im/element-android/pull/5398/commits/63389418853c1db018cc5e868c2c64113361ecc4) and [added an issue](https://github.com/vector-im/element-android/issues/5437) to get back to this as soon as I'm able to.

Lastly, this PR also fixes a bug where SSO login after a soft logout would crash the app. I would believe the SSO journey to be fixed but currently there's no true way to test this with both the client and server knowing that the user had soft logged out. The closest we can achieve is to throw a global error on the client side as per the test instructions below.

I attached a gif showing this journey which ultimately ends back in the soft logout screen (assumably because the global error gets thrown again or there's no log in actually happen because the user is already logged in according to the server) so we can only base this fix off of assumptions here

## Screenshots / GIFs

| Password | SSO |
|---|---|
| image | image |

**SSO Journey:**

SVID_20220309_121237_1

## Tests

There is no current way to properly test this without mocking the situation with a few code changes.

1. In SoftLogoutController above line 155 add this line:
globalErrorReceiver.handleGlobalError(GlobalError.InvalidToken(softLogout = true))

2. Run the app and login. You'll immediately be taken to the Soft Logout screen as shown above (though the server doesn't know of any logout happening, it only shows in the app)

3. Test this with a password account and an SSO account

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 10

## Checklist


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DefaultLoginWizard.kt
```diff
@@ -90,15 +90,15 @@ internal class DefaultLoginWizard(
...
+        return sessionCreator.createSession(credentials, pendingSessionData.homeServerConnectionConfig, LoginType.SSO)
     }
 
     override suspend fun loginCustom(data: JsonDict): Session {
         val credentials = executeRequest(null) {
             authAPI.login(data)
         }
 
-        return sessionCreator.createSession(credentials, pendingSessionData.homeServerConnectionConfig)
+        return sessionCreator.createSession(credentials, pendingSessionData.homeServerConnectionConfig, LoginType.UNSUPPORTED)
```
BillCarsonFr, line 101: Could we call it LoginType.CUSTOMN

vector/src/main/java/im/vector/app/features/login/LoginActivity.kt
```diff
@@ -252,7 +253,7 @@ open class LoginActivity : VectorBaseActivity<ActivityLoginBinding>(), UnlockedA
                 // It depends on the LoginMode
                 when (state.loginMode) {
                     LoginMode.Unknown,
-                    is LoginMode.Sso      -> error("Developer error")
+                    is LoginMode.Sso -> launchSsoFlow()
```
BillCarsonFr, line 256: Unknown and sso are doing the same thing?

BillCarsonFr (changes requested):
Thx eric, good work. I added a few comments after a static review.

It should be tested against a real setup before beeing merged, please reach out in internal room to see how to setup a sandbox homeserver with SSO and ways to revocate tokens.

From the user point of view, I think that we should probably remember the social login used if possible instead and just show that social login button instead of sending back to the raw fallback. Do you think it's possible?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DirectLoginTask.kt
```diff
@@ -77,7 +77,7 @@ internal class DefaultDirectLoginTask @Inject constructor(
             }
         }
 
-        return sessionCreator.createSession(credentials, params.homeServerConnectionConfig)
+        return sessionCreator.createSession(credentials, params.homeServerConnectionConfig, LoginType.UNSUPPORTED)
```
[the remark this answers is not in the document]
ericdecanini: This must've been a temporary thing I added. Good catch! Changing this to unknown as its not based on any of our known standard login types

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/registration/DefaultRegistrationWizard.kt
```diff
@@ -63,7 +64,7 @@ internal class DefaultRegistrationWizard(
 
     override suspend fun getRegistrationFlow(): RegistrationResult {
         val params = RegistrationParams()
-        return performRegistrationRequest(params)
+        return performRegistrationRequest(params, LoginType.PASSWORD)
```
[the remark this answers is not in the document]
ericdecanini: I thought this too. The ideal solution would be we get this value everytime from the backend but we don't. Another solution would be to persist this either in storage or statically but I think that's more confusing

vector/src/main/java/im/vector/app/features/login/LoginActivity.kt
ericdecanini: Unintentional. Good catch!

ericdecanini:
> Thx eric, good work. I added a few comments after a static review.
> 
> It should be tested against a real setup before beeing merged, please reach out in internal room to see how to setup a sandbox homeserver with SSO and ways to revocate tokens.
> 
> From the user point of view, I think that we should probably remember the social login used if possible instead and just show that social login button instead of sending back to the raw fallback. Do you think it's possible?

Thanks for the comments, I pushed some changes addressing them.

Regarding testing against a real setup, I agree that this would be ideal to do before merging. I'll try tomorrow if I can get this sandbox homeserver set up.

As for remembering social login used, I tried this but this needs much more consideration from a technical standpoint. We could persist which social was used in the database but I think this is the wrong way to go about it. I can see in the code that a recover session through SSO has been set up but not fully implemented. I tried to use this but the implementation was too unclear for me to extend

Mikhail5555:
@ericdecanini I tried this out, and it no longer crashes but it also doesnt solve the issue. It now continues in a login loop.
I did find a very easy way of testing it, set `session_lifetime: 1m` an re-login. This will give you a session which expires within one minute.
Screenrecord: https://cloud.mehome.dev/s/2jP7p6wHSzGK6tG

ericdecanini:
> @ericdecanini I tried this out, and it no longer crashes but it also doesnt fully solve the issue. It now continues in a login loop. I did find a very easy way of testing it, set `session_lifetime: 1m` an re-login. This will give you a session which expires within one minute. Screenrecord: https://cloud.mehome.dev/s/2jP7p6wHSzGK6tG
> 
> edit: I just also read your last comment, it seems that the `not actually login in` is somewhat intended. I think this PR already helpes to solve a lot (Aka the not crashing) and the login issue itself might ned fixing in a seperate PR.

Thanks for testing it! Yeah setting a low `session_lifetime` is exactly the way to test this, and your screen record does give very much good insight. Unfortunately I'm still struggling to look into this deeper as setting up a test sso server on our side is taking ages (you'd think it would be quicker for us). Is there any way you could provide a test sso account on your server, or allow me to register one of my own on your server? Or would you rather we continue trying to set up our own server to test this? (I can't guarantee how quick that would happen though)

Mikhail5555:
@ericdecanini can you contact me on matrix at `@mikhail:mehome.dev`, I will do my best to help you where ever i can, and maybe just setup a second instance with SSO.

Mikhail5555:
@ericdecanini I managed to setup a test server for you which i can provide you with a test account. With the only (major) config difference being the the session time is extreme short (1m). Please contact me through matrix and we can exchange some credentials and other info ;)

ericdecanini:
It must be known that nobody in the team has the capacity at the moment to take care of the issue of SSO login after soft logout causing an infinite loop. After a discussion with @bmarty we determined that this is at least a better outcome than the previous crashing the app had (which also covered soft logout with password which should now be working on this branch) so we'll work towards getting this merged.

We can then open a new issue to address the sso infinite loop in isolation

vector/src/main/java/im/vector/app/features/login/LoginActivity.kt
bmarty: yes, good catch @BillCarsonFr !

bmarty (approved):
LGTM, thanks.
@BillCarsonFr can you also review so that we can merge this PR? Thanks!

BillCarsonFr (changes requested):
Ok to merge after rebase

ericdecanini:
Force pushed after failed rebase (resulting in 1.8k files changed)

BillCarsonFr (approved):

PR #5404, opened 2022-03-02
Voice Message Playback Scrolling Support

### LINKED ISSUE #5426: Allow scrolling position of Voice Message playback
### Your use case

Users should be able to change the current playback position by touching different locations on the waveform of a voice message.

_No response_

### Additional context

_No response_

onurays:
Fixes #5426.

https://user-images.githubusercontent.com/1254401/156786596-3fd9e7f7-3eaf-43bd-9214-15bbd19d4ccf.mp4

Voice Message related UI and state classes became to be messy. We should plan to refactor them by rethinking all states and flow between them.


vector/src/main/java/im/vector/app/features/home/room/detail/composer/voice/VoiceMessageViews.kt
```diff
@@ -69,6 +83,8 @@ class VoiceMessageViews(
         observeMicButton(actions)
     }
 
+    private fun getTouchedPositionPercentage(motionEvent: MotionEvent, view: View) = motionEvent.x / view.width
```
bmarty, line 86: Not tested, but show probably use `coerceIn(0,1)` to avoid overflow.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -695,8 +709,7 @@ class MessageItemFactory @Inject constructor(
         return this
                 ?.filterNotNull()
                 ?.map {
-                    // Value comes from AudioRecordView.maxReportableAmp, and 1024 is the max value in the Matrix spec
```
bmarty, line 698: This comment is still valuable, no? Maybe just update the part about `AudioRecordView.maxReportableAmp` to `AudioWaveformView.MAX_FFT`

vector/src/main/res/values/audio_waveform_attr.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bmarty, line 1: This file should be moved to `https://github.com/vector-im/element-android/tree/main/library/ui-styles/src/main/res/values` and its name should start with `stylable_`

bmarty (commented):
Some minor remark.
Static review OK!

vector/src/main/java/im/vector/app/features/home/room/detail/composer/voice/VoiceMessageViews.kt
onurays: More secure, done.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
onurays: Right, done.

vector/src/main/res/values/audio_waveform_attr.xml
onurays: Done.

ouchadam:
when scrubbing the play head, should the audio continue to play? at the moment we're pausing each time we change the play head position :thinking:

vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
```diff
@@ -169,11 +171,19 @@ class VoiceMessageHelper @Inject constructor(
     }
 
     fun stopPlayback() {
-        playbackTracker.stopPlayback(VoiceMessagePlaybackTracker.RECORDING_ID)
+        playbackTracker.pausePlayback(VoiceMessagePlaybackTracker.RECORDING_ID)
         mediaPlayer?.stop()
         stopPlaybackTicker()
     }
 
+    fun movePlaybackTo(id: String, percentage: Float, totalDuration: Int) {
```
ouchadam, line 179: if we use the `mediaPlayer?.duration` then we'll be able to avoid inferring the duration from the UI countdown ticker 
would that be possible? (it would assume the media player is in the prepared state)

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageVoiceItem.kt
```diff
@@ -93,33 +120,41 @@ abstract class MessageVoiceItem : AbsMessageItem<MessageVoiceItem.Holder>() {
         holder.voicePlaybackLayout.backgroundTintList = ColorStateList.valueOf(backgroundTint)
         holder.voicePlaybackControlButton.setOnClickListener { playbackControlButtonClickListener?.invoke(it) }
 
-        voiceMessagePlaybackTracker.track(attributes.informationData.eventId, object : VoiceMessagePlaybackTracker.Listener {
-            override fun onUpdate(state: VoiceMessagePlaybackTracker.Listener.State) {
-                when (state) {
-                    is VoiceMessagePlaybackTracker.Listener.State.Idle    -> renderIdleState(holder)
-                    is VoiceMessagePlaybackTracker.Listener.State.Playing -> renderPlayingState(holder, state)
-                    is VoiceMessagePlaybackTracker.Listener.State.Paused  -> renderPausedState(holder, state)
+        // Don't track and don't try to update UI before view is present
```
ouchadam, line 123: is this because we're relying on [`holder.voicePlaybackWaveform.post`](https://github.com/vector-im/element-android/pull/5404/files#diff-986f29a475d699ca21a7504d49eba50b6567a25b3ea9284b82bc17bd854db924R93)?  if so we should move the tracker logic to the same `post` body to better couple the logic together (it seems that we're relying on the audio view to be fully drawn because it only updates its internal state as a side effect of the drawing)

onurays:
> when scrubbing the play head, should the audio continue to play? at the moment we're pausing each time we change the play head position 🤔

Ideally yes, I couldn't find an easy way to implement it and I am not sure this is a blocker.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
onurays: We have a single mediaplayer which is initialized when users start the playback. But users are able to scroll idle recordings and the mediaplayer is null at this stage.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageVoiceItem.kt
onurays: Yes we need `post` because the width of the view is 0 (not calculated yet) when it is bind.

ouchadam (approved):
thanks for the update! :100:

DoM1niC:
@onurays the Player visual events freeze for me after skip play and stop a Voice Message and play a another one.

onurays:
> @onurays the Player visual events freeze for me after skip play and stop a Voice Message and play a another one above the timeline.

Good catch, thank you. I will fix this in another PR.

PR #5416, opened 2022-03-03
Run ImageUtils requests async

### LINKED ISSUE #5236: Crash when sharing files stored on nextcloud
### Steps to reproduce

Trying to share a file/image from within a room or file manager (with destination element)

#### What did you expect?
File shared

#### What happened instead?
Crash

[log.txt](https://github.com/vector-im/element-android/files/8067004/log.txt)

### Your phone model

PixelXL

### Operating system version

Android 10 (Lineageos, no Gapps)

Element 1.4.0 from F-Droid

### Homeserver

matrix.org

### Will you send logs?

No

akallabeth:
Fixes #5236: Run file requests in background to not block ui thread

 
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Run operations in `ImageUtils` async to avoid blocking the UI thread and therefore crashing the app if the file chosen must be downloaded from some location

## Tests

- Step 1: Installed debug build and logged in to existing account
- Step 2: Shared some files

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [x ] Pull request is based on the develop branch

### DIFF
--- a/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/ImageUtils.kt
+++ b/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/ImageUtils.kt
@@ -23,35 +23,51 @@ import android.graphics.ImageDecoder
 import android.net.Uri
 import android.os.Build
 import androidx.exifinterface.media.ExifInterface
+import kotlinx.coroutines.async
+import kotlinx.coroutines.runBlocking
 import timber.log.Timber
 
 object ImageUtils {
 
     fun getBitmap(context: Context, uri: Uri): Bitmap? {
-        return try {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-                ImageDecoder.decodeBitmap(ImageDecoder.createSource(context.contentResolver, uri))
-            } else {
-                context.contentResolver.openInputStream(uri)?.use { inputStream ->
-                    BitmapFactory.decodeStream(inputStream)
+        var bitmap: Bitmap? = null
+        runBlocking {
+            val job = async {
+                try {
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                        ImageDecoder.decodeBitmap(ImageDecoder.createSource(context.contentResolver, uri))
+                    } else {
+                        context.contentResolver.openInputStream(uri)?.use { inputStream ->
+                            BitmapFactory.decodeStream(inputStream)
+                        }
+                    }
+                } catch (e: Exception) {
+                    Timber.e(e, "Cannot decode Bitmap: %s", uri.toString())
+                    null
                 }
             }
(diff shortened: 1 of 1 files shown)


library/multipicker/src/main/java/im/vector/lib/multipicker/utils/ImageUtils.kt
```diff
@@ -23,35 +23,51 @@ import android.graphics.ImageDecoder
...
     fun getBitmap(context: Context, uri: Uri): Bitmap? {
-        return try {
-            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-                ImageDecoder.decodeBitmap(ImageDecoder.createSource(context.contentResolver, uri))
-            } else {
-                context.contentResolver.openInputStream(uri)?.use { inputStream ->
-                    BitmapFactory.decodeStream(inputStream)
+        var bitmap: Bitmap? = null
+        runBlocking {
+            val job = async {
```
bmarty, line 35: I do not really understand what will be the diff with the previous version. Since there is a `runBlocking` block, this is still synchronous, no?

bmarty (commented):
Thanks for the PR.
I have seen the issue #5236, and you tell that it fixes your problem, but I do not really understand why.

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/ImageUtils.kt
ouchadam: this might be acting more as workaround by tricking the main thread guard into thinking it's a generic task than a network task :thinking: 

we probably want to mark the functions as `suspend` and then ensure the entire chain :sob:  is also `suspend` and finally called by a coroutine scope of some sort (most likely a view model scope)

akallabeth: @ouchadam is right. I´ve just did rerun my tests, it seems it only partially fixes the issue, so I was a bit too rash. (There are still other parts in the chain called on main thread I did not run into the first time I tested)

akallabeth: Each error/problem in `contentResolver.openInputStream` (e.g. `FileNotFoundException` due to some network timeout) still crashes the app :/

akallabeth:
Ok, this is indeed only working by chance.
Main problem seems to be that `contentResolver.*` stuff should not be run on the UI thread, as these operations require access to the file in question and might trigger network/... requests go complete.

Only successful workaround (that did not crash under some conditions) so far is
```
 StrictMode.setThreadPolicy(StrictMode.ThreadPolicy.Builder().permitAll().build())
```
which is obviously not a good idea.

PR #5419, opened 2022-03-03
Giving avatar/display name error dialogs human readable error messages

### LINKED ISSUE #5418: Unfriendly message in display name error dialog
Avoid printing raw exceptions as user facing messages

To improve the message readability

By reusing the error dialog which translates exceptions to human readable messages

_No response_

### Additional context

Screenshot_20220303_155653

ouchadam:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Reuses the error dialog logic which translates exceptions to human readable strings in the settings screen

## Motivation and context

Fixes #5418 - The error messages for display name/avatar changing are making use of the raw exception instead of the nicely formatted message included within the exception

## Screenshots / GIFs

| BEFORE | AFTER |
| --- | --- |
|Screenshot_20220303_155653|Screenshot_20220303_160945

## Tests

- Via the settings
- Attempt to change the display name to one longer than 256 characters

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 31 (Sv2)

### DIFF
--- a/vector/src/main/java/im/vector/app/features/settings/VectorSettingsBaseFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/settings/VectorSettingsBaseFragment.kt
@@ -148,24 +148,6 @@ abstract class VectorSettingsBaseFragment : PreferenceFragmentCompat(), Maverick
         }
     }
 
-    /**
-     * A request has been processed.
-     * Display a toast if there is a an error message
-     *
-     * @param errorMessage the error message
-     */
-    protected fun onCommonDone(errorMessage: String?) {
-        if (!isAdded) {
-            return
-        }
-        activity?.runOnUiThread {
-            if (errorMessage != null && errorMessage.isNotBlank()) {
-                displayErrorDialog(errorMessage)
-            }
-            hideLoadingView()
-        }
-    }
-
     protected fun displayErrorDialog(throwable: Throwable) {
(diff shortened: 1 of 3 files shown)


vector/src/main/java/im/vector/app/features/settings/VectorSettingsBaseFragment.kt
```diff
@@ -148,24 +148,6 @@ abstract class VectorSettingsBaseFragment : PreferenceFragmentCompat(), Maverick
...
     }
 
-    /**
-     * A request has been processed.
-     * Display a toast if there is a an error message
-     *
-     * @param errorMessage the error message
-     */
-    protected fun onCommonDone(errorMessage: String?) {
-        if (!isAdded) {
```
bmarty, line 158: This check is lost. Now we should probably use `launchWhenResumed`. If the operation takes time and the user leave the screen before it ends we can have crashes.

**EDIT** My bad, we are already in the `lifecycleScope`, so I guess it's fine

bmarty (approved):
Thanks!

vector/src/main/java/im/vector/app/features/settings/VectorSettingsBaseFragment.kt
ouchadam: yep exactly ^^^ seemed like left over code from before coroutines were introduced

PR #5436, opened 2022-03-07
change selected space highlight

### LINKED ISSUE #5346: Fix highlighting showing which Space is selected
On Android, the current Space is only highlighted with a green bar. Nique says that the designs show it as having a background highlighting as well.

This is done correctly on iOS.

### Tasks

- [x] @niquewoodhouse to attach link to figma/share design
- [ ] @fedrunov implement the design

fedrunov:
fixes #5346 

| Light theme  | Dark theme |
| ------------- | ------------- |
| image | image |

### DIFF
--- a/changelog.d/5346.misc
+++ b/changelog.d/5346.misc
@@ -1,0 +1,1 @@
+Selected space highlight changed in left panel
--- a/vector/src/main/res/drawable/bg_space_item.xml
+++ b/vector/src/main/res/drawable/bg_space_item.xml
@@ -8,11 +8,12 @@
                     <solid android:color="@android:color/transparent" />
                 </shape>
             </item>
-            <item android:gravity="center_vertical|start" >
+            <item android:gravity="center_vertical|start"
+                android:left="8dp"
+                android:right="11dp">
                 <shape>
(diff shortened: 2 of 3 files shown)


niquewoodhouse (approved):
@fedrunov just some questions to clarify some of the details, please. Looking between the designs and the build, I can see what you mean about differences.

- What's the size of the space avatars?
- What's the size of the space name next to the avatar?

It would be nice if these kind of details did match the design but I think making the details match could/should be a separate issue, so I'm approving this as it does change the highlight well while in keeping with the existing look/feel.

fedrunov:
@niquewoodhouse 

Entire line (list item) height: 65dp
Avatar size: 42x42 dp
Space name text style: textSize = 16sp, leterSpacing = 0, textStyle=bold
Space description text style: textSize = 14sp, leterSpacing = 0, lineHeight = 20sp

fedrunov:
I agree that  we need to file a separate PR to align current implementation with design in figma if we want to

ouchadam:
to double check, these layours/drawables are used in other places, are we also okay for them to update?

bmarty (approved):
LGTM, thanks!

Issue #5040, opened 2022-01-24
Voting an ended poll sends a `org.matrix.msc3381.poll.response` event


bmarty:
- Create a poll
- End it
- Vote on this poll

It does not have an effect on the UI, but an event `org.matrix.msc3381.poll.response` is sent to the timeline (which is then correctly ignored). So we send noise to the timeline.

Voting on closed poll could be disabled on the UI and it should also be filtered out SDK side.

bmarty:
Still there in 1.4.2

PR #5473, opened 2022-03-09
Fixes ended poll voting

ericdecanini:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes polls being votable after being ended

## Motivation and context

Closes https://github.com/vector-im/element-android/issues/5040

The cause of the issue is found in `EventRelationsAggregationProcessor:388`.

When voting, the SDK checks that the eventTimestamp is before the closed time and if it is, the vote response can go through, despite the poll having already been ended (determined by the non-null presence of a closed time).

The fix on the SDK was to simply remove that extra condition. I checked that this shouldn't have any side-effects. I tried to approach this differently, i.e. trying to fix the event timestamp (`event.originServerTs`) set during the poll ending so that it matches the poll closed time but to no avail.

Though I should be able to investigate this with a bit more time if that's the preferrable course of action

I also added an extra layer of protection on the client side to check that the poll is in an ended state.

## Tests

- Set your device system time to 1 hour ahead of current time
- In any timeline, create a new poll
- Immediately end it
- Try to vote on it and see that it cannot be voted

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 10

## Checklist


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -385,7 +385,7 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
                 }
 
         val closedTime = existingPollSummary?.closedTime
-        if (closedTime != null && eventTimestamp > closedTime) {
+        if (closedTime != null) {
```
ouchadam, line 388: is there a case where the poll could have a closed timestamp but still be valid to casting votes? cc @onurays

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/PollItem.kt
```diff
@@ -73,13 +72,21 @@ abstract class PollItem : AbsMessageItem<PollItem.Holder>() {
...
 
+    private fun onPollItemClick(optionViewState: PollOptionViewState) {
+        val relatedEventId = eventId
+
+        if (isPollActive(optionViewState) && relatedEventId != null) {
+            callback?.onTimelineItemAction(RoomDetailAction.VoteToPoll(relatedEventId, optionViewState.optionId))
+        }
+    }
+
+    private fun isPollActive(optionViewState: PollOptionViewState) = optionViewState !is PollOptionViewState.PollEnded
```
bmarty, line 88: There is too much logic here. Better to move that to the controller.
Maybe replace `var pollSent: Boolean = false` which is unused by something like `var canVote: Boolean = false`

bmarty (commented):
One remark.

You may have seen it, but this is where the SDK ignore late vote FWIW: https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt#L388

ericdecanini:
Update: refactored MessageItemFactory's `buildPollItem` method and a few other poll-related classes

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
ericdecanini: @onurays bump

ouchadam (approved):
thanks for all the cleaning up :100: 

would be good to double check with @onurays for the original timestamp condition in case there's a scenario we aren't aware of

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
onurays: I think we should keep this condition. Even if we prevent it previously, there is always a chance to trigger a poll response to an ended poll from another client or via rest api.

vector/src/main/java/im/vector/app/features/poll/PollState.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.features.poll
+
+sealed class PollState {
```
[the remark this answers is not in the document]
ericdecanini: This has never crossed my mind, but I totally agree! Changes pushed

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -179,42 +191,44 @@ class MessageItemFactory @Inject constructor(
...
-            is MessageVideoContent               -> buildVideoMessageItem(messageContent, informationData, highlight, callback, attributes)
-            is MessageFileContent                -> buildFileMessageItem(messageContent, highlight, attributes)
-            is MessageAudioContent               -> {
+            is MessageEmoteContent -> buildEmoteMessageItem(messageContent, informationData, highlight, callback, attributes)
+            is MessageTextContent -> buildItemForTextContent(messageContent, informationData, highlight, callback, attributes)
+            is MessageImageInfoContent -> buildImageMessageItem(messageContent, informationData, highlight, callback, attributes)
+            is MessageNoticeContent -> buildNoticeMessageItem(messageContent, informationData, highlight, callback, attributes)
+            is MessageVideoContent -> buildVideoMessageItem(messageContent, informationData, highlight, callback, attributes)
+            is MessageFileContent -> buildFileMessageItem(messageContent, highlight, attributes)
+            is MessageAudioContent -> {
```
onurays, line 203: The arrows of when conditions should be at the same column. Could you please check your formatter? (We had this problem on a previous Android Studio).

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -482,46 +482,39 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
...
 
         // we have it
-        val existingPollSummary = existing.pollResponseSummary
+        val existingPollSummary = existingPoll.pollResponseSummary
                 ?: realm.createObject(PollResponseAggregatedSummaryEntity::class.java).also {
-                    existing.pollResponseSummary = it
+                    existingPoll.pollResponseSummary = it
                 }
 
-        if (existingPollSummary.closedTime != null) {
```
ericdecanini, line 509: The source of the bug is here. A local echo being processed before the remote echo meant that the remote echo never reaches past this point. `closedTime` thus never gets set from the remote and always turns out to be the current time millis of the local device that ended the poll. If the device has time that's out of sync with unix current time, this bug can occur

cc @onurays

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
ericdecanini: Reverted this change and pushed alternative solution. See below comment for details

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
onurays: I see, thanks for the clarification.

onurays (approved):

PR #5523, opened 2022-03-11
Use different tags for unit tests and android test compilation.

michaelkaye:
Otherwise we will cancel one in favour of the other.

see eg https://github.com/vector-im/element-android/runs/5513822281?check_suite_focus=true


bmarty (approved):
LGTM, thanks

PR #5528, opened 2022-03-13
Fix missing/swapped/duplicated messages due to wrong TimelineChunk modifications or insertions

SpiritCroc:
I was observing cases where `builtEvents[modificationIndex]` was not
having the same `eventId` as the `udpatedEntity` in `handleDatabaseChangeSet()`.

In particular, I observed both cases that
- there was no item in the list yet with the same `eventId` as the updated
  one
- there was an item with the same `eventId` already in the list, but at a
  different position.

Whenever this happened, the timeline would render missing, duplicated,
or swapped messages in the timeline.

Instead of relying on `modificationIndex` to be the same for both the
change set and `builtEvents`, look up the proper index by `eventId`.

~~Note: I'm not entirely sure if this is the correct way to fix this.
Is the previously used assumption that `displayIndex == modificationIndex` something that should hold?
If yes, this PR is likely not the correct fix.
If no, could the insertions in `handleDatabaseChangeSet()` also be broken, since I'm only addressing modification's indices here?~~

For insertions, inconsistencies could also occur, if the chunk was not fully loaded yet, so check this as well.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Look up indices for event modifications, instead of using the ones from the change set.

## Motivation and context

Messages were sometimes showing swapped, duplicated, or missing.

## Tests

Observe if messages are rendered correctly. I also used following commits to further debug this:
- https://github.com/SchildiChat/SchildiChat-android/commit/7bfd3c8dca6ca0c63706c27569cae427420f1952
- https://github.com/SchildiChat/SchildiChat-android/commit/12e4853c59f038cff11c320df25c849a0f3eb008
- https://github.com/SchildiChat/SchildiChat-android/commit/4b12f162a07904061a645b40589385dfd98e1287

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 11

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineChunk.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineChunk.kt
@@ -430,6 +430,31 @@ internal class TimelineChunk(private val chunkEntity: ChunkEntity,
     private fun handleDatabaseChangeSet(results: RealmResults<TimelineEventEntity>, changeSet: OrderedCollectionChangeSet) {
         val insertions = changeSet.insertionRanges
         for (range in insertions) {
+            // Check if the insertion's displayIndices match our expectations - or skip this insertion.
+            // Inconsistencies (missing messages) can happen otherwise if we get insertions before having loaded all timeline events of the chunk.
+            if (builtEvents.isNotEmpty()) {
+                // Check consistency to item before insertions
+                if (range.startIndex > 0) {
+                    val firstInsertion = results[range.startIndex]!!
+                    val lastBeforeInsertion = builtEvents[range.startIndex-1]
+                    if (firstInsertion.displayIndex+1 != lastBeforeInsertion.displayIndex) {
+                        Timber.i("handleDatabaseChangeSet: skip insertion at ${range.startIndex}/${builtEvents.size}, " +
+                                "displayIndex mismatch at ${range.startIndex}: ${firstInsertion.displayIndex} -> ${lastBeforeInsertion.displayIndex}")
+                        continue
+                    }
+                }
+                // Check consistency to item after insertions
+                if (range.startIndex < builtEvents.size) {
+                    val lastInsertion = results[range.startIndex+range.length-1]!!
+                    val firstAfterInsertion = builtEvents[range.startIndex]
+                    if (firstAfterInsertion.displayIndex+1 != lastInsertion.displayIndex) {
+                        Timber.i("handleDatabaseChangeSet: skip insertion at ${range.startIndex}/${builtEvents.size}, " +
+                                "displayIndex mismatch at ${range.startIndex+range.length}: " +
+                                "${firstAfterInsertion.displayIndex} -> ${lastInsertion.displayIndex}")
+                        continue
+                    }
+                }
+            }
             val newItems = results
                     .subList(range.startIndex, range.startIndex + range.length)
                     .map { it.buildAndDecryptIfNeeded() }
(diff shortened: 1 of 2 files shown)


SpiritCroc:
Example of what this is trying to fix: see missing message 97 with `displayIndex` 353, but duplicated message 98 with `displayIndex` 354:

Screenshot_20220309-132319_Beeper_dbg

SpiritCroc:
I think I just observed a case which fits this prediction of mine:

> If no, could the insertions in `handleDatabaseChangeSet()` also be broken, since I'm only addressing modification's indices here?

I.e., I just had a case where the chunk had all events until displayIndex 11 loaded, and then `handleDatabaseChangeSet()` was called to insert an event at displayIndex 14, leading to a gap (missing messages 12 and 13).

SpiritCroc:
> I think I just observed a case which fits this prediction of mine:
> 
> > If no, could the insertions in `handleDatabaseChangeSet()` also be broken, since I'm only addressing modification's indices here?
> 
> I.e., I just had a case where the chunk had all events until displayIndex 11 loaded, and then `handleDatabaseChangeSet()` was called to insert an event at displayIndex 14, leading to a gap (missing messages 12 and 13). Any ideas how to best handle this case?

ok, found a way to reproduce: just reduce initial loaded chunk size and pagination size, artificially delay chunk loading from code, and open the timeline somewhere in the past to get some slow forwards loading going on. Now if new messages arrive, these are fast-forwarded before the rest of the chunk is loaded, resulting in skipped messages. Pushing a fix as another commit.

bmarty:
Thanks for the PR. I defer to @ganfra to review it.

ganfra (approved):
LGTM! Probably caused by using coroutines...
Hopefully this won't degrade performance as you need to lookup.

Can you just fix the lint error?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineChunk.kt
```diff
@@ -430,6 +430,31 @@ internal class TimelineChunk(private val chunkEntity: ChunkEntity,
     private fun handleDatabaseChangeSet(results: RealmResults<TimelineEventEntity>, changeSet: OrderedCollectionChangeSet) {
         val insertions = changeSet.insertionRanges
         for (range in insertions) {
+            // Check if the insertion's displayIndices match our expectations - or skip this insertion.
+            // Inconsistencies (missing messages) can happen otherwise if we get insertions before having loaded all timeline events of the chunk.
+            if (builtEvents.isNotEmpty()) {
+                // Check consistency to item before insertions
+                if (range.startIndex > 0) {
+                    val firstInsertion = results[range.startIndex]!!
+                    val lastBeforeInsertion = builtEvents[range.startIndex - 1]
```
ganfra, line 439: If there is no built event yet, this could crash

SpiritCroc: There is `builtEvents.isNotEmpty()` in line 435

ganfra: Oh yeah; I'm still wandering if we should try to fix the underneath issue here instead of doing some weird checks :/

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineChunk.kt
```diff
@@ -430,6 +430,31 @@ internal class TimelineChunk(private val chunkEntity: ChunkEntity,
     private fun handleDatabaseChangeSet(results: RealmResults<TimelineEventEntity>, changeSet: OrderedCollectionChangeSet) {
         val insertions = changeSet.insertionRanges
         for (range in insertions) {
+            // Check if the insertion's displayIndices match our expectations - or skip this insertion.
+            // Inconsistencies (missing messages) can happen otherwise if we get insertions before having loaded all timeline events of the chunk.
+            if (builtEvents.isNotEmpty()) {
+                // Check consistency to item before insertions
+                if (range.startIndex > 0) {
+                    val firstInsertion = results[range.startIndex]!!
```
ariskotsomitopoulos, line 438: I would avoid `!!` even if we are sure it cannot be null

ariskotsomitopoulos (approved):
I didn't test the logic in the code. Just a minor comment

Issue #5092, opened 2022-01-28
Timeline is doing weird things


Linerly:
### Steps to reproduce

1. Open a (big) room, like Matrix HQ

#### What did you expect?
Everything to be fine

#### What happened instead?

https://user-images.githubusercontent.com/47475676/151550614-a51f712d-da87-445e-8028-02f13c526982.mp4

### Your phone model

Mito W1

### Operating system version

Android 9

### Application version and app store

1.3.16 [40103162] (G-b5744)

### Homeserver

matrix.org

### Will you send logs?

Yes

T-bond:
I had a very similar issue a few days ago, and I sent a regshake about it. I have an own home server, and the problem appeared in a federated room.

SpiritCroc:
The new timeline mechanisms seems broken to me as well. When opening chats using permalinks, I often find missing messages or messages in the wrong order. Here an example from v1.3.16, where my bot has in fact counted correctly, but upon entering the room using a permalink to the "1 - spam" message, there are lots of messages missing (all messages from 2 to 76):

![Screenshot_20220202-115346_Element_dbg](https://user-images.githubusercontent.com/12481742/152141038-277dee43-fb75-4e36-bf44-20cf6edbba02.png)

Element did show the missing messages before, but did not show them upon entering using the permalink to the message which I clicked in a different room. It might be worth to note that my bot was still spamming me in this chat while I clicked on the link to the message. I think I experienced similar behavior in inactive chats as well though.

SpiritCroc:
I think I'm somewhat close to fixing the issue I described above. Seems to be a little different then the original issue here, so moved it to https://github.com/vector-im/element-android/issues/5448

PR #5554, opened 2022-03-16
Fix endless loading timeline due to conflicting chunks

SpiritCroc:
I ran into following scenario recently: at some point, scrolling down to more recent messages wasn't working anymore, it would endlessly show the loading animation. This was reproducible, so I started debugging. It turned out:

- The TimelineChunk I was viewing had `chunkEntity.nextChunk == null`.
- When trying to download a new chunk, `TokenChunkEventPersistor` aborted, as the next chunk was already in the database.

It turned out that I had multiple chunks in my DB, so only one had the correct reference to `nextChunk`, while the other one, which I was viewing, was stuck at loading forever.

This commit works around this by "repairing" links in this case. I assume there is a better way/place to fix it, but this works for me for now. I figured it might still be interesting for you, maybe you have some suggestions how to better fix it.

## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 11

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TokenChunkEventPersistor.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TokenChunkEventPersistor.kt
@@ -109,8 +109,14 @@ internal class TokenChunkEventPersistor @Inject constructor(
                         this.nextChunk = nextChunk
                         this.prevChunk = prevChunk
                     }
-                    nextChunk?.prevChunk = currentChunk
-                    prevChunk?.nextChunk = currentChunk
+                    val allNextChunks = ChunkEntity.findAll(realm, roomId, prevToken = nextToken)
+                    val allPrevChunks = ChunkEntity.findAll(realm, roomId, nextToken = prevToken)
+                    allNextChunks?.forEach {
+                        it.prevChunk = currentChunk
+                    }
+                    allPrevChunks?.forEach {
+                        it.nextChunk = currentChunk
+                    }
                     if (receivedChunk.events.isEmpty() && !receivedChunk.hasMore()) {
                         handleReachEnd(roomId, direction, currentChunk)
                     } else {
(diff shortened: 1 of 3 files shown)


ganfra:
Thats weird you got multiple same chunk in the db with only one valid...

SpiritCroc:
This seems to happen more often though, we got a few user reports of similar endless loading issues. With the first commit, such issues still happened, but got fixed after re-opening the room. Since the second commit, I haven't observed any new reports that sound like this issue.

ganfra:
Thanks, I'll check a bit! 
Are you working at Beeper now? :D

SpiritCroc:
Yes, (part-time) since around April, Beeper-Android was technically a fork of SchildiChat initially :)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TokenChunkEventPersistor.kt
```diff
@@ -80,7 +81,26 @@ internal class TokenChunkEventPersistor @Inject constructor(
 
                     val existingChunk = ChunkEntity.find(realm, roomId, prevToken = prevToken, nextToken = nextToken)
                     if (existingChunk != null) {
-                        Timber.v("This chunk is already in the db, returns")
+                        Timber.v("This chunk is already in the db, checking if this might be caused by broken links")
```
ganfra, line 84: Can you just create a method to keep this a bit smaller

ganfra: By the way, is this still necessary with your second commit?

SpiritCroc: I can probably do that this weekend or next week.
It's probably only needed if we want to retrospectively fix existing cases of the issue without requiring users to do a fresh initial sync to fix it. Commit two should avoid this scenario, if I haven't missed any case.

ganfra: Yeah, this should, even if there is probably another issue underneath as I don't think we should have multiple same chunks in the db :/

SpiritCroc: We still have some issues with timeline chunk loops, like in this issue: https://github.com/vector-im/element-android/issues/5092. With some added debugging, I've seen loops starting with 4 chunks, and up to 81 chunks forming a loop (timeline chunks linking forwards and backwards indirectly to the same chunk). Seems more likely to happen while viewing a room in which spam is going on. I wonder if those loops are caused by a related bug to this one.

ganfra: it's kind of hard to replicate, but there is something wrong definitively. 
Nevermind, let's merge this to mitigate. Thanks!

ganfra (approved):

PR #5576, opened 2022-03-18
Threads Migration

### LINKED ISSUE #5451: Inform labs threads users that their old threads have moved to the main timeline
When we update the client to consider m.thread as the threads prefix in https://github.com/vector-im/element-android/issues/5364, the user will see their old threads move into the main timeline.

We need to inform users that have labs enabled in threads that this has occurred.

ariskotsomitopoulos:
This PR aims to smoothly migrate users to threads live implementation

- For users with **threads enabled** from lab settings will automatically enable live threads, so no action needed from their side. They will also be notified about the threads transition and what will happen with their old messages
- Preparation to enable threads for everyone. Threads will remain a lab setting until the next RC

Closes #5451

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/lightweight/LightweightSettingsStorage.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/lightweight/LightweightSettingsStorage.kt
@@ -24,13 +24,13 @@ import javax.inject.Inject
 
 /**
  * The purpose of this class is to provide an alternative and lightweight way to store settings/data
- * on the sdi without using the database. This should be used just for sdk/user preferences and
+ * on the sdk without using the database. This should be used just for sdk/user preferences and
  * not for large data sets
  */
 
 class LightweightSettingsStorage  @Inject constructor(
         context: Context,
-        val matrixConfiguration: MatrixConfiguration
+        private val matrixConfiguration: MatrixConfiguration
 ) {
 
     private val sdkDefaultPrefs = PreferenceManager.getDefaultSharedPreferences(context.applicationContext)
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/lightweight/LightweightSettingsStorage.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/lightweight/LightweightSettingsStorage.kt
@@ -19,6 +19,7 @@ package org.matrix.android.sdk.internal.database.lightweight
 import android.content.Context
 import androidx.core.content.edit
 import androidx.preference.PreferenceManager
+import org.matrix.android.sdk.api.MatrixConfiguration
 import javax.inject.Inject
 
 /**
@@ -27,7 +28,10 @@ import javax.inject.Inject
  * not for large data sets
  */
 
-class LightweightSettingsStorage  @Inject constructor(context: Context) {
+class LightweightSettingsStorage  @Inject constructor(
+        context: Context,
+        val matrixConfiguration: MatrixConfiguration
+) {
 
     private val sdkDefaultPrefs = PreferenceManager.getDefaultSharedPreferences(context.applicationContext)
 
@@ -38,7 +42,7 @@ class LightweightSettingsStorage  @Inject constructor(context: Context) {
     }
 
     fun areThreadMessagesEnabled(): Boolean {
-        return sdkDefaultPrefs.getBoolean(MATRIX_SDK_SETTINGS_THREAD_MESSAGES_ENABLED, false)
+        return sdkDefaultPrefs.getBoolean(MATRIX_SDK_SETTINGS_THREAD_MESSAGES_ENABLED, matrixConfiguration.threadMessagesEnabledDefault)
     }
 
     companion object {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/relation/threads/FetchThreadSummariesTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/relation/threads/FetchThreadSummariesTask.kt
@@ -44,7 +44,7 @@ internal interface FetchThreadSummariesTask : Task<FetchThreadSummariesTask.Para
     data class Params(
             val roomId: String,
             val from: String = "",
-            val limit: Int = 100,
+            val limit: Int = 500,
             val isUserParticipating: Boolean = true
     )
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
@@ -75,7 +75,7 @@ internal class SyncResponseHandler @Inject constructor(
     suspend fun handleResponse(syncResponse: SyncResponse,
                                fromToken: String?,
                                reporter: ProgressReporter?) {
-        val isInitialSync = fromToken == null
+        var isInitialSync = fromToken == null
         Timber.v("Start handling sync, is InitialSync: $isInitialSync")
 
         measureTimeMillis {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
@@ -75,7 +75,7 @@ internal class SyncResponseHandler @Inject constructor(
     suspend fun handleResponse(syncResponse: SyncResponse,
                                fromToken: String?,
                                reporter: ProgressReporter?) {
-        var isInitialSync = fromToken == null
+        val isInitialSync = fromToken == null
         Timber.v("Start handling sync, is InitialSync: $isInitialSync")
 
         measureTimeMillis {
--- a/vector/src/main/java/im/vector/app/features/MainActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/MainActivity.kt
@@ -241,7 +241,7 @@ class MainActivity : VectorBaseActivity<ActivityMainBinding>(), UnlockedActivity
                 // We have a session.
                 // Check it can be opened
                 if (sessionHolder.getActiveSession().isOpenable) {
-                    HomeActivity.newIntent(this)
+                    HomeActivity.newIntent(this, existingSession = true)
                 } else {
                     // The token is still invalid
                     navigator.softLogout(this)
--- a/vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
@@ -90,6 +90,7 @@ import javax.inject.Inject
 data class HomeActivityArgs(
         val clearNotification: Boolean,
         val accountCreation: Boolean,
+        val existingSession: Boolean = false,
         val inviteNotificationRoomId: String? = null
 ) : Parcelable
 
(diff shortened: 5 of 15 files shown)


vector/src/main/java/im/vector/app/features/MainActivity.kt
```diff
@@ -241,7 +241,7 @@ class MainActivity : VectorBaseActivity<ActivityMainBinding>(), UnlockedActivity
                 // We have a session.
                 // Check it can be opened
                 if (sessionHolder.getActiveSession().isOpenable) {
-                    HomeActivity.newIntent(this)
+                    HomeActivity.newIntent(this, existingSession = true)
```
ariskotsomitopoulos, line 244: Here we want to invoke threads migration (clear cache) only if there is no initial sync already applied. For example when a user login for the first time existing session will be false, so no migration will happen

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt
```diff
@@ -75,7 +75,7 @@ internal class SyncResponseHandler @Inject constructor(
     suspend fun handleResponse(syncResponse: SyncResponse,
                                fromToken: String?,
                                reporter: ProgressReporter?) {
-        val isInitialSync = fromToken == null
+        var isInitialSync = fromToken == null
```
ouchadam, line 78: looks like the `isInitialSync` isn't changed, could it stay as a `val`?

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -130,6 +133,47 @@ class HomeActivityViewModel @AssistedInject constructor(
...
 
+    /**
+     * Handle threads migration. The migration includes:
+     * - Notify users that had io.element.thread enabled from labs
+     * - Re-Enable m.thread to those users (that they had enabled labs threads)
+     * - Handle migration when threads are enabled by default
+     */
+    private fun initThreadsMigration() {
+        // Notify users
+        if (vectorPreferences.shouldNotifyUserAboutThreads() && vectorPreferences.areThreadMessagesEnabled()) {
```
ouchadam, line 144: if each block has a return we could use a `when` and avoid multiple exit points 

```kotlin
when {
 vectorPreferences.shouldNotifyUserAboutThreads() && vectorPreferences.areThreadMessagesEnabled() -> ...
 vectorPreferences.shouldNotifyUserAboutThreads() && vectorPreferences.shouldMigrateThreads() -> ...
}
```

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -130,6 +133,47 @@ class HomeActivityViewModel @AssistedInject constructor(
                 .launchIn(viewModelScope)
     }
 
+    /**
+     * Handle threads migration. The migration includes:
+     * - Notify users that had io.element.thread enabled from labs
+     * - Re-Enable m.thread to those users (that they had enabled labs threads)
+     * - Handle migration when threads are enabled by default
+     */
+    private fun initThreadsMigration() {
```
ouchadam, line 142: how would you feel about extracting this logic to its own `ThreadsMigrator` class?  I'm wondering if in the future we would want to add more migrations to this `ViewModel`

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
```diff
@@ -269,6 +272,48 @@ class HomeActivity :
...
+        vectorPreferences.threadsMigrated()
+        MainActivity.restartApp(this, MainActivityArgs(clearCache = true))
+    }
+
+    private fun handleNotifyUserForThreadsMigration() {
+        MaterialAlertDialogBuilder(this)
+                .setTitle(R.string.threads_notice_migration_title)
+                .setMessage(R.string.threads_notice_migration_message)
+                .setCancelable(true)
+                .setPositiveButton(R.string.sas_got_it) { _, _ -> }
```
ouchadam, line 313: should/does something happen after pressing `got it`?

ariskotsomitopoulos: Nothing, it is just informational

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
ariskotsomitopoulos: I agree 100%

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
ariskotsomitopoulos: Sounds good !

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
ouchadam: ah, so the next time they launch the app they'll fall in to the other branch which does the migration?

not a blocker for the PR but I wonder if the dialog could be "migrate now or maybe later"

ariskotsomitopoulos: It will not let the user decide to migrate, the notice is just informative. Users that had threads enable will receiver this notice, that will inform them that old messages will be visible in the main timeline and will be automatically migrated to live threads.

ouchadam: this is the bit that confuses me, when does the automatic migration happen for these users?

ariskotsomitopoulos: At once they open the updated app.
- I am a user and I have lab threads enabled
- I update the app and the migration will happen on the first app run

Is it more clear now?

ouchadam: :+1: yep thanks! the message is telling users they have been migrated

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
ariskotsomitopoulos: I am not sure if it would be more clear to extract the logic while it mainly invokes  `_viewEvents.post ` events, and I am afraid it might be confusing

ouchadam: I was imagining split between the view event creation and the logic of deciding the migration status
not a blocker for me as it's not something we do often in the current codebase 

```kotlin
sealed interface Result { 
  object Migrated: Result
  data class Migrate(migrateExistingSession: Boolean)
}

class ThreadsMigator {
  fun migrate() : Result { }
}


// viewmodel
fun initThreadsMigration() {
   val event = when(val result = threadsMigrator.migrate()) {
     Migrated -> NotifyUserForThreadsMigration
     ...
   }
   post(event)
}

```

has the benefit of writing tests simpler for both migration and viewmodel ^^^ but does introduce extra boilerplate for bridging the layers

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -130,6 +133,47 @@ class HomeActivityViewModel @AssistedInject constructor(
...
+            // instead of io.element.thread so old thread messages will be displayed as normal timeline messages
+            _viewEvents.post(HomeActivityViewEvents.NotifyUserForThreadsMigration)
+            vectorPreferences.userNotifiedAboutThreads()
+            return
+        }
+
+        // Migrate users with enabled lab settings
+        if (vectorPreferences.shouldNotifyUserAboutThreads() && vectorPreferences.shouldMigrateThreads()) {
+            Timber.i("----> Migrate threads with enabled labs")
+            // If user had io.element.thread enabled then enable the new thread support,
```
ouchadam: these comments look like great candidates for unit test cases 

```
given user has labs io.element.thread enabled, 
when checking migration 
then sets threads enabled and returns Migrate without clearing cache result
```

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -130,6 +133,47 @@ class HomeActivityViewModel @AssistedInject constructor(
...
+            // clear cache to sync messages appropriately
+            vectorPreferences.setThreadMessagesEnabled()
+            lightweightSettingsStorage.setThreadMessagesEnabled(vectorPreferences.areThreadMessagesEnabled())
+            // Clear Cache
+            _viewEvents.post(HomeActivityViewEvents.MigrateThreads(checkSession = false))
+            return
+        }
+        // Enable all users
+        // When we would to enable threads for all
```
ouchadam: rather than commented code, this could be behind a feature flag https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/VectorFeatures.kt#L21

ouchadam (approved):
LGTM :100:  might be worth a peek from @ganfra or @bmarty  in case there's some other examples of migration we've done for similar features

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/lightweight/LightweightSettingsStorage.kt
```diff
@@ -27,7 +28,10 @@ import javax.inject.Inject
  * not for large data sets
  */
 
-class LightweightSettingsStorage  @Inject constructor(context: Context) {
+class LightweightSettingsStorage  @Inject constructor(
+        context: Context,
+        val matrixConfiguration: MatrixConfiguration
```
ganfra, line 33: private?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/relation/threads/FetchThreadSummariesTask.kt
```diff
@@ -44,7 +44,7 @@ internal interface FetchThreadSummariesTask : Task<FetchThreadSummariesTask.Para
     data class Params(
             val roomId: String,
             val from: String = "",
-            val limit: Int = 100,
+            val limit: Int = 500,
```
ganfra, line 47: Why 500 as default?

vector/src/main/java/im/vector/app/features/settings/VectorPreferences.kt
```diff
@@ -1006,7 +1012,56 @@ class VectorPreferences @Inject constructor(private val context: Context) {
...
+                .edit()
+                .putBoolean(SETTINGS_LABS_ENABLE_THREAD_MESSAGES_OLD_CLIENTS, false)
+                .apply()
+    }
+
+    /**
+     * Indicates whether or not we should clear cache for threads migration.
+     * Default value is true, for fresh installs and updates
+     */
+    fun shouldMigrateThreads(): Boolean {
```
ganfra, line 1054: We will have initial sync here for most people then?

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
```diff
@@ -90,6 +90,7 @@ import javax.inject.Inject
 data class HomeActivityArgs(
         val clearNotification: Boolean,
         val accountCreation: Boolean,
+        val existingSession: Boolean = false,
```
ganfra, line 93: maybe rename 'hasExistingSession" or sth like that?

ganfra (commented):
Some remarks & questions

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/relation/threads/FetchThreadSummariesTask.kt
ariskotsomitopoulos: Until backend will support pagination, we concluded to increase the limit here

vector/src/main/java/im/vector/app/features/settings/VectorPreferences.kt
ariskotsomitopoulos: Yes

ariskotsomitopoulos: In this RC only users with labs threads enabled, in the next RC all users

PR #5583, opened 2022-03-18
Add a setting to be able to always appear offline

### LINKED ISSUE #5582: Move presence enabling to an offline mode user setting
### Your use case

Remove build config for Presence enabling as its not really needed and replace it with a user setting / preference

No

### Additional context

No

aringenbach:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Presence enabling has been moved to an offline mode user setting. When enabled, user will always appear offline, even when using the application.

## Motivation and context

Fixes #5582 

## Screenshots / GIFs

Screenshot_20220412_101940

## Tests

- Step 1: Have contacts on servers that show presence (e.g. local synapse)
- Step 2: Toggle setting

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): API 32

## Checklist


vector/src/main/java/im/vector/app/features/settings/VectorSettingsPreferencesFragment.kt
```diff
@@ -71,6 +71,14 @@ class VectorSettingsPreferencesFragment @Inject constructor(
             }
         }
 
+        findPreference<VectorSwitchPreference>(VectorPreferences.SETTINGS_SHOW_PRESENCE)!!.let { pref ->
+            pref.isChecked = vectorPreferences.showPresence()
+            pref.setOnPreferenceChangeListener { _, _ ->
+                MainActivity.restartApp(requireActivity(), MainActivityArgs(clearCache = false))
```
aringenbach, line 77: I wonder if we can do better than restarting the MainActivity to take changes into account

aringenbach (commented):
This might require input from product team, as this is adding a setting we might not want for some reason.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/PresenceSyncHandler.kt
```diff
@@ -31,26 +31,24 @@ import javax.inject.Inject
 internal class PresenceSyncHandler @Inject constructor(private val matrixConfiguration: MatrixConfiguration) {
 
     fun handle(realm: Realm, presenceSyncResponse: PresenceSyncResponse?) {
-        if (matrixConfiguration.presenceSyncEnabled) {
```
ariskotsomitopoulos, line 34: Here you removed the enable/disable logic. So now even with presence disabled the code will run.

vector/src/main/java/im/vector/app/features/settings/VectorSettingsPreferencesFragment.kt
ariskotsomitopoulos: Why do we need an activity restart here ? vectorPreferences.showPresence() should have the updated value after user changes the setting, it doesn't ?

vector/src/main/res/xml/vector_settings_preferences.xml
```diff
@@ -198,6 +198,16 @@
 
     </im.vector.app.core.preference.VectorPreferenceCategory>
 
+    <im.vector.app.core.preference.VectorPreferenceCategory android:title="@string/settings_presence">
+
+        <im.vector.app.core.preference.VectorSwitchPreference
+            android:title="@string/settings_show_presence"
+            android:summary="@string/settings_show_presence_summary"
+            android:key="SETTINGS_SHOW_PRESENCE"
+            android:defaultValue="true" />
```
ariskotsomitopoulos, line 207: The default value here should be the default value you defined in vector-config I guess

ariskotsomitopoulos (commented):
Thanks for the PR, I added some comments

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/PresenceSyncHandler.kt
aringenbach: Yes this is on purpose. Since using a user setting, I wanted Realm to store Presence data regardless, so toggling is just a matter of refreshing the UI rather than having to force re-sync just to get Presence data we ignored before.

ariskotsomitopoulos: If its the expected behaviour its fine, thanks

aringenbach:
> This might require input from product team, as this is adding a setting we might not want for some reason.

Done. It's ok to have this setting available, and it will be added on iOS as well as part of the presence display improvement there.

vector/src/main/java/im/vector/app/features/settings/VectorSettingsPreferencesFragment.kt
ariskotsomitopoulos: If the problem is for example the room list recyclerview, maybe you can try to call `requestModelBuild`  to request a model update so that epoxy will rebuild the UI for the current updated data

ariskotsomitopoulos (approved):
It would be good to avoid the activity restart if possible. I added a comment regarding that

vector-config/src/main/res/values/config-settings.xml
```diff
@@ -28,6 +28,8 @@
 
     <bool name="settings_interface_bubble_visible">true</bool>
     <bool name="settings_interface_bubble_default">false</bool>
+    <bool name="settings_presence_visible">false</bool>
+    <bool name="settings_presence_offline_mode_default">false</bool>
```
aringenbach, line 32: This feels worthless, I wonder if we could simply pass on this, since it kinda doubles on the default value in MatrixConfiguration 🤔

vector/src/main/java/im/vector/app/features/settings/VectorSettingsPreferencesFragment.kt
```diff
@@ -71,6 +76,17 @@ class VectorSettingsPreferencesFragment @Inject constructor(
             }
         }
 
+        findPreference<VectorSwitchPreference>(LightweightSettingsStorage.MATRIX_SDK_SETTINGS_PRESENCE_OFFLINE_MODE_ENABLED)!!.let { pref ->
+            pref.isChecked = lightweightSettingsStorage.getPresenceOfflineModeEnabled()
+            pref.setOnPreferenceChangeListener { _, newValue ->
+                val presenceOfflineModeEnabled = newValue as? Boolean ?: false
+                lifecycleScope.launch {
```
aringenbach, line 83: I wonder if there is a better place to start this but right now we don't have a viewModel or interactor on Preferences

vector/src/main/java/im/vector/app/features/settings/VectorSettingsPreferencesFragment.kt
```diff
@@ -71,6 +76,17 @@ class VectorSettingsPreferencesFragment @Inject constructor(
...
+                    session.setMyPresence(if (presenceOfflineModeEnabled) PresenceEnum.OFFLINE else PresenceEnum.ONLINE)
```
ariskotsomitopoulos, line 84: Keep in mind there is also a 3rd value `PresenceEnum.UNAVAILABLE`, that I think we currently we are not using

ariskotsomitopoulos (commented):
Thanks for the update, some comments

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixConfiguration.kt
```diff
@@ -61,14 +61,14 @@ data class MatrixConfiguration(
...
-         */
-        val presenceSyncEnabled: Boolean = true,
         /**
          * Thread messages default enable/disabled value
          */
         val threadMessagesEnabledDefault: Boolean = false,
+        /**
+         * Presence offline mode default enabled/disabled value
+         */
+        val presenceOfflineModeEnabledDefault: Boolean = false,
```
bmarty, line 71: I think naming could maybe be a bit clearer.
What about: `userAlwaysAppearsOfflineByDefault`?

Other names could be updated using the same terminology.

vector/src/main/java/im/vector/app/features/settings/VectorSettingsPreferencesFragment.kt
```diff
@@ -71,6 +76,17 @@ class VectorSettingsPreferencesFragment @Inject constructor(
             }
         }
 
+        findPreference<VectorSwitchPreference>(LightweightSettingsStorage.MATRIX_SDK_SETTINGS_PRESENCE_OFFLINE_MODE_ENABLED)!!.let { pref ->
```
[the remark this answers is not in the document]
bmarty: Not ideal that the app set a SharedPref that the SDK is using (IIUC). We should use the SDK API to pass settings to the SDK.

vector/src/main/res/values/strings.xml
```diff
@@ -1049,6 +1049,11 @@
     <string name="media_saving_period_1_month">1 month</string>
     <string name="media_saving_period_forever">Forever</string>
 
+    <!-- Presence settings -->
+    <string name="settings_presence">Presence</string>
+    <string name="settings_presence_offline_mode">Offline mode</string>
+    <string name="settings_presence_offline_mode_summary">If enabled, this client won\'t send presence online status on sync</string>
```
bmarty, line 1055: It's a bit technical for the users.
I would rephrase to "If enabled, you will always appear offline to other users, even when using the application."

bmarty (changes requested):
Some minor and less minor remarks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixConfiguration.kt
bmarty: (But I think this should be removed anyway)

vector/src/main/java/im/vector/app/features/settings/VectorSettingsPreferencesFragment.kt
bmarty: This call to `PresenceService` should be enough to tell the SDK to use `Offline` or `Online` for every sync request when on foreground

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/presence/service/DefaultPresenceService.kt
```diff
@@ -20,18 +20,22 @@ package org.matrix.android.sdk.internal.session.presence.service
...
 internal class DefaultPresenceService @Inject constructor(
         @UserId private val userId: String,
         private val setPresenceTask: SetPresenceTask,
-        private val getPresenceTask: GetPresenceTask
+        private val getPresenceTask: GetPresenceTask,
+        private val lightweightSettingsStorage: LightweightSettingsStorage
 ) : PresenceService {
 
     override suspend fun setMyPresence(presence: PresenceEnum, statusMsg: String?) {
+        lightweightSettingsStorage.setSyncPresenceStatus(SyncPresence.from(presence))
```
bmarty, line 38: Not a big deal, but storing the value could be handled by `setPresenceTask`. Not sure what is best, so let's keep it like that.

changelog.d/5582.change
```diff
@@ -0,0 +1 @@
+Move presence enabling to an offline mode user setting
```
bmarty, line 1: The file extension `change` is not correct and will be ignored by towncrier. Please refer to https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#changelog

In this case, I think it's best to use the `feature` extension. For the content, since it's more user facing, I will write "Add a setting to be able to always appear offline"

bmarty (changes requested):
Thanks for the update.

Just a last remark about the changelog.

Also the PR description is quite out of date. Can you update it please, for posterity?

Issue #2525, opened 2020-12-11
Add an embedded player for Audio files


bmarty:
As per Element Web:

<img width="331" alt="image" src="https://user-images.githubusercontent.com/3940906/101908003-ab12ef00-3bbb-11eb-8132-b4136edd1971.png">

Currently the user has to download the file and play it with a third party application, and it's rendered like that in the timeline:

<img width="343" alt="image" src="https://user-images.githubusercontent.com/3940906/101908118-df86ab00-3bbb-11eb-8588-315ff05fc1ca.png">

Edit: From @daniellekirkwood Can we aim to use the same component as voice messages when implementing the solve for this?

DanHakimi:
Highly in favor of this.

For the record, the Element Web player is just the default HTML5 audio player for the browser. I'm not a fan of this; I think Element should have a consistent audio experience across all element apps, which I think demands a dedicated player style.

532910:
An embedded player must have a speed control change. Should a separate issue be opened or it's a part of this one?

DanHakimi:
... why would it have a speed control change? That sounds like extreme clutter. Most other apps (whatsapp, telegram) don't even have a volume button on the embedded player, because they value design. It's an open source cliche to stuff niche functionality in and clutter the design. I might _never_ use a speed control changer for audio messages I receive. If you want one, download the file and open it in your audio player of choice.

532910:
I'm not a tg user, but even I know that it HAS speed control change! Youtube also has speed control change.

Could you explain, why should I waste my time listening at 1x speed?

notramo:
Having a "Skip silence" button is also useful for voice messages.

532910:
Maybe it would be better just not create silence during the recording phase.

iav:
Perhaps it would be better if everyone around us became as convenient for us.
But so far this is not the case.

DanHakimi:
> I'm not a tg user, but even I know that it HAS speed control change!

[I don't see it here](https://core.telegram.org/file/811140719/1/ZMopdSMCNRc/a4b95d3fc0c26b93d9). Oh, and I just checked fluffychat -- they don't have it either, and they don't have volume control either. So I am very firmly of the opinion that it should _just_ be a seek bar + play/pause button.

> Youtube also has speed control change.

[...]

PR #5586, opened 2022-03-19
Audio files in the timeline now appear with the audio player

ericdecanini:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Instead of appearing as a simple downloadable file, audio files uploaded in the timeline will now be displayed with the audio player so they are playable from the timeline

I took this opportunity to improve our UX with the way the timeline handles playing multiple audio files (by that I mean playing one while another one is playing so the other pauses, etc)

The seekbar is a little fiddly though. Clicking on it works nicely to seek and so does swiping right, but swiping left doesn't work (not before already being in a seeking state) because it conflicts with the `ItemTouchHelper` of the recycler view which already has a swipe left action being handled. Due to this being on the level of the recycler view, I couldn't find a way to handle this from within the context of the audio item

## Motivation and context

Closes https://github.com/vector-im/element-android/issues/2525

## Screenshots / GIFs

| Light Mode | Dark Mode |
|---|---|
| image | image |

## Tests

**Audio Files**
- Upload an audio file in the timeline
- See that it displays differently from a regular file attachment and plays the audio
- See that the seekbar works
- Try uploading a second audio file and see that they interact nicely with each other
- Try backgrounding and foregrounding the app and see that no strange behaviour is found

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Android 10, 12

## Checklist


ericdecanini:
@daniellekirkwood @amshakal can I ask you guys to take a look at this new UI?

amshakal:
Tagging @gaelledel who worked on voice messages UI.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageAudioItem.kt
```diff
@@ -0,0 +1,159 @@
...
+        renderSendState(holder.rootLayout, null)
+        bindFilenameViewAttributes(holder)
+        bindUploadState(holder)
+        applyLayoutTint(holder)
+        holder.audioPlaybackControlButton.setOnClickListener { playbackControlButtonClickListener?.invoke(it) }
+        renderStateBasedOnAudioPlayback(holder)
+    }
+
+    private fun bindUploadState(holder: Holder) {
+        if (!attributes.informationData.sendState.hasFailed()) {
```
mnaturel, line 77: Same remark about the condition, it is better to use positive condition when possible, in that case, we would have:
```
if (attributes.informationData.sendState.hasFailed()) {
    holder.audioPlaybackControlButton.setImageResource(R.drawable.ic_cross)
    holder.audioPlaybackControlButton.contentDescription =
    holder.view.context.getString(R.string.error_audio_message_unable_to_play, filename)
    holder.progressLayout.isVisible = false
} else {
     contentUploadStateTrackerBinder.bind(attributes.informationData.eventId, isLocalFile, holder.progressLayout)
}
```

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageVoiceItem.kt
```diff
@@ -55,10 +55,12 @@ abstract class MessageVoiceItem : AbsMessageItem<MessageVoiceItem.Holder>() {
     var waveform: List<Int> = emptyList()
 
     @EpoxyAttribute
-    var izLocalFile = false
+    @JvmField
+    var isLocalFile = false
```
mnaturel, line 59: Should we really keep these fields which you annotated with `@JvmFields`. We could use `private const` instead or directly the boolean value where it is called. Same remark for the field `isLocalFile` in `MessageAudioItem`. What do you think?

mnaturel (commented):
Good work! I have only small remarks. When testing, I have seen something strange. I play an audio file in a room. Then I press back which leads me to previous screen. It pauses the audio but does not reset it to the start. It may appear strange to users since they have no way to restart at the beginning of the audio except restarting the app. It may be a side effect of the bug you fixed when going in background.

I have also a question about UI: Could we indicate somewhere near the filename the total duration of the audio track? When we start playing we loose that information and since there is currently no seekbar we don't really know where we are in the audio track.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -485,13 +475,24 @@ class MessageItemFactory @Inject constructor(
             .iconRes(R.drawable.ic_paperclip)
     }
 
+    private fun buildAudioContent(
+            params: TimelineItemFactoryParams,
+            messageContent: MessageAudioContent,
+            informationData: MessageInformationData,
+            highlight: Boolean,
+            attributes: AbsMessageItem.Attributes,
+    ) = if (messageContent.voiceMessageIndicator != null) {
```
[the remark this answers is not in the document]
ericdecanini: I can see the Uncle Bob here ✨ Clean Code never mentions this in the case of null checks (and I can't find anything online about it), but imo I think `!= null` is actually positive, because it's akin to "if this data exists"

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -409,14 +405,8 @@ class MessageItemFactory @Inject constructor(
...
-            .izLocalFile(localFilesHelper.isLocalFile(fileUrl))
-            .izDownloaded(session.fileService().isFileInCache(
-                fileUrl,
-                messageContent.getFileName(),
-                messageContent.mimeType,
-                messageContent.encryptedFileInfo?.toElementToDecrypt())
-            )
+            .audioMessagePlaybackTracker(audioMessagePlaybackTracker)
+            .isLocalFile(localFilesHelper.isLocalFile(fileUrl))
             .mxcUrl(fileUrl)
```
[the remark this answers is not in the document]
ericdecanini: Good spot! `isDownloaded` in `MessageVoiceItem` isn't actually being used. I'll push changes to remove it from that class.

It's used only in `MessageFileItem` to determine the icon that's shown, but in the case of voice and audio items this is replaced by the play button

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageVoiceItem.kt
ericdecanini: These can't be `const` as they're set via epoxy, but the reason for adding `@JvmField` here is Epoxy returns a compile error without it.

What I find weird is that this only happened after renaming it to fix the spelling. The reasoning for that is something that completely baffles me 🤷

ericdecanini:
> Good work! I have only small remarks. When testing, I have seen something strange. I play an audio file in a room. Then I press back which leads me to previous screen. It pauses the audio but does not reset it to the start. It may appear strange to users since they have no way to restart at the beginning of the audio except restarting the app. It may be a side effect of the bug you fixed when going in background.
> 
> I have also a question about UI: Could we indicate somewhere near the filename the total duration of the audio track? When we start playing we loose that information and since there is currently no seekbar we don't really know where we are in the audio track.

Good catch with that bug! It should be an easy fix so I'll push that in a bit.

With that UI suggestion, that is a good point. I would like to know what design thinks about it

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageVoiceItem.kt
mnaturel: I agree. I prefer good naming as well.

mnaturel (approved):
Good for me. I will let you see with design team if you want to add total duration of the audio track somewhere near the title.

gaelledel (changes requested):
Cool! Just need to follow the specs for audio files available in compound here https://www.figma.com/file/X4XTH9iS2KGJ2wFKDqkyed/Compound?node-id=2039%3A26421

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerAction.kt
```diff
@@ -42,4 +42,5 @@ sealed class MessageComposerAction : VectorViewModelAction {
     data class EndAllVoiceActions(val deleteRecord: Boolean = true) : MessageComposerAction()
     data class VoiceWaveformTouchedUp(val eventId: String, val duration: Int, val percentage: Float) : MessageComposerAction()
     data class VoiceWaveformMovedTo(val eventId: String, val duration: Int, val percentage: Float) : MessageComposerAction()
+    data class AudioSeekBarMovedTo(val eventId: String, val duration: Int, val percentage: Float) : MessageComposerAction()
```
mnaturel, line 45: I am wondering if this is really an action of the `MessageComposerView`? Does it handle the move of the seek bar inside the timeline or when composing a new message? The question is the same for `VoiceWaveformTouchedUp` and `VoiceWaveformMovedTo`. And maybe others but I don't have all the understanding of the feature.

mnaturel (commented):
Nice work! We now have a real audio player, hope it will be appreciated by users! I have just a question about the place of some `ViewActions`.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerAction.kt
ericdecanini: You make a strong point! It doesn't sound like this and some of the other voice actinos like the ones you mentioned and ones concerning play/pause functionality should belong here. Perhaps I could work it in another PR to separate these out a little bit

vector/src/main/java/im/vector/app/features/home/room/detail/composer/AudioMessageHelper.kt
```diff
@@ -171,17 +174,24 @@ class VoiceMessageHelper @Inject constructor(
...
-        stopPlayback()
-        playbackTracker.pausePlayback(id)
+        if (currentPlayingId == id) {
+            mediaPlayer?.seekTo(toMillisecond)
+            playbackTracker.updatePlayingAtPlaybackTime(id, toMillisecond, percentage)
+        } else {
+            mediaPlayer?.pause()
+            playbackTracker.updatePausedAtPlaybackTime(id, toMillisecond, percentage)
+            stopPlaybackTicker()
+        }
```
bmarty, line 194: 👀  thanks!

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageVoiceItem.kt
bmarty: OK

vector/src/main/res/values/strings.xml
```diff
@@ -2860,6 +2860,14 @@
     <string name="error_voice_message_cannot_reply_or_edit">Cannot reply or edit while voice message is active</string>
     <string name="voice_message_reply_content">Voice Message (%1$s)</string>
 
+    <string name="a11y_audio_message_item">%1$s, %2$s, %3$s</string> <!-- filename, duration, file size -->
```
bmarty, line 2863: Comments have to be above the string for Weblate to handle it properly.

bmarty (commented):
My 2 cents

vector/src/main/res/values/strings.xml
ericdecanini: Added this to https://github.com/vector-im/element-android/issues/5714

PR #5592, opened 2022-03-21
Avoid accessing realm instance from suspend functions

ariskotsomitopoulos:
Add runBlocking when decrypt events to avoid thread switching when accessing the realm instance (thread local)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineEventDecryptor.kt
```diff
@@ -101,9 +101,7 @@ internal class TimelineEventDecryptor @Inject constructor(
         executor?.execute {
             Realm.getInstance(realmConfiguration).use { realm ->
                 try {
-                    runBlocking {
```
ouchadam, line 104: if the aim is to ensure the realm instance is only executed on a single thread it might be safer to keep the `runBlocking` here next to the instance creation

ouchadam:
it's a bit tricky to tell from the diff, where does the thread jumping happen? 

I was expecting to see a `withContext` or `scope.launch` being replaced/removed

ariskotsomitopoulos:
> scope.launch

From my understanding without `suspend` function, no thread hopping can be done. Maybe @BillCarsonFr has a better input on that.

BillCarsonFr (changes requested):
I would just wrap cryptoService.decryptEvent(XXX) with runBlocking instead of whole functions

BillCarsonFr (approved):

PR #5594, opened 2022-03-21
Improve permalink on thread root message

### LINKED ISSUE #5567: Following a link to a thread takes the user to the timeline
### Steps to reproduce

 -I'm in an existing thread
2. What do you click?
 -I tap on "Copy link to thread" in options
3. More steps…
 -I follow the link

#### What did you expect?
I expected to be taken to that thread
#### What happened instead?
I landed on the timeline instead

### Your phone model

Samsung Galaxy A40

### Operating system version

Android 11

1.4.6-dev [40104060] (G-b6874) feature/aris/thread_live_thread_list

### Homeserver

matrix.org

### Will you send logs?

No

ariskotsomitopoulos:
Permalinks to root thread messages will now navigate the user within the thread timeline 

Closes #5567

### DIFF
--- a/vector/src/main/java/im/vector/app/features/permalink/PermalinkHandler.kt
+++ b/vector/src/main/java/im/vector/app/features/permalink/PermalinkHandler.kt
@@ -38,6 +38,7 @@ import org.matrix.android.sdk.api.session.permalinks.PermalinkService
 import org.matrix.android.sdk.api.session.room.model.Membership
 import org.matrix.android.sdk.api.session.room.model.RoomSummary
 import org.matrix.android.sdk.api.session.room.model.RoomType
+import org.matrix.android.sdk.api.session.room.timeline.isRootThread
 import javax.inject.Inject
 
 class PermalinkHandler @Inject constructor(private val activeSessionHolder: ActiveSessionHolder,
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/permalink/PermalinkHandler.kt
```diff
@@ -89,7 +90,13 @@ class PermalinkHandler @Inject constructor(private val activeSessionHolder: Acti
...
                 val rootThreadEventId = permalinkData.eventId?.let { eventId ->
                     val room = roomId?.let { session?.getRoom(it) }
-                    room?.getTimelineEvent(eventId)?.root?.getRootThreadEventId()
+
+                    val rootThreadEventId = room?.getTimelineEvent(eventId)?.root?.getRootThreadEventId()
+                    rootThreadEventId ?: if (room?.getTimelineEvent(eventId)?.isRootThread() == true) {
+                        eventId
+                    } else {
+                        null
+                    }
```
bmarty, line 99: I understand the code and it should work, hence it's a bit confusing to have multiple declaration of `val rootThreadEventId`. Not a blocker anyway.

bmarty (approved):
LGTM. The code miss a bit of documentation, especially about the param `rootThreadEventId` of `openJoinedRoomScreen`. Could be nice to now the effect if its null, if it's not null, etc.

PR #5595, opened 2022-03-21
Live Location Sharing - Foreground Service

onurays:
We need to start a foreground service and show a sticky notification to be able to share live location if the application is in background and even if it is killed.

- [x] Start a foreground service when users start sharing live location
- [x] Show a sticky notification to inform users that the app continues to share location
- [x] Support sharing location in multiple rooms at the same time
- [x] We need to set a timer per room and stop foreground service once all times are up
- [x] Navigate to room list screen when users click on the notification (we will discuss showing room list in which live location is active)
- [x] Check if we can stop the location manager when the service is destroyed
- [x] Check if the service is not recreated when users share live location in another room

**Live Location Sharing is disabled by a BuildConfig. We are creating small PRs to make them easily reviewable.**

live_location_sharing_notification


vector/src/main/AndroidManifest.xml
```diff
@@ -46,6 +46,7 @@
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
```
ouchadam, line 49: we already have this permission :smile:  https://github.com/vector-im/element-android/blob/develop/vector/src/main/AndroidManifest.xml#L13

vector/src/main/java/im/vector/app/features/location/LocationSharingFragment.kt
```diff
@@ -176,6 +179,17 @@ class LocationSharingFragment @Inject constructor(
         views.mapView.zoomToLocation(event.userLocation.latitude, event.userLocation.longitude)
     }
 
+    private fun handleStartLiveLocationService(event: LocationSharingViewEvents.StartLiveLocationService) {
+        Intent(requireContext(), LocationSharingService::class.java)
+                .apply {
```
ouchadam, line 184: as the `putX` methods on `Intent` are builder style, we could avoid the apply 

```java
Intent()
  .putExtra(foo, bar)
  .also { ContextCompat.startForegroundService(requireContext(), it) }
``` 

what do you think?

vector/src/main/java/im/vector/app/features/location/LocationSharingFragment.kt
```diff
@@ -221,7 +235,9 @@ class LocationSharingFragment @Inject constructor(
     }
 
     private fun startLiveLocationSharing() {
-        viewModel.handle(LocationSharingAction.StartLiveLocationSharing)
+        // TODO. Get duration from user
+        val duration = 30 * 1000L
```
ouchadam, line 239: might be worth a constant? 

not for this PR - _thinking out of loud about todos_ I'm wondering if they are helpful or add noise, would a `User can set live location duration` ticket that references the class provide the same information :thinking:

vector/src/main/java/im/vector/app/features/location/LocationTracker.kt
```diff
@@ -38,18 +40,17 @@ class LocationTracker @Inject constructor(
...
     }
 
-    private var callback: Callback? = null
+    private var callbacks = mutableListOf<Callback>()
 
     private var hasGpsProviderLiveLocation = false
 
     @RequiresPermission(anyOf = [Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_FINE_LOCATION])
-    fun start(callback: Callback?) {
+    fun start() {
```
ouchadam, line 48: now that we're calling `start` from multiple places, do we need to guard the...

```kotlin
                    locationManager.requestLocationUpdates(
                            provider,
                            MIN_TIME_TO_UPDATE_LOCATION_MILLIS,
                            MIN_DISTANCE_TO_UPDATE_LOCATION_METERS,
                            this
                    )
```  

to avoid registering duplicate updates?

vector/src/main/java/im/vector/app/features/location/LocationTracker.kt
```diff
@@ -88,7 +89,24 @@ class LocationTracker @Inject constructor(
     fun stop() {
         Timber.d("## LocationTracker. stop()")
         locationManager?.removeUpdates(this)
-        callback = null
+        callbacks.clear()
+    }
+
+    fun addCallback(callback: Callback) {
+        synchronized(callbacks) {
```
ouchadam, line 96: if we only call `addCallback/removeCallback` from the main thread, we could avoid needing to use `synchronized`

if we are calling from multiple threads, we'll probably want to make all the callback interactions synchronised :cry:

vector/src/main/AndroidManifest.xml
onurays: Oh no! Good catch, done.

vector/src/main/java/im/vector/app/features/location/LocationSharingFragment.kt
onurays: Exactly, forgot that `puExtra` returns `Intent`. Done.

vector/src/main/java/im/vector/app/features/location/LocationSharingFragment.kt
onurays: I already started implementing time-limiting UI. I will create another quick PR today.

vector/src/main/java/im/vector/app/features/location/LocationTracker.kt
onurays: It is protected internally and doesn't create multiple requests.

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
```diff
@@ -0,0 +1,120 @@
...
+
+            // Schedule a timer to stop sharing
+            scheduleTimer(roomArgs.roomId, roomArgs.durationMillis)
+        }
+
+        return START_STICKY
+    }
+
+    private fun scheduleTimer(roomId: String, durationMillis: Long) {
+        Timer().schedule(object : TimerTask() {
```
ouchadam, line 75: we might also want to keep a reference to the timer and cancel it if it's still running when the service is destroyed

vector/src/main/java/im/vector/app/features/location/LocationTracker.kt
ouchadam: good to know, thanks!

vector/src/main/java/im/vector/app/features/location/LocationTracker.kt
onurays: You are right. It is called in a single UI thread.

ouchadam (approved):
LGTM! Would be handy to see a screenshot of the notification in action :pray: 
(and a tiny comment about the timer instance maybe leaking~)

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
onurays: Nice catch, done.

Issue #5465, opened 2022-03-08
Improve CI for release branches


michaelkaye:
We run the nightly integration tests against release branches as well as nightly against develop.

 - SonarQube test failed with an exception of some kind
 - Notification to the channel doesn't make it clear it's a release branch build and not a nightly.

michaelkaye:
If updating the notification; perhaps also include the changes to colour format and add icons for success/failure

PR #5617, opened 2022-03-23
Skip sonar on release branches, be clearer in text why nightly run is running.

michaelkaye:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [X] Technical
- [ ] Other :

## Content

Update text printed to channel when test run completes.

Don't run sonarqube other than on scheduled nightly runs.

## Motivation and context

Sonarqube needs GITHUB_TOKEN for branch pushes (why?) and the messages in the channel were causing people to ask questions.


michaelkaye:
Helps with #5465

Issue #5321, opened 2022-02-23
Wrong display name in search result for user


kittykat:
### Steps to reproduce

Not sure how to reproduce:

![Screenshot_20220223-122439.png](https://user-images.githubusercontent.com/51663/155319750-c025e8aa-6160-455c-91c5-ae8673f9897f.png)

### Expected outcome

Danielle is called "Danielle"

### Application version and app store

1.4.0 / SDK 1.3.19 / olm 3.2.10

### Homeserver

matrix.org

### Will you send logs?

Yes

daniellekirkwood:
For context... I set my DisplayName to be Jano in the Threads P0 Delivery room (which @kittykat is a member of). My Display name is **only** changed in that room (using /myroomnick on web) and so I believe the correct behaviour here is to show my regular nickname/display name **not** the one that is specific only to the Threads P0 Delivery room

kittykat:
I can confirm that I was trying to invite Danielle from an unrelated room in the Element space

daniellekirkwood:
The behaviour of room-only display names should be such that it only applies _inside_ of that room. 

@bmarty do you have any background on this? Do you agree that the behaviour should be that we do not show room specific display names in the search results?

I'm testing on iOS now...

ericdecanini:
I found this happening on the Desktop client
![image](https://user-images.githubusercontent.com/20701752/159715724-eff7aeea-e1f5-488c-8145-dd4f41498b45.png)

Are we sure it's not a backend issue?

ericdecanini:
![image](https://user-images.githubusercontent.com/20701752/159764350-2400678a-3344-4a3e-9377-2049e206c808.png)

Fix is on the way, but I should also highlight that this bug in its entirety is that using `/myroomnick` currently changes your display name across the entire app. This affects not only your search result name, but also your name beside your avatar in the navigation drawer and such. 

This fortunately doesn't affect your display name in other rooms (existing or new), but another weird fact about this bug is that your display name will be set to whatever your display name is in the room you last sent a message in

bmarty:
Fixed in 1.4.12

PR #5618, opened 2022-03-23
Fixes myRoomNick changing Display Name

ericdecanini:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Removes a user entity update when changing your nickname for a single room using the command `/myRoomNick`

## Motivation and context

Closes https://github.com/vector-im/element-android/issues/5321

When entering the above command to change the nickname, it updates both the Room Member Entity and the User Entity. It's clear why the Room Member Entity update needs to be made (this actually updates the nickname of that room member), but it's very unclear if the User Entity update is needed at all. This being updated is the source of this bug.

Additionally, changing your nickname in the room changed your display name across the app. You would even see this reflected beside your avatar in the navigation drawer.

## Tests

- In any room type the command `/myRoomNick [name]` to change your nickname for that room
- Send a message in the room and see that your nickname in that room is successfully changed
- See that your display name on the navigation drawer is as normal and not your newly set room nickname
- On another account, create a new room and go to the invite page and search for your account that set the nickname in a room. See that the display name is as normal and not the previously set room nickname

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
@@ -26,7 +26,6 @@ import org.matrix.android.sdk.internal.database.query.where
 import org.matrix.android.sdk.internal.di.UserId
 import org.matrix.android.sdk.internal.session.events.getFixedRoomMemberContent
 import org.matrix.android.sdk.internal.session.sync.SyncResponsePostTreatmentAggregator
-import org.matrix.android.sdk.internal.session.user.UserEntityFactory
 import javax.inject.Inject
 
 internal class RoomMemberEventHandler @Inject constructor(
@@ -58,10 +57,6 @@ internal class RoomMemberEventHandler @Inject constructor(
                 // but we want to preserve presence record value and not replace it with null
                 getExistingPresenceState(realm, roomId, userId))
         realm.insertOrUpdate(roomMemberEntity)
-        if (roomMember.membership.isActive()) {
-            val userEntity = UserEntityFactory.create(userId, roomMember)
-            realm.insertOrUpdate(userEntity)
-        }
 
         // check whether this new room member event may be used to update the directs dictionary in account data
         // this is required to handle correctly invite by email in DM


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
```diff
@@ -58,10 +57,6 @@ internal class RoomMemberEventHandler @Inject constructor(
                 // but we want to preserve presence record value and not replace it with null
                 getExistingPresenceState(realm, roomId, userId))
         realm.insertOrUpdate(roomMemberEntity)
-        if (roomMember.membership.isActive()) {
-            val userEntity = UserEntityFactory.create(userId, roomMember)
-            realm.insertOrUpdate(userEntity)
-        }
```
ericdecanini, line 64: @ganfra was there any purpose to this being here?

ganfra: Yes, this is the way to have local User created from RoomMember state event. If you remove this, we'll loose the avatar/display name of lots of users (which is the same than a RoomMember but outside the context of a room.)
If we remove this, we should take care of fetching profile of users everywhere, which is not that great...

ganfra (changes requested):
This will fix this case, but will break others, see my comment

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
ericdecanini: Ahh I see now

ganfra:
So the idea now, is to try fetching profile when having incremental sync using the SyncResponsePostTreatmentAggregator to track userIds to fetch.
In practice, we have 2 cases:
- InitialSync: just save the RoomMember content into the UserEntity, except for our own userId (as it already have a fetched profile)
- IncrementalSync: check if avatar/displayName of the RoomMember is different from the avatar/DisplayName of the UserEntity. If yes, add this userId to the SyncResponsePostTreatmentAggregator. Then at the end of the sync, you can fetch all the profiles from the aggregator.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
```diff
@@ -33,27 +33,31 @@ internal class RoomMemberEventHandler @Inject constructor(
...
-        saveRoomMemberEntityLocally(realm, roomId, userId, roomMember)
-        saveUserEntityLocallyIfNecessary(realm, userId, roomMember, prevContent)
+    private fun handleInitialSync(realm: Realm,
+                                  roomId: String,
+                                  currentUserId: String,
+                                  eventUserId: String,
+                                  roomMember: RoomMemberContent,
+                                  aggregator: SyncResponsePostTreatmentAggregator?): Boolean {
+        if (currentUserId != eventUserId) {
+            saveRoomMemberEntityLocally(realm, roomId, eventUserId, roomMember)
```
ganfra, line 57: Always save RoomMember even if its our userId, but don't save User

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -68,4 +76,27 @@ internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
...
+            it.insertOrUpdate(userEntity)
+        }
+    }
+
+    private suspend fun fetchUsers(usersToFetch: MutableList<String>) = usersToFetch.map {
+        val profileJson = profileService.getProfile(it)
+
+        MatrixItem.UserItem(
+                id = it,
+                displayName = profileJson[ProfileService.DISPLAY_NAME_KEY] as? String,
```
ganfra, line 98: Create a method to reuse these json keys as it's used in several places

ganfra (changes requested):
Some remarks

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
```diff
@@ -33,27 +33,31 @@ internal class RoomMemberEventHandler @Inject constructor(
         @UserId private val myUserId: String
 ) {
 
-    fun handle(realm: Realm,
-               roomId: String,
-               event: Event,
-               aggregator: SyncResponsePostTreatmentAggregator? = null): Boolean {
+    fun handleInitialSync(realm: Realm,
+                          roomId: String,
+                          currentUserId: String,
```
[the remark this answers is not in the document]
ericdecanini: image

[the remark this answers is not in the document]
ericdecanini: 👁️👄👁️

ganfra (approved):
Ok looks like it's ok now, thanks for this! (haven't tested)

ganfra:
Just fix the conflict and build failing

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/MatrixItem.kt
```diff
@@ -37,11 +38,20 @@ sealed class MatrixItem(
...
             MatrixItem(id, displayName?.removeSuffix(IRC_PATTERN), avatarUrl) {
+
         init {
             if (BuildConfig.DEBUG) checkId()
         }
 
         override fun updateAvatar(newAvatar: String?) = copy(avatarUrl = newAvatar)
+
+        companion object {
+            fun fromJson(userId: String, profileJson: JsonDict) = UserItem(
```
bmarty, line 49: I would prefer that this does not exist.

You can use instead `User.fromJson(...).toMatrixItem()`

But `MatrixItem` is more something that the SDK expose to the app. The SDK should not use `MatrixItem` internally. You should use `User` instead.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -68,4 +76,22 @@ internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
...
+    private fun saveUserLocally(monarchy: Monarchy, userItem: MatrixItem.UserItem) {
+        val userEntity = UserEntityFactory.create(userItem)
+        monarchy.doWithRealm {
+            it.insertOrUpdate(userEntity)
+        }
+    }
+
+    private suspend fun fetchUsers(usersToFetch: MutableList<String>) = usersToFetch.map {
+        val profileJson = profileService.getProfile(it)
+        MatrixItem.UserItem.fromJson(it, profileJson)
```
bmarty, line 95: Better to use `User.fromJson(it, profileJson)` here. (see my other remark)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/RoomSyncHandler.kt
```diff
@@ -206,7 +206,7 @@ internal class RoomSyncHandler @Inject constructor(private val readReceiptHandle
                                          insertType: EventInsertType,
                                          syncLocalTimestampMillis: Long,
                                          aggregator: SyncResponsePostTreatmentAggregator): RoomEntity {
-        Timber.v("Handle join sync for room $roomId")
```
bmarty, line 209: This comment is useful when debugging sync request. Can you restore it please?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/DefaultUserService.kt
```diff
@@ -36,18 +35,10 @@ internal class DefaultUserService @Inject constructor(private val userDataSource
...
-            return known
-        } else {
-            val params = GetProfileInfoTask.Params(userId)
-            val data = getProfileInfoTask.execute(params)
-            return User(
-                    userId,
-                    data[ProfileService.DISPLAY_NAME_KEY] as? String,
-                    data[ProfileService.AVATAR_URL_KEY] as? String)
-        }
+    override suspend fun resolveUser(userId: String) = getUser(userId) ?: run {
```
bmarty, line 38: When reading the code in GitHub, it's far better to keep explicit type `: User {`
Short `fun` form with `=` should be reserved for very small functions.

vector/src/main/java/im/vector/app/features/roommemberprofile/RoomMemberProfileViewModel.kt
```diff
@@ -327,14 +327,8 @@ class RoomMemberProfileViewModel @AssistedInject constructor(
...
-            session.getProfileAsUser(initialState.userId)
-                    .let {
-                        MatrixItem.UserItem(
-                                id = initialState.userId,
-                                displayName = it.displayName,
-                                avatarUrl = it.avatarUrl
-                        )
-                    }
+            session.getProfile(initialState.userId)
+                    .let { MatrixItem.UserItem.fromJson(initialState.userId, it) }
```
bmarty, line 331: So we will have:
```kotlin
            session.getProfile(initialState.userId)
                    .let { User.fromJson(initialState.userId, it) }
                    .toMatrixItem()
```

bmarty (changes requested):
I think this PR will fix the issue. Thanks!
Some remarks on the form.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/LoadRoomMembersTask.kt
```diff
@@ -123,7 +123,7 @@ internal class DefaultLoadRoomMembersTask @Inject constructor(
                     eventId = roomMemberEvent.eventId
                     root = eventEntity
                 }
-                roomMemberEventHandler.handle(realm, roomId, roomMemberEvent)
+                roomMemberEventHandler.handleIncrementalSync(realm, roomId, roomMemberEvent)
```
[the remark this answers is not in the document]
ericdecanini: I'm generally not a fan of passing boolean flags as the argument here is if you need to pass a boolean that changes the actions of the functions, you may as well make it another function which should be kept as small as possible (again, from Uncle Bob's Clean Code)

But tbh in this case I'm not too bothered either way. I'll make the changes!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -92,6 +94,6 @@ internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
 
     private suspend fun fetchUsers(usersToFetch: MutableList<String>) = usersToFetch.map {
         val profileJson = profileService.getProfile(it)
-        MatrixItem.UserItem.fromJson(it, profileJson)
+        User.fromJson(it, profileJson).toMatrixItem()
```
bmarty, line 97: As I said, we should not use `MatrixItem` internally.
Can you change to:
```kotlin
User.fromJson(it, profileJson)
```
update the code to make it compile and remove `UserEntityFactory.create(userItem: MatrixItem.UserItem): UserEntity`

Thanks!

bmarty (changes requested):
Thanks for the update. One last remark.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -68,4 +78,22 @@ internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
             updateUserAccountDataTask.execute(UpdateUserAccountDataTask.DirectChatParams(directMessages = directChats))
         }
     }
+
+    private suspend fun fetchAndUpdateUsers(usersToFetch: MutableList<String>) {
+        val userProfiles = fetchUsers(usersToFetch)
+        userProfiles.forEach { saveUserLocally(monarchy, it) }
+        usersToFetch.clear()
```
bmarty, line 85: This is probably not necessary to clear the list.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
ericdecanini: Ahh I see what you mean! Sorry I misunderstood the first time

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
ericdecanini: If we don't though, wouldn't we keep fetching the same users even when we don't need to fetch them anymore? I think it could increase bandwith and performance usage the longer the app is kept on

bmarty: Nope. A brand new `SyncResponsePostTreatmentAggregator` is created for each sync request. See https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt#L103

bmarty: `usersToFetch` parameter could probably be of type `List` (instead of `MutableList`)

ericdecanini: Ahhh this wasn't clear to me at first. Changes pushed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -68,4 +76,21 @@ internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
...
+        val userProfiles = fetchUsers(usersToFetch)
+        userProfiles.forEach { saveUserLocally(monarchy, it) }
+    }
+
+    private suspend fun fetchUsers(usersToFetch: List<String>) = usersToFetch.map {
+        val profileJson = profileService.getProfile(it)
+        User.fromJson(it, profileJson)
+    }
+
+    private fun saveUserLocally(monarchy: Monarchy, user: User) {
```
bmarty, line 90: monarchy is a class member, there is no need to add it as a method parameter.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -16,22 +16,30 @@
...
 import org.matrix.android.sdk.internal.session.user.accountdata.DirectChatsHelper
 import org.matrix.android.sdk.internal.session.user.accountdata.UpdateUserAccountDataTask
 import javax.inject.Inject
 
 internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
         private val directChatsHelper: DirectChatsHelper,
         private val ephemeralTemporaryStore: RoomSyncEphemeralTemporaryStore,
-        private val updateUserAccountDataTask: UpdateUserAccountDataTask
+        private val updateUserAccountDataTask: UpdateUserAccountDataTask,
+        private val profileService: ProfileService,
```
bmarty, line 36: You should not inject Service in SDK classes. Services are meant to be used by the client app.
Instead you can inject `getProfileInfoTask` in this case.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -68,4 +76,21 @@ internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
...
+        val profileJson = profileService.getProfile(it)
+        User.fromJson(it, profileJson)
+    }
+
+    private fun saveUserLocally(monarchy: Monarchy, user: User) {
+        val userEntity = UserEntityFactory.create(user)
+        monarchy.doWithRealm {
+            it.insertOrUpdate(userEntity)
+        }
+    }
```
bmarty, line 95: It's good to split code into several smaller methods, but it's maybe harder to read in this particular case.

What about replacing those 3 methods with a single one:

```kotlin
    private suspend fun fetchAndUpdateUsers(usersToFetch: List<String>) {
        usersToFetch
                .map { userId ->
                    val profileJson = profileService.getProfile(userId)
                    User.fromJson(userId, profileJson)
                }
                .map { user -> UserEntityFactory.create(user) }
                .let { userEntities ->
                    monarchy.doWithRealm {
                        it.insertOrUpdate(userEntities)
                    }
                }
    }
```

Also has the advantage to do on single Realm transaction instead of N transactions with your code.

WDYT?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
```diff
@@ -68,4 +76,21 @@ internal class SyncResponsePostTreatmentAggregatorHandler @Inject constructor(
...
         }
     }
+
+    private suspend fun fetchAndUpdateUsers(usersToFetch: List<String>) {
+        val userProfiles = fetchUsers(usersToFetch)
+        userProfiles.forEach { saveUserLocally(monarchy, it) }
+    }
+
+    private suspend fun fetchUsers(usersToFetch: List<String>) = usersToFetch.map {
+        val profileJson = profileService.getProfile(it)
```
bmarty, line 86: What will happen if the request fails for any reason, such as no network? I think it will crash the app. Maybe protect by using `tryOrNull`. We will just ignore updating user in that case, this is not a big deal I think.

bmarty (changes requested):
Sorry, I have further remarks for you.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
ericdecanini: istg i'm actually blind

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/SyncResponsePostTreatmentAggregatorHandler.kt
ericdecanini: A very good point

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
```diff
@@ -58,26 +84,62 @@ internal class RoomMemberEventHandler @Inject constructor(
...
-     * by NULL value.
-     */
+    private fun handleIncrementalSync(realm: Realm,
+                                      roomId: String,
+                                      eventUserId: String,
+                                      roomMember: RoomMemberContent,
+                                      prevContent: Map<String, Any>?,
+                                      aggregator: SyncResponsePostTreatmentAggregator?): Boolean {
+        val previousDisplayName = prevContent?.get("displayname")
+        val previousAvatar = prevContent?.get("avatar_url")
```
bmarty, line 128: Maybe add ` as? String` on those 2 lines, so `isDifferentFrom` can just be removed and replaced by `!=`, WDYT?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
```diff
@@ -58,26 +84,62 @@ internal class RoomMemberEventHandler @Inject constructor(
...
 
-    /**
-     * Get the already existing presence state for a specific user & room in order NOT to be replaced in RoomMemberSummaryEntity
-     * by NULL value.
-     */
+    private fun handleIncrementalSync(realm: Realm,
+                                      roomId: String,
+                                      eventUserId: String,
+                                      roomMember: RoomMemberContent,
+                                      prevContent: Map<String, Any>?,
```
bmarty, line 125: Could be of type `Content?`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
```diff
@@ -58,26 +84,62 @@ internal class RoomMemberEventHandler @Inject constructor(
...
+                                      aggregator: SyncResponsePostTreatmentAggregator?): Boolean {
+        val previousDisplayName = prevContent?.get("displayname")
+        val previousAvatar = prevContent?.get("avatar_url")
 
-    private fun getExistingPresenceState(realm: Realm, roomId: String, userId: String): UserPresenceEntity? {
-        return RoomMemberSummaryEntity.where(realm, roomId, userId).findFirst()?.userPresenceEntity
+        if (previousDisplayName.isDifferentFrom(roomMember.displayName) ||
+                previousAvatar.isDifferentFrom(roomMember.avatarUrl)) {
+            aggregator?.userIdsToFetch?.add(eventUserId)
```
bmarty, line 132: If aggregator is null, there is no need to do all this treatment, maybe replace by:
```kotlin
        if (aggregator != null) {
            val previousDisplayName = prevContent?.get("displayname") as? String
            val previousAvatar = prevContent?.get("avatar_url") as? String

            if (previousDisplayName != roomMember.displayName ||
                    previousAvatar != roomMember.avatarUrl) {
                aggregator.userIdsToFetch.add(eventUserId)
            }
        }
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
```diff
@@ -58,26 +84,62 @@ internal class RoomMemberEventHandler @Inject constructor(
...
+
+        saveRoomMemberEntityLocally(realm, roomId, eventUserId, roomMember)
+        // At the end of the sync, fetch all the profiles from the aggregator
+        updateDirectChatsIfNecessary(roomId, roomMember, aggregator)
+        return true
+    }
+
+    private fun Any?.isDifferentFrom(value: Any?) = when {
+        this == null || this == value -> false
+        else                          -> true
```
bmarty, line 143: Can be removed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/RoomMemberEventHandler.kt
ericdecanini: I tested this and it does work!

bmarty (commented):
Latest remark I think. Can you squash all the commits and force push?

bmarty (approved):
Approved!
Could be nice to squash all the commits to avoid polluting the git history with all the little change / revert. A "squash and merge" can also be done.
Thanks!

PR #5629, opened 2022-03-24
Use gradle plugin for cache

michaelkaye:


michaelkaye:
So far I can't see evidence that this is actually using the cache (?) that is being written - no dependencies or even gradlew are being loaded from gha's cache.

ouchadam:
@michaelkaye the project currently doesn't opt in to the gradle build cache, we need to add `org.gradle.caching=true` to https://github.com/vector-im/element-android/blob/develop/gradle.properties or explicitly use the `--build-cache` command line flag

with that said, the `.gradle/*` should at least be coming through! the gradle cache plugin also has the benefit of caching some of the $GRADLE_HOME caches

I've been using the flag locally for a while without too much issue~ although there isn't a massive amount of benefit for incremental builds due to the monolithic nature of our modules, would be interesting to see if it helps on the CI for our case! :crossed_fingers:

ouchadam (approved):
:100: for swapping out the manual caching with the plugin

bmarty:
This PR has been forgotten, and now there are conflict to fix. @michaelkaye does it worth fixing the conflicts and merge it?

michaelkaye:
If you still want this; i might open a fresh branch rather than trying to merge in all the other changes; the actual diffs are fairly simple. to apply.

michaelkaye:
Closing in favour of https://github.com/vector-im/element-android/pull/7944

Issue #5596, opened 2022-03-21
Update UISI Auto-Reporting to match Web/iOS implementation


langleyd:
Bring the UISI auto-reporting implementation in line with web and iOS( recently added in https://github.com/vector-im/element-ios/pull/5743)

More Specifically:
- Implementation of android differed slightly with web  in just monitoring decryption attempts after initial sync.
- Current implementation may be sending false positives as is checking isEncrypted but not clear event/error [here](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/UISIDetector.kt#L79)
- Web also allows 60 a minimum of seconds between reports. This is what was also set on iOS and seems like another reasonable measure to avoid user impact and sending to many reports. At the moment on android I think it is 2 seconds.

PR #5639, opened 2022-03-25
Align with web implementation

langleyd:
Fixes https://github.com/vector-im/element-android/issues/5596

### DIFF
--- a/vector/src/main/java/im/vector/app/AutoRageShaker.kt
+++ b/vector/src/main/java/im/vector/app/AutoRageShaker.kt
@@ -17,23 +17,31 @@
 package im.vector.app
 
 import android.content.SharedPreferences
+import androidx.lifecycle.asFlow
 import im.vector.app.core.di.ActiveSessionHolder
 import im.vector.app.features.rageshake.BugReporter
 import im.vector.app.features.rageshake.ReportType
+import im.vector.app.features.session.coroutineScope
 import im.vector.app.features.settings.VectorPreferences
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.delay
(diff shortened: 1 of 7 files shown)


bmarty (commented):
I'll defer to @BillCarsonFr review for the content of the change.
Just one remark about the changelog (a PR can have multiple changelog files)

vector/src/main/java/im/vector/app/AutoRageShaker.kt
```diff
@@ -261,7 +259,14 @@ class AutoRageShaker @Inject constructor(
             return
         }
         this.currentActiveSessionId = sessionId
-        this.detector.enabled = _enabled
+
+        hasSynced = session.hasAlreadySynced()
+        session.getSyncStatusLive()
+                .asFlow()
+                .onEach {
+                    hasSynced = it !is SyncStatusService.Status.Progressing
```
BillCarsonFr, line 267: This is not exaclty what we want no? Looks a bit strange. Also we are subscribing to the flow for the all session lifecycle no?

langleyd: "For the lifetime of the session, confirm we are not in an initial sync", Being outside an initial sync I thought would be reasonably equivalent to after an initial sync.

Web uses the sync token, so equivalent to `hasAlreadySynced`. We could keep a reference to the active sessions and call `hasAlreadySynced` synchronously in `enabled`?

BillCarsonFr: @ganfra or @bmarty what would be the best way to check that an initial sync has already been done?

bmarty: I think this is acceptable.
Maybe we should rename `SyncStatusService.Status.Progressing` to `SyncStatusService.Status.InitialSyncProgressing` for clarity.

It's historical, previously we did not track the incremental sync status.

bmarty (approved):
LGTM

PR #5644, opened 2022-03-28
Code Style Change

ericdecanini:
Draft until entire project is reformatted. See bottom of description for more detail

Open for discussion!

[Proposal Doc](https://docs.google.com/document/u/2/d/1xLtbinj-2VKgAhz8EFF1GiiZeGYYcMC5uOquSstaVxU/)

**Summary:**

We want to make changes to the code style for better readability and to stick closer to Kotlin code conventions.

The changes include:
- Function and class parameters in new lines
- Disable `when` branches in column
- Disable all continuation indentation
- Enable trailing commas
- Enable place ‘)’ on new line for if statements
- Disable keeping comments at first column

These changes have been decided on through chats and polls.

The changes were made on a `.editorconfig` file which automatically generates a `project.xml` file for projects that include it, which then automatically applies the code style to the IDE. This will then affect the style applied when auto formatting code.

The aforementioned `.editorconfig` file was generated through selecting all the necessary options in Android Studio's kotlin code style and exporting it as such. The exported file did also include style config for other languages (Java, C++, etc.) but I removed those from the config as we don't have a concensus on that yet.

<strike>I'm currently testing if this works too with the linter, both locally as well as through the CI</strike>

Edit: This affects the CI linter so the entire project needs to be reformatted. Turning this PR into a draft until that's done


ericdecanini:
image

Poll: "Do we want to start using trailing commas?"
Majority vote: Yes, let's start using them (5/8 votes)

ericdecanini:
image

Poll: Do we want to placing ')' on new line for if statments?
Majority vote: Enable it (3/7)

.editorconfig
```diff
@@ -1,14 +1,106 @@
-# For ktlint configuration. Ref: https://ktlint.github.io/
+[*]
+charset = utf-8
+end_of_line = crlf
+indent_size = 4
+indent_style = space
+insert_final_newline = false
```
bmarty, line 6: it was true, I think it should stay true

library/attachment-viewer/src/main/java/im/vector/lib/attachmentviewer/AnimatedImageViewHolder.kt
```diff
@@ -20,7 +20,7 @@ import android.view.View
 import im.vector.lib.attachmentviewer.databinding.ItemAnimatedImageAttachmentBinding
 
 class AnimatedImageViewHolder constructor(itemView: View) :
-        BaseViewHolder(itemView) {
+    BaseViewHolder(itemView) {
```
bmarty, line 23: is it the "continuation indent" thing?
I do not like the new format :/

```kotlin
class AnimatedImageViewHolder constructor(itemView: View) :
    BaseViewHolder(itemView) {
    val views = ItemAnimatedImageAttachmentBinding.bind(itemView)
```

is more confusing than

```kotlin
class AnimatedImageViewHolder constructor(itemView: View) :
        BaseViewHolder(itemView) {
    val views = ItemAnimatedImageAttachmentBinding.bind(itemView)
```
if I am reading it correctly.

library/attachment-viewer/src/main/java/im/vector/lib/attachmentviewer/AttachmentsAdapter.kt
```diff
@@ -42,21 +42,21 @@ class AttachmentsAdapter : RecyclerView.Adapter<BaseViewHolder>() {
         val inflater = LayoutInflater.from(parent.context)
         val itemView = inflater.inflate(viewType, parent, false)
         return when (viewType) {
-            R.layout.item_image_attachment          -> ZoomableImageViewHolder(itemView)
+            R.layout.item_image_attachment -> ZoomableImageViewHolder(itemView)
             R.layout.item_animated_image_attachment -> AnimatedImageViewHolder(itemView)
-            R.layout.item_video_attachment          -> VideoViewHolder(itemView)
-            else                                    -> UnsupportedViewHolder(itemView)
+            R.layout.item_video_attachment -> VideoViewHolder(itemView)
+            else -> UnsupportedViewHolder(itemView)
```
bmarty, line 48: You will not convince me that the new format is easier to read. (but I know the advantages).
Applicable to nearly all the `when` statements.
Probably because to me code readability is top priority.

library/attachment-viewer/src/main/res/layout/activity_attachment_viewer.xml
```diff
@@ -1,45 +1,45 @@
 <?xml version="1.0" encoding="utf-8"?>
 <FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    android:id="@+id/rootContainer"
-    android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    tools:context=".AttachmentViewerActivity">
+        xmlns:tools="http://schemas.android.com/tools"
```
bmarty, line 3: I do not understand why we now have 8 spaces for an indent.

bmarty (commented):
Some thoughts. Also it will be a nightmare to handle so many (formatting) diff for the forks with lots of changes :/. Does it worth it?

image

.editorconfig
ericdecanini: Good catch!

library/attachment-viewer/src/main/java/im/vector/lib/attachmentviewer/AttachmentsAdapter.kt
ericdecanini: It is mine too, but personally I find this easier to read than if they're all aligned. It also may well be the case that this is easier or harder to read based on what you're used to and falls into the realm of subjective, hence, the earlier poll

library/attachment-viewer/src/main/res/layout/activity_attachment_viewer.xml
ericdecanini: A mistake. Good spot!

ericdecanini:
> Some thoughts. Also it will be a nightmare to handle so many (formatting) diff for the forks with lots of changes :/. Does it worth it?
> 
> image

I agree that we have to give this some proper thought before pushing through with it. For the branches of the main fork, us all autoformatting on our own branches should work but this could be a lot more involved when it comes to the forks. We might have to look at alternate approaches to this, but I do very much think it's worth it as it boosts the code's readability up a lot and opens avenues for even more improvements

bmarty:
Have you considered re-enabling disabled rules? See https://github.com/vector-im/element-android/blob/main/build.gradle#L107

ericdecanini:
> Have you considered re-enabling disabled rules? See https://github.com/vector-im/element-android/blob/main/build.gradle#L107

Thanks for pointing this out! We'd most probably want to

ericdecanini:
Closing as we decided on the approach of tackling these changes one at a time.

PR #5645, opened 2022-03-28
Fix regression from #4471

bmarty:
`is Loading` has been replaced by `Uninitialized, is Loading` in `when` statements, which is not strictly equivalent

This commit revert those changes.

Related: #4471

This PR fixes regressions on:
- create DM screen
- search message screen

Other places were not regression (`is Incomplete` may have been used previously instead of `is Loading`), but I prefer PR #4471 be strictly equivalent to the previous code.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomActivity.kt
@@ -28,7 +28,6 @@ import com.airbnb.mvrx.Async
 import com.airbnb.mvrx.Fail
 import com.airbnb.mvrx.Loading
 import com.airbnb.mvrx.Success
-import com.airbnb.mvrx.Uninitialized
 import com.airbnb.mvrx.viewModel
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import dagger.hilt.android.AndroidEntryPoint
@@ -167,10 +166,10 @@ class CreateDirectRoomActivity : SimpleFragmentActivity() {
 
     private fun renderCreateAndInviteState(state: Async<String>) {
         when (state) {
-            Uninitialized,
             is Loading -> renderCreationLoading()
             is Success -> renderCreationSuccess(state())
             is Fail    -> renderCreationFailure(state.error)
+            else       -> Unit
         }
     }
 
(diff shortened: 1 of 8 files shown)


vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomActivity.kt
```diff
@@ -167,10 +166,10 @@ class CreateDirectRoomActivity : SimpleFragmentActivity() {
 
     private fun renderCreateAndInviteState(state: Async<String>) {
         when (state) {
-            Uninitialized,
```
ouchadam, line 170: not a strong preference, I would be for keeping the case rather than relying on the else (has the benefit of catching areas that would change if another state was introduced)

```kotlin
when(state) {
  Uninitialized -> Unit
}
```

ouchadam (approved):
small comment about keeping the case vs using else, will leave up to you to decide! 

LGTM :+1:

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomActivity.kt
bmarty: I agree, but using `else` has the advantage to not change the format of the other cases (arrow alignment would have changed :/).

Also there is a few chance that another case is introduced to `Async`, this is pretty stable.

vector/src/main/java/im/vector/app/features/discovery/DiscoverySettingsController.kt
```diff
@@ -71,6 +70,7 @@ class DiscoverySettingsController @Inject constructor(
                     buildMsisdnSection(data.phoneNumbersList)
                 }
             }
+            else       -> Unit
```
ericdecanini, line 73: It would be better to explicitly define `Uninitialized -> Unit` than use `else -> Unit`. Reason being if a new subclass of `Async`, the compiler would catch that everywhere we use it.

In this case it's unlikely that airbnb will add one, but in general it's good practice

ericdecanini (approved):

vector/src/main/java/im/vector/app/features/createdirect/CreateDirectRoomActivity.kt
ericdecanini: Case and point against arrow alignment 👀

Issue #3119, opened 2021-04-05
Sending an image from the gallery crashes Element


dkasak:
#### Describe the bug

Sending an image from a gallery into a room crashes Element.

#### To Reproduce
1. Open a room
2. Click `+` in lower left.
3. Select "Gallery".
4. Select an image.
5. A crash happens.

#### Expected behavior
Element should send the image to the room successfully.

#### details.log.gz

This is rageshake 2021-04-05/090607.

```


--------------------------------- crash call stack ---------------------------------
Element Build : 40101032
Element Version : 1.1.3 [40101032] (G-b1729)
SDK Version : 0.0.1 (a6c43f46)
Phone : MI 3W (V11.0.8.0.QFQEUXM 10 REL)
Memory statuses 
usedSize   38 MB
freeSize   3 MB
totalSize   41 MB
Thread: main, Exception: java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:502)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:935)
Caused by: java.lang.reflect.InvocationTargetException
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
	... 1 more
Caused by: java.io.FileNotFoundException: open failed: ENOENT (No such file or directory)
	at android.database.DatabaseUtils.readExceptionWithFileNotFoundExceptionFromParcel(DatabaseUtils.java:149)
	at android.content.ContentProviderProxy.openTypedAssetFile(ContentProviderNative.java:705)
	at android.content.ContentResolver.openTypedAssetFileDescriptor(ContentResolver.java:1689)
	at android.content.ContentResolver.openAssetFileDescriptor(ContentResolver.java:1505)
	at android.content.ContentResolver.openInputStream(ContentResolver.java:1189)
	at im.vector.lib.multipicker.utils.ImageUtils.getOrientation(ImageUtils.kt:1)
	at im.vector.lib.multipicker.ImagePicker.getSelectedFiles(ImagePicker.kt:12)
	at im.vector.app.features.attachments.AttachmentsHelper.onImageResult(AttachmentsHelper.kt:3)
[...]

PR #5652, opened 2022-03-28
Potentially fixes intermittent crashing from sending gallery images

ericdecanini:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Edit: This PR's title and description have been updated as an entirely new solution has been implemented

The suspected cause of the error is that depending on the way the device's gallery caches images to be sent to the Element app, the content resolver may need to access IO operations (network, etc)

Currently, we're accessing the content resolver on the main thread

The way the content resolver works, any exception thrown during this operation would be rethrown as a `FileNotFoundException`, the exception we see in the rageshakes

On this PR, I forced the accessing of the io thread on the content resolver operations that send media from gallery. This does of course affect many other files with coroutine scope and suspend chains.

## Motivation and context

Potentially fixes https://github.com/vector-im/element-android/issues/3119

## Screenshots / GIFs

N/A

## Tests

Test with devices pre and post Android 11
- Go to the timeline and send different media types (image, video, audio). See that they send normally and the app doesn't crash

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Android 10, 12 (emulator)

## Checklist

### DIFF
--- a/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
+++ b/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
@@ -22,7 +22,7 @@ import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
 
-fun createTemporaryMediaFile(context: Context, mediaType: MediaType): File {
+internal fun createTemporaryMediaFile(context: Context, mediaType: MediaType): File {
     val timeStamp: String = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.getDefault()).format(Date())
     val storageDir: File = context.filesDir.also { it.mkdirs() }
     val fileSuffix = when (mediaType) {
--- a/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
+++ b/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2022 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.lib.multipicker.utils
+
+import android.content.Context
+import java.io.File
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+
+object MediaFileUtils {
+
+    fun createTemporaryMediaFile(context: Context, mediaType: MediaType): File {
+        val timeStamp: String = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.getDefault()).format(Date())
+        val storageDir: File = context.filesDir.also { it.mkdirs() }
+        val fileSuffix = when (mediaType) {
+            MediaType.IMAGE -> ".jpg"
+            MediaType.VIDEO -> ".mp4"
+        }
+
+        return File.createTempFile(
+                "${timeStamp}_",
+                fileSuffix,
+                storageDir
+        )
+    }
+
+    enum class MediaType {
+        IMAGE, VIDEO
+    }
+}
--- a/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
+++ b/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
@@ -37,6 +37,6 @@ internal fun createTemporaryMediaFile(context: Context, mediaType: MediaType): F
     )
 }
 
-enum class MediaType {
+internal enum class MediaType {
     IMAGE, VIDEO
 }
--- a/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
+++ b/library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
@@ -22,24 +22,21 @@ import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
 
-object MediaFileUtils {
-
-    fun createTemporaryMediaFile(context: Context, mediaType: MediaType): File {
-        val timeStamp: String = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.getDefault()).format(Date())
-        val storageDir: File = context.filesDir.also { it.mkdirs() }
-        val fileSuffix = when (mediaType) {
-            MediaType.IMAGE -> ".jpg"
-            MediaType.VIDEO -> ".mp4"
-        }
-
-        return File.createTempFile(
-                "${timeStamp}_",
-                fileSuffix,
-                storageDir
-        )
+fun createTemporaryMediaFile(context: Context, mediaType: MediaType): File {
+    val timeStamp: String = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.getDefault()).format(Date())
+    val storageDir: File = context.filesDir.also { it.mkdirs() }
+    val fileSuffix = when (mediaType) {
+        MediaType.IMAGE -> ".jpg"
+        MediaType.VIDEO -> ".mp4"
     }
 
-    enum class MediaType {
-        IMAGE, VIDEO
-    }
+    return File.createTempFile(
+            "${timeStamp}_",
+            fileSuffix,
+            storageDir
+    )
+}
+
+enum class MediaType {
+    IMAGE, VIDEO
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
@@ -28,10 +28,8 @@ internal class TemporaryFileCreator @Inject constructor(
 ) {
     suspend fun create(): File {
         return withContext(Dispatchers.IO) {
-            runCatching {
-                File.createTempFile(UUID.randomUUID().toString(), null, context.cacheDir)
-                        .apply { mkdirs() }
-            }.getOrThrow()
+            File.createTempFile(UUID.randomUUID().toString(), null, context.cacheDir)
+                    .apply { mkdirs() }
         }
     }
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
@@ -28,7 +28,10 @@ internal class TemporaryFileCreator @Inject constructor(
 ) {
     suspend fun create(): File {
         return withContext(Dispatchers.IO) {
-            File.createTempFile(UUID.randomUUID().toString(), null, context.cacheDir)
+            runCatching {
+                File.createTempFile(UUID.randomUUID().toString(), null, context.cacheDir)
+                        .apply { mkdirs() }
+            }.getOrThrow()
         }
     }
 }
--- a/changelog.d/5652.bugfix
+++ b/changelog.d/5652.bugfix
@@ -1,0 +1,1 @@
+Tentative fix of images crashing when being sent or shared from gallery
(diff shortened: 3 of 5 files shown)


ericdecanini:
I'm deliberately not including a changelog for this PR . While we can't reproduce the issue, we also can't confirm if this fixes the issue and so a changelog file could be misleading.

@bmarty are you ok with this?

library/multipicker/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="im.vector.lib.multipicker">
 
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
```
bmarty, line 4: This permission is already in the main manifest (https://github.com/vector-im/element-android/blob/main/vector/src/main/AndroidManifest.xml#L17) so I do not think it will change anything to add it in this module. The merged manifest will be the same.

For the changelog you could add a line like `Tentative fix of ...` or something similar.

ariskotsomitopoulos (commented):
As Benoit mentioned, we can see that the merged manifest contains `android.permission.READ_EXTERNAL_STORAGE`. Each .APK can can have **only** one manifest and it's the `merged manifest`. So your change here will not affect the final APK

vector/src/main/java/im/vector/app/features/home/room/detail/composer/VoiceMessageHelper.kt
```diff
@@ -76,7 +76,7 @@ class VoiceMessageHelper @Inject constructor(
         startRecordingAmplitudes()
     }
 
-    fun stopRecording(convertForSending: Boolean): MultiPickerAudioType? {
+    suspend fun stopRecordingAndGetAudio(convertForSending: Boolean): MultiPickerAudioType? {
```
ericdecanini, line 79: I renamed these functions as the name `stopRecording` doesn't imply a return value. It also becomes more confusing with its original name as to why it needs to be a suspend functions, but I think this makes it clearer

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/ContentResolverUtil.kt
```diff
@@ -113,7 +116,14 @@ internal fun Uri.toMultiPickerVideoType(context: Context): MultiPickerVideoType?
     }
 }
 
-fun Uri.toMultiPickerAudioType(context: Context): MultiPickerAudioType? {
+private suspend fun ContentResolver.openFileDescriptorOnIo(uri: Uri, mode: String) =
+        withContext(Dispatchers.IO) {
```
ariskotsomitopoulos, line 120: Not a blocker, It would be better especially for tests, if the value is not hardcoded as  `Dispatchers.IO`, it could be injected or passed from the constructor.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -817,25 +817,26 @@ class MessageComposerViewModel @AssistedInject constructor(
...
-            audioMessageHelper.stopRecording(convertForSending = true)?.let { audioType ->
-                if (audioType.duration > 1000) {
-                    room.sendMedia(
-                            attachment = audioType.toContentAttachmentData(isVoiceMessage = true),
-                            compressBeforeSending = false,
-                            roomIds = emptySet(),
-                            rootThreadEventId = rootThreadEventId)
-                } else {
+    private fun handleEndRecordingVoiceMessage(isCancelled: Boolean, rootThreadEventId: String? = null) =
+            viewModelScope.launch {
```
ariskotsomitopoulos, line 821: Keep in mind that` viewModelScope.launch` default thread is the Main. If we want to run it on IO for example we should do something like this` viewModelScope.launch(Dispatchers.IO)`

ariskotsomitopoulos (approved):
Thanks for the PR, I have some comments. Also, adding `mkdirs` maybe will eliminate the crash, check [here](https://stackoverflow.com/questions/36088699/error-open-failed-enoent-no-such-file-or-directory). With a quick look I saw we skip it in some cases maybe thats the problem.

bmarty (commented):
Sorry, I forget to post my remark, they were pending :/

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/ImageUtils.kt
```diff
@@ -16,23 +16,28 @@
...
 object ImageUtils {
 
-    fun getBitmap(context: Context, uri: Uri): Bitmap? {
+    suspend fun getBitmap(context: Context, uri: Uri): Bitmap? {
         return try {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-                ImageDecoder.decodeBitmap(ImageDecoder.createSource(context.contentResolver, uri))
+                runCatching {
+                    ImageDecoder.decodeBitmap(ImageDecoder.createSource(context.contentResolver, uri))
+                }.getOrThrow()
```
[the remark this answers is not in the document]
ericdecanini: Without this, we get the warning `Inappropriate blocking method call` which is caused because the method in question blocks the current thread and doesn't allow the coroutine to suspend and has the potential to throw. Adding this block makes our intent more explicit thus removing the warning

ericdecanini:
@ariskotsomitopoulos @bmarty I reapproached this following Aris' suggestion. I tried doing this without losing the entire commit history up till now but ultimately failed, but I think it's a much better solution

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
```diff
@@ -28,7 +28,10 @@ internal class TemporaryFileCreator @Inject constructor(
 ) {
     suspend fun create(): File {
         return withContext(Dispatchers.IO) {
-            File.createTempFile(UUID.randomUUID().toString(), null, context.cacheDir)
+            runCatching {
+                File.createTempFile(UUID.randomUUID().toString(), null, context.cacheDir)
+                        .apply { mkdirs() }
```
ariskotsomitopoulos, line 33: I am not sure that `mkdirs()` here will have any effect while we store the files in the `cache directory`. (should always exists). 

Also its interesting that using cache directory (from the doc):

> Apps are strongly encouraged to keep their usage of cache space below the quota returned by StorageManager.getCacheQuotaBytes(java.util.UUID). If your app goes above this quota, your cached files will be some of the first to be deleted when additional disk space is needed. Conversely, if your app stays under this quota, your cached files will be some of the last to be deleted when additional disk space is needed.

So whatever we store there, there are cases that system will delete our files! Just an idea that this may cause some errors!

ariskotsomitopoulos (approved):

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/ImageUtils.kt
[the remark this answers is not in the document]
bmarty: I do not have the warning on my side, strange. Thanks for the update.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
```diff
@@ -28,7 +28,10 @@ internal class TemporaryFileCreator @Inject constructor(
 ) {
     suspend fun create(): File {
         return withContext(Dispatchers.IO) {
-            File.createTempFile(UUID.randomUUID().toString(), null, context.cacheDir)
+            runCatching {
+                File.createTempFile(UUID.randomUUID().toString(), null, context.cacheDir)
```
bmarty, line 32: I understand that this is now the only change brought by this PR.
What about the 2 other occurences of `File.createTempFile` in our codebase? Maybe add the same protection, or create an extensions?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
ericdecanini: It sounds weird to me too, but could it be possible that in some devices even the cache can run into the problem described here? Perhaps it's something we need to pay attention after we release this fix. The only other possible cause I can think of is the image is being acquired from the main thread, but to fix that is a long coroutine that covers lots of areas that really would be better without it.

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+
+package im.vector.lib.multipicker.utils
+
+import android.content.Context
+import java.io.File
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+
+object MediaFileUtils {
```
bmarty, line 25: I am not sure, but I am always wondering about the RAM long life footprint of `object`. It's maybe negligible, hence, maybe it could be a class, taking the context as constructor `val` parameter, WDYT?

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+            MediaType.IMAGE -> ".jpg"
+            MediaType.VIDEO -> ".mp4"
+        }
+
+        return File.createTempFile(
+                "${timeStamp}_",
+                fileSuffix,
+                storageDir
+        )
+    }
```
bmarty, line 40: Thanks for the code duplication hunting!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
```diff
@@ -28,7 +28,10 @@ internal class TemporaryFileCreator @Inject constructor(
...
+                        .apply { mkdirs() }
+            }.getOrThrow()
```
bmarty, line 34: runCatching / getOrThrow thing :)

bmarty (commented):
Thanks for the update and the code deduplication!

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
bmarty: or just convert to a simple `fun`.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
ericdecanini: In this case I'm not pro for making it `getOrNull`. Trying it out, it leads to a long chain of nullable changes which could lead to some weird bugs that are harder for us to identify than a crash.

I doubt that we should be crashing here though. given that it's effectively the same as `File.createTempFile` on its own. Just without the warning shown below
image

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
ariskotsomitopoulos: Yes, let's track it after the release with the changes

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/TemporaryFileCreator.kt
bmarty: Using `runCatching / getOrThrow` just to get ride of the warning is quite strange to me. We do other people think?

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+
+package im.vector.lib.multipicker.utils
+
+import android.content.Context
+import java.io.File
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
+
+fun createTemporaryMediaFile(context: Context, mediaType: MediaType): File {
```
bmarty, line 25: maybe declare it `internal`?

library/multipicker/src/main/java/im/vector/lib/multipicker/utils/MediaFileUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    return File.createTempFile(
+            "${timeStamp}_",
+            fileSuffix,
+            storageDir
+    )
+}
+
+enum class MediaType {
```
bmarty, line 40: not a big deal but could be internal too I think.

bmarty (changes requested):
Thanks for the update! Let's cross fingers it will fix the issue!

Can you add a changelog as per https://github.com/vector-im/element-android/pull/5652#discussion_r837207551 please?

bmarty (approved):
thanks!

PR #5660, opened 2022-03-29
Show a banner in timeline while location sharing service is running

onurays:
We have already implemented the banner previously. This PR handles the visibility of the banner by observing the location service lifecycle.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
@@ -482,6 +482,8 @@ class TimelineFragment @Inject constructor(
                 RoomDetailViewEvents.StopChatEffects                     -> handleStopChatEffects()
                 is RoomDetailViewEvents.DisplayAndAcceptCall             -> acceptIncomingCall(it)
                 RoomDetailViewEvents.RoomReplacementStarted              -> handleRoomReplacement()
+                RoomDetailViewEvents.ShowLocationSharingIndicator        -> handleShowLocationSharingIndicator()
+                RoomDetailViewEvents.HideLocationSharingIndicator        -> handleHideLocationSharingIndicator()
             }
         }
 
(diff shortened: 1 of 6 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -482,6 +482,8 @@ class TimelineFragment @Inject constructor(
                 RoomDetailViewEvents.StopChatEffects                     -> handleStopChatEffects()
                 is RoomDetailViewEvents.DisplayAndAcceptCall             -> acceptIncomingCall(it)
                 RoomDetailViewEvents.RoomReplacementStarted              -> handleRoomReplacement()
+                RoomDetailViewEvents.ShowLocationSharingIndicator        -> handleShowLocationSharingIndicator()
```
ariskotsomitopoulos, line 485: To simplify the code and not to add more lines in TImelineFragment while its already huge, I would have one Event `ShowLocationSharingIndicator` instead of two: 
```

handleShowLocationSharingIndicator(show: Boolean){
    views.locationLiveStatusIndicator.isVisible = show
}
```

ariskotsomitopoulos (commented):
Thanks for the PR, I added some comments. Keep in mind `TimelineFragment` is also used for threads, so you should handle that appropriately. You might want to disable that on threads? or an other business request.

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ariskotsomitopoulos: Nice!

ariskotsomitopoulos (approved):
Thanks for the change, please check my comment about threads, while they use the same timeline and maybe will interfere with the location service.

onurays:
> Thanks for the change, please check my comment about threads, while they use the same timeline and maybe will interfere with the location service.

Thank you, we will create another PR to support threads.

PR #5661, opened 2022-03-29
#5581: live location sharing feature flag

### LINKED ISSUE #5581: [Location Sharing] - Add feature flag for live location sharing
> for the testing it could be helpful to the [`VectorOverrides`](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/DefaultVectorOverrides.kt#L22) to enable the feature for debug builds
> 
> eventually I imagine the full feature could be controlled by the [`VectorFeatures`](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/VectorFeatures.kt) if we want to develop behind a feature flag

In fact, we have a created a `buildConfigField` for that purpose: `ENABLE_LIVE_LOCATION_SHARING`. But using `VectorOverrides` is a better approach. I will create an issue.

_Originally posted by @mnaturel in https://github.com/vector-im/element-android/issues/5579#issuecomment-1072508976_

mnaturel:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Adding a way to enable/disable the live location feature from private settings.

## Motivation and context

Closes #5581 

## Tests

- Go to private debug options
- Click on features
- Enable the live location feature
- Go to a room
- Start sharing your location
- Check the live location sharing is visible when the pin is on the current user location
- Go to private debug options
- Click on features
- Disable the live location feature
- Go to a room
- Start sharing your location
- Check the live location sharing is not visible when the pin is on the current user location

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 12

## Checklist

### DIFF
--- a/vector/src/debug/java/im/vector/app/features/debug/settings/DebugPrivateSettingsViewModel.kt
+++ b/vector/src/debug/java/im/vector/app/features/debug/settings/DebugPrivateSettingsViewModel.kt
@@ -62,6 +62,9 @@ class DebugPrivateSettingsViewModel @AssistedInject constructor(
                     avatar = homeserverCapabilityOverrides.avatar.copy(activeOption = activeAvatarOption),
             ))
         }
+        debugVectorOverrides.forceEnableLiveLocationSharing.setOnEach {
+            copy(forceEnableLiveLocationSharing = it)
+        }
     }
 
     override fun handle(action: DebugPrivateSettingsViewActions) {
@@ -70,6 +73,7 @@ class DebugPrivateSettingsViewModel @AssistedInject constructor(
             is DebugPrivateSettingsViewActions.SetForceLoginFallbackEnabled -> handleSetForceLoginFallbackEnabled(action)
             is SetDisplayNameCapabilityOverride                             -> handSetDisplayNameCapabilityOverride(action)
             is SetAvatarCapabilityOverride                                  -> handSetAvatarCapabilityOverride(action)
+            is DebugPrivateSettingsViewActions.SetEnableLiveLocationSharing -> handleSetEnableLiveLocationSharingOverride(action)
         }
     }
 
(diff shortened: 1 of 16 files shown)


vector/src/main/java/im/vector/app/features/location/LocationSharingViewState.kt
```diff
@@ -39,7 +41,8 @@ data class LocationSharingViewState(
 
     constructor(locationSharingArgs: LocationSharingArgs) : this(
             roomId = locationSharingArgs.roomId,
-            mode = locationSharingArgs.mode
+            mode = locationSharingArgs.mode,
+            isLiveLocationSharingEnabled = BuildConfig.ENABLE_LIVE_LOCATION_SHARING
```
Johennes, line 45: Maybe a stupid question but should that be reading from `DebugVectorFeatures` instead?

mnaturel: This is a good question but here, unfortunately we cannot use dependency injection to create the corresponding `VectorFeatures` implementation. Another solution would be to declare the boolean as nullable and initialize it to `null`. When `null`, we would consider the feature as disabled. Do you think it is a better approach?

Johennes: Oh, if that's not possible technically, that's fine. I only meant to verify whether or not usage of the build config flag was intentional here.

ouchadam: we could mark this as `nullable` rather than mix the build config and feature flags, what do you think?

thinking out loud, it would be great if the assisted viewmodel initial state could use the dagger graph something like 

```kotlin
@Provides fun provideInitial(features: VectorFeatures) = LocationSharingViewState(
  features.isLiveLocationEnabled()
)
```

not sure if this is possible with our current setup~

vector/src/main/java/im/vector/app/features/location/LocationSharingFragment.kt
```diff
@@ -243,7 +242,7 @@ class LocationSharingFragment @Inject constructor(
         // first, update the options view
         val options: Set<LocationSharingOption> = when (state.areTargetAndUserLocationEqual) {
             true  -> {
-                if (BuildConfig.ENABLE_LIVE_LOCATION_SHARING) {
+                if (state.isLiveLocationSharingEnabled) {
```
ouchadam, line 245: I wouldn't be opposed to injecting the flags directly to the fragment if it helps makes checking them simpler or another option could be to store the `set` in the state instead (and generate it based on the flag)

mnaturel: I guess injecting the `VectorFeatures` into the `Fragment` would simplified things. I will go this way.

vector/src/main/java/im/vector/app/features/location/LocationSharingViewState.kt
mnaturel: I removed the parameter from the state to simplify.

ouchadam (approved):

vector/src/debug/java/im/vector/app/features/debug/settings/DebugPrivateSettingsViewModel.kt
```diff
@@ -85,14 +85,14 @@ class DebugPrivateSettingsViewModel @AssistedInject constructor(
         }
     }
 
-    private fun handSetDisplayNameCapabilityOverride(action: SetDisplayNameCapabilityOverride) {
+    private fun handleSetDisplayNameCapabilityOverride(action: SetDisplayNameCapabilityOverride) {
```
ouchadam, line 88: great catch!

PR #5666, opened 2022-03-30
Add loader to thread list

### LINKED ISSUE #5562: No loading spinner is shown while threads list is loading
### Steps to reproduce

1. Go to a room with lots of threads (e.g. Threads internal)
2. View the thread list on a fresh login

#### What did you expect?

To either see the thread list or a loading spinner telling me whats up

#### What happened instead?

The threads list looked empty for about 20 seconds, I assumed it was broken somehow

### Your phone model

Galaxy S20

### Operating system version

Android 11

1.4.6-dev [40104060] (G-b6874) feature/aris/thread_live_thread_list

### Homeserver

matrix.org

### Will you send logs?

No

ariskotsomitopoulos:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Add loader when user open/load the thread list

## Motivation and context

Especially on large rooms and on initial launch of the thread list, we should show an indicator to the user about the progress

## Screenshots / GIFs
|Before|After|
|---|---|
|Screenshot_20220330-121034_Element dbg|Screenshot_20220330-121026_Element dbg|

Closes #5562

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/threads/list/viewmodel/ThreadListViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/threads/list/viewmodel/ThreadListViewModel.kt
@@ -113,7 +113,15 @@ class ThreadListViewModel @AssistedInject constructor(@Assisted val initialState
 
     private fun fetchThreadList() {
         viewModelScope.launch {
+            isLoading(true)
             room?.fetchThreadSummaries()
+            isLoading(false)
+        }
+    }
+
+    private fun isLoading(show: Boolean) {
+        setState {
+            copy(isLoading = show)
         }
     }
 
--- a/vector/src/main/java/im/vector/app/features/home/room/threads/list/viewmodel/ThreadListViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/threads/list/viewmodel/ThreadListViewModel.kt
@@ -113,15 +113,15 @@ class ThreadListViewModel @AssistedInject constructor(@Assisted val initialState
 
     private fun fetchThreadList() {
         viewModelScope.launch {
-            isLoading(true)
+            setLoading(true)
             room?.fetchThreadSummaries()
-            isLoading(false)
+            setLoading(false)
         }
     }
 
-    private fun isLoading(show: Boolean) {
+    private fun setLoading(isLoading: Boolean) {
         setState {
-            copy(isLoading = show)
+            copy(isLoading = isLoading)
         }
     }
 
--- a/vector/src/main/java/im/vector/app/features/home/room/threads/list/views/ThreadListFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/threads/list/views/ThreadListFragment.kt
@@ -104,6 +104,11 @@ class ThreadListFragment @Inject constructor(
     override fun invalidate() = withState(threadListViewModel) { state ->
         renderEmptyStateIfNeeded(state)
         threadListController.update(state)
+        renderLoaderIfNeeded(state)
+    }
+
+    private fun renderLoaderIfNeeded(state: ThreadListViewState) {
+        views.threadListProgressBar.isVisible = state.isLoading
     }
 
     private fun renderToolbar() {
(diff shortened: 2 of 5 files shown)


vector/src/main/java/im/vector/app/features/home/room/threads/list/viewmodel/ThreadListViewModel.kt
```diff
@@ -113,7 +113,15 @@ class ThreadListViewModel @AssistedInject constructor(@Assisted val initialState
 
     private fun fetchThreadList() {
         viewModelScope.launch {
+            isLoading(true)
             room?.fetchThreadSummaries()
+            isLoading(false)
+        }
+    }
+
+    private fun isLoading(show: Boolean) {
```
onurays, line 122: I think `private fun setLoading(isLoading: Boolean)` is more convenient.

vector/src/main/res/layout/fragment_thread_list.xml
```diff
@@ -37,6 +37,21 @@
...
+
+        <ProgressBar
+            android:id="@+id/threadListProgressBar"
+            style="@style/Widget.Vector.ProgressBar.Horizontal"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center_vertical"
+            android:indeterminate="true"
+            android:visibility="gone"
+            android:layout_marginTop="-6dp"
```
onurays, line 49: I see why we need this but we have a lint rule for negative margins. Is it a problem to show the tiny progressbar on the first message? If it is we need `tools:ignore="NegativeMargin"` at least.

onurays: Or `app:layout_constraintBottom_toBottomOf="@id/threadListAppBarLayout"` solves the problem without negative margin?

onurays (changes requested):
Nice improvement! I have some minor remarks.

vector/src/main/res/layout/fragment_thread_list.xml
ariskotsomitopoulos: For that known problem I used to wrap it around a fragment, that eliminates the problem, but this adds an extra layer and decrease code readability.  So I prefer to use `-6dp` as an exception in my projects, never had an issue with that.

Regarding `app:layout_constraintBottom_toBottomOf="@id/threadListAppBarLayout"` will not have any difference while the padding is either on top or on the bottom.

P.S. On Jetpack Compose they fixed that issues the horizontal progressBar ! 🥇

vector/src/main/java/im/vector/app/features/home/room/threads/list/viewmodel/ThreadListViewModel.kt
ariskotsomitopoulos: Refactored!

onurays (approved):
Thank you for the update.

vector/src/main/java/im/vector/app/features/home/room/threads/list/views/ThreadListFragment.kt
```diff
@@ -104,6 +104,11 @@ class ThreadListFragment @Inject constructor(
     override fun invalidate() = withState(threadListViewModel) { state ->
         renderEmptyStateIfNeeded(state)
         threadListController.update(state)
+        renderLoaderIfNeeded(state)
+    }
+
+    private fun renderLoaderIfNeeded(state: ThreadListViewState) {
```
bmarty, line 110: `IfNeeded` suffix is not necessary here I guess.

vector/src/main/res/layout/fragment_thread_list.xml
bmarty: You could have reused what we are doing in the room list / room timeline: https://github.com/vector-im/element-android/blob/main/vector/src/main/res/layout/view_sync_state.xml#L38

bmarty (commented):
Some post merge remarks :)

vector/src/main/java/im/vector/app/features/home/room/threads/list/views/ThreadListFragment.kt
ariskotsomitopoulos: Correct, its there from a previous approach!

vector/src/main/res/layout/fragment_thread_list.xml
ariskotsomitopoulos: Yes I saw it but, nesting affects performance and makes code less readable, I cannot see any benefit from that. Do you have a strong point?

PR #5672, opened 2022-03-31
Move to using hookshot to send notifications, rather than using matrix client directly.

michaelkaye:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [X] Other notifications

## Content

With v1.0.0 of matrix-hookshot-action we can use hookshoot to send notifications via generic webhook, rather than embedding an access token with read/write access to the room. Otherwise there's no changes in the new hookshot-action version.

## Motivation and context

Neatness; avoiding risk of access token to the room leaking.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -330,10 +330,9 @@ jobs:
     if: always() && github.event_name != 'workflow_dispatch'
     # No concurrency required, runs every time on a schedule.
     steps:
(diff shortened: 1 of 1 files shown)


michaelkaye:
`ELEMENT_ANDROID_HOOKSHOT_URL` has been configured in secrets already

bmarty (approved):
LGTM, I let you merge it?

PR #5688, opened 2022-04-04
Improved thread rendering in the main timeline

### LINKED ISSUE #5151: Implement improved rendering for threads in the main timeline
**Changelog**

Root messages should form a standalone group with their respective thread summaries.

This way, we establish a clear relationship between the thread summary and the root message it belongs to. Failure to do it results in an ambiguous relationship that only gets worse when message bubbles are enabled.

**Source**

→ https://www.figma.com/file/T309ztx0sNyOOK6NKVLHsK/Threads?node-id=3085%3A305790

**Snapshot**

Threads – Figma

ariskotsomitopoulos:
Currently, when there are multiple consecutive messages by the same author we group them as a single block in the main timeline.

If any of those messages, except for the last one, gets converted to a thread then there's no clear relationship between thread summaries and root messages anymore. This affects all layout modes, but can be especially problematic when Message Bubbles are enabled.

Below you can see the expected result, each root message should have the icon displayed

|Before | After |
|---|---|
|Screenshot_20220404-115145_Element dbg|Screenshot_20220404-120227_Element dbg|

Closes #5151

### DIFF
--- a/changelog.d/5151.misc
+++ b/changelog.d/5151.misc
@@ -1,0 +1,1 @@
+Improve main timeline thread summary rendering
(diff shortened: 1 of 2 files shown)


ganfra (approved):
LGTM

PR #5692, opened 2022-04-04
Threads Beta opt-in mechanism

ariskotsomitopoulos:
### Threads Beta Opt-In

If threads are disabled in Labs:

Hide the thread button in the room header. Keep the “**Reply in thread**” action in the bottom sheet, which shows a BETA badge until the stable version is released. On tapping on the “Reply in thread” action a modal is shown to enable threads in Labs. Ideally, this would be a one-step action (refer to “Beta opt-in — Modal (direct)”).

|Screenshot_20220404-195113_Element dbg|Screenshot_20220405-133840_Element dbg|
|---|---|

### DIFF
--- a/vector-config/src/main/res/values/urls.xml
+++ b/vector-config/src/main/res/values/urls.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- This file contains url values-->
+
+    <string name="threads_learn_more_url" translatable="false">https://element.io/help#threads</string>
+</resources>
--- a/vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetActionItem.kt
+++ b/vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetActionItem.kt
@@ -71,6 +71,9 @@ abstract class BottomSheetActionItem : VectorEpoxyModel<BottomSheetActionItem.Ho
     @EpoxyAttribute
     var destructive = false
 
+    @EpoxyAttribute
+    var showBetaLabel = false
+
     @EpoxyAttribute(EpoxyAttribute.Option.DoNotHash)
     lateinit var listener: ClickListener
 
@@ -106,6 +109,7 @@ abstract class BottomSheetActionItem : VectorEpoxyModel<BottomSheetActionItem.Ho
         } else {
             holder.text.setCompoundDrawablesWithIntrinsicBounds(null, null, null, null)
         }
+        holder.betaLabel.isVisible =  showBetaLabel
     }
 
     class Holder : VectorEpoxyHolder() {
@@ -113,5 +117,6 @@ abstract class BottomSheetActionItem : VectorEpoxyModel<BottomSheetActionItem.Ho
         val icon by bind<ImageView>(R.id.actionIcon)
         val text by bind<TextView>(R.id.actionTitle)
         val selected by bind<ImageView>(R.id.actionSelected)
+        val betaLabel by bind<TextView>(R.id.actionBetaTextView)
     }
 }
--- a/vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetActionItem.kt
+++ b/vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetActionItem.kt
@@ -109,7 +109,7 @@ abstract class BottomSheetActionItem : VectorEpoxyModel<BottomSheetActionItem.Ho
         } else {
             holder.text.setCompoundDrawablesWithIntrinsicBounds(null, null, null, null)
         }
-        holder.betaLabel.isVisible =  showBetaLabel
+        holder.betaLabel.isVisible = showBetaLabel
     }
 
     class Holder : VectorEpoxyHolder() {
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
@@ -2393,7 +2393,6 @@ class TimelineFragment @Inject constructor(
                         .setCancelable(true)
                         .setNegativeButton(R.string.action_not_now) { _, _ -> }
                         .setPositiveButton(R.string.action_try_it_out) { _, _ ->
-                            timelineViewModel.onCleared() // We should first clear our viewModel
                             threadsManager.enableThreadsAndRestart(it)
                         }
                         .show()
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
@@ -2362,7 +2362,7 @@ class TimelineFragment @Inject constructor(
         if (vectorPreferences.areThreadMessagesEnabled()) {
             navigateToThreadTimeline(action.eventId, action.startsThread)
         } else {
-            displayThreadsBetaNotice()
+            displayThreadsBetaOptInDialog()
         }
     }
 
@@ -2385,11 +2385,7 @@ class TimelineFragment @Inject constructor(
         }
     }
 
-    /**
-     * Display a dialog that will let the user to enable threads
-     */
-
-    private fun displayThreadsBetaNotice() =
+    private fun displayThreadsBetaOptInDialog() =
             activity?.let {
                 MaterialAlertDialogBuilder(it)
                         .setTitle(R.string.threads_beta_enable_notice_title)
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
@@ -25,6 +25,7 @@ import android.os.Build
 import android.os.Bundle
 import android.text.Spannable
 import android.text.format.DateUtils
+import android.text.method.LinkMovementMethod
 import android.view.HapticFeedbackConstants
 import android.view.KeyEvent
 import android.view.LayoutInflater
@@ -169,6 +170,7 @@ import im.vector.app.features.home.room.detail.timeline.url.PreviewUrlRetriever
 import im.vector.app.features.home.room.detail.upgrade.MigrateRoomBottomSheet
 import im.vector.app.features.home.room.detail.views.RoomDetailLazyLoadedViews
 import im.vector.app.features.home.room.detail.widget.RoomWidgetsBottomSheet
+import im.vector.app.features.home.room.threads.ThreadsManager
 import im.vector.app.features.home.room.threads.arguments.ThreadTimelineArgs
 import im.vector.app.features.html.EventHtmlRenderer
 import im.vector.app.features.html.PillImageSpan
@@ -251,6 +253,7 @@ class TimelineFragment @Inject constructor(
         private val notificationDrawerManager: NotificationDrawerManager,
         private val eventHtmlRenderer: EventHtmlRenderer,
         private val vectorPreferences: VectorPreferences,
+        private val threadsManager: ThreadsManager,
         private val colorProvider: ColorProvider,
         private val dimensionConverter: DimensionConverter,
         private val userPreferencesProvider: UserPreferencesProvider,
@@ -2199,7 +2202,7 @@ class TimelineFragment @Inject constructor(
             }
             is EventSharedAction.ReplyInThread              -> {
                 if (withState(messageComposerViewModel) { it.isVoiceMessageIdle }) {
-                    navigateToThreadTimeline(action.eventId, action.startsThread)
+                    onReplyInThreadClicked(action)
                 } else {
                     requireActivity().toast(R.string.error_voice_message_cannot_reply_or_edit)
                 }
@@ -2355,6 +2358,14 @@ class TimelineFragment @Inject constructor(
                 .show()
     }
 
+    private fun onReplyInThreadClicked(action: EventSharedAction.ReplyInThread) {
+        if (vectorPreferences.areThreadMessagesEnabled()) {
+            navigateToThreadTimeline(action.eventId, action.startsThread)
+        } else {
+            displayThreadsBetaNotice()
+        }
+    }
+
     /**
      * Navigate to Threads timeline for the specified rootThreadEventId
      * using the ThreadsActivity
@@ -2373,6 +2384,29 @@ class TimelineFragment @Inject constructor(
             navigator.openThread(it, roomThreadDetailArgs)
         }
     }
+
+    /**
+     * Display a dialog that will let the user to enable threads
+     */
+
+    private fun displayThreadsBetaNotice() =
+            activity?.let {
+                MaterialAlertDialogBuilder(it)
+                        .setTitle(R.string.threads_beta_enable_notice_title)
+                        .setMessage(threadsManager.getBetaEnableThreadsMessage())
+                        .setCancelable(true)
+                        .setNegativeButton(R.string.action_not_now) { _, _ -> }
+                        .setPositiveButton(R.string.action_try_it_out) { _, _ ->
+                            timelineViewModel.onCleared() // We should first clear our viewModel
+                            threadsManager.enableThreadsAndRestart(it)
+                        }
+                        .show()
+                        ?.findViewById<TextView>(android.R.id.message)
+                        ?.apply {
+                            linksClickable = true
+                            movementMethod = LinkMovementMethod.getInstance()
+                        }
+            }
 
     /**
      * Navigate to Threads list for the current room
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
@@ -2385,7 +2385,7 @@ class TimelineFragment @Inject constructor(
         }
     }
 
-    private fun displayThreadsBetaOptInDialog() =
+    private fun displayThreadsBetaOptInDialog() {
             activity?.let {
                 MaterialAlertDialogBuilder(it)
                         .setTitle(R.string.threads_beta_enable_notice_title)
@@ -2403,6 +2403,7 @@ class TimelineFragment @Inject constructor(
                             movementMethod = LinkMovementMethod.getInstance()
                         }
             }
+        }
 
     /**
      * Navigate to Threads list for the current room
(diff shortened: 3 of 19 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -2374,6 +2385,29 @@ class TimelineFragment @Inject constructor(
         }
     }
 
+    /**
```
ouchadam, line 2388: we could avoid this comment by including the context in the function name `displayThreadsBetaOptInDialog`

vector/src/main/res/layout/item_bottom_sheet_action.xml
```diff
@@ -82,4 +82,27 @@
         tools:ignore="MissingPrefix"
         tools:visibility="visible" />
 
+
+    <TextView
```
ouchadam, line 86: would be great to get a screenshot of this view :pray:

vector/src/main/res/layout/item_bottom_sheet_action.xml
```diff
@@ -82,4 +82,27 @@
...
+        android:gravity="center"
+        android:paddingStart="10dp"
+        android:paddingTop="3dp"
+        android:paddingEnd="10dp"
+        android:layout_marginEnd="1dp"
+        android:paddingBottom="3dp"
+        android:visibility="gone"
+        tools:visibility="visible"
+        android:text="@string/beta_title_bottom_sheet_action"
+        android:textColor="@color/palette_white"
```
ouchadam, line 103: to double check, should this use the theme colours? I'm guessing we're not using them because the background colour is hardcoded blue

vector/src/main/java/im/vector/app/features/home/room/threads/ThreadsManager.kt
```diff
@@ -0,0 +1,61 @@
...
+        vectorPreferences.setThreadMessagesEnabled()
+        lightweightSettingsStorage.setThreadMessagesEnabled(vectorPreferences.areThreadMessagesEnabled())
+        MainActivity.restartApp(activity, MainActivityArgs(clearCache = true))
+    }
+
+    /**
+     * Generates and return an Html spanned string to be rendered especially in dialogs
+     */
+    fun getBetaEnableThreadsMessage(): Spanned {
+        val href = "<a href='$learnMoreUrl'>Learn more</a>.<br><br>"
```
ouchadam, line 57: if we want to allow translations here we'll need to move this `Learn more` to the resources

ouchadam: thinking a little more, assuming the `learnMoreUrl` doesn't need to be dynamic, it could be in the string resource (we've done something similar for the EMS copy in the onboarding)

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -2374,6 +2385,29 @@ class TimelineFragment @Inject constructor(
...
+
+    private fun displayThreadsBetaNotice() =
+            activity?.let {
+                MaterialAlertDialogBuilder(it)
+                        .setTitle(R.string.threads_beta_enable_notice_title)
+                        .setMessage(threadsManager.getBetaEnableThreadsMessage())
+                        .setCancelable(true)
+                        .setNegativeButton(R.string.action_not_now) { _, _ -> }
+                        .setPositiveButton(R.string.action_try_it_out) { _, _ ->
+                            timelineViewModel.onCleared() // We should first clear our viewModel
```
ouchadam, line 2400: do we need to clear the viewmodel if we're restarting the app? I'm wondering if we'll get it for free~

ariskotsomitopoulos: If we don't there is a crash in the live events observers

ariskotsomitopoulos: This is only happening if we RR the app from the Timeline, while the observers are active

vector/src/main/res/layout/item_bottom_sheet_action.xml
ariskotsomitopoulos: Yeap, we are not using them, its the same for both themes!

vector/src/main/res/layout/item_bottom_sheet_action.xml
ariskotsomitopoulos: Will do!

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ouchadam: ah that is strange, I expected restarting the app to also call `timelineViewModel.onCleared()`, perhaps there's a existing issue in the screen with the way the listeners are being set/released (if they're tied to the viewmodel lifecycle rather than the fragment)

not a blocker for this PR, just something to bear in mind for future

vector/src/main/java/im/vector/app/features/home/room/threads/ThreadsManager.kt
ariskotsomitopoulos: The easiest way to implement that functionality was to include `<a href='$learnMoreUrl'>Learn more</a>.<br><br>` in a string resources but I didn't want to do that. Thats why I decoupled the html part with the rest.

1. Yes I agree` learn more` can be also a string resource, will update it accordingly
2. I have not seen any `http` links in any of our string resources so I tried to avoid it. And if we do that we should mark it as not translatable to avoid trying to translate them. Where exactly we do that in the onboarding?

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ariskotsomitopoulos: The truth is thats a better name !

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ariskotsomitopoulos: Yes, definitely its something wrong here, manually clearing the viewModel it's a workaround

vector/src/main/java/im/vector/app/features/home/room/threads/ThreadsManager.kt
bmarty: We already have several instance of "learn more" in the string, maybe create a generic one, and we can do some cleanup during the next welbate sync.

ouchadam: for `2.`, I've just realised it hasn't been pushed yet whilst we figure out the exact url :facepalm:  the benefit of including the url in the translatable strings is that we could also change the domain based on locale eg DE could use the .de domain myurl.de 

although this only works if the url is static :thinking:

vector/src/main/java/im/vector/app/core/epoxy/bottomsheet/BottomSheetActionItem.kt
```diff
@@ -106,12 +109,14 @@ abstract class BottomSheetActionItem : VectorEpoxyModel<BottomSheetActionItem.Ho
         } else {
             holder.text.setCompoundDrawablesWithIntrinsicBounds(null, null, null, null)
         }
+        holder.betaLabel.isVisible =  showBetaLabel
```
bmarty, line 112: extra space! (ktlint should complain about it now)

ouchadam (approved):
tiny comment about the Learn more copy, otherwise LGTM! :100:

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -2374,6 +2385,29 @@ class TimelineFragment @Inject constructor(
         }
     }
 
+    /**
+     * Display a dialog that will let the user to enable threads
+     */
+
+    private fun displayThreadsBetaNotice() =
```
bmarty, line 2392: I do not like using `=` when the fun should not return anything. Can you use `{` instead?

Here it will return a `TextView?` which is highly unexpected.

image

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
bmarty: I agree this is quite strange to call `onCleared` manually (there is no other case like this in the codebase). There is probably another bug for your unexpected behavior.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
```diff
@@ -450,7 +450,8 @@ class MessageActionsViewModel @AssistedInject constructor(
     private fun canReplyInThread(event: TimelineEvent,
                                  messageContent: MessageContent?,
                                  actionPermissions: ActionPermissions): Boolean {
-        if (!vectorPreferences.areThreadMessagesEnabled()) return false
+/* We let reply in thread visible even if threads are not enabled, with an enhanced flow to attract users */
+//        if (!vectorPreferences.areThreadMessagesEnabled()) return false
```
bmarty, line 454: I would prefer to add some indentation here and for the comment above.

vector/src/main/java/im/vector/app/features/home/room/threads/ThreadsManager.kt
```diff
@@ -0,0 +1,61 @@
...
+ * The class is responsible for handling thread specific tasks
+ */
+class ThreadsManager @Inject constructor(
+        private val vectorPreferences: VectorPreferences,
+        private val lightweightSettingsStorage: LightweightSettingsStorage,
+        private val context: Context
+) {
+
+    companion object {
+        const val learnMoreUrl = "https://element.io/help#threads"
```
bmarty, line 40: Probably something to move to the `vector-config` module, maybe in a new `urls.xml` resource file

vector/src/main/res/layout/item_bottom_sheet_action.xml
```diff
@@ -82,4 +82,27 @@
...
+        android:background="@drawable/notification_badge"
+        android:backgroundTint="@color/palette_azure"
+        android:gravity="center"
+        android:paddingStart="10dp"
+        android:paddingTop="3dp"
+        android:paddingEnd="10dp"
+        android:layout_marginEnd="1dp"
+        android:paddingBottom="3dp"
+        android:visibility="gone"
+        tools:visibility="visible"
```
bmarty, line 101: please rearrange attributes

vector/src/main/res/values/strings.xml
```diff
@@ -734,6 +735,8 @@
     <string name="search_thread_from_a_thread">From a Thread</string>
     <string name="threads_notice_migration_title">Threads Approaching Beta 🎉</string>
     <string name="threads_notice_migration_message">We’re getting closer to releasing a public Beta for Threads.\n\nAs we prepare for it, we need to make some changes: threads created before this point will be displayed as regular replies.\n\nThis will be a one-off transition as Threads are now part of the Matrix specification.</string>
+    <string name="threads_beta_enable_notice_title">Threads Beta</string>
+    <string name="threads_beta_enable_notice_message">Threads help keep your conversations on-topic and easy to track. %sEnabling threads will refresh the app. This may take longer for some accounts.</string>
```
bmarty, line 739: Maybe add a comment above to explain with what `%s` will be replaced.

vector/src/main/java/im/vector/app/features/home/room/threads/ThreadsManager.kt
ariskotsomitopoulos: @bmarty I also observed it will do a clean up there.

@ouchadam I would suggest to create a separate string resource file for non translatable strings. And when we have dynamic URLs with multiple domains for sure we can add them in our translatable strings

bmarty (changes requested):

vector/src/main/java/im/vector/app/features/home/room/threads/ThreadsManager.kt
bmarty: If we want to translate the URL, this is fine, but please let the formatting out of the string resource.
Also I think the URL can be translated, but out of Weblate, since this is quite technical.

bmarty: example of url in the resources: https://github.com/vector-im/element-android/blob/main/vector-config/src/main/res/values/config.xml#L7

vector/src/main/res/layout/item_bottom_sheet_action.xml
ariskotsomitopoulos: Yes, also maybe we can force `ktlintformat` to do that if possible

vector/src/main/java/im/vector/app/features/home/room/threads/ThreadsManager.kt
ariskotsomitopoulos: Yes, thats what we discussed with adam

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ariskotsomitopoulos: I agree, and I prefer that than return Unit manually

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/action/MessageActionsViewModel.kt
ariskotsomitopoulos: Done

vector/src/main/res/layout/item_bottom_sheet_action.xml
bmarty: ktlint only works on Kotlin files AFAIK, not XML resources.
But if you found a solution for this, it could be very nice.

vector-config/src/main/res/values/urls.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- This file contains url values-->
+
+    <string name="threads_learn_more_url" translatable="false">https://element.io/help#threads</string>
```
bmarty, line 5: This anchor `threads` does not exist (yet), and today this page does not mention thread the user experience will be quite bad. But @ariskotsomitopoulos told me that the webpage will be ready today or tomorrow.
So OK to merge the PR.

vector/src/main/res/values/strings_login_v2.xml
```diff
@@ -43,7 +43,7 @@
     <string name="login_set_msisdn_title_2">Associate a phone number</string>
     <string name="login_set_msisdn_notice_2">Associate a phone number to optionally allow people you know to discover you.</string>
     <string name="login_set_msisdn_mandatory_notice_2">The server %s requires you to associate a phone number to create an account.</string>
-    <!-- %S will be replaced by the value of login_server_modular_learn_more ("Learn more" in English)-->
+    <!-- %S will be replaced by the value of action_learn_more ("Learn more" in English)-->
```
bmarty, line 46:
```suggestion
    <!-- %s will be replaced by the value of action_learn_more ("Learn more" in English)-->
```

bmarty (approved):
`timelineViewModel.onCleared()` is still annoying me :/

vector/src/main/res/values/strings.xml
```diff
@@ -2137,7 +2142,6 @@
     <string name="direct_room_profile_encrypted_subtitle">Messages here are end-to-end encrypted.\n\nYour messages are secured with locks and only you and the recipient have the unique keys to unlock them.</string>
     <string name="room_profile_section_security">Security</string>
     <string name="room_profile_section_restore_security">Restore Encryption</string>
-    <string name="room_profile_section_security_learn_more">Learn more</string>
```
bmarty, line 2140: This is not the proper way to remove string. Please refer to this: https://github.com/vector-im/element-android/blob/main/CONTRIBUTING.md#removing-existing-strings

bmarty (changes requested):
We should another way to avoid calling `onCleared()`

vector-config/src/main/res/values/urls.xml
ariskotsomitopoulos: Yeap, it should be deployed before the release as we discussed. All platforms will release that url

ariskotsomitopoulos:
Update: We achieved to avoid manually calling viewModel `onCleared`, with an alternative approach. Thanks @ouchadam  for that!

vector/src/main/res/values/strings_login_v2.xml
```diff
@@ -43,7 +43,7 @@
     <string name="login_set_msisdn_title_2">Associate a phone number</string>
     <string name="login_set_msisdn_notice_2">Associate a phone number to optionally allow people you know to discover you.</string>
     <string name="login_set_msisdn_mandatory_notice_2">The server %s requires you to associate a phone number to create an account.</string>
-    <!-- %s will be replaced by the value of action_learn_more ("Learn more" in English)-->
+    <!-- %S will be replaced by the value of login_server_modular_learn_more ("Learn more" in English)-->
```
bmarty, line 46: I think you have been a bit fast with the "replace all" command :/

bmarty (changes requested):

vector/src/main/res/values/strings_login_v2.xml
ariskotsomitopoulos: I never use that command

vector/src/main/java/im/vector/app/features/home/room/threads/ThreadsManager.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.matrix.android.sdk.internal.database.lightweight.LightweightSettingsStorage
+import javax.inject.Inject
+
+/**
+ * The class is responsible for handling thread specific tasks
+ */
+class ThreadsManager @Inject constructor(
+        private val vectorPreferences: VectorPreferences,
+        private val lightweightSettingsStorage: LightweightSettingsStorage,
+        private val context: Context
```
bmarty, line 36: could be a `stringProvider` here

ariskotsomitopoulos: Nice point, will update asap

vector/src/main/res/values/strings_login_v2.xml
bmarty: OK

vector-config/src/main/res/values/urls.xml
bmarty: Seems to be live now 👍

bmarty: image

bmarty (approved):
Thanks for the update! Happy to merge it now.

PR #5697, opened 2022-04-05
Live Location Sharing - Send location data

onurays:
Sends beacon pulse events with a reference to the initial beacon info event.

```
{
  "type": "org.matrix.msc3672.beacon",
  "sender": "@user:matrix.org",
  "room_id": "!abc123:matrix.org",
  "content": {
    "m.relates_to": {
      "event_id": "<m.beacon_info event_id>",
      "rel_type": "m.reference"
    },
    "org.matrix.msc3488.location": {
      "uri": "geo:53.99803101552848,-8.25347900390625;u=10"
    },
    "org.matrix.msc3488.ts": 1648820683092
  }
}
```

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/EventType.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/EventType.kt
@@ -49,7 +49,8 @@ object EventType {
     const val STATE_ROOM_JOIN_RULES = "m.room.join_rules"
     const val STATE_ROOM_GUEST_ACCESS = "m.room.guest_access"
     const val STATE_ROOM_POWER_LEVELS = "m.room.power_levels"
-    private const val STATE_ROOM_BEACON_INFO_PREFIX = "org.matrix.msc3489.beacon_info."
+    val STATE_ROOM_BEACON_INFO = listOf("org.matrix.msc3672.beacon_info", "m.beacon_info")
+    val BEACON_LOCATION_DATA = listOf("org.matrix.msc3672.beacon", "m.beacon")
 
     const val STATE_SPACE_CHILD = "m.space.child"
 
@@ -120,13 +121,5 @@ object EventType {
                 type == CALL_NEGOTIATE ||
                 type == CALL_REJECT ||
                 type == CALL_REPLACES
-    }
-
-    /**
-     * Returns an event type like org.matrix.msc3489.beacon_info.@userid:matrix.org.1648814272273
-     */
-    fun generateBeaconInfoStateEventType(userId: String): String {
-        val uniqueId = System.currentTimeMillis()
-        return "$STATE_ROOM_BEACON_INFO_PREFIX$userId.$uniqueId"
     }
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLiveLocationContent.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLiveLocationContent.kt
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2022 The Matrix.org Foundation C.I.C.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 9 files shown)


vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
```diff
@@ -143,6 +142,17 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
 
     override fun onLocationUpdate(locationData: LocationData) {
         Timber.i("### LocationSharingService.onLocationUpdate. Uncertainty: ${locationData.uncertainty}")
+        roomArgsList.forEach { roomArg ->
+            val room = activeSessionHolder.getSafeActiveSession()?.getRoom(roomArg.roomId)
+            room?.getStateEvent(EventType.STATE_ROOM_BEACON_INFO.first())?.let { beaconInfoEvent ->
+                room.sendLiveLocation(
+                        beaconInfoEventId = beaconInfoEvent.eventId!!,
```
Johennes, line 149: Nit: Would it make sense to chain this before the `let` to avoid the force unwrap? Looks like you're not reading anything else from the event here anyway.

onurays: Nice catch, done. (Although I don't know why event id is nullable).

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/EventType.kt
```diff
@@ -49,7 +49,8 @@ object EventType {
     const val STATE_ROOM_JOIN_RULES = "m.room.join_rules"
     const val STATE_ROOM_GUEST_ACCESS = "m.room.guest_access"
     const val STATE_ROOM_POWER_LEVELS = "m.room.power_levels"
-    private const val STATE_ROOM_BEACON_INFO_PREFIX = "org.matrix.msc3489.beacon_info."
+    val STATE_ROOM_BEACON_INFO = listOf("org.matrix.msc3672.beacon_info", "m.beacon_info")
```
ouchadam, line 52: out of interest, how come we've turned these into a list? looks like we're always using the first item~

onurays: Spec introduced unstable prefixes recently. We use them until MSCs are merged. Then we will send stable ones, so we also need to support previously sent unstable ones.

I am thinking about a better approach. We have similar events for polls and there will be more coming.

ouchadam: thanks for the explanation! could be nice to have way to say `STATE_ROOM_BEACON_INFO`.stable/unstable() instead of `STATE_ROOM_BEACON_INFO`.first()

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
```diff
@@ -143,6 +142,26 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
 
     override fun onLocationUpdate(locationData: LocationData) {
         Timber.i("### LocationSharingService.onLocationUpdate. Uncertainty: ${locationData.uncertainty}")
+
+        // Emit location update to all rooms in which live location sharing is active
+        roomArgsList.forEach { roomArg ->
+            sendLiveLocation(roomArg.roomId, locationData)
```
ouchadam, line 148: is `sendLiveLocation` a blocking network call? it might need a io dispatchers~ anddd it might be safer to copy the `roomArgsList` to avoid also blocking the synchronised calls 

```kotlin
val inputs = roomArgsList.toList()

scope.launch(io) {
  inputs.forEach { sendLiveLocation() }
}
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageLiveLocationContent.kt
```diff
@@ -0,0 +1,52 @@
...
+data class MessageLiveLocationContent(
+        /**
+         * Local message type, not from server
+         */
+        @Transient
+        override val msgType: String = MessageType.MSGTYPE_LIVE_LOCATION,
+
+        @Json(name = "body") override val body: String = "",
+        @Json(name = "m.relates_to") override val relatesTo: RelationDefaultContent? = null,
+        @Json(name = "m.new_content") override val newContent: Content? = null,
```
Johennes, line 34: Stupid question, sorry: This doesn't seem to be read or set anywhere in this diff. I also haven't seen this on any of the location sharing MSCs. What is this field used for?

PR #5707, opened 2022-04-06
Prune the content of deleted events completely

### LINKED ISSUE #5682: Deleting a location event show a message deleted w/o respecting the format, and the location can still be accessed from both sides.
### Steps to reproduce

1. Share location, current or other.
2. Delete the event.
3. The Message Deleted event is displayed with errors, and when selecting it and seeing their delails she location can still be seen, from mine and nd from the recipient side.

### Outcome

The Message Deleted should have the same format than others deleted events.
The deleted location shouldn't be accessible from any party.

### Your phone model

Samsung S21

### Operating system version

Android 12

Element Android v 1.4.8, from the Play Store

### Homeserver

matrix.org, from both sides

### Will you send logs?

Yes

ariskotsomitopoulos:
We can prune completely redacted events again.

The reason why we kept the content of redacted events was due to the fact that relations are within the content. So after an event redaction we couldn't know if the event was a thread or a normal event.

Hopefully threads are now dynamic, so the problem is minimised. To further optimise this behaviour backend actions are needed to exclude relations from within the content.

Related [MSC](https://github.com/matrix-org/matrix-spec-proposals/pull/3389)

Closes #5682

### DIFF
--- a/changelog.d/5707.bugfix
+++ b/changelog.d/5707.bugfix
@@ -1,0 +1,1 @@
+Redacted events are no longer visible.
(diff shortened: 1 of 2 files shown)


bmarty (approved):
LGTM, thanks!

PR #5709, opened 2022-04-06
The string `ftue_auth_carousel_workplace_body` was declared not translatable

### LINKED ISSUE #5262: `template_ftue_auth_carousel_body_workplace` string appears to be read-only on Weblate

bmarty:
Fixes #5262 

I do not know why the tag had this attribute, probably some historical reason.


ouchadam (approved):

bmarty:
(there is no other cases like this in this file)

PR #5711, opened 2022-04-06
Live Location Sharing - Attach location data to beacon info state event

onurays:
First `m.beacon_info` state event is sent and then `m.beacon` event with location data is sent by referencing the state event. Aggregating beacon info state event with the last location will be robust for UI rendering.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
@@ -32,7 +32,6 @@ import org.matrix.android.sdk.api.session.room.model.PowerLevelsContent
 import org.matrix.android.sdk.api.session.room.model.ReferencesAggregatedContent
 import org.matrix.android.sdk.api.session.room.model.VoteInfo
 import org.matrix.android.sdk.api.session.room.model.VoteSummary
-import org.matrix.android.sdk.api.session.room.model.livelocation.LiveLocationBeaconContent
 import org.matrix.android.sdk.api.session.room.model.message.MessageContent
 import org.matrix.android.sdk.api.session.room.model.message.MessageEndPollContent
 import org.matrix.android.sdk.api.session.room.model.message.MessageLiveLocationContent
@@ -49,7 +48,6 @@ import org.matrix.android.sdk.internal.crypto.verification.toState
 import org.matrix.android.sdk.internal.database.helper.findRootThreadEvent
 import org.matrix.android.sdk.internal.database.mapper.ContentMapper
 import org.matrix.android.sdk.internal.database.mapper.EventMapper
-import org.matrix.android.sdk.internal.database.model.CurrentStateEventEntity
 import org.matrix.android.sdk.internal.database.model.EditAggregatedSummaryEntity
 import org.matrix.android.sdk.internal.database.model.EditionOfEvent
 import org.matrix.android.sdk.internal.database.model.EventAnnotationsSummaryEntity
@@ -63,11 +61,11 @@ import org.matrix.android.sdk.internal.database.model.TimelineEventEntity
 import org.matrix.android.sdk.internal.database.model.TimelineEventEntityFields
 import org.matrix.android.sdk.internal.database.query.create
 import org.matrix.android.sdk.internal.database.query.getOrCreate
-import org.matrix.android.sdk.internal.database.query.getOrNull
 import org.matrix.android.sdk.internal.database.query.where
 import org.matrix.android.sdk.internal.di.SessionId
 import org.matrix.android.sdk.internal.di.UserId
 import org.matrix.android.sdk.internal.session.EventInsertLiveProcessor
+import org.matrix.android.sdk.internal.session.room.aggregation.livelocation.LiveLocationAggregationProcessor
 import org.matrix.android.sdk.internal.session.room.state.StateEventDataSource
 import timber.log.Timber
 import javax.inject.Inject
@@ -76,7 +74,8 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
         @UserId private val userId: String,
         private val stateEventDataSource: StateEventDataSource,
         @SessionId private val sessionId: String,
-        private val sessionManager: SessionManager
+        private val sessionManager: SessionManager,
+        private val liveLocationAggregationProcessor: LiveLocationAggregationProcessor
 ) : EventInsertLiveProcessor {
 
     private val allowedTypes = listOf(
@@ -191,7 +190,7 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
                             }
                             in EventType.BEACON_LOCATION_DATA -> {
                                 event.content.toModel<MessageLiveLocationContent>(catchError = true)?.let {
-                                    handleLiveLocation(realm, event, it, roomId, isLocalEcho)
+                                    liveLocationAggregationProcessor.handleLiveLocation(realm, event, it, roomId, isLocalEcho)
                                 }
                             }
                         }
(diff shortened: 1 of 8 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -532,6 +546,49 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
         }
     }
 
+    private fun handleLiveLocation(realm: Realm,
```
mnaturel, line 549: In my opininion, we clearly need to stop adding code to this class. It is too big. To separate easily the code, I propose using delegation pattern. The idea is:

- create a dedicated package named `aggregation.livelocation`. 
- In this package, create an interface as follow:
```kotlin
interface LiveLocationAggregationProcessor {
    fun handleLiveLocation(realm: Realm,
                           event: Event,
                           content: MessageLiveLocationContent,
                           roomId: String,
                           isLocalEcho: Boolean)
}
```
- In the same package create the class to implement the aggregation (moved from `EventRelationsAggregationProcessor`)
- Finally make `EventRelationsAggregationProcessor` implements the `LiveLocationAggregationProcessor` interface using delegation:
```kotlin
EventRelationsAggregationProcessor(...) : LiveLocationAggregationProcessor by DefaultLiveLocationAggregationProcessor() {...}
```

This way we have a class dedicated to aggregating a single event type and it is easier to add unit tests on it.
What do you think?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -532,6 +546,49 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
...
+                                   isLocalEcho: Boolean) {
+        val locationSenderId = event.senderId ?: return
+
+        // We shouldn't process local echos
+        if (isLocalEcho) {
+            return
+        }
+
+        // A beacon info state event has to be sent before sending location
+        val beaconInfoEntity = CurrentStateEventEntity.getOrNull(realm, roomId, locationSenderId, EventType.STATE_ROOM_BEACON_INFO.first())
```
mnaturel, line 562: Would it be possible to check for both the stable and unstable prefixes for compaibility reason? And using `first()`, we are not sure to use the correct prefix since the list may change over time.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -532,6 +546,49 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
...
+
+        // Update last location info of the beacon state event
+        beaconInfoContent.lastLocationContent = content
+        beaconInfoEntity.root?.content = ContentMapper.map(beaconInfoContent.toContent())
+    }
+
+    private fun isBeaconInfoOutdated(beaconInfoContent: LiveLocationBeaconContent,
+                                     liveLocationContent: MessageLiveLocationContent): Boolean {
+        val beaconInfoStartTime = beaconInfoContent.getBestTimestampAsMilliseconds() ?: 0
+        val liveLocationEventTime = liveLocationContent.getBestTs() ?: 0
```
mnaturel, line 587: Is it possible to do some renamings on `MessageLiveLocationContent` for the timestamps fields and the method `getBestTs()` to precise the unit and not use abreviation? For the method we would have `getBestTimestampAsMilliseconds()`.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -532,6 +546,49 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
...
+        }
+
+        // Update last location info of the beacon state event
+        beaconInfoContent.lastLocationContent = content
+        beaconInfoEntity.root?.content = ContentMapper.map(beaconInfoContent.toContent())
+    }
+
+    private fun isBeaconInfoOutdated(beaconInfoContent: LiveLocationBeaconContent,
+                                     liveLocationContent: MessageLiveLocationContent): Boolean {
+        val beaconInfoStartTime = beaconInfoContent.getBestTimestampAsMilliseconds() ?: 0
```
mnaturel, line 586: I guess we should first check if the beacon is still live before checking the timeout, what do you think?

mnaturel (commented):
I have some remarks to improve separation of code and about the process of the timestamps. How can I test if the events are well agregatted in database?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -532,6 +546,49 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
...
+
+        // Check if beacon info is outdated
+        if (isBeaconInfoOutdated(beaconInfoContent, content)) {
+            Timber.v("## LIVE LOCATION. Beacon info content is invalid")
+            return
+        }
+
+        // Update last location info of the beacon state event
+        beaconInfoContent.lastLocationContent = content
+        beaconInfoEntity.root?.content = ContentMapper.map(beaconInfoContent.toContent())
```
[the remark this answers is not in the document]
onurays: I debugged on `TimelineEventViewModel.init()` like `room.getStateEvent(EventType.STATE_ROOM_BEACON_INFO.first()).content`. I don't want to block your work in timeline for now. But I am going to create integration tests next.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
onurays: Right, done. It is also easy to test now.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
onurays: Oh, nice catch! Done.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
onurays: Done.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
onurays: Right, so we need to set `live=false` to stop sharing live location.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/livelocation/DefaultLiveLocationAggregationProcessor.kt
```diff
@@ -0,0 +1,80 @@
...
+    override fun handleLiveLocation(realm: Realm, event: Event, content: MessageLiveLocationContent, roomId: String, isLocalEcho: Boolean) {
+        val locationSenderId = event.senderId ?: return
+
+        // We shouldn't process local echos
+        if (isLocalEcho) {
+            return
+        }
+
+        // A beacon info state event has to be sent before sending location
+        val beaconInfoEntity = CurrentStateEventEntity.getOrNull(realm, roomId, locationSenderId, EventType.STATE_ROOM_BEACON_INFO.first())
```
mnaturel, line 44: I guess we should check all values of the list of `EventType.STATE_ROOM_BEACON_INFO` in case the size of the list change in the future. Something like that:
```kotlin
var beaconInfoEntity: CurrentStateEventEntity? = null
val eventTypesIterator = EventType.STATE_ROOM_BEACON_INFO.iterator()
while (beaconInfoEntity == null && eventTypesIterator.hasNext()) {
    beaconInfoEntity = CurrentStateEventEntity.getOrNull(realm, roomId, locationSenderId, eventTypesIterator.next())
}
```

mnaturel (commented):
Thanks for having extracted the aggregation into a dedicated component! I have added a small remaining comment.

PR #5719, opened 2022-04-07
Do not cancel the current sync request when the app goes to background

### LINKED ISSUE #5621: Sync request parsing takes a long time (can be more than 30s)
### Steps to reproduce

 1. Under **Settings** -> **Advanced Settings** -> enable **Developer Mode**
 1. And enable **Show debug info on screen**
 1. Open Element after some time has passed and activity has occurred
 1. Notice how the `/sync` network request is actually pretty fast but the parsing rooms and `toDevices` step takes a considerable amount longer

#### What did you expect?

I would expect the network to be the bigger bottleneck in seeing new messages over parsing JSON locally.

It just feels like there is some optimization to be had here. [GTA load times style](https://nee.lv/2021/02/28/How-I-cut-GTA-Online-loading-times-by-70/)

#### What happened instead?

Network `/sync` request is fast but the parsing step is slow.

### Your phone model

Pixel 4a

### Operating system version

Android 12

Element 1.4.4, olm version 3.2.10

### Homeserver

matrix.org

### Will you send logs?

No

bmarty:
The important change is in 7cfa38870b5854f733ee4324b4afe2916c3555fc : we do not cancel the current sync request when the app goes to background, to give a chance to the current sync, or the current sync response treatment to finish.

Test:
- on a big account, go to bg
- start the app
- observe that a incr sync is started (show debug info setting can help)
- go the bg
- previously the current incr sync was cancelled
- now the treatment continues
- if the user put the app in fg, we can see the current sync treatment is still in progress
- if the user wait a long time, the current sync treatment can end in bg
- the sync thread does not start another sync request when the app stay in bg

According to my tests, this is all fine.

This PR also contains some other minor changes regarding logs.

Closes #5621 . We will not investigate more the sync duration since it will be replaced by the Rust implementation in a near future.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
@@ -584,11 +584,11 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
                 sum.key = reaction
                 sum.firstTimestamp = event.originServerTs ?: 0
                 if (isLocalEcho) {
-                    Timber.v("Adding local echo reaction $reaction")
+                    Timber.v("Adding local echo reaction")
                     sum.sourceLocalEcho.add(txId)
                     sum.count = 1
                 } else {
-                    Timber.v("Adding synced reaction $reaction")
+                    Timber.v("Adding synced reaction")
                     sum.count = 1
                     sum.sourceEvents.add(reactionEventId)
                 }
@@ -600,16 +600,16 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
                     // check if it's not the sync of a local echo
                     if (!isLocalEcho && sum.sourceLocalEcho.contains(txId)) {
                         // ok it has already been counted, just sync the list, do not touch count
-                        Timber.v("Ignoring synced of local echo for reaction $reaction")
+                        Timber.v("Ignoring synced of local echo for reaction")
                         sum.sourceLocalEcho.remove(txId)
                         sum.sourceEvents.add(reactionEventId)
                     } else {
                         sum.count += 1
                         if (isLocalEcho) {
-                            Timber.v("Adding local echo reaction $reaction")
+                            Timber.v("Adding local echo reaction")
                             sum.sourceLocalEcho.add(txId)
                         } else {
-                            Timber.v("Adding synced reaction $reaction")
+                            Timber.v("Adding synced reaction")
                             sum.sourceEvents.add(reactionEventId)
                         }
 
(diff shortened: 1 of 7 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -584,11 +584,11 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
                 sum.key = reaction
                 sum.firstTimestamp = event.originServerTs ?: 0
                 if (isLocalEcho) {
-                    Timber.v("Adding local echo reaction $reaction")
+                    Timber.v("Adding local echo reaction")
```
ganfra, line 587: Why do you remove reaction here? it can be helpful to see which one is being synced

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/job/SyncThread.kt
```diff
@@ -107,7 +107,11 @@ internal class SyncThread @Inject constructor(private val syncTask: SyncTask,
             Timber.tag(loggerTag.value).d("Pause sync...")
             isStarted = false
             retryNoNetworkTask?.cancel()
-            syncScope.coroutineContext.cancelChildren()
+            // Do not cancel the current incremental sync.
+            // Incremental sync can be long and it requires the user to wait for the treatment to end,
+            // else all is restarted from the beginning each time the user moves the app to foreground.
+            Timber.tag(loggerTag.value).d("Pause sync... Not cancelling incremental sync")
+            // syncScope.coroutineContext.cancelChildren()
```
ericdecanini, line 114: Fix looks good but I'd rather remove the offending line than simply comment it out. We always have git history if we need to bring it back, but commented code is a code smell

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
bmarty: For privacy reason

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/job/SyncThread.kt
bmarty: I agree on the principle, also sonar for instance complain about commented code. But git history is less visible than a commented out line of code. Also sometimes the history can be lost during rework (if not done correctly).
So I prefer to comment the line.
Also in case of trouble it's easier to see the change and to revert.

ouchadam: I would also be in the _removing the line_ camp, if we want to protect against regressions a unit test to replace the comment which asserts that pauses do not cancel inprogress syncs would be stronger 

although with the upcoming rust SDK changes I'm not sure it would be the best use of time to add tests here~

if context/history is important I would be for linking to the PR rather than commenting the line

bmarty: OK. I have force push the last commit, so that it reduces noise. Also note that the commit contains the changelog file, which links to the PR :).

PR #5758, opened 2022-04-13
Live Location Sharing - Update beacon info state event when sharing is ended

onurays:


vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
```diff
@@ -140,6 +145,25 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
         }
     }
 
+    private fun updateStoppedBeaconInfo(roomId: String) {
```
mnaturel, line 148: To clarify namings what do you think about renaming this method to `sendStoppedBeaconInfo()` and the method `sendBeaconInfo()` to `sendLivedBeaconInfo()` ?

vector/src/main/java/im/vector/app/features/location/LocationSharingServiceConnection.kt
```diff
@@ -51,7 +52,12 @@ class LocationSharingServiceConnection @Inject constructor(
         callback = null
     }
 
+    fun stopLiveLocationSharing(roomId: String) {
+        locationSharingService?.stopSharingLocation(roomId)
+    }
+
     override fun onServiceConnected(className: ComponentName, binder: IBinder) {
+        locationSharingService = (binder as LocationSharingService.LocalBinder).getService()
```
mnaturel, line 60: Should we reset `locationSharingService` to `null` in `onServiceDisconnected` for safety reason?

mnaturel (commented):
I have just added small comments to improve code.

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
```diff
@@ -140,6 +145,25 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
...
+        activeSessionHolder
+                .getSafeActiveSession()
+                ?.let { session ->
+                    session.coroutineScope.launch(session.coroutineDispatchers.io) {
+                        val room = session.getRoom(roomId)
+                        EventType
```
mnaturel, line 154: I think we also need to add a filter to get all state events with `lived=true`. Normally there should be only a single live beacon per user per room. Maybe we need to extract it into a dedicated method to use it also in `sendLiveLocation`. I see in this last method 2 missing filters: the state_key=userId and lived=true which will be solved by this new method.

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
onurays: Done.

vector/src/main/java/im/vector/app/features/location/LocationSharingServiceConnection.kt
onurays: Good catch, done.

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
onurays: Moved to an helper function.

Issue #5726, opened 2022-04-11
Setup Dokka and cleanup SDK API


bmarty:
- [x] Setup Dokka to generate documentation of our Matrix SDK API
- [x] Host generated documentation somewhere (**where?**). Do something like https://matrix-org.github.io/matrix-rust-sdk/ if possible ?
- [x] Cleanup SDK api. Add `internal` keyword, move `internal` classes used by the app to `api` package (it will fix a bunch of TODOs!). Generated doc should not contain documented classes from the `internal` package.
- [ ] Consider usage of `explicitApi`. See https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors
- [x] Ensure there is a changelog `.sdk` file to explain what has been moved/updated
- [ ] Add missing API documentation
- [x] Ensure that the CI checks that app modules do not import classes from `internal` package (update our check tool)

bmarty:
Documentation is hosted here: https://matrix-org.github.io/matrix-android-sdk2/

bmarty:
missing documentation is added on a regular basis

PR #5762, opened 2022-04-14
Make some data class immutable

bmarty:
Following #5726 where some TODOs has been added, some data class was using `var` instead of `val`. This PR fixes that.

Hopefully it will not break anything, but please review carefully.

### DIFF
--- a/dependencies_groups.gradle
+++ b/dependencies_groups.gradle
@@ -50,6 +50,7 @@ ext.groups = [
                         'com.beust',
                         'com.davemorrissey.labs',
                         'com.dropbox.core',
+                        'com.soywiz.korlibs.korte',
                         'com.facebook.fbjni',
                         'com.facebook.fresco',
                         'com.facebook.infer.annotation',
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/keysbackup/KeysBackupVersionTrustSignature.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/keysbackup/KeysBackupVersionTrustSignature.kt
@@ -20,23 +20,21 @@ import org.matrix.android.sdk.api.session.crypto.model.CryptoDeviceInfo
 
 /**
  * A signature in a `KeysBackupVersionTrust` object.
- * TODO Make it a data class with only val
  */
-class KeysBackupVersionTrustSignature {
+data class KeysBackupVersionTrustSignature(
+        /**
+         * The id of the device that signed the backup version.
+         */
+        val deviceId: String?,
 
-    /**
-     * The id of the device that signed the backup version.
-     */
-    var deviceId: String? = null
+        /**
+         * The device that signed the backup version.
+         * Can be null if the device is not known.
+         */
+        val device: CryptoDeviceInfo?,
 
-    /**
-     * The device that signed the backup version.
-     * Can be null if the device is not known.
-     */
-    var device: CryptoDeviceInfo? = null
-
-    /**
-     * Flag to indicate the signature from this device is valid.
-     */
-    var valid = false
-}
+        /**
+         * Flag to indicate the signature from this device is valid.
+         */
+        val valid: Boolean,
+)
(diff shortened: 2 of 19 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -316,14 +316,16 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
...
-                        winnerVoteCount = 0
-                        votes = emptyList()
-                        votesSummary = emptyMap()
-                    }
-                    ?.apply {
-                        eventAnnotationsSummaryEntity.pollResponseSummary?.aggregatedContent = ContentMapper.map(toContent())
+                    ?.let { existingPollSummaryContent ->
+                        eventAnnotationsSummaryEntity.pollResponseSummary?.aggregatedContent = ContentMapper.map(
+                                PollSummaryContent(
+                                        myVote = existingPollSummaryContent.myVote,
```
bmarty, line 322: This code is quite weird to men the goal is to only keep `myVote`? CC @onurays

dependencies_groups.gradle
```diff
@@ -50,6 +50,7 @@ ext.groups = [
                         'com.beust',
                         'com.davemorrissey.labs',
                         'com.dropbox.core',
+                        'com.soywiz.korlibs.korte',
```
ouchadam, line 53: was this missing from the previous PR? :thinking:  thought the CI would catch those errors

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
onurays: Yes, it is to help reach myVote easily while rendering it.

bmarty: OK, thanks

dependencies_groups.gradle
bmarty: The CI does not call (yet) `./gradlew dokkaHtml` which support has been added by the previous PR, and this addition is to be able to call `./gradlew dokkaJavadoc` (as per the commit message)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/keysbackup/KeysBackupVersionTrustSignature.kt
```diff
@@ -20,23 +20,21 @@ import org.matrix.android.sdk.api.session.crypto.model.CryptoDeviceInfo
...
-
-    /**
-     * Flag to indicate the signature from this device is valid.
-     */
-    var valid = false
-}
+        /**
+         * Flag to indicate the signature from this device is valid.
+         */
+        val valid: Boolean,
```
onurays, line 39: Should we rename it as `isValid`?

ouchadam (approved):
from the `var` to `val` perspective everything looks fine, seems that we were just using them to accumulate data for constructors

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
ouchadam: not for this PR, might be better suited as a separate type if possible :thinking:

bmarty: Can you clarify? You mean having something like `data class MutablePollSummaryContent` with `var` and mutable types and a `toPollSummaryContent()` method, or is it something else?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/keysbackup/KeysBackupVersionTrustSignature.kt
bmarty: Yes, you are right, something for a future PR :)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
ouchadam: I meant a dedicated `MyVotePollContent(myVote)`, avoiding the need for the other values to be set to default values

Issue #1899, opened 2020-08-12
Immediatly hide messages from just ignored user


bmarty:
When ignoring a user, the timeline still display the previous messages sent by the user. Element-Web do a local filter of such messages.

Clearing cache fix that, but it's not ideal.

Current side effect: it's still possible to open room member profile screen of an ignored user, with the possibility to ignore it again. There is no need to implement an unignore action here, if we properly fix this issue.

bmarty:
Fixed by #5772, in EA 1.4.14

PR #5772, opened 2022-04-15
Delete Events from ignored users and trigger a clear cache request dialog when unignoring user(s)

bmarty:
## Ignoring users

When users is ignored, all the Timeline Events from this user are deleted from the database.

<del>The only remaining issue is that the timeline has to be reloaded for the change to take effect. Related (but not really closing it): #1899</del>. Handled in https://github.com/vector-im/element-android/pull/5772/commits/903e226e5934a852f910ea5861de357bb33b4e2b

Closing #1899

## Unignoring user

When users are unignored, an initial sync has to be performed. Current code warn the user and perform an initial sync.

This PR handle the case where users are unignored from another session.

So:
- The warnings about an initial sync when unignoring a user has been removed
- A dialog is displayed when a user is unignored. This information is coming from the sync, so it can be a consequence of unignoring locally of from another session.
- It's possible to does the init sync later, so several users can be unignored without having to perform an init sync for each user.

The dialog looks like:

image


bmarty:
Assigned to @ariskotsomitopoulos manually, since you have reviewed the previous PR #5767 and this PR reverts part of the previous change.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/UserAccountDataSyncHandler.kt
```diff
@@ -184,12 +195,39 @@ internal class UserAccountDataSyncHandler @Inject constructor(
...
+        currentIgnoredUsers.deleteAllFromRealm()
         // And save the new received list
         userIds.forEach { realm.createObject(IgnoredUserEntity::class.java).apply { userId = it } }
-        // TODO If not initial sync, we should execute a init sync
+
+        // Delete all the TimelineEvents for all the ignored users
+        // See https://spec.matrix.org/latest/client-server-api/#client-behaviour-22 :
+        // "Once ignored, the client will no longer receive events sent by that user, with the exception of state events"
+        // So just delete all non-state events from our local storage.
+        realm.where(TimelineEventEntity::class.java)
```
ariskotsomitopoulos, line 209: Might be better to extract this functionality in `TimelineEventEntityQueries.kt` or `TimelineEventHelper.kt`. Not a blocker

ariskotsomitopoulos (approved):
Thanks, approved, I didn't run the code. Just a minor comment

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/UserAccountDataSyncHandler.kt
bmarty: You are right. Will be handled in a dedicated PR later.

PR #5773, opened 2022-04-15
Move "Ignored users" setting section into "Security & Privacy"

bmarty:
as per Element Web

|Before|After|
|-|-|
|image|imageimage|


onurays (approved):
LGTM, thanks!

PR #5774, opened 2022-04-15
Add a picto for ignored users in the room member list screen

### LINKED ISSUE #5764: Ignored users should be more visible in the room member list screen

I would like to be able to see which users are ignored in a specific room, even if the users are ignored for all the rooms.

I would like to be able to easily fix an issue when I suspect I do not see all the messages.

In the room member list, "Room detail/X people", I would like ignored users, who are still displayed there, to be highlighted in a way or another:
- Either by adding a "Ignored" label
- or by moving them to a dedicated "Ignored members" section at the bottom of the list.

Today it's possible to see the list of ignored users, from "Settings/Ignored users", but it's not accessible when a timeline is displayed, or from the room settings

### Additional context

Sometimes users are ignored by mistake and this change could help users of Element Android to fix the problem

bmarty:
Proposal for #5764

Closes #5764

|Before|After|
|-|-|
|image|image|


fedrunov (approved):
LGTM

PR #5780, opened 2022-04-15
Remove unused dependencies

ofalvai:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ buildscript {
 // ktlint Plugin
 plugins {
     id "org.jlleitschuh.gradle.ktlint" version "10.2.1"
+    id 'com.autonomousapps.dependency-analysis' version "1.0.0"
 }
 
 // https://github.com/jeremylong/DependencyCheck
@@ -193,3 +194,19 @@ project(":library:diff-match-patch") {
 //        }
 //    }
 //}
+
+dependencyAnalysis {
+    dependencies {
+        bundle("kotlin-stdlib") {
+            includeGroup("org.jetbrains.kotlin")
+        }
+        bundle("react") {
+            includeGroup("com.facebook.react")
+        }
+    }
+    issues {
+        all {
+            ignoreKtx(true) // default is false
+        }
+    }
+}
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -81,7 +81,7 @@ ext.libs = [
                 'hiltCompiler'            : "com.google.dagger:hilt-compiler:$dagger"
         ],
         squareup    : [
-                'moshi'                  : "com.squareup.moshi:moshi-adapters:$moshi",
+                'moshi'                  : "com.squareup.moshi:moshi:$moshi",
                 'moshiKotlin'            : "com.squareup.moshi:moshi-kotlin-codegen:$moshi",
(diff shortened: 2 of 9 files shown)


build.gradle
```diff
@@ -31,6 +31,7 @@ buildscript {
 // ktlint Plugin
 plugins {
     id "org.jlleitschuh.gradle.ktlint" version "10.2.1"
+    id 'com.autonomousapps.dependency-analysis' version "1.0.0"
```
bmarty, line 34: After a discussion with the team, wwe are wondering of the overhead of adding a generally unused plugin. We think this is still acceptable.

Also maybe we could try to add a CI check using this plugin, and ensure it fails when unused dependencies are detected, ignoring false positive.

bmarty (changes requested):
Ah sorry, I will need a sign-off to be able to merge the PR.
See https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off

ofalvai:
@bmarty I added the missing sign-off as well.

build.gradle
ofalvai: The full dependency analysis takes a long time, so I think it's a better idea to run it in parallel to all the other checks, but let me know if it's not feasible.
Here is an example workflow run: https://github.com/vector-im/element-android/runs/6505420170?check_suite_focus=true

.github/workflows/quality.yml
```diff
@@ -97,6 +102,25 @@ jobs:
               comment_id: ${{ steps.fc.outputs.comment-id }}
             })
 
+# Gradle dependency analysis using https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin
+  dependency-analysis:
+    name: Dependency analysis
+    runs-on: ubuntu-latest
+    # Allow all jobs on main and develop. Just one per PR.
+    concurrency:
+      group: ${{ github.ref == 'refs/heads/main' && format('ktlint-main-{0}', github.sha) || github.ref == 'refs/heads/develop' && format('ktlint-develop-{0}', github.sha) || format('ktlint-{0}', github.ref) }}
```
bmarty, line 111: I think you need to replace all occurence of `ktlint` by `dep` or something similar. I am not an expert on GHA though.

bmarty (changes requested):
Thanks for adding the GHA. One remark on it.

bmarty:
Also why all the advices from the report are not applied? Are they false positive? (Sorry if you mentioned it before)
If they were all applied, we could enforce the GHA to be mandatory for each PR.

.github/workflows/quality.yml
ofalvai: Good catch, thank you! Fixed now.

ofalvai:
@bmarty I marked all the other advice types optional because I'm not familiar with the module hierarchy of this codebase and I was afraid I would break something by changing those `api` vs. `implementation` dependency declarations. I think it's a good start to flag the unused dependencies first, then the rules can be tweaked later.

bmarty (approved):
Thanks for the update and the latest comments.

PR #5781, opened 2022-04-17
create fallback turn server option

### LINKED ISSUE #2329: allow to use fallbackcall assist server (turn.matrix.org)

### LINKED ISSUE #3063: ability to switch on using turn server turn.matrix.org from Android device

CicadaCinema:
Fixes #2329 
Fixes #3063

# PR Moved from https://github.com/vector-im/element-android/pull/4391

## Screenshot
You need to enable the fallback option in `Settings > Voice & Video` to use this feature.

### DIFF
--- a/vector/src/main/res/values/strings.xml
+++ b/vector/src/main/res/values/strings.xml
@@ -548,6 +548,8 @@
     <string name="settings_call_category">Calls</string>
     <string name="settings_call_show_confirmation_dialog_title">Prevent accidental call</string>
     <string name="settings_call_show_confirmation_dialog_summary">Ask for confirmation before starting a call</string>
+    <string name="settings_call_ringtone_use_default_stun_title">Allow fallback call assist server</string>
+    <string name="settings_call_ringtone_use_default_stun_summary">Will use %s as assist when your homeserver does not offer one (your IP address will be seen by the stun server during a call)</string>
     <!-- Note to translators: the translation MUST contain the string "${app_name}", which will be replaced by the application name -->
     <string name="settings_call_ringtone_use_app_ringtone">Use default ${app_name} ringtone for incoming calls</string>
     <string name="settings_call_ringtone_title">Incoming call ringtone</string>
--- a/changelog.d/2329.feature
+++ b/changelog.d/2329.feature
@@ -1,0 +1,1 @@
+Add setting to enable using turn.matrix.org as a fallback measure.
--- a/vector-config/src/main/res/values/config.xml
+++ b/vector-config/src/main/res/values/config.xml
@@ -5,6 +5,7 @@
 
     <!-- server urls -->
     <string name="matrix_org_server_url" translatable="false">https://matrix.org</string>
+    <string name="fallback_stun_server_url" translatable="false">turn.matrix.org</string>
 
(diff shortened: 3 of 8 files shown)


CicadaCinema:
Signed-off-by: Arseny Uskov <arsenyuskov@gmail.com>

CicadaCinema:
I've addressed the comments in https://github.com/vector-im/element-android/pull/4391 (apart from proactively asking the user to turn this option on if their homeserver has no ice server config) and moved this to a new branch - hence the new PR.

CicadaCinema:
@BillCarsonFr @bmarty

vector/src/main/res/values/strings.xml
```diff
@@ -548,6 +548,8 @@
     <string name="settings_call_category">Calls</string>
     <string name="settings_call_show_confirmation_dialog_title">Prevent accidental call</string>
     <string name="settings_call_show_confirmation_dialog_summary">Ask for confirmation before starting a call</string>
+    <string name="settings_call_ringtone_use_default_stun_title">Allow fallback call assist server</string>
+    <string name="settings_call_ringtone_use_default_stun_summary">Will use %s as assist when your homeserver does not offer one (your IP address will be seen by the stun server during a call)</string>
```
bmarty, line 552: `ringtone` is a bit strange here, probably coming from a copy/paste :)

bmarty (commented):
I think this is OK now, WDYT @BillCarsonFr ?

vector/src/main/res/values/strings.xml
CicadaCinema: > `ringtone` is a bit strange here (for the 2 string keys), probably coming from a copy/paste :)

Yes, I copied the string keys from the old riot-android repo:

https://github.com/vector-im/riot-android/blob/a763655c54c4e465d2ae81e3ecd6a5146564771f/vector/src/main/res/values/strings.xml#L349-L350

PR #5785, opened 2022-04-19
FTUE - Registration step ordering

### LINKED ISSUE #5783: FTUE - Update sign up steps ordering
The ideal ordering is 

- Create account 
- Enter and verify email
- Privacy policy / T&Cs
- Captcha

ouchadam:
## Type of change

- [x] WIP Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes #5783 

Reorders the FTUE onboarding steps to prioritise Email first, instead of consuming the first step given to us by the SDK. 

## Motivation and context

To match the flow with the designs

## Screenshots / GIFs

No gifs due to needing real email accounts

## Tests

- Enable the combined register flag
- Start account creation
- Notice email is asked first (if the homeserver requires it)

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 29

### DIFF
--- a/changelog.d/5783.wip
+++ b/changelog.d/5783.wip
@@ -1,0 +1,1 @@
+Reorders the registration steps to prioritise email, then terms for the FTUE onboarding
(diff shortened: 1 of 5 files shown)


ericdecanini (approved):
Idk how it is on the sdk side of things but it feels unideal that we have to perform this logic on the client side.

Anyway as far as the implementation goes, LGTM!

ouchadam:
> Idk how it is on the sdk side of things but it feels unideal that we have to perform this logic on the client side.
> 
> Anyway as far as the implementation goes, LGTM!

totally agree, I'm wondering if the sdk should have a parameter to customise the ordering :thinking:  

for some extra context, the main reason for not doing this change within the SDK is that Element android has different orders (depending on the new or legacy onboarding flow) with is client behaviour rather than at the spec/protocol level~

ouchadam:
I've created an issue https://github.com/vector-im/element-android/issues/5859 to unblock this PR

PR #5815, opened 2022-04-21
Enhance crypto posthog errors with more details

ariskotsomitopoulos:
This PR aims to break down E2EE analytics errors into more detailed codes. 

Analytics `Error.context` schema will be used to temporary store a more detailed exception. The plan is to create a crypto specific schema/type to appropriately store crypto errors. 

`mxc_crypto_error_type` will be added in the context to differentiate crypto errors

### DIFF
--- a/vector/src/main/java/im/vector/app/features/analytics/DecryptionFailureTracker.kt
+++ b/vector/src/main/java/im/vector/app/features/analytics/DecryptionFailureTracker.kt
@@ -29,6 +29,7 @@ import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
 import org.matrix.android.sdk.api.session.crypto.MXCryptoError
 import org.matrix.android.sdk.api.session.room.timeline.TimelineEvent
+import timber.log.Timber
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -38,7 +39,7 @@ private data class DecryptionFailure(
         val failedEventId: String,
         val error: MXCryptoError.ErrorType
 )
-
+private typealias DetailedErrorName  = Pair<String, Error.Name>
 private const val GRACE_PERIOD_MILLIS = 4_000
 private const val CHECK_INTERVAL = 2_000L
 
@@ -112,7 +113,7 @@ class DecryptionFailureTracker @Inject constructor(
 
     private fun checkFailures() {
         val now = clock.epochMillis()
-        val aggregatedErrors: Map<Error.Name, List<String>>
+        val aggregatedErrors: Map<DetailedErrorName, List<String>>
         synchronized(failures) {
             val toReport = mutableListOf<DecryptionFailure>()
             failures.removeAll { failure ->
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/analytics/DecryptionFailureTracker.kt
```diff
@@ -136,20 +136,19 @@ class DecryptionFailureTracker @Inject constructor(
...
-                        analyticsTracker.capture(Error(failedEventId, Error.Domain.E2EE, aggregation.key))
+                        analyticsTracker.capture(Error(aggregation.key.first, Error.Domain.E2EE, aggregation.key.second))
                         alreadyReported.add(failedEventId)
                     }
         }
     }
 
-    private fun MXCryptoError.ErrorType.toAnalyticsErrorName(): Error.Name {
+    private fun MXCryptoError.ErrorType.toAnalyticsErrorName(): DetailedErrorName {
+        val detailed = "$name | mxc_crypto_error_type"
```
BillCarsonFr, line 146: Why add the `| mxc_crypto_error_type` and not just $name?

vector/src/main/java/im/vector/app/features/analytics/DecryptionFailureTracker.kt
```diff
@@ -136,20 +136,19 @@ class DecryptionFailureTracker @Inject constructor(
...
         return when (this) {
-            MXCryptoError.ErrorType.UNKNOWN_INBOUND_SESSION_ID -> Error.Name.OlmKeysNotSentError
-            MXCryptoError.ErrorType.OLM                        -> {
-                Error.Name.OlmUnspecifiedError
-            }
-            MXCryptoError.ErrorType.UNKNOWN_MESSAGE_INDEX      -> Error.Name.OlmIndexError
-            else                                               -> Error.Name.UnknownError
+            MXCryptoError.ErrorType.UNKNOWN_INBOUND_SESSION_ID -> Pair(detailed, Error.Name.OlmKeysNotSentError)
+            MXCryptoError.ErrorType.OLM                        -> Pair(detailed, Error.Name.OlmUnspecifiedError)
+            MXCryptoError.ErrorType.UNKNOWN_MESSAGE_INDEX      -> Pair(detailed, Error.Name.OlmIndexError)
```
BillCarsonFr, line 150: I think we can add MXCryptoError.ErrorType.KEYS_WITHHELD -> OlmKeysNotSentError
even if it could be a OlmIndexError

vector/src/main/java/im/vector/app/features/analytics/DecryptionFailureTracker.kt
ariskotsomitopoulos: For two reasons:
- To differentiate/filter out already existing values from that field. While context is already used.
- To filter only crypto errors, while someone else may use this field

vector/src/main/java/im/vector/app/features/analytics/DecryptionFailureTracker.kt
ariskotsomitopoulos: Done, lets try that

vector/src/main/java/im/vector/app/features/analytics/DecryptionFailureTracker.kt
```diff
@@ -136,20 +137,20 @@ class DecryptionFailureTracker @Inject constructor(
...
-            MXCryptoError.ErrorType.OLM                        -> {
-                Error.Name.OlmUnspecifiedError
-            }
-            MXCryptoError.ErrorType.UNKNOWN_MESSAGE_INDEX      -> Error.Name.OlmIndexError
-            else                                               -> Error.Name.UnknownError
+            MXCryptoError.ErrorType.UNKNOWN_INBOUND_SESSION_ID,
+            MXCryptoError.ErrorType.KEYS_WITHHELD         -> Pair(detailed, Error.Name.OlmKeysNotSentError)
+            MXCryptoError.ErrorType.OLM                   -> Pair(detailed, Error.Name.OlmUnspecifiedError)
+            MXCryptoError.ErrorType.UNKNOWN_MESSAGE_INDEX -> Pair(detailed, Error.Name.OlmIndexError)
+            else                                          -> Pair(detailed, Error.Name.UnknownError)
```
Florian14, line 153: Just about the syntax, I think you can keep and store the previous code into a val `errorName` and at the end return the pair `detailed to errorName`. Otherwise, as you defined a typealias, you can probably use it instead of calling the Pair?

Florian14 (approved):

vector/src/main/java/im/vector/app/features/analytics/DecryptionFailureTracker.kt
ariskotsomitopoulos: Thanks, I refactored that part

PR #5816, opened 2022-04-21
Implement a more scoped API to the `Session` interface

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

`Session` does not directly implement the various services. Now the services are accessible with a getter. It improves the SDK api, and make it homogeneous.

A few extensions has been added to still be able to do `session.getRoom(roomId)` for instance.

Also the `search` method from the `Room` interface has been removed. It was a shortcut.

## Motivation and context

Have a nicer scoped completion when using the SDK, and make the SDK api more user friendly:

image

image

## Tests

- Smoke test

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):


fedrunov:
I don't understand purpose of session in this implementation: why would I inject `session` and call `session.roomService()` if i can just inject `roomService` and use its methods?

bmarty:
> I don't understand purpose of session in this implementation: why would I inject `session` and call `session.roomService()` if i can just inject `roomService` and use its methods?

`Session` is the current entry point to a Matrix account. It's still possible to inject app side a `RoomService`, by providing a method `providesCurrentRoomService`  into `VectorStaticModule` as per `providesCurrentSession` for instance (this is out of scope of this PR).

We can discuss and make some changes in the future when we will be supporting multi-account.

fedrunov (approved):
LGTM, thanks for explanation

PR #5835, opened 2022-04-25
Ensure `getRootSpaceSummaries()` is not called on the main thread.

bmarty:
Draft for a hotfix

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
@@ -18,6 +18,8 @@ package org.matrix.android.sdk.internal.session.space
 
 import android.net.Uri
 import androidx.lifecycle.LiveData
+import kotlinx.coroutines.withContext
+import org.matrix.android.sdk.api.MatrixCoroutineDispatchers
 import org.matrix.android.sdk.api.query.QueryStringValue
 import org.matrix.android.sdk.api.session.events.model.Event
 import org.matrix.android.sdk.api.session.events.model.EventType
@@ -64,7 +66,8 @@ internal class DefaultSpaceService @Inject constructor(
         private val stateEventDataSource: StateEventDataSource,
         private val peekSpaceTask: PeekSpaceTask,
         private val resolveSpaceInfoTask: ResolveSpaceInfoTask,
-        private val leaveRoomTask: LeaveRoomTask
+        private val leaveRoomTask: LeaveRoomTask,
+        private val coroutineDispatchers: MatrixCoroutineDispatchers,
 ) : SpaceService {
 
     override suspend fun createSpace(params: CreateSpaceParams): String {
@@ -105,8 +108,10 @@ internal class DefaultSpaceService @Inject constructor(
         return roomSummaryDataSource.getSpaceSummaries(spaceSummaryQueryParams, sortOrder)
     }
 
-    override fun getRootSpaceSummaries(): List<RoomSummary> {
-        return roomSummaryDataSource.getRootSpaceSummaries()
+    override suspend fun getRootSpaceSummaries(): List<RoomSummary> {
+        return withContext(coroutineDispatchers.io) {
+            roomSummaryDataSource.getRootSpaceSummaries()
+        }
     }
 
     override suspend fun peekSpace(spaceId: String): SpacePeekResult {
--- a/vector/src/main/java/im/vector/app/features/spaces/SpaceListViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/spaces/SpaceListViewModel.kt
@@ -212,7 +212,8 @@ class SpaceListViewModel @AssistedInject constructor(@Assisted initialState: Spa
         }
         session.coroutineScope.launch {
             orderCommands.forEach {
-                session.getRoom(it.spaceId)?.updateAccountData(RoomAccountDataTypes.EVENT_TYPE_SPACE_ORDER,
+                session.getRoom(it.spaceId)?.updateAccountData(
+                        RoomAccountDataTypes.EVENT_TYPE_SPACE_ORDER,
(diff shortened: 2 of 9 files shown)


vector/src/main/java/im/vector/app/features/spaces/SpaceListViewModel.kt
```diff
@@ -296,7 +296,7 @@ class SpaceListViewModel @AssistedInject constructor(@Assisted initialState: Spa
             communityGroups
         }
                 .execute { async ->
-                    val rootSpaces = session.spaceService().getRootSpaceSummaries()
+                    val rootSpaces = async.invoke().orEmpty().filter { it.flattenParentIds.isEmpty() }
```
bmarty, line 299: On my side this is equivalent, but I do not have any spaces with child space. I will create some

bmarty: Ok, it works as expected

Boogieman2022 (approved):

vector/src/main/java/im/vector/app/features/spaces/SpaceListViewModel.kt
```diff
@@ -296,7 +296,7 @@ class SpaceListViewModel @AssistedInject constructor(@Assisted initialState: Spa
             communityGroups
```
BillCarsonFr, line 296: should rename this to spaceList

bmarty: I agree, but I wanted to limit the number of changes for the hotfix

BillCarsonFr (approved):
LGTM, it's fixing freeze issues on my side

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
```diff
@@ -105,8 +108,10 @@ internal class DefaultSpaceService @Inject constructor(
         return roomSummaryDataSource.getSpaceSummaries(spaceSummaryQueryParams, sortOrder)
     }
 
-    override fun getRootSpaceSummaries(): List<RoomSummary> {
-        return roomSummaryDataSource.getRootSpaceSummaries()
+    override suspend fun getRootSpaceSummaries(): List<RoomSummary> {
+        return withContext(coroutineDispatchers.io) {
```
ouchadam, line 112: :+1: 

not for this PR as it would introduce lots of changes, but worth highlighting in case we want guard against other cases

from my understanding the general approach of coroutines scoping is to define the context switching at the lowest possible level, which in this case would be 

```
monarchy.fetchAllMappedSync
```
 
we already have a similar wrapper for [awaitTransaction](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/AsyncTransaction.kt#L38)

this would also force non coroutine based usages to rely on `runBlocking`, which would have the benefit of make finding the usages a bit easier

vector/src/main/java/im/vector/app/features/spaces/SpaceListViewModel.kt
```diff
@@ -278,31 +278,23 @@ class SpaceListViewModel @AssistedInject constructor(@Assisted initialState: Spa
...
-                        },
-                flowSession
+                session.flow()
                         .liveSpaceSummaries(params),
-                session
-                        .accountDataService()
+                session.accountDataService()
                         .getLiveRoomAccountDataEvents(setOf(RoomAccountDataTypes.EVENT_TYPE_SPACE_ORDER))
                         .asFlow()
-        ) { _, communityGroups, _ ->
```
ouchadam, line 295: to double check, were the other flows being used as some sort of predicate? / Do we need `.getLiveRoomAccountDataEvents(...).asFlow()`

it's a bit strange to see us submit flows to `combine` and then ignore their result :thinking:

vector/src/main/java/im/vector/app/features/spaces/SpaceListViewModel.kt
```diff
@@ -278,31 +278,23 @@ class SpaceListViewModel @AssistedInject constructor(@Assisted initialState: Spa
...
-            communityGroups
+        ) { spaces, _ ->
+            spaces
         }
                 .execute { async ->
-                    val rootSpaces = session.spaceService().getRootSpaceSummaries()
-                    val orders = rootSpaces.map {
+                    val rootSpaces = async.invoke().orEmpty().filter { it.flattenParentIds.isEmpty() }
+                    val orders = rootSpaces.associate {
```
ouchadam, line 292: :100:

vector/src/main/java/im/vector/app/features/spaces/SpaceListViewModel.kt
bmarty: It's true that we ignore the result, but we want to be notified when the account data `EVENT_TYPE_SPACE_ORDER` is updated.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
bmarty: I agree, good point!

ouchadam (approved):
from the fix perspective LGTM! :100:  should this PR point to `main` ?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/space/DefaultSpaceService.kt
ganfra: Actually we might want to dispatch querying db to a background thread, to avoid blocking the dispatcher, as Retrofit does for request (and probably Room)

bmarty:
The PR could have pointed to `main`, but it was not aimed to be merged from GH interface. We use gitflow, which will merge this branch on `main` and on `develop`

PR #5843, opened 2022-04-26
Baseline XML formatting 

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Formats XML files using the existing .editorconfig (with new file end of line temporarily disabled)

- Cleans up the XML files giving us a consistent baseline to apply changes on top of.
- Does not include any strings.xml as they're synced with weblate and could cause conflicts

## Motivation and context

To create a consistent baseline 

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):


bmarty:
I would like the CI to check about formatting, using what's in `.editorconfig`, else we will keep on not following the rules, for instance when adding a new vector drawable using the Android Studio tool. Is it possible?

ouchadam:
> I would like the CI to check about formatting, using what's in `.editorconfig`, else we will keep on not following the rules, for instance when adding a new vector drawable using the Android Studio tool. Is it possible?

`./gradlew ktlintCheck` should cover us on the kotlin side but for xml it looks like we might have to use another tool like prettier with an xml plugin https://github.com/prettier/plugin-xml

changelog.d/5843.misc
```diff
@@ -0,0 +1 @@
+Formatting XML resources to a create baseline in preparation for future changes
```
bmarty, line 1:
```suggestion
Formatting XML resources to create a baseline in preparation for future changes
```

(I may be wrong, you are the native English speaker :) )

ouchadam: :facepalm: one day I'll learn English....

bmarty:
> Does not include any strings.xml as they're synced with weblate and could cause conflicts

Are the file generated by Weblate not properly formatted? We can consider formatting them during a Weblate sync, depending on the changes.

bmarty:
OK, then I think we can format the translations during the next weblate think. You could add a TODO about that in the main `strings.xml`

ouchadam:
> OK, then I think we can format the translations during the next weblate think. You could add a TODO about that in the main `strings.xml`

742dd2c6a0eae5c807d5af466e0b2703ea2e3620

mnaturel:
> > I would like the CI to check about formatting, using what's in `.editorconfig`, else we will keep on not following the rules, for instance when adding a new vector drawable using the Android Studio tool. Is it possible?
> 
> `./gradlew ktlintCheck` should cover us on the kotlin side but for xml it looks like we might have to use another tool like prettier with an xml plugin https://github.com/prettier/plugin-xml

Maybe we could give a try to this gradle plugin? https://github.com/ec4j/editorconfig-gradle-plugin

ouchadam:
going to close, this PR will need rebasing and the formatting reapplied

Boogieman2022:
👍

PR #5853, opened 2022-04-27
Share Megolm session keys when inviting a new user 

### LINKED ISSUE #4153: Sharing room keys for past messages
From element-ios created by [manuroe](https://github.com/manuroe): vector-im/element-ios#4947  

As per [MSC3061](https://github.com/matrix-org/matrix-doc/pull/3061)

Web PRs:
- https://github.com/matrix-org/matrix-js-sdk/pull/1640
- https://github.com/matrix-org/matrix-react-sdk/pull/5763

ariskotsomitopoulos:
Closes #4153 and relates to [MSC3061](https://github.com/matrix-org/matrix-spec-proposals/pull/3061)

This PR aims to improve users experience when they are first invited to a room. Users will be able to decrypt and view previous messages

PR includes:
- Share megolm sessions extracted from the latest messages. We decided not to share the whole history for performance reasons
- Share inbound sessions to **all** user devices
- Will store an extra boolean flag `sharedHistory` in inbound Megolm sessions. That flag will determine whether or not the inbound session will be shared. That flag respects the room history visibility settings at the time: 
  `world_readable` -> sharedHistory = **true**
  `shared`                -> sharedHistory = **true**
  `invited`                -> sharedHistory = **false**
  `joined`                 ->  sharedHistory = **false**
- Will share keys appropriately even if the user is unknown (will download the appropriate keys for that user)
- Rotate session with respect to room history visibility
- Integration Tests

### Update

Added MSC3061 as a lab flag (off by default).
image

Enabling it will trigger an initial sync.

Crypto database have been migrated to stop using frozen serializable java objects for session metadata (using json now)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -1371,6 +1376,54 @@ internal class DefaultCryptoService @Inject constructor(
         }
     }
 
+    override fun sendSharedHistoryKeysToLastChunk(roomId: String, userId: String, sessionInfoSet: Set<SessionInfoPair>?) {
+        cryptoCoroutineScope.launch(coroutineDispatchers.crypto) {
+            runCatching {
+                deviceListManager.downloadKeys(listOf(userId), false)
+            }.mapCatching {
+                val userDevices = cryptoStore.getUserDevices(userId)
+                userDevices?.forEach {
```
ouchadam, line 1385: if one of the devices fails to export/send then the rest of the devices will be skipped, is that okay/expected?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -1371,6 +1376,54 @@ internal class DefaultCryptoService @Inject constructor(
...
+                        val decryptor = roomDecryptorProvider.getRoomDecryptor(roomId, algorithm)
+                        decryptor?.shareKeysWithDevice(exportedKeys, deviceId, userId)
+                        Timber.i("## CRYPTO | Sharing inbound session")
+                    }
+                }
+            }
+        }
+    }
+
+    override fun sendSharedHistoryKeys(roomId: String, userId: String) {
```
ouchadam, line 1407: the implementation of `sendSharedHistoryKeys` and `sendSharedHistoryKeysToLastChunk` looks very similar, can we share some of the logic?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -1371,6 +1376,54 @@ internal class DefaultCryptoService @Inject constructor(
...
     }
 
+    override fun sendSharedHistoryKeysToLastChunk(roomId: String, userId: String, sessionInfoSet: Set<SessionInfoPair>?) {
+        cryptoCoroutineScope.launch(coroutineDispatchers.crypto) {
+            runCatching {
+                deviceListManager.downloadKeys(listOf(userId), false)
+            }.mapCatching {
+                val userDevices = cryptoStore.getUserDevices(userId)
+                userDevices?.forEach {
+                    // Lets share the provided inbound sessions for every user device
```
ouchadam, line 1386: general comment for the comments, they look like they could be replaced with functions

```kotlin
// Get inbound session from sessionId and sessionKey
sessionInfoSet?.inboundSessions()

// Filter only sessions with sharedHistory enabled
?.onlySessionsWithSharedHistory()

// Share the session to userId with deviceId
fun shareSessionWithUserDevice(session, roomId, userId, deviceId)
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/IMXDecrypting.kt
```diff
@@ -60,6 +61,8 @@ internal interface IMXDecrypting {
      */
     fun shareKeysWithDevice(request: IncomingRoomKeyRequest) {}
 
+    fun shareKeysWithDevice(exportedKeys: MegolmSessionData?, deviceId: String, userId: String) {}
```
ouchadam, line 64: it feels a little strange to allow a nullable keys here, or is it possible to share null keys with a device?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
```diff
@@ -341,6 +342,46 @@ internal class MXMegolmDecryption(private val userId: String,
...
     }
 
+    override fun shareKeysWithDevice(exportedKeys: MegolmSessionData?, deviceId: String, userId: String) {
+        exportedKeys ?: return
+        cryptoCoroutineScope.launch(coroutineDispatchers.crypto) {
+            runCatching { deviceListManager.downloadKeys(listOf(userId), false) }
+                    .mapCatching {
+                        val deviceInfo = cryptoStore.getUserDevice(userId, deviceId)
+                        if (deviceInfo == null) {
+                            throw RuntimeException()
```
ouchadam, line 352: is it worth a dedicated exception with more information or an error code?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
```diff
@@ -341,6 +342,46 @@ internal class MXMegolmDecryption(private val userId: String,
...
+
+                            val encodedPayload = messageEncrypter.encryptMessage(payloadJson, listOf(deviceInfo))
+                            val sendToDeviceMap = MXUsersDevicesMap<Any>()
+                            sendToDeviceMap.setObject(userId, deviceId, encodedPayload)
+                            Timber.tag(loggerTag.value).i("shareKeysWithDevice() : sending ${exportedKeys.sessionId} to $userId:$deviceId")
+                            val sendToDeviceParams = SendToDeviceTask.Params(EventType.ENCRYPTED, sendToDeviceMap)
+                            try {
+                                sendToDeviceTask.execute(sendToDeviceParams)
+                            } catch (failure: Throwable) {
+                                Timber.tag(loggerTag.value).e(failure, "shareKeysWithDevice() : Failed to send ${exportedKeys.sessionId} to $userId:$deviceId")
```
ouchadam, line 378: we can avoid this inner `try/catch` and rely on the `runCatching` chain by using 

```kotlin
.mapCatching { .. }
.onFailure { 
  Timber.tag(loggerTag.value).e(failure, "shareKeysWithDevice() : Failed to send ...")
}

File: matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmInboundGroupSessionWrapper2.kt [hunk 7]
```diff
@@ -43,6 +43,10 @@ internal class OlmInboundGroupSessionWrapper2 : Serializable {
     // Devices which forwarded this session to us (normally empty).
     var forwardingCurve25519KeyChain: List<String>? = ArrayList()
 
+    // Flag that indicates whether or not the current inboundSession will be shared to
+    // invited users to decrypt past messages
+    var sharedHistory: Boolean = false
```
ouchadam, line 48: cc @bmarty does dokka support comments `//` or should we start using javadoc `/**` instead?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/DefaultMembershipService.kt
```diff
@@ -127,6 +131,10 @@ internal class DefaultMembershipService @AssistedInject constructor(
     }
 
     override suspend fun invite(userId: String, reason: String?) {
+        val sessionInfoSet = Realm.getInstance(monarchy.realmConfiguration).use {
```
ouchadam, line 134: should we only share the keys if the invite succeeds?

ariskotsomitopoulos: No we should share them even if the invite fails

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
ariskotsomitopoulos: Again, I copied the existing code style. A dedicated exception should be much better I agree

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
BillCarsonFr: I'd say it's probably ok to use cache in this case. Or we should propagate exception to caller so he can do something about it.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/helper/ChunkEntityHelper.kt
```diff
@@ -236,3 +240,14 @@ internal fun ChunkEntity.isMoreRecentThan(chunkToCheck: ChunkEntity): Boolean {
...
+internal fun ChunkEntity.Companion.findLatestSessionInfo(realm: Realm, roomId: String): Set<SessionInfoPair>? =
+        ChunkEntity.findLastForwardChunkOfRoom(realm, roomId)?.timelineEvents?.mapNotNull { timelineEvent ->
+            timelineEvent?.root?.asDomain()?.content?.toModel<EncryptedEventContent>()?.let { content ->
+                content.sessionId ?:  return@mapNotNull null
+                content.senderKey ?:  return@mapNotNull null
+                Pair(content.sessionId, content.senderKey)
+            }
+        }?.toSet()
+
+internal typealias SessionInfoPair = Pair<String, String>
```
BillCarsonFr, line 253: I think it's exposed in the service, so shouldn't be internal. Can it be a simple data class also, to avoid confusion with first/second.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
```diff
@@ -162,4 +163,9 @@ interface CryptoService {
      * send, in order to speed up sending of the message.
      */
     fun prepareToEncrypt(roomId: String, callback: MatrixCallback<Unit>)
+
+    /**
+     * Share all inbound sessions of the last chunk messages to the provided userId devices
+     */
+    fun sendSharedHistoryKeysToLastChunk(roomId: String, userId: String, sessionInfoSet: Set<SessionInfoPair>?)
```
BillCarsonFr, line 170: Can we use same name as js-sdk? would be easier when investigating/checking code.
Also it's probably possible to invite several people at once, so it could be a list of userId?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
BillCarsonFr: Can we also log something on error

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
```diff
@@ -341,6 +342,46 @@ internal class MXMegolmDecryption(private val userId: String,
         return olmDevice.hasInboundSessionKeys(roomId, senderKey, sessionId)
     }
 
+    override fun shareKeysWithDevice(exportedKeys: MegolmSessionData?, deviceId: String, userId: String) {
```
BillCarsonFr, line 345: would name that forwardKeys instead of shareKeys

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -1371,6 +1376,35 @@ internal class DefaultCryptoService @Inject constructor(
...
+                    sessionInfoSet?.mapNotNull { sessionInfoPair ->
+                        // Get inbound session from sessionId and sessionKey
+                        cryptoStore.getInboundGroupSession(
+                                sessionId = sessionInfoPair.first,
+                                senderKey = sessionInfoPair.second,
+                                sharedHistory = true
+                        )
+                    }?.forEach { inboundGroupSession ->
+                        // Share the sharable session to userId with deviceId
+                        val exportedKeys = inboundGroupSession.exportKeys(sharedHistory = true)
```
BillCarsonFr, line 1397: We need to check the roomId (that it matches the one in parameter)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
BillCarsonFr: Oh we are doing that in existing code? feels bad

BillCarsonFr: Also there is no indication that this method throws, could end up crashing somewhere

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/helper/ChunkEntityHelper.kt
```diff
@@ -236,3 +240,14 @@ internal fun ChunkEntity.isMoreRecentThan(chunkToCheck: ChunkEntity): Boolean {
     // We don't know, so we assume it's false
     return false
 }
+
+internal fun ChunkEntity.Companion.findLatestSessionInfo(realm: Realm, roomId: String): Set<SessionInfoPair>? =
+        ChunkEntity.findLastForwardChunkOfRoom(realm, roomId)?.timelineEvents?.mapNotNull { timelineEvent ->
+            timelineEvent?.root?.asDomain()?.content?.toModel<EncryptedEventContent>()?.let { content ->
+                content.sessionId ?:  return@mapNotNull null
+                content.senderKey ?:  return@mapNotNull null
```
BillCarsonFr, line 248: This field is in the process of beeing phased out https://github.com/matrix-org/matrix-spec-proposals/pull/3700
Refrain from using it

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -953,6 +957,7 @@ internal class DefaultCryptoService @Inject constructor(
         val eventContent = event.content.toModel<RoomHistoryVisibilityContent>()
         eventContent?.historyVisibility?.let {
             cryptoStore.setShouldEncryptForInvitedMembers(roomId, it != RoomHistoryVisibility.JOINED)
+            cryptoStore.setShouldShareHistory(roomId, it.shouldShareHistory())
```
BillCarsonFr, line 960: I'd a bit paranoid here and put this to false if eventContent?.historyVisibility is null

BillCarsonFr (changes requested):
Thx for the PR 👏
Please see my comments. Also it would be good to add a bit more log (.debug level to keep track on what hapenned in RS, and .verbose to help debugging).
I think the part in MXMegolmDecryption would need some rework because I refactored part of it in my latest PR, we could check that together when it will be merged.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/CryptoService.kt
ariskotsomitopoulos: It will be triggered for every userId, so there is no need for a list here

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/algorithms/megolm/MXMegolmDecryption.kt
ouchadam: I'm happy with either but would lean towards avoiding a mix (at least within the same function block)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/membership/DefaultMembershipService.kt
ariskotsomitopoulos: Yes its a good point. Though, we have already discussed it with the crypto team and sharing prior to a user invite is accepted. It has more pros than cons like some performance benefits. Also keep in mind that the proper/robust implementation for that feature will be implemented through the rust SDK. So it's not a permanent solution.

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/E2eeSanityTests.kt
```diff
@@ -145,16 +145,15 @@ class E2eeSanityTests : InstrumentedTest {
             delay(3_000)
         }
 
-        // check that messages are encrypted (uisi)
+        // Due to the new shared keys implementation, invited user should be able to decrypt messages
```
BillCarsonFr, line 148: I would to that as a separate test?
I still want to test that in normal scenario you don't get the past history. Two tests would be better

ariskotsomitopoulos: There is also a separate test for that. If we do not change that the test will fail while keys will be shared, so user will be able to decrypt.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/DefaultCryptoService.kt
```diff
@@ -953,7 +957,8 @@ internal class DefaultCryptoService @Inject constructor(
         val eventContent = event.content.toModel<RoomHistoryVisibilityContent>()
         eventContent?.historyVisibility?.let {
             cryptoStore.setShouldEncryptForInvitedMembers(roomId, it != RoomHistoryVisibility.JOINED)
-        }
+            cryptoStore.setShouldShareHistory(roomId, it.shouldShareHistory())
+        } ?: cryptoStore.setShouldShareHistory(roomId, false)
```
BillCarsonFr, line 961: I am not sure of the syntax. if setShouldEncryptForInvitedMembers signature change we could end up calling both test and elvis?

BillCarsonFr:
This PR https://github.com/vector-im/element-android/pull/5559 should be merged first

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmInboundGroupSessionWrapper2.kt
bmarty: Dokka only understand `/**`. I have added a note about that for the coming weekly sync.
But note the `Dokka` is not generating doc for internal classes, so not a big deal here.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/helper/ChunkEntityHelper.kt
bmarty: if it can be `internal`, the compiler will complain.

BillCarsonFr:
Pushed an update with changed needed to have polyjuice test passing.
`org.matrix.msc3061.shared_history` was not properly transmitted and use from/to `m.room_key` and `m.forwarded_room_key`

Migrated the `OlmInboundGroupSessionWrapper2`serializable class (frozen) to a `MXInboundMegolmSessionWrapper` composed of the actual session (olm) and a data class serialized in json.

Moved things around a bit and rebased after latest SDK PR #5492 
Updated the test to use less internal access (crypto store)

@ariskotsomitopoulos I think we should review the way we get the last messages of the timeline. We could add a new API in timeline service. This could be usefull for other features, like redact the n last message of someone (when moderating)

BillCarsonFr:
@ariskotsomitopoulos One last thing to add is the export/import to backup and file:
> the session_data field in key backups of this key has a shared_history property set to true in the decrypted JSON structure, and the SessionData type used in key exports has a shared_history property that is set to true for this key.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/crypto/MXCryptoConfig.kt
```diff
@@ -38,4 +38,10 @@ data class MXCryptoConfig constructor(
          * You can limit request only to your sessions by turning this setting to `true`
          */
         val limitRoomKeyRequestsToMyDevices: Boolean = false,
+
+        /**
+         * Flag that indicates whether or not key history will be shared to invited
+         * users with respect to room visibility
+         */
+        val shouldShareKeyHistory: Boolean = true,
```
BillCarsonFr, line 46: The flag should be used also when sending room_key, forwarded_room_key as well as when uploading to backup.

BillCarsonFr (approved):

rubo77:
> * We decided not to share the whole history for performance

Is there a way to enable sharing the whole history? On my server this performance is not so important

filipesmedeiros:
Just to make sure, any updates on this? The Labs feature is disabled for be, not sure exactly why. Are there plans to move this forward? Are there any replacements in the meantime?

Thanks so much for your work! ❤️

PR #5856, opened 2022-04-27
Use fixed text size in read receipt counter

### LINKED ISSUE #5367: Read by "+(X)" appear cut off in larger fonts
The read by +(Number of people who read the message) appear cut off in the next font sizes

- Larger
- Largest
- Huge

Also in "Large" font size the icon is very close to the inferior border

Screenshot_20220225-210751_Element_1
*Large*

Screenshot_20220225-210248_Element_1
*Larger*

Screenshot_20220225-210713_Element_1
*Largest*

Screenshot_20220225-210725_Element_1
*Huge*

Element version: 1.4.2
Android version: 11
Device: Galaxy A51

ofalvai:
## Type of change

- [ ] Feature
- [X] Bugfix: Fixes #5367 
- [ ] Technical
- [ ] Other :

## Content

Use a fixed `12dp` text size for the read recept counter number instead of `12sp` that would scale with the font scale setting. It's important that text should fit inside the circle and the circle has a fixed size regardless of the font scale.

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot_20220427-201459|Screenshot_20220427-201526|

Screenshots taken on "Largest" font setting.

## Tests

- Use the font scale setting to set something above "Normal"
- Visit a room timeline with more than 5 read receipts

## Tested devices

- [X] Physical
- [X] Emulator
- OS version(s): 11

## Checklist

- [X] Accessibility has been taken into account. See https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#accessibility
- [X] Pull request includes screenshots or videos if containing UI changes

Signed-off-by: Olivér Falvai <ofalvai@gmail.com>

### DIFF
--- a/vector/src/main/res/layout/view_read_receipts.xml
+++ b/vector/src/main/res/layout/view_read_receipts.xml
@@ -7,7 +7,7 @@
 
     <TextView
         android:id="@+id/receiptMore"
-        style="@style/Widget.Vector.TextView.Caption"
+        style="@style/TimelineFixedSizeCaptionStyle"
         android:layout_width="wrap_content"
         android:layout_height="@dimen/item_event_message_state_size"
         android:background="@drawable/pill_receipt"
@@ -16,8 +16,6 @@
         android:paddingStart="4dp"
         android:paddingEnd="4dp"
         android:textColor="?vctr_content_primary"
-        android:textSize="12dp"
-        tools:ignore="SpUsage"
         tools:text="999+" />
 
     <ImageView
--- a/vector/src/main/res/layout/view_read_receipts.xml
+++ b/vector/src/main/res/layout/view_read_receipts.xml
@@ -16,6 +16,8 @@
         android:paddingStart="4dp"
         android:paddingEnd="4dp"
         android:textColor="?vctr_content_primary"
+        android:textSize="12dp"
+        tools:ignore="SpUsage"
         tools:text="999+" />
 
     <ImageView
--- a/changelog.d/5856.bugfix
+++ b/changelog.d/5856.bugfix
@@ -1,0 +1,1 @@
+Use fixed text size in read receipt counter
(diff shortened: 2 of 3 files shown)


vector/src/main/res/layout/view_read_receipts.xml
```diff
@@ -16,6 +16,8 @@
         android:paddingStart="4dp"
         android:paddingEnd="4dp"
         android:textColor="?vctr_content_primary"
+        android:textSize="12dp"
```
ouchadam, line 19: to double check, the container views in this area look like they're `wrap_content` and the drawable has no hardcoded heights which means removing the text view's [fixed height](https://github.com/vector-im/element-android/pull/5856/files#diff-b4585bb9d47cfa459cbc36c3397fcfbfe131ce5e790ec7865baaf8d24d89b68dR12) might be enough :crossed_fingers: 

it has the benefit of allowing us to remain friendly to large text size users

ofalvai: @ouchadam Just to double-check, you'd like the TextView to be taller than the avatar ImageViews? (I'm fine with it)

ouchadam: Yep! Ideally the profile pictures would also scale with the text size by avoiding the hardcoded sizes and constraining to the height of the `TextView` itself but that could be another ticket/second iteration.

For the original issue symptom of clipped text I would lean towards it being better to have the text readable for a user with large fonts enabled at the cost of some UI imperfections, but not a strong opinion since this particular text isn't critical to the page

bmarty: My 2 cents: I agree with the original fix, since having a big font here is maybe not necessary for a11y.

Maybe a better approach could be to enlarge the Read Receipt avatar too, which is probably a better solution. We can have avatar size in `sp` to do that.

ouchadam (approved):
I'm happy with the current fix, there's always room to iterate the large font experience later 

will avoid blocking from my side!

bmarty:
Looks like we have to add an exception for this error:
`❌ Error: 'android:textSize' detected 1 time, in file './vector/src/main/res/layout/dialog_select_text_size.xml', at line 85.`

Or maybe better create a dedicated style with size in `dp`

bmarty (approved):
Thanks for the update!

PR #5866, opened 2022-04-28
Fix sanity test

bmarty:
image

(does not deserve a changelog I think)

Room header with one room member:

image

Previously only the video call was displayed


yostyle (approved):
LGTM

PR #5873, opened 2022-04-29
Rename OlmInboundGroupSessionWrapper2 to OlmInboundGroupSessionWrappe…

ariskotsomitopoulos:
Rename `OlmInboundGroupSessionWrapper2` to `OlmInboundGroupSessionWrapper`
Rename  `OlmInboundGroupSessionWrapper` to `OlmInboundGroupSessionWrapperOld` 

Improve the readability of the code and prevent developers from using the wrong wrapper

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmInboundGroupSessionWrapper.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmInboundGroupSessionWrapper.kt
@@ -48,15 +48,12 @@ internal class OlmInboundGroupSessionWrapper : Serializable {
      */
     val firstKnownIndex: Long?
         get() {
-            if (null != olmInboundGroupSession) {
-                try {
-                    return olmInboundGroupSession!!.firstKnownIndex
-                } catch (e: Exception) {
-                    Timber.e(e, "## getFirstKnownIndex() : getFirstKnownIndex failed")
-                }
+            return try {
+                olmInboundGroupSession?.firstKnownIndex
+            } catch (e: Exception) {
+                Timber.e(e, "## getFirstKnownIndex() : getFirstKnownIndex failed")
+                null
             }
-
-            return null
         }
 
     /**
@@ -77,6 +74,10 @@ internal class OlmInboundGroupSessionWrapper : Serializable {
         }
     }
 
+    constructor() {
+        // empty
+    }
+
     /**
      * Create a new instance from the provided keys map.
      *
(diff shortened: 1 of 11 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmInboundGroupSessionWrapperOld.kt
```diff
@@ -26,7 +26,7 @@ import java.io.Serializable
  * This class adds more context to a OlmInboundGroupSession object.
  * This allows additional checks. The class implements Serializable so that the context can be stored.
  */
-internal class OlmInboundGroupSessionWrapper2 : Serializable {
+internal class OlmInboundGroupSessionWrapperOld : Serializable {
```
ouchadam, line 29: not a blocker, my personal preference would be for `*Legacy` instead of `*Old`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmInboundGroupSessionWrapper.kt
```diff
@@ -138,14 +146,11 @@ internal class OlmInboundGroupSessionWrapper : Serializable {
      * @return the exported data
      */
     fun exportSession(messageIndex: Long): String? {
```
ouchadam, line 148: I wonder if it's worth using the `orNull` convention the standard lib uses for this case `exportSessionOrNull()`

what do you think?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/model/OlmInboundGroupSessionWrapperOld.kt
```diff
@@ -119,18 +115,14 @@ internal class OlmInboundGroupSessionWrapper2 : Serializable {
...
             MegolmSessionData(
                     senderClaimedEd25519Key = keysClaimed?.get("ed25519"),
-                    forwardingCurve25519KeyChain = forwardingCurve25519KeyChain?.toList().orEmpty(),
+                    forwardingCurve25519KeyChain = ArrayList(forwardingCurve25519KeyChain!!),
                     senderKey = senderKey,
                     senderClaimedKeys = keysClaimed,
                     roomId = roomId,
-                    sessionId = safeOlmInboundGroupSession.sessionIdentifier(),
-                    sessionKey = safeOlmInboundGroupSession.export(wantedIndex),
+                    sessionId = olmInboundGroupSession!!.sessionIdentifier(),
```
ouchadam, line 124: :100: it will be interesting to see if we were swallowing errors by returning null earlier without any logging

BillCarsonFr (changes requested):
I am not sure this would work. I think that OlmInboundGroupSessionEntity would throw a bad class exception when trying to deserialize the old class using the new one.

Maybe we should get rid of the java serialization and do a proper export() of the session and store that in the Entity (instead of serialized class). And see what is missing and what we should add also to the entity, and convert all in a database migration.

But I think this might impact the export code @ganfra has been doing. Ganfra do you think it's worth it?

ouchadam (approved):
some tiny naming remarks but not blockers, assuming the empty constructor isn't needed LGTM!

EDIT: just saw @BillCarsonFr 's comment about the java deserializer

with a serialVersionUid set, the name change should still work? :thinking:

ariskotsomitopoulos:
@BillCarsonFr @ouchadam Thanks for your comments but after some investigation I will drop the PR. Dropping the the serialisation completely as Valere suggests and move it to Entity seems the best approach 

@ouchadam with a `serialVersionUid` set will still not work while it will have a different id than the previous version. I tested it. :/

PR #5886, opened 2022-05-02
Fix UISIDetector grace period bug

BillCarsonFr:
## Type of change

- [x] Bugfix

## Content

Grace period for UISIDetection was not working properly, causing errors to be reported even if the event was decrypted during grace period.

Added Unit test on UISIDetector.

### DIFF
--- a/vector/src/test/java/im/vector/app/features/crypto/UISIDetectorTest.kt
+++ b/vector/src/test/java/im/vector/app/features/crypto/UISIDetectorTest.kt
@@ -38,7 +38,8 @@ class UISIDetectorTest {
 
     @Test
     fun `trigger detection after grace period`() {
-        val uisiDetector = UISIDetector()
+        val gracePeriod = 5_000L
+        val uisiDetector = UISIDetector(gracePeriod)
         var detectedEvent: E2EMessageDetected? = null
 
         uisiDetector.callback = object : UISIDetector.UISIDetectorCallback {
@@ -60,7 +61,7 @@ class UISIDetectorTest {
         uisiDetector.onEventDecryptionError(event, fakeCryptoError())
 
         runBlocking {
-            delay(40_000)
+            delay((gracePeriod * 1.2).toLong())
         }
         Assert.assertEquals(eventId, detectedEvent?.eventId)
     }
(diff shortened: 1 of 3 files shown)


vector/src/test/java/im/vector/app/features/crypto/UISIDetectorTest.kt
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+
+        // report a decryption error
+        val eventId = "0001"
+        val event = fakeEncryptedEvent(eventId, "s1", "r1")
+        uisiDetector.onEventDecryptionError(event, fakeCryptoError())
+
+        runBlocking {
+            delay(40_000)
```
bmarty, line 63: The test will last at least 40 seconds? I think there is a way for test case like this much faster with coroutines (sort of fake time / accelerated time)

vector/src/test/java/im/vector/app/features/crypto/UISIDetectorTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        uisiDetector.onEventDecryptionError(event, fakeCryptoError())
+
+        // the grace period is 30s
+        scope.launch(Dispatchers.Default) {
+            delay(10_000)
+            uisiDetector.onEventDecrypted(event, emptyMap())
+        }
+
+        runBlocking {
+            delay(60_000)
```
bmarty, line 97: same

vector/src/test/java/im/vector/app/features/crypto/UISIDetectorTest.kt
BillCarsonFr: The grace period is 30s. I could make it configurable and then reduce the test duration, but is it really an issue if it's taking 40s?

bmarty: we want test (and so the CI) to be fast!

BillCarsonFr: I made the grace period configurable, and reduced the test duration.

vector/src/test/java/im/vector/app/features/crypto/UISIDetectorTest.kt
BillCarsonFr: done

bmarty (approved):
Thanks for the update.

Florian14 (approved):
LGTM

PR #5854, opened 2022-04-27
Documentation update

bmarty:
Improve the documentation of the project regarding default language.
Small cleanup
Add headers to SDK documentation. It will render like this:

image

(sadly I did not find a way to include links)

PR #5887, opened 2022-05-02
Improve Matrix class

bmarty:
Small change in the Matrix class: deprecated methods have been removed and the constructor is now public. Also the fun `workerFactory()` has been renamed to `getWorkerFactory()`

Documentation has been improved to.

Knit has been set up, also on the CI. For now we are not doing a big usage of it, but can be more useful in the future.

Also generated documentation header will have working link:

image

(see previous version without links in #5854)

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
@@ -43,6 +43,9 @@ import javax.inject.Inject
 /**
  * This is the main entry point to the matrix sdk.
  * <br/>
+ *
+ * The constructor creates a new instance of Matrix, it's recommended to manage this instance as a singleton.
+ *
  * @param context the application context
  * @param matrixConfiguration global configuration that will be used for every [org.matrix.android.sdk.api.session.Session]
  */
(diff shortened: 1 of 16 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/Matrix.kt
```diff
@@ -61,85 +64,70 @@ class Matrix private constructor(context: Context, matrixConfiguration: MatrixCo
...
+    /**
+     * Return the User Agent used for any request that the SDK is making to the homeserver.
+     * There is no way to change the user agent at the moment
+     */
     fun getUserAgent() = userAgentHolder.userAgent
 
+    /**
+     * Return the AuthenticationService
+     */
     fun authenticationService() = authenticationService
```
ouchadam, line 89: not for this PR, we have a mix of `get${object}` and `${object}`, is it something we want to align on in the public api?

bmarty: Yes, we have a mix, the (un-official!) rule so far is to not use the `get` prefix for the SDK services, and I think this is the case for all services, to avoid having long name for the methods.

Happy to discuss this more, and/or to change the rule.

I have renamed `workerFactory()` because this is not a SDK service.

ouchadam (approved):

Claire1817 (approved):

michaelkaye:
https://github.com/vector-im/element-android/runs/6450886478?check_suite_focus=true#step:8:392

I think this has caused a regression in the integration tests (this also happened to me locally) - looks like we've changed how we instantiate the TestMatrix, and the android WorkManagerImpl is complaining.

michaelkaye:
```
        val delegate = WorkManagerImpl(
                context,
                configuration,
                WorkManagerTaskExecutor(configuration.taskExecutor)
        )
        WorkManagerImpl.setDelegate(delegate)
```

Might work (so using the unit test delegate thing to handle it) to avoid repeated initialization BUT, it's still a global and I think this means we need to prevent tests running in parallel otherwise race conditions will happen.

PR #5888, opened 2022-05-02
Add a final question to know if the issue submitter will provide a PR

bmarty:
Small improvement on the issue template.

Inspired from https://github.com/Kotlin/dokka/blob/master/.github/ISSUE_TEMPLATE/bug.md

@kittykat WDYT?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -73,3 +73,14 @@ body:
         - 'No'
     validations:
       required: true
+  - type: dropdown
+    id: pr
+    attributes:
+      label: Are you willing to provide a PR?
+      description:  |
+        Providing a PR can drastically speed up the process of fixing this bug. Don't worry, it's still OK to answer 'No' :).
+      options:
+        - 'Yes'
+        - 'No'
+    validations:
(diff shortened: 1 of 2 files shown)


onurays (approved):
LGTM! (This seems like a very good idea to me)

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -73,3 +73,14 @@ body:
...
+    id: pr
+    attributes:
+      label: Are you willing to provide a PR?
+      description:  |
+        Providing a PR can drastically speed up the process of fixing this bug. Don't worry, it's still OK to answer 'No' :).
+      options:
+        - 'Yes'
+        - 'No'
+    validations:
+      required: true
```
kittykat, line 86: What is your logic for making this a required question?

bmarty: Was just copy pasting the value from the question above `Will you send logs?`. To me this is the same level of requirement, and I guess we want the issue submitter to answer every questions.

But happy to set this one to `false` if you prefer.

bmarty:
Let's merge and iterate if we want to.

PR #5911, opened 2022-05-03
Screen sharing over WebRTC

onurays:
screen_sharing_pr

1. Start a video call between 2 or more devices
2. Once the call is connected tap the menu icon (...)
3. You should see the "Start sharing screen" option, select it
4. A system permission dialog should appear, accept it
5. Your screen will be streamed to other devices (in ~5 seconds)
6. A sticky push notification should appear for Android 10 or later versions
7. "Start sharing screen" button should turn to "Stop sharing screen"

### DIFF
--- a/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
@@ -651,7 +651,7 @@ class VectorCallActivity : VectorBaseActivity<ActivityCallBinding>(), CallContro
     private fun startScreenSharing(activityResult: ActivityResult) {
         val videoCapturer = ScreenCapturerAndroid(activityResult.data, object : MediaProjection.Callback() {
             override fun onStop() {
-                Timber.v("User revoked the screen capturing permission")
+                Timber.i("User revoked the screen capturing permission")
             }
         })
         callViewModel.handle(VectorCallViewActions.StartScreenSharing(videoCapturer))
--- a/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
@@ -164,6 +164,9 @@ class VectorCallActivity : VectorBaseActivity<ActivityCallBinding>(), CallContro
                 }
             }
         }
+
+        // Bind to service in case of user killed the app while there is an ongoing call
+        bindToScreenCaptureService()
     }
 
     override fun onNewIntent(intent: Intent?) {
@@ -662,9 +665,13 @@ class VectorCallActivity : VectorBaseActivity<ActivityCallBinding>(), CallContro
                 this,
                 Intent(this, ScreenCaptureService::class.java)
         )
+        bindToScreenCaptureService(activityResult)
+    }
+
+    private fun bindToScreenCaptureService(activityResult: ActivityResult? = null) {
         screenCaptureServiceConnection.bind(object : ScreenCaptureServiceConnection.Callback {
             override fun onServiceConnected() {
-                startScreenSharing(activityResult)
+                activityResult?.let { startScreenSharing(it) }
             }
         })
     }
--- a/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
@@ -24,6 +24,7 @@ import android.content.Intent
 import android.content.Intent.FLAG_ACTIVITY_CLEAR_TOP
 import android.content.res.Configuration
 import android.graphics.Color
+import android.media.projection.MediaProjection
 import android.media.projection.MediaProjectionManager
 import android.os.Build
 import android.os.Bundle
@@ -32,6 +33,7 @@ import android.util.Rational
 import android.view.MenuItem
 import android.view.View
 import android.view.WindowManager
+import androidx.activity.result.ActivityResult
 import androidx.annotation.StringRes
 import androidx.core.content.ContextCompat
 import androidx.core.content.getSystemService
@@ -76,6 +78,7 @@ import org.matrix.android.sdk.api.session.call.TurnServerResponse
 import org.matrix.android.sdk.api.session.room.model.call.EndCallReason
 import org.webrtc.EglBase
 import org.webrtc.RendererCommon
+import org.webrtc.ScreenCapturerAndroid
 import timber.log.Timber
 import javax.inject.Inject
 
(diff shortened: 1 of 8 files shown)


vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
```diff
@@ -636,18 +639,36 @@ class VectorCallActivity : VectorBaseActivity<ActivityCallBinding>(), CallContro
...
+            } else {
+                startScreenSharing(activityResult)
             }
         }
     }
 
+    private fun startScreenSharing(activityResult: ActivityResult) {
+        val videoCapturer = ScreenCapturerAndroid(activityResult.data, object : MediaProjection.Callback() {
+            override fun onStop() {
+                Timber.v("User revoked the screen capturing permission")
```
ericdecanini, line 654: If we're keeping this in production I would make this `Timber.i`

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
```diff
@@ -770,12 +775,37 @@ class WebRtcCall(
...
+
+        // Show screen preview locally
+        localVideoTrack = factory.createVideoTrack(VIDEO_TRACK_ID, videoSource).apply { setEnabled(true) }
+        localMediaStream?.addTrack(localVideoTrack)
+        localSurfaceRenderers.forEach { it.get()?.let { localVideoTrack?.addSink(it) } }
+
+        // Remove camera stream
+        peerConnection?.removeTrack(videoSender)
+
+        screenSender = peerConnection?.addTrack(localVideoTrack, listOf(STREAM_ID))
```
ericdecanini, line 802: Each of these "regions" (as separated by the comments) can be made more readable by each being their own private function. This method can then read like a story without any comments

vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
onurays: Done. Yes, we have to implement this dummy callback.

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
onurays: Done. This class needs refactoring.

vector/src/main/java/im/vector/app/features/call/VectorCallViewModel.kt
```diff
@@ -147,6 +147,7 @@ class VectorCallViewModel @AssistedInject constructor(
                     setState { copy(otherKnownCallInfo = null) }
                 }
             }
+            _viewEvents.post(VectorCallViewEvents.StopScreenSharingService)
```
mnaturel, line 150: Here I think the set of the `isSharingScreen` to false in the state is missing. Maybe we should reuse an mutualize what is done in `handleToggleScreenSharing()` when screen is shared:
```
if (isSharingScreen) {
    call?.stopSharingScreen()
    setState {
        copy(isSharingScreen = false)
    }
    _viewEvents.post(VectorCallViewEvents.StopScreenSharingService)
}
```
What do you think?

vector/src/main/java/im/vector/app/features/call/webrtc/ScreenCaptureServiceConnection.kt
```diff
@@ -27,10 +27,17 @@ class ScreenCaptureServiceConnection @Inject constructor(
...
+    interface Callback {
+        fun onServiceConnected()
+    }
+
     private var isBound = false
     private var screenCaptureService: ScreenCaptureService? = null
+    private var callback: Callback? = null
+
+    fun bind(callback: Callback) {
+        this.callback = callback
```
mnaturel, line 39: I think it would be better to reset the callback to `null` when unbound to the service in `onServiceDisconnected` for example.

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
```diff
@@ -770,12 +775,49 @@ class WebRtcCall(
         return currentCaptureFormat
     }
 
-    fun startSharingScreen() {
-        // TODO. Will be handled within the next PR.
+    fun startSharingScreen(videoCapturer: VideoCapturer) {
```
mnaturel, line 778: Have you tried to extract the screen sharing capabilities into a dedicated component to avoid adding new responsabilities to `WebRtcCall`?

mnaturel (commented):
When testing on a real device (Nokia 7.2, Android 11), it seems the sharing is not working. I click on share screen then I see a dialog saying I will share all the content of my screen. And then I see a notification saying a screen sharing is running. But the sharing option stays with "Share my screen" and not "Stop sharing" like I would expect. And I don't see the share on Web. It seems like it does not transfer the video stream.

Also, could you please add a bit more of context and explanation in the PR description (for example how to test, to see what is expected).

vector/src/main/java/im/vector/app/features/call/VectorCallViewModel.kt
onurays: Done, although we finish the call activity as soon as the call ended.

vector/src/main/java/im/vector/app/features/call/webrtc/ScreenCaptureServiceConnection.kt
onurays: Right, done.

vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt
onurays: I don't think it is easily possible. This class has many local fields. We will start implementing Element Call soon, so we probably won't invest in this class more.

onurays:
> When testing on a real device (Nokia 7.2, Android 11), it seems the sharing is not working. I click on share screen then I see a dialog saying I will share all the content of my screen. And then I see a notification saying a screen sharing is running. But the sharing option stays with "Share my screen" and not "Stop sharing" like I would expect. And I don't see the share on Web. It seems like it does not transfer the video stream.
> 
> Also, could you please add a bit more of context and explanation in the PR description (for example how to test, to see what is expected).

Thank you for testing by turning off the camera :) I will fix this.

mnaturel:
I tested again with physical device (Android 11, Nokia 7.2), here is what I discovered:
- in audio call, when I approach my fingers from the front camera or status bar, the screen becomes black (I guess it behaves like this is several devices since you are supposed to talk to the phone). It may be annoying when screen sharing is running, don't know if we can do something about it.
- in audio call, after stopping a screen sharing, the last image stays in the call preview in the web. I don't know if the problem comes from our side. It does not occur in a video call (even if we disable the video after the video call has started)

onurays:
> I tested again with physical device (Android 11, Nokia 7.2), here is what I discovered:
> 
> * in audio call, when I approach my fingers from the front camera or status bar, the screen becomes black (I guess it behaves like this is several devices since you are supposed to talk to the phone). It may be annoying when screen sharing is running, don't know if we can do something about it.
> * in audio call, after stopping a screen sharing, the last image stays in the call preview in the web. I don't know if the problem comes from our side. It does not occur in a video call (even if we disable the video after the video call has started)

Proximity sensor issue is a good catch and fixed. Second one seems the responsibility of web, because we stop sending streams.

mnaturel:
Thanks for fixing the proximity sensor issue, it is okay now. I discovered a new thing:
- in a video call, I start sharing screen. When I disabled the video it disable the screen sharing as well. I think the user expects to only turn off the camera and not both camera and screen sharing.

mnaturel (approved):

Issue #5724, opened 2022-04-08
Client should notify others of the accepted verification request


langleyd:
Similar to https://github.com/vector-im/element-ios/issues/4470

The Android client should notify other devices when one accepts a verification request. All other devices (other than the current device and the accepting device should cancel the existing verification request prompts)

Reproduced this issue on the android client.

Steps:
1. Login + verify on a web client with alice
2. Login + verify on a second web client with alice
3. Login on Android with alice and click "verify this login"
4.  2 verification requests appear, 1 on each web client.
5. Accept 1 request.
Expected: The request on the second web client should be dismissed
Observed: The request on the second web client is not dismissed.

PR #5916, opened 2022-05-03
Notify other devices of acceptance of verification request

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

When a verification request sent from a client is answered, this same client should notify other clients that verification is no longer necessary.

## Motivation and context

Issue: #5724 .

## Tests

Using the same account:

- Log into a client, verify session (I used the web client).
- Log into another client.
- Log into the Android client.
- A 'verify your session' banner should appear, tap on it.
- On one of the other clients, accept the verification request.
- Check that the verification popup disappears on the other client.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 10

## Checklist

- [ ] Accessibility has been taken into account.
- [x] Pull request includes a new file under ./changelog.d.
- [ ] Pull request includes a [sign off]

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/CancelCode.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/CancelCode.kt
@@ -28,7 +28,8 @@ enum class CancelCode(val value: String, val humanReadable: String) {
     MismatchedKeys("m.key_mismatch", "Key mismatch"),
     UserError("m.user_error", "User error"),
     MismatchedUser("m.user_mismatch", "User mismatch"),
-    QrCodeInvalid("m.qr_code.invalid", "Invalid QR code")
+    QrCodeInvalid("m.qr_code.invalid", "Invalid QR code"),
+    AcceptedByAnotherDevice("m.accepted", "Verification request accepted by another device")
 }
 
 fun safeValueOf(code: String?): CancelCode {
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
@@ -939,9 +939,25 @@ internal class DefaultVerificationService @Inject constructor(
 
         updatePendingRequest(
                 existingRequest.copy(
-                        readyInfo = readyReq
+                    readyInfo = readyReq
                 )
         )
+
+        notifyOthersOfAcceptance(readyReq.transactionId, readyReq.fromDevice)
+    }
+
+    /**
+     * Gets a list of device ids excluding the current one.
+     */
+    private fun getMyOtherDeviceIds(): List<String> = cryptoStore.getUserDevices(userId)?.keys?.filter { it != deviceId }.orEmpty()
+
+    /**
+     * Notifies other devices that the current verification transaction is being handled by [acceptedByDeviceId].
+     */
+    private fun notifyOthersOfAcceptance(transactionId: String, acceptedByDeviceId: String) {
+        val deviceIds = getMyOtherDeviceIds().filter { it != acceptedByDeviceId }
+        val transport = verificationTransportToDeviceFactory.createTransport(null)
+        transport.cancelTransaction(transactionId, userId, deviceIds, CancelCode.AcceptedByAnotherDevice)
     }
 
     private fun createQrCodeData(requestId: String?, otherUserId: String, otherDeviceId: String?): QrCodeData? {
--- a/changelog.d/5724.sdk
+++ b/changelog.d/5724.sdk
@@ -1,0 +1,1 @@
+- Notifies other devices when a verification request sent from an Android device is accepted.`
--- a/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/verification/qrcode/VerificationTest.kt
+++ b/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/verification/qrcode/VerificationTest.kt
@@ -27,11 +27,13 @@ import org.matrix.android.sdk.api.auth.UIABaseAuth
 import org.matrix.android.sdk.api.auth.UserInteractiveAuthInterceptor
 import org.matrix.android.sdk.api.auth.UserPasswordAuth
 import org.matrix.android.sdk.api.auth.registration.RegistrationFlowResponse
+import org.matrix.android.sdk.api.session.crypto.verification.CancelCode
 import org.matrix.android.sdk.api.session.crypto.verification.PendingVerificationRequest
 import org.matrix.android.sdk.api.session.crypto.verification.VerificationMethod
 import org.matrix.android.sdk.api.session.crypto.verification.VerificationService
 import org.matrix.android.sdk.common.CommonTestHelper
 import org.matrix.android.sdk.common.CryptoTestHelper
+import org.matrix.android.sdk.common.SessionTestParams
 import org.matrix.android.sdk.common.TestConstants
 import java.util.concurrent.CountDownLatch
 import kotlin.coroutines.Continuation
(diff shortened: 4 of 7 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/CancelCode.kt
```diff
@@ -28,7 +28,8 @@ enum class CancelCode(val value: String, val humanReadable: String) {
     MismatchedKeys("m.key_mismatch", "Key mismatch"),
     UserError("m.user_error", "User error"),
     MismatchedUser("m.user_mismatch", "User mismatch"),
-    QrCodeInvalid("m.qr_code.invalid", "Invalid QR code")
+    QrCodeInvalid("m.qr_code.invalid", "Invalid QR code"),
+    Accepted("m.accepted", "Verification request accepted by another device")
```
ganfra, line 32: Naming should probably reflect it has been accepted on another device, otherwise it's a bit weird

ganfra (commented):
I don't see anything wrong here, but I defect to Valere to be sure

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
```diff
@@ -927,6 +927,24 @@ internal class DefaultVerificationService @Inject constructor(
...
                 readyInfo = readyReq
         ))
+
+        notifyOthersOfAcceptance(readyReq.transactionId, senderId, readyReq.fromDevice)
+    }
+
+    /**
+     * Gets a list of device ids excluding the current one.
+     */
+    private fun getOtherDeviceIds(): List<String> = cryptoStore.getUserDevices(userId)?.keys?.filter { it != deviceId }.orEmpty()
```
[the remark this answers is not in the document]
langleyd: Yea as you mention the iOS client gets the other device ids with a `forceDownload` flag set to true. I really just re-used that function as it was conveniently in the same class. In the other case it is called at the start of the verification flow. So using the `forceDownload` flag when we are doing the `m.accepted` probably ins't necessary( as we have just downloaded). If android also refreshes at the start of the flow then I'm guessing you probably don't need the refresh. @BillCarsonFr may have more context.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
```diff
@@ -927,6 +927,24 @@ internal class DefaultVerificationService @Inject constructor(
...
+
+    /**
+     * Notifies other devices that the current verification transaction is being handled by [fromDeviceId].
+     */
+    private fun notifyOthersOfAcceptance(transactionId: String, otherUserId: String, fromDeviceId: String) {
+        val deviceIds = getOtherDeviceIds().filter { it != fromDeviceId }
+        val transport = verificationTransportToDeviceFactory.createTransport(null)
+        deviceIds.forEach { deviceId ->
+            transport.cancelTransaction(transactionId, otherUserId, deviceId, CancelCode.Accepted)
```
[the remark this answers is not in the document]
langleyd: On iOS we take the existing verification requests `transactionId` which seems to make sense to me. It looks like what you have should match that according to [this](https://github.com/vector-im/element-android/pull/5916/files#diff-6dba983c4fd6b2410349c45f7a2d356e2c7b923df092bf56ddfc1324b1e064feR893) line. But on iOS it doesn't appear that it's over the form `local.12345` as you report on android. @BillCarsonFr Could you confirm it's ok for the transaction to be of that form?

langleyd, line 946: I think this line will send multiple to-device requests(1 per device). On iOS we were able to add all the cancellations to the same contentMap and send a single request. Web does the same. I think it should be possible to change [cancelRequest](https://github.com/vector-im/element-android/blob/2761b35b6f95979be685fb270c10ed71213e33aa/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/VerificationTransportToDevice.kt#L173) signature to allow you pass multiple deviceId targets.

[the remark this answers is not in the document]
jmartinesp: On Android, both the existing verification request and the 'ready' request share the same transaction id, and it's local in both cases 😕.

Florian14 (approved):
Code seems good to me

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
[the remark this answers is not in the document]
BillCarsonFr: don't think you need to force download for self verification

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
```diff
@@ -938,9 +938,25 @@ internal class DefaultVerificationService @Inject constructor(
...
+
+    /**
+     * Gets a list of device ids excluding the current one.
+     */
+    private fun getOtherDeviceIds(): List<String> = cryptoStore.getUserDevices(userId)?.keys?.filter { it != deviceId }.orEmpty()
+
+    /**
+     * Notifies other devices that the current verification transaction is being handled by [fromDeviceId].
+     */
+    private fun notifyOthersOfAcceptance(transactionId: String, otherUserId: String, fromDeviceId: String) {
```
BillCarsonFr, line 956: why otherUserId, it will be always you right?

BillCarsonFr (changes requested):
LGTM, some minor comments.
Would be a good idea to add an integration test.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/verification/CancelCode.kt
BillCarsonFr: SHould be AcceptedByAnotherDevice as per doc 
> was accepted by a different device. The device receiving this error can ignore the verification request.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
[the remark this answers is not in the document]
BillCarsonFr: transaction id is always something created locally, WDYM?
Only room verification would need an eventID there (as it's tracked as an event relation)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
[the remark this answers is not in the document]
BillCarsonFr: > The iOS counterpart of this PR first refreshes the device ids. I couldn't do that here since adding `CryptoService` creates a DI cycle, but I'm not entirely sure this is needed.

FTR, you can use the DeviceListManager to download keys if needed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
jmartinesp: Yes, you're right.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
```diff
@@ -939,9 +939,25 @@ internal class DefaultVerificationService @Inject constructor(
...
                 )
         )
+
+        notifyOthersOfAcceptance(readyReq.transactionId, readyReq.fromDevice)
+    }
+
+    /**
+     * Gets a list of device ids excluding the current one.
+     */
+    private fun getOtherDeviceIds(): List<String> = cryptoStore.getUserDevices(userId)?.keys?.filter { it != deviceId }.orEmpty()
```
BillCarsonFr, line 952: getMyOtherDevices?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
```diff
@@ -939,9 +939,25 @@ internal class DefaultVerificationService @Inject constructor(
...
+
+    /**
+     * Notifies other devices that the current verification transaction is being handled by [fromDeviceId].
+     */
+    private fun notifyOthersOfAcceptance(transactionId: String, fromDeviceId: String) {
```
BillCarsonFr, line 957: can we have a better name for `fromDeviceId`? here it would be `acceptedFromDevice` or

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
jmartinesp: Changed to `getMyOtherDeviceIds`.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/verification/DefaultVerificationService.kt
jmartinesp: Now it's `acceptedByDeviceId`.

BillCarsonFr (approved):
LGTM !

PR #5947, opened 2022-05-05
Small changes

bmarty:
Another tiny PR, does not deserve a longer description, nor a changelog

### DIFF
--- a/library/jsonviewer/src/main/java/org/billcarsonfr/jsonviewer/ValueItem.kt
+++ b/library/jsonviewer/src/main/java/org/billcarsonfr/jsonviewer/ValueItem.kt
@@ -18,11 +18,11 @@ package org.billcarsonfr.jsonviewer
 
 import android.content.ClipData
 import android.content.ClipboardManager
-import android.content.Context
 import android.view.ContextMenu
 import android.view.View
 import android.widget.LinearLayout
 import android.widget.TextView
+import androidx.core.content.getSystemService
 import com.airbnb.epoxy.EpoxyAttribute
 import com.airbnb.epoxy.EpoxyHolder
 import com.airbnb.epoxy.EpoxyModelClass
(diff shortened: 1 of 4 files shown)


library/jsonviewer/src/main/java/org/billcarsonfr/jsonviewer/ValueItem.kt
```diff
@@ -77,8 +77,7 @@ internal abstract class ValueItem : EpoxyModelWithHolder<ValueItem.Holder>() {
         ) {
             if (copyValue != null) {
                 val menuItem = menu?.add(R.string.copy_value)
-                val clipService =
-                        v?.context?.getSystemService(Context.CLIPBOARD_SERVICE) as? ClipboardManager
+                val clipService = v?.context?.getSystemService<ClipboardManager>()
```
ariskotsomitopoulos, line 80: Maybe here we could use [this function](https://github.com/vector-im/element-android/blob/5efe1f4bd8e065f9dc33d46edc53e38d2a9ba85f/vector/src/main/java/im/vector/app/core/utils/SystemUtils.kt#L88)

ariskotsomitopoulos (approved):
Minor comment, not related to the PR. Lets merge it

library/jsonviewer/src/main/java/org/billcarsonfr/jsonviewer/ValueItem.kt
bmarty: no, we are in a submodule. Or this fun must be move to a deeper module in the hierarchy.

tools/check/check_code_quality.sh
```diff
@@ -67,6 +67,9 @@ echo "Search for forbidden patterns in code..."
 ${searchForbiddenStringsScript} ./tools/check/forbidden_strings_in_code.txt \
     ./matrix-sdk-android/src/main/java \
     ./matrix-sdk-android-flow/src/main/java \
+    ./library/core-utils/src/main/java \
```
ouchadam, line 70: I don't know the internals of `searchForbiddenStringsScript` but perhaps we could glob this to automatically pickup source directories? `**/src/main/java` (it looks like we want to check all the modules~)

ouchadam (approved):

tools/check/check_code_quality.sh
bmarty: yes, it's a pain to add all the modules one by one...

PR #5949, opened 2022-05-05
Correct public_baseurl for nightly tests.

michaelkaye:
The tests seem to function against a synapse with an incorrect public_baseurl - some parts of the client use the public_baseurl for information and timeout, others use the configured http endpoint (which is hardcoded in the tests to be correct).

Making this change reduces timeouts waiting for requests to these incorrect endpoints.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -43,11 +43,12 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Start synapse server
-        uses: michaelkaye/setup-matrix-synapse@v1.0.2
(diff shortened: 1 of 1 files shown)


michaelkaye:
So the change in fb19d6b seems to have been powerful in itself; the baseurl was wrong for a number of the tests (compared to running locally).

bmarty (commented):
LGTM so far (the PR is still a draft)

michaelkaye:
We may as well merge this as-is.

ouchadam (approved):

ouchadam:
I have this change locally as it's needed for the sanity tests to login in after creating an account

PR #5952, opened 2022-05-05
Sdk user story

bmarty:
This PR add some cleanup on the SDK documentation and on the API. See commit messages for the details.

Also it introduced the SDK user Story, with their documentation:

image

and for **Us100SignIn** for instance:

image

This is still WIP, but we can merge as it is and update later, to first check that the format is OK.


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/keysbackup/KeysBackupService.kt
```diff
@@ -204,10 +202,13 @@ interface KeysBackupService {
...
     val keysBackupVersion: KeysVersionResult?
+
     val currentBackupVersion: String?
-    val isEnabled: Boolean
-    val isStucked: Boolean
-    val state: KeysBackupState
+        get() = keysBackupVersion?.version
+
+    fun isEnabled(): Boolean
+    fun isStucked(): Boolean
```
mnaturel, line 210: I guess it should be `isStuck()` instead. The version with "ed" seems to not exist.

matrix-sdk-android/docs/packages.md
```diff
@@ -1,3 +1,7 @@
+# Package org.matrix.android.sdk._userstories
```
mnaturel, line 1: The "_" should be removed since you renamed it without it.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/userstories/Us000InitMatrix.kt
```diff
@@ -0,0 +1,28 @@
...
+package org.matrix.android.sdk.userstories
+
+/**
+ * ### Title
+ * Init a Matrix object
+ *
+ * ### Required APIs:
+ * - [org.matrix.android.sdk.api.Matrix] constructor
+ */
+object Us000InitMatrix
```
mnaturel, line 28: I have 2 questions:
- do the user stories really need to be `object`? I think simple `class` could work and avoid loading of non used US at runtime.
- Am I missing something or the US are not implemented yet? If not implemented, would't be better to make it `internal` or at least mention it in the doc?

matrix-sdk-android/docs/packages.md
bmarty: Good catch, thanks. Sadly Dokka does not complain :/

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/crypto/keysbackup/KeysBackupService.kt
bmarty: Ah yes, will add.

bmarty: eb576808b97179969a95ff81a0b2daf7277dcb00

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/registration/RegistrationResult.kt
```diff
@@ -18,13 +18,31 @@ package org.matrix.android.sdk.api.auth.registration
 
 import org.matrix.android.sdk.api.session.Session
 
-// Either a session or an object containing data about registration stages
+/**
+ * Either a session or an object containing data about registration stages.
+ */
```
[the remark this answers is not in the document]
bmarty: This is definitely more compact. I will think about it for next changes.
With the current format, it's easier to add new lines.

mnaturel (approved):

PR #5962, opened 2022-05-06
Run the PR test after merge and report to channel if it fails

michaelkaye:
Additionally move sonarqube as the only nightly run; and again report to the channel only if it fails.

This should reduce noise in the channel and also provide feedback on who merged a PR with a failing test in it.

### DIFF
--- a/.github/workflows/post-pr.yml
+++ b/.github/workflows/post-pr.yml
@@ -1,28 +1,36 @@
-name: Nightly Tests
+name: Integration Tests
+
+# This runs for all closed pull requests against main, including those closed without merge.
+# Further filtering occurs in 'should-i-run'
 
 on:
-  push:
-    branches: [ release/* ]
-  schedule:
-    # At 20:00 every day UTC
-    - cron: '0 20 * * *'
-  workflow_dispatch:
+  pull-request:
+    types: [closed]
+    branches: [develop]
 
 # Enrich gradle.properties for CI/CD
 env:
   CI_GRADLE_ARG_PROPERTIES: >
     -Porg.gradle.jvmargs=-Xmx4g
     -Porg.gradle.parallel=false
+
 jobs:
+  should-i-run:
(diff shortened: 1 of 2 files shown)


.github/workflows/sonarqube.yml
```diff
@@ -0,0 +1,81 @@
...
+          ORG_GRADLE_PROJECT_SONAR_LOGIN: ${{ secrets.SONAR_TOKEN }}
+
+# Notify the channel about sonarqube failures
+  notify:
+    name: Notify matrix
+    runs-on: ubuntu-latest
+    needs:
+      - sonarqube
+      - codecov-units
+    if: always() && needs.sonarqube.result != "success" and needs.codecov-units.result != "success"
```
mnaturel, line 74: For this condition, I would have expect something like: 

`if: always() && (needs.sonarqube.result != "success" || needs.codecov-units.result != "success")`

I guess we should notify in case either the sonarqube job or the code coverage job fails. And also I am not sure we can use `and` instead of `&&`? What do you think?

.github/workflows/post-pr.yml
```diff
@@ -1,28 +1,43 @@
...
 
 # Enrich gradle.properties for CI/CD
 env:
   CI_GRADLE_ARG_PROPERTIES: >
     -Porg.gradle.jvmargs=-Xmx4g
     -Porg.gradle.parallel=false
+
 jobs:
+
+  # More info on should-i-run:
```
mnaturel, line 19: Thanks for the comment, it helps me to understand how this works.

.github/workflows/sonarqube.yml
michaelkaye: Argh, yes; I made this fix in post-pr.yml but not here - good catch.

mnaturel (approved):
LGTM but I see there are some conflicts to be resolved.

PR #5965, opened 2022-05-06
Handling SSL/TLS errors during WellKnown lookup

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical - SDK
- [ ] Other :

## Content

Applies the SSL/TLS certificate handling logic to the `sign in with matrix id` flow

- Due to the flow inferring the homeserver url from the matrix id domain it meant we were passing a null `HomeserverConnectionConfig` which in turn meant the SSL/TLS socket factory was never applied 

## Motivation and context

To make the `sign in with matrix id` and `other` sign in flows consistent

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot_20220506_154304|after-certs|
 
## Tests

- change the device time/date to a month in the past
- attempt to sign in via `sign in with matrix id`
- Notice invalid home server error, however when signing in via `other` a certificate error is shown

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 31

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/DefaultAuthenticationService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/DefaultAuthenticationService.kt
@@ -382,10 +382,15 @@ internal class DefaultAuthenticationService @Inject constructor(
         return getWellknownTask.execute(
                 GetWellknownTask.Params(
                         domain = matrixId.getDomain(),
-                        homeServerConnectionConfig = homeServerConnectionConfig
+                        homeServerConnectionConfig = homeServerConnectionConfig.orWellKnownDefaults()
                 )
         )
     }
+
+    private fun HomeServerConnectionConfig?.orWellKnownDefaults() = this ?: HomeServerConnectionConfig.Builder()
+            // server uri is ignored when doing a wellknown lookup as we use the matrix id domain instead
+            .withHomeServerUri("https://dummy.org")
+            .build()
 
     override suspend fun directAuthentication(homeServerConnectionConfig: HomeServerConnectionConfig,
                                               matrixId: String,
--- a/changelog.d/5965.sdk
+++ b/changelog.d/5965.sdk
@@ -1,0 +1,1 @@
+Including SSL/TLS error handing when doing WellKnown lookups without a custom HomeServerConnectionConfig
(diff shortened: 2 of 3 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/DefaultAuthenticationService.kt
```diff
@@ -382,11 +382,16 @@ internal class DefaultAuthenticationService @Inject constructor(
         return getWellknownTask.execute(
                 GetWellknownTask.Params(
                         domain = matrixId.getDomain(),
-                        homeServerConnectionConfig = homeServerConnectionConfig
+                        homeServerConnectionConfig = homeServerConnectionConfig.orWellKnownDefaults()
                 )
         )
     }
 
+    private fun HomeServerConnectionConfig?.orWellKnownDefaults() = this ?: HomeServerConnectionConfig.Builder()
```
ouchadam, line 390: I wasn't too sure about this, I had originally planned to make the [service entry point](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/AuthenticationService.kt#L102) non null but it feels strange to pass the `matrix id` and a homeserver config (with the url already calculated from the matrix id) 

Ideally the `getWellKnownData` could take the certificate parts of the homeserver config rather than the entire model

onurays: I also tried with non-null parameter in my local but it produces ugly code as so said.

onurays (approved):
LGTM!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/DefaultAuthenticationService.kt
ouchadam: I'm happy to add a build method and then remove the `matrixId` from the `getWellKnownData` :+1:

ouchadam: I had a go at using the builder this however the direct usage of Android's `Uri` breaks the OnboardingViewModel tests
https://github.com/vector-im/element-android/compare/feature/adm/matrix-id-via-builder

will merge _as is_ for the time being

PR #5997, opened 2022-05-09
Remove spec v1.3 check for threads

turt2live:
Citation: https://matrix.to/#/!ewdjhNcPcEmYNKzlWp:t2l.io/$CkPuvKdFZyFL547JCy5J3MfvLaWUo_a1XEdmiop1PKc?via=matrix.org&via=element.io&via=envs.net

 
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/Versions.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/Versions.kt
@@ -74,8 +74,8 @@ internal fun Versions.isLoginAndRegistrationSupportedBySdk(): Boolean {
  * Indicate if the homeserver support MSC3440 for threads
  */
 internal fun Versions.doesServerSupportThreads(): Boolean {
-    return getMaxVersion() >= HomeServerVersion.v1_3_0 ||
-            unstableFeatures?.get(FEATURE_THREADS_MSC3440_STABLE) ?: false
+    // TODO: Check for v1.3 or whichever spec version formally specifies MSC3440.
+    return unstableFeatures?.get(FEATURE_THREADS_MSC3440_STABLE) ?: false
 }
 
 /**
--- a/matrix-sdk-android/src/test/java/org/matrix/android/sdk/api/auth/data/VersionsKtTest.kt
+++ b/matrix-sdk-android/src/test/java/org/matrix/android/sdk/api/auth/data/VersionsKtTest.kt
@@ -68,6 +68,6 @@ class VersionsKtTest {
         Versions(supportedVersions = listOf("r0.6.0"), unstableFeatures = mapOf("org.matrix.msc3440.stable" to true)).doesServerSupportThreads() shouldBe true
         Versions(supportedVersions = listOf("v1.2.1"), unstableFeatures = mapOf("org.matrix.msc3440.stable" to true)).doesServerSupportThreads() shouldBe true
         Versions(supportedVersions = listOf("r0.6.0"), unstableFeatures = mapOf("org.matrix.msc3440.stable" to false)).doesServerSupportThreads() shouldBe false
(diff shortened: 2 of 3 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/Versions.kt
```diff
@@ -74,8 +74,8 @@ internal fun Versions.isLoginAndRegistrationSupportedBySdk(): Boolean {
  * Indicate if the homeserver support MSC3440 for threads
  */
 internal fun Versions.doesServerSupportThreads(): Boolean {
-    return getMaxVersion() >= HomeServerVersion.v1_3_0 ||
-            unstableFeatures?.get(FEATURE_THREADS_MSC3440_STABLE) ?: false
+    // TODO: Check for v1.3 or whichever spec version formally specifies MSC3440.
```
bmarty, line 77: Does this TODO means that it has to be done before merging this PR?

bmarty (commented):
Just a small remark, also a file for towncrier will have to be added.
I let @ariskotsomitopoulos take over this PR.

bmarty:
Also I think the version parser should support version without patch number like "v1.3".

The code to update is here: https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt#L41

Some unit test could be added too.

ariskotsomitopoulos (approved):
Thanks for the PR. 

We should keep in mind that already existing clients will have a bad experience. For instance:
- Bob is a user with the latest Android app
- Server release v1_3_0 without threads support
- Bob will try to use dynamic threads instead of the local implementation 👎 
- When bob (and if) updates the app, will see the expected user experience

I think the only solution to that problem would be if the server waits for threads to go to `v1_3_0` and name the new version in something like `v1_2_15 `

turt2live:
@bmarty yes - the versioning will need fixing, but probably left to a different PR to handle. Will get that changelog added.

@ariskotsomitopoulos thanks for the review! We're *removing* the v1.3 check because from the spec side we can't guarantee that the server will have support in v1.3, but all known implementations of threads should be available using the stable flag check and thus be fine. This is more a formality for spec compliance as threads will be released in a future spec version.

fwiw, the modern versions (not-`r` versions) for the spec do not have patch versions.

matrix-sdk-android/src/test/java/org/matrix/android/sdk/api/auth/data/VersionsKtTest.kt
```diff
@@ -62,9 +62,9 @@ class VersionsKtTest {
         Versions(supportedVersions = listOf("r0.6.0")).doesServerSupportThreads() shouldBe false
         Versions(supportedVersions = listOf("r0.9.1")).doesServerSupportThreads() shouldBe false
         Versions(supportedVersions = listOf("v1.2.0")).doesServerSupportThreads() shouldBe false
-        Versions(supportedVersions = listOf("v1.3.0")).doesServerSupportThreads() shouldBe true
-        Versions(supportedVersions = listOf("v1.3.1")).doesServerSupportThreads() shouldBe true
-        Versions(supportedVersions = listOf("v1.5.1")).doesServerSupportThreads() shouldBe true
+        Versions(supportedVersions = listOf("v1.3.0")).doesServerSupportThreads() shouldBe false
+        Versions(supportedVersions = listOf("v1.3.1")).doesServerSupportThreads() shouldBe false
+        Versions(supportedVersions = listOf("v1.5.1")).doesServerSupportThreads() shouldBe false
```
bmarty, line 67: Line 71 (so below, but I cannot comment this line) is failing:

image

This is the test
```kotlin
Versions(supportedVersions = listOf("v1.4.0"), unstableFeatures = mapOf("org.matrix.msc3440.stable" to false)).doesServerSupportThreads() shouldBe true
```

and I think that the test has to be updated to `shouldBe false`

bmarty (approved):

PR #6022, opened 2022-05-11
Gh pages

bmarty:
Publish current SDK documentation to GH pages, with an extra note, see below:

image

A change will have to be done on the SDK to remove the note before generated the documentation there.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,22 @@
+name: Documentation
+
+on:
+  push:
+    branches: [ develop ]
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,23 @@
+name: Documentation
+
+on:
+  pull_request: { }
```
bmarty, line 4: temporary line

bmarty:
It's working: https://vector-im.github.io/element-android/

ouchadam (approved):

PR #6025, opened 2022-05-11
Improve test handling in CI

michaelkaye:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [X] Technical
- [ ] Other :

## Content

~All tests I found in the most recent test runs that are failing have been marked `@Ignore`.~

Having merged other PRs and discussed; increasing timeouts seems to make the tests work (reflecting that the underlying everything is being slow). Have removed the @ ignores for now, replaced with increased timeouts and retries.

Additional test failures that have shown up as part of this have been fixed afterwards.

This should allow new failures of integration tests in CI to be instantly visible when the PR run fails.

## Motivation and context

As discussed in a retro a few weeks ago, we want to know when the longer tests fail due to new code. As they're slow to build/run and need macos runners, we're not running on every commit, but we will verify after each PR is merged, as a reasonable intemediate between "too much" and "not enough".

## Checklist

### DIFF
--- a/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/CryptoStoreTest.kt
+++ b/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/CryptoStoreTest.kt
@@ -39,7 +39,7 @@ private const val DUMMY_DEVICE_KEY = "DeviceKey"
 @RunWith(AndroidJUnit4::class)
 class CryptoStoreTest : InstrumentedTest {
 
-    @get:Rule var rule = RetryTestRule(3)
+    @get:Rule val rule = RetryTestRule(3)
 
     private val cryptoStoreHelper = CryptoStoreHelper()
     private val clock = DefaultClock()
--- a/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/CryptoStoreTest.kt
+++ b/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/CryptoStoreTest.kt
@@ -41,7 +41,6 @@ class CryptoStoreTest : InstrumentedTest {
 
     @get:Rule var rule = RetryTestRule(3)
 
-
     private val cryptoStoreHelper = CryptoStoreHelper()
     private val clock = DefaultClock()
 
--- a/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/CryptoStoreTest.kt
+++ b/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/CryptoStoreTest.kt
@@ -22,9 +22,11 @@ import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotEquals
 import org.junit.Assert.assertNull
 import org.junit.Before
+import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.matrix.android.sdk.InstrumentedTest
+import org.matrix.android.sdk.common.RetryTestRule
 import org.matrix.android.sdk.internal.crypto.model.OlmSessionWrapper
 import org.matrix.android.sdk.internal.crypto.store.IMXCryptoStore
 import org.matrix.android.sdk.internal.util.time.DefaultClock
@@ -36,6 +38,9 @@ private const val DUMMY_DEVICE_KEY = "DeviceKey"
 
 @RunWith(AndroidJUnit4::class)
 class CryptoStoreTest : InstrumentedTest {
+
+    @get:Rule var rule = RetryTestRule(3)
+
 
     private val cryptoStoreHelper = CryptoStoreHelper()
     private val clock = DefaultClock()
--- a/vector/src/androidTest/java/im/vector/app/ui/robot/ElementRobot.kt
+++ b/vector/src/androidTest/java/im/vector/app/ui/robot/ElementRobot.kt
@@ -120,6 +120,26 @@ class ElementRobot {
                                 .perform(ViewActions.closeSoftKeyboard(), click())
                     }
                 }
+                // at this point we are in a race with the app restarting. The steps that happen are:
+                // - (initially) app has started, app has initial synched
+                // - (restart) app has strted, app has not initial synched
+                // - (racey) app shows some UI but overlays with initial sync ui
+                // - (initial sync finishes) app has started, has initial synched
+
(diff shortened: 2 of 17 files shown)


michaelkaye:
TODO: after merge, create an issue to track these have been `@ignored` (or add to an existing one)

vector/src/androidTest/java/im/vector/app/ui/robot/ElementRobot.kt
```diff
@@ -120,6 +120,26 @@ class ElementRobot {
...
+                // so we can racily have an IllegalStateException that we have transitioned from busy -> idle
+                // but never having sent the signal.
+
+                // So we need to not start waiting for an initial sync until we have restarted
+                // then we do need to wait for the sync to complete.
+
+                // Which is convoluted especially as it involves the app state refreshing
+                // so; in order to make this be more stable
+                // I hereby cheat and write:
+                Thread.sleep(30_000)
```
michaelkaye, line 142: to be clear: i feel bad about this but i experimented with various options for ~1day and none were more likely to work than this - handling the restart requirement just makes it hard to test with.

ouchadam:
out of interest, do these tests pass locally but fail on the CI?

vector/src/androidTest/java/im/vector/app/ui/robot/ElementRobot.kt
ouchadam: this seems fair, it sounds like we're missing registering the sync as an `IdlingResource` in order for espresso to be able to wait for the sync to finish 

there's a helper for `withRetry` if we wanted to poll for the home appearing instead of a blanket 30 second sleep (assuming the original check wasn't providing false positives)

```kotlin
fun waitForHome() {
  // each attempt has a 500ms delay
  withRetry(times = 60)
  ... original logic
  }
}
```

ouchadam (approved):
LGTM :+1:  small comment about potentially reducing the 30 second wait but not a blocker (would prefer the CI and local to be passing)

vector/src/androidTest/java/im/vector/app/ui/robot/ElementRobot.kt
michaelkaye: The problem is that waitForHome() is using an idling resource on initial sync; but it's idle at two points: Immediately (+/-  thread racing) after we return from getting espresso to click the toggle, and then it's busy for 10-15s until after the app has restarted and done the initial sync.

michaelkaye:
> out of interest, do these tests pass locally but fail on the CI?

They fail for me locally (on a linux machine) and on GHA CI, and also when using firebase test lab and a remote synapse server to run the UI test code.

michaelkaye:
With the above changes (recently rebased against develop) I'm now getting zero test fails locally (and I've generally been able to replicate all CI test fails locally).

I'm a little concerned that 895... is going to be slowing down the tests generally because of lots of initializations, but if we want the separation I'm not sure how else to do it.

michaelkaye:
specifcally asking for review from @bmarty for 8950aa3 related to a recentl change, 

and @BillCarsonFr to have a confirm whether a19c1d6 which changes the actual logic of a crypto test is right.

bmarty:
> specifcally asking for review from @bmarty for [8950aa3](https://github.com/vector-im/element-android/commit/8950aa38eabac4aaf87d0ca1ffc9f6fc7166543f) related to a recentl change,

Having double checked, I do not see what recent change may require this change. But if it fixes the tests, it's OK for me.

Asking for @ganfra advice.

> and @BillCarsonFr to have a confirm whether [a19c1d6](https://github.com/vector-im/element-android/commit/a19c1d6fbebaee014d59d91895919d0a6974cafa) which changes the actual logic of a crypto test is right.

I do not want to answer in place of @BillCarsonFr , but the change is now matching the message in the checks.

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/CryptoStoreTest.kt
```diff
@@ -37,6 +39,8 @@ private const val DUMMY_DEVICE_KEY = "DeviceKey"
 @RunWith(AndroidJUnit4::class)
 class CryptoStoreTest : InstrumentedTest {
 
+    @get:Rule var rule = RetryTestRule(3)
```
bmarty, line 42: This could be `val` instead of `val` (= immutable). Applicable to all the occurrences on this PR

bmarty (commented):
Small remarks

BillCarsonFr:
> > specifcally asking for review from @bmarty for [8950aa3](https://github.com/vector-im/element-android/commit/8950aa38eabac4aaf87d0ca1ffc9f6fc7166543f) related to a recentl change,
> 
> Having double checked, I do not see what recent change may require this change. But if it fixes the tests, it's OK for me.
> 
> Asking for @ganfra advice.
> 
> > and @BillCarsonFr to have a confirm whether [a19c1d6](https://github.com/vector-im/element-android/commit/a19c1d6fbebaee014d59d91895919d0a6974cafa) which changes the actual logic of a crypto test is right.
> 
> I do not want to answer in place of @BillCarsonFr , but the change is now matching the message in the checks.

Yes that is correct. Wonder why it was pushed like that :/

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/CryptoStoreTest.kt
michaelkaye: :+1: Have updated these.

michaelkaye:
> > specifcally asking for review from @bmarty for [8950aa3](https://github.com/vector-im/element-android/commit/8950aa38eabac4aaf87d0ca1ffc9f6fc7166543f) related to a recentl change,
> 
> Having double checked, I do not see what recent change may require this change. But if it fixes the tests, it's OK for me.
> 
> Asking for @ganfra advice.

It's the change merged here:

https://github.com/vector-im/element-android/pull/5887/files#diff-7ab56d26588e03ae923b3a9b5a37b0a3e6bdb1c7d560104072426c0e0361118bR68

that will create multiple TestMatrix instances; each eventually does a WorkManager.initialize() further on in the logic:

https://github.com/vector-im/element-android/pull/5887/files#diff-2a024565d457bef75c1870ebbf3454ed8f6f134bdd6a28cf29c71910e0f2697cR69

Which is causing the errors.

bmarty:
I see, thanks @michaelkaye for the explanation!

PR #6038, opened 2022-05-12
Detekt

bmarty:
Let the CI fails first!

Now the CI should not fail.

This PR setup detekt and fix a few issues detected by detekt.

Some rule have been disabled temporarily, to avoid having too many changes.

Change on generated analytics file has been reported in https://github.com/matrix-org/matrix-analytics-events/pull/62


build.gradle
```diff
@@ -140,6 +143,15 @@ allprojects {
                 "experimental:kdoc-wrapping",
         ]
     }
+
+    detekt {
+        // preconfigure defaults
+        buildUponDefaultConfig = true
+        // activate all available (even unstable) rules.
+        allRules = false
```
ouchadam, line 151: we can generate a baseline if we want to enable more rules without having to fix them https://detekt.dev/docs/introduction/baseline/

tools/detekt/detekt.yml
```diff
@@ -0,0 +1,73 @@
...
+  VariableNaming:
+    # TODO Enable it
+    active: false
+  TopLevelPropertyNaming:
+    # TODO Enable it
+    active: false
+
+performance:
+  SpreadOperator:
+    active: false
```
ouchadam, line 73: I don't mind enabling this for production code but it can be handy to be able to use them in tests

ouchadam:
detekt can be quite quick to run, on previous projects I've included it as part of the prebuild tasks to trigger detekt to run on every compile (makes the feedback loop super quick)

```
// module build.gradle
preBuild.dependsOn(project.tasks.getByName("detekt"))
```

ouchadam (approved):

bmarty:
Detekt can also analyses documentation issues 🎉 , but this is disabled by default. I was looking for such tools when setting up `Dokka`!

To add to detekt.yml:

```yml
comments:
  OutdatedDocumentation:
    active: true
  UndocumentedPublicClass:
    active: true
  UndocumentedPublicFunction:
    active: true
  UndocumentedPublicProperty:
    active: true
```

I will probably add those rules in the file now, but disabled by default, so that it can be handled in dedicated PRs (there are many issues :/). The top priority one will be `OutdatedDocumentation`

build.gradle
bmarty: Yes, thanks, I have seen that. Good to prevent new issues. I would prefer to enable the disabled rules one by one and fix the existing issues though.

bmarty:
> detekt can be quite quick to run, on previous projects I've included it as part of the prebuild tasks to trigger detekt to run on every compile (makes the feedback loop super quick)
> 
> ```
> // module build.gradle
> preBuild.dependsOn(project.tasks.getByName("detekt"))
> ```

It can be painful when developing to be blocked by for instance a temporary unused private fun (even if this specific rule is disabled now).

On the other side, it's painful to not see the error locally, but only after the CI checks.

We could have some `check_code_before_creating_pr.sh` which will run all that can be run locally and will be anyway run by the CI. I'm thinking about compiling android tests, run ktlint, run detekt, etc. WDYT?

ouchadam:
> We could have some check_code_before_creating_pr.sh which will run all that can be run locally and will be anyway run by the CI. I'm thinking about compiling android tests, run ktlint, run detekt, etc. WDYT?

having a script to run all our tools detekt, ktlint and the other code quality script would be helpful
there's already a check that causes the build to fail for unused variables :thinking:

Issue #5865, opened 2022-04-28
One way audio after resuming (coming back from holding)


chagai95:
### Steps to reproduce

1. Call someone.
2. Have them answer.
3. Put them on hold.
4. Resume.

#### What did you expect?
To hear them.

#### What happened instead?
You can now not hear them anymore.

### Your phone model

Mi 11 Lite

### Operating system version

MIUI Global 13.0.4 Stable, Android 12

### Application version and app store

from version 1.3.8 and onwards

### Homeserver

matrix.org

### Will you send logs?

Yes

chagai95:
I added some debug logs and here is what I am seeing in the logs when testing myself, but I also just sent logs from the official app:

```
I/ViewRootImpl@11267c8[VectorCallActivity]: ViewPostIme pointer 0
I/MSHandlerLifeCycle: isMultiSplitHandlerRequested: windowingMode=1 isFullscreen=true isPopOver=false isHidden=false skipActivityType=false isHandlerType=true this: DecorView@7ec5e33[VectorCallActivity]
I/ViewRootImpl@11267c8[VectorCallActivity]: ViewPostIme pointer 1
I/MSHandlerLifeCycle: isMultiSplitHandlerRequested: windowingMode=1 isFullscreen=true isPopOver=false isHidden=false skipActivityType=false isHandlerType=true this: DecorView@7ec5e33[VectorCallActivity]
D/VectorCallViewModel$handle: ToggleHoldResume isRemoteOnHold: true
D/WebRtcCall$updateRemoteOnHold: onHold: false
V/VOIP/WebRtcCallManager: WebRtcPeerConnectionManager onCall active: 1651146902437nQRA0bdwbMK3z6fL(3)
D/WebRtcCallManager: Setting audioManager mode
I/CallAudioManager: Update audio route for mode: AUDIO_CALL
D/Analytics: capture(CallStarted(isVideo=false, numParticipants=2, placed=false))
D/WebRtcCall$updateRemoteOnHold: updateRemoteOnHold changing direction from SEND_ONLY to SEND_RECV
V/VOIP/PeerConnectionObserver: StreamObserver onRenegotiationNeeded
D/VectorCallViewModel$callListener: call.isLocalOnHold: false call?.isRemoteOnHold: false
V/VOIP/WebRtcCall: creating offer...
[...]

PR #6039, opened 2022-05-12
Direction and track need to be set for hold/unhold

langleyd:
Fixes https://github.com/vector-im/element-android/issues/5865

As per https://www.w3.org/TR/webrtc/#hold-functionality track and direction need to be set for hold/unhold music. 
iOS client was setting tracks.


bmarty (approved):
LGTM, assuming you have tested the change!

PR #6064, opened 2022-05-16
Detekt: fix end period

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : documentation and code quality

## Content

Add missing period in Kdoc. Enable rule `EndOfSentenceFormat`

Will require https://github.com/matrix-org/matrix-analytics-events/pull/63

## Motivation and context

Improve code quality and generated documentation quality

## Reviewers

Sorry for the big PR. There is no code change hopefully. Maybe using https://patch-diff.githubusercontent.com/raw/vector-im/element-android/pull/6064.diff can help. Maybe not.

### DIFF
--- a/.github/workflows/post-pr.yml
+++ b/.github/workflows/post-pr.yml
@@ -325,5 +325,5 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           hookshot_url: ${{ secrets.ELEMENT_ANDROID_HOOKSHOT_URL }}
-          text_template: "Post-merge validation of ${{ github.head_ref }} into ${{ github.base_ref }} by ${{ github.event.merged_by }} failed: {{#each job_statuses }}{{#with this }}{{#if completed }}  {{name}} {{conclusion}} at {{completed_at}}, {{/if}}{{/with}}{{/each}}"
-          html_template: "Post-merge validation of ${{ github.head_ref }} into ${{ github.base_ref }} by ${{ github.event.merged_by }} failed: {{#each job_statuses }}{{#with this }}{{#if completed }}<br />{{icon conclusion}} {{name}} <font color='{{color conclusion}}'>{{conclusion}} at {{completed_at}} <a href=\"{{html_url}}\">[details]</a></font>{{/if}}{{/with}}{{/each}}"
+          text_template: "Post-merge validation of ${{ github.head_ref }} into ${{ github.base_ref }} by ${{ github.event.pull_request.merged_by }} failed: {{#each job_statuses }}{{#with this }}{{#if completed }}  {{name}} {{conclusion}} at {{completed_at}}, {{/if}}{{/with}}{{/each}}"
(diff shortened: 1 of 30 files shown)


ouchadam (approved):
LGTM :+1: (Didn't check every file...)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/UrlUtils.kt
```diff
@@ -39,7 +39,7 @@ internal fun String.ensureProtocol(): String {
 }
 
 /**
- * Ensure string has trailing /
+ * Ensure string has trailing /.
```
fedrunov, line 42: NIT, but at first I thought that this sentence is just cut off by mistake. It would be cool to take / in quotes to make it clear, but I'm not sure how to escape a single quote in kDoc. Possible option is Html escape like 

&lsquo / &lsquo

but it looks ugly when you read doc in file, not by viewer. Maybe taking in double quotes inside markdown block will be better choice, like this

\`"/"\`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/integrationmanager/AllowedWidgetsContent.kt
```diff
@@ -22,18 +22,20 @@ import com.squareup.moshi.JsonClass
...
-         * Map of native widgetType to a map of domain to Allowed
+         * Map of native widgetType to a map of domain to Allowed.
+         * <pre>
          * {
          *      "jitsi" : {
          *            "jitsi.domain.org"  : true,
          *            "jitsi.other.org"  : false
          *      }
          * }
+         * </pre>
```
fedrunov, line 38: Missing dot here? Or it's not required when sentence starts and ends with same tag

fedrunov (changes requested):
I've checked every file and I feel like I need a pair of new eyes now :) There is one place, where I think dot might be missing and small NIT (not important thing) which I think doesn't deserve to make a commit just for it

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/util/UrlUtils.kt
bmarty: Clarified in dcbc85484c2f321367244ece4e480296bed436d3

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/integrationmanager/AllowedWidgetsContent.kt
bmarty: I think this is OK for detekt since the first sentence `Map of native widgetType to a map of domain to Allowed.` ends with a period.

fedrunov (approved):
LGTM

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/integrationmanager/AllowedWidgetsContent.kt
fedrunov: I just thought that every sentence in doc have to end with dot, since you actually do so in other places, with many sentences, like here:

 https://github.com/vector-im/element-android/blob/12eb23b198254191274601f0f933ceb67c7cb251/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/query/QueryStringValue.kt#L50

It doesn't matter much, if it's ok for detekt, just curious

PR #6074, opened 2022-05-17
Poll refactoring with unit tests (PSF-1020)

onurays:
Aggregating poll editing, poll response and poll end events were handled in `EventRelationsAggregationProcessor` class which was over crowded. Furthermore it wasn't easy to write unit test for this class.

Within this PR, poll implementation moved to a separate class and unit tests are written.

Test coverage:
| Method | Line |
|-------|-------|
| 100% (8/8) | 75.5% (77/102) |


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/poll/DefaultPollAggregationProcessor.kt
```diff
@@ -0,0 +1,210 @@
...
+            aggregatedPollSummaryEntity.sourceLocalEchoEvents.remove(txId)
+            aggregatedPollSummaryEntity.sourceEvents.add(event.eventId)
+        }
+
+        return true
+    }
+
+    private fun getPollEvent(session: Session, roomId: String, eventId: String): TimelineEvent? {
+        return session.roomService().getRoom(roomId)?.getTimelineEvent(eventId) ?: return null.also {
+            Timber.v("## POLL target poll event $eventId not found in room $roomId")
```
ericdecanini, line 186: Optional: If these are error logs we want to keep in prod, don't we want to increase the log level to info or warning? I think our verbose and debug logs are very overcrowded and it can make it harder to debug normal development tasks

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/room/aggregation/poll/PollAggregationProcessorTest.kt
```diff
@@ -0,0 +1,301 @@
...
+
+    @Before
+    fun setup() {
+        mockEventAnnotationsSummaryEntity()
+        mockRoom(A_ROOM_ID, AN_EVENT_ID)
+        every { session.myUserId } returns A_USER_ID_1
+    }
+
+    @Test
+    fun `given a poll start event which is not a replace is not processed by poll aggregator`() {
```
ericdecanini, line 203: In the test names I think we should reflect the expected outcome, not simply the condition setup.

Following a given-when-then structure, I think the then clause is the most important as it shows us what is expected (personally I always use when-then with given as an optional if there are multiple paths/inputs to a function, but curious to know how other people go about this)

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/room/aggregation/poll/PollAggregationProcessorTest.kt
```diff
@@ -0,0 +1,301 @@
...
+import org.matrix.android.sdk.internal.database.model.EventAnnotationsSummaryEntity
+import org.matrix.android.sdk.internal.database.model.EventAnnotationsSummaryEntityFields
+import org.matrix.android.sdk.internal.database.model.PollResponseAggregatedSummaryEntity
+import org.matrix.android.sdk.test.fakes.FakeRealm
+
+private const val A_USER_ID_1 = "@user_1:matrix.org"
+private const val A_ROOM_ID = "!sUeOGZKsBValPTUMax:matrix.org"
+private const val AN_EVENT_ID = "\$vApgexcL8Vfh-WxYKsFKCDooo67ttbjm3TiVKXaWijU"
+
+private val A_POLL_CONTENT = MessagePollContent(
```
ericdecanini, line 56: It also feels weird to have so much test data on top of the test class so much so that the class itself is so far down. I would suggest creating a separate object file containing all this test data and making it internal

ericdecanini (changes requested):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/poll/DefaultPollAggregationProcessor.kt
onurays: Removed. We don't need these logs since they are heavily tested.

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/room/aggregation/poll/PollAggregationProcessorTest.kt
onurays: Done. Let's discuss this on Friday with team.

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/room/aggregation/poll/PollAggregationProcessorTest.kt
onurays: Right, moved to a separate file.

matrix-sdk-android/src/test/java/org/matrix/android/sdk/test/fakes/FakeRealm.kt
```diff
@@ -0,0 +1,36 @@
...
+import io.realm.RealmModel
+import io.realm.RealmQuery
+import io.realm.kotlin.where
+
+internal class FakeRealm {
+
+    val instance = mockk<Realm>(relaxed = true)
+
+    inline fun <reified T : RealmModel> givenWhereReturns(result: T?): RealmQuery<T> {
+        val queryResult = mockk<RealmQuery<T>>(relaxed = true)
```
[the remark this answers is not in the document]
onurays: Removed. We don't need it at all, for now.

ericdecanini (approved):
Ty for the changes! Just conflicts to resolve

PR #6077, opened 2022-05-17
Feature/aris/crypto replay attack

ariskotsomitopoulos:
This PR aims to reduce the UISI `DUPLICATED_MESSAGE_INDEX`.

- With respect to this [issue](https://github.com/matrix-org/internal-config/issues/1145), the code is enhanced to check whether a given key (ratcheted to some particular index) was used to decrypt more than one different event. Two events are considered different if their eventId differs
- Tests are added to ensure that decrypting the same event with the same message index will prevent UISI
- Adds a custom `timelineId` when decryption is called from `RoomSyncHandler`


matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/replayattack/ReplayAttackTest.kt
```diff
@@ -0,0 +1,109 @@
...
+        val sentEvents = testHelper.sendTextMessage(aliceRoomPOV, "Hello I will be decrypted twice", 1)
+        Assert.assertTrue("Message should be sent", sentEvents.size == 1)
+
+        testHelper.runBlockingTest {
+            // Lets assume we are from the main timelineId
+            val timelineId = "timelineId"
+            // Lets decrypt the original event
+            aliceSession.cryptoService().decryptEvent(sentEvents[0].root, timelineId)
+            // Lets try to decrypt the same event
+            aliceSession.cryptoService().decryptEvent(sentEvents[0].root, timelineId)
```
BillCarsonFr, line 105: I'd do a try {.. decrypt } catch (error) { fail("Shouldn't throw a decryption error for same event") }

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -763,59 +765,71 @@ internal class MXOlmDevice @Inject constructor(
...
+        val messageIndexKey = senderKey + "|" + sessionId + "|" + roomId + "|" + decryptResult.mIndex
+        Timber.tag(loggerTag.value).d("##########################################################")
+        Timber.tag(loggerTag.value).d("## decryptGroupMessage() timeline: $timeline")
+        Timber.tag(loggerTag.value).d("## decryptGroupMessage() senderKey: $senderKey")
+        Timber.tag(loggerTag.value).d("## decryptGroupMessage() sessionId: $sessionId")
+        Timber.tag(loggerTag.value).d("## decryptGroupMessage() roomId: $roomId")
+        Timber.tag(loggerTag.value).d("## decryptGroupMessage() eventId: $eventId")
+        Timber.tag(loggerTag.value).d("## decryptGroupMessage() mIndex: ${decryptResult.mIndex}")
+
+        if (timeline?.isNotBlank() == true) {
```
BillCarsonFr, line 800: i think we might need some concurrency issues when accessing the timelineset. Maybe use a mutex for that?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -763,59 +765,71 @@ internal class MXOlmDevice @Inject constructor(
...
+            val timelineSet = inboundGroupSessionMessageIndexes.getOrPut(timeline) { mutableSetOf() }
+            if (timelineSet.contains(messageIndexKey) && messageIndexKey.alreadyUsed(eventId)) {
+                val reason = String.format(MXCryptoError.DUPLICATE_MESSAGE_INDEX_REASON, decryptResult.mIndex)
+                Timber.tag(loggerTag.value).e("## decryptGroupMessage() timelineId=$timeline: $reason")
+                throw MXCryptoError.Base(MXCryptoError.ErrorType.DUPLICATED_MESSAGE_INDEX, reason)
+            }
+            timelineSet.add(messageIndexKey)
+        }
+        replayAttackMap[messageIndexKey] = eventId
```
BillCarsonFr, line 809: It feels strange to have the replayAttackMap global and shared by everyone.
I would have made inboundGroupSessionMessageIndexes a Map<timelineId, replayAttackMap>

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/RoomSyncHandler.kt
```diff
@@ -537,6 +539,11 @@ internal class RoomSyncHandler @Inject constructor(
         }
     }
 
+    private fun generateTimelineId(roomId: String, event: Event): String {
+        val threadIndicator = if (event.isThread()) "_thread_" else "_"
```
BillCarsonFr, line 543: why do we make a different timeline id between thread and normal?

ariskotsomitopoulos: For the fact that it's technically a different timeline, that can contain the same events with the main timeline in parallel. But maybe re-using the main timelineId is enough, will update it

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
ariskotsomitopoulos: Yeap, its a good point will refactor it a bit

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
ariskotsomitopoulos: Its the way the current code works, I will have a look

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/MXOlmDevice.kt
```diff
@@ -763,59 +764,63 @@ internal class MXOlmDevice @Inject constructor(
...
+            sessionHolder.mutex.withLock {
+                inboundGroupSession.decryptMessage(body)
+            }
+        } catch (e: OlmException) {
+            Timber.tag(loggerTag.value).e(e, "## decryptGroupMessage () : decryptMessage failed")
+            throw MXCryptoError.OlmError(e)
+        }
+
+        val messageIndexKey = senderKey + "|" + sessionId + "|" + roomId + "|" + decryptResult.mIndex
+        Timber.tag(loggerTag.value).d("##########################################################")
```
BillCarsonFr, line 791: Reduce log visibility to verbose!

BillCarsonFr (approved):
LGTM, could you just reduce the log visibility to verbose

matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/internal/crypto/replayattack/ReplayAttackTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        val aliceRoomPOV = aliceSession.roomService().getRoom(e2eRoomID)!!
+
+        // Bob
+        val bobSession = cryptoTestData.secondSession
+        val bobRoomPOV = bobSession!!.roomService().getRoom(e2eRoomID)!!
+        assertEquals(bobRoomPOV.roomSummary()?.joinedMembersCount, 2)
+
+        // Alice will send a message
+        val sentEvents = testHelper.sendTextMessage(aliceRoomPOV, "Hello I will be decrypted twice", 1)
+        Assert.assertTrue("Message should be sent", sentEvents.size == 1)
```
bmarty, line 98: Can be better to use `assertEqual` for better test failure report (for instance "expected 1, actual 2", rather than "expected true, actual false" when using `assertTrue`

ariskotsomitopoulos: Updated

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/room/RoomSyncHandler.kt
```diff
@@ -537,6 +538,10 @@ internal class RoomSyncHandler @Inject constructor(
         }
     }
 
+    private fun generateTimelineId(roomId: String): String {
+        return "${RoomSyncHandler::class.java.simpleName}$roomId"
```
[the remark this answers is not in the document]
ariskotsomitopoulos: Yes, it doesn't really matters. I hardcoded the prefix

PR #6075, opened 2022-05-17
Translations update from Weblate

RiotTranslateBot:
Translations update from [Weblate](https://translate.element.io/projects/element-android/element-app/) for Element Android/Element Android App.

Current translation status:

Weblate translation status

PR #6079, opened 2022-05-17
Fix translation

bmarty:
Following #6075


bmarty:
Can be merged if CI is happy

PR #6088, opened 2022-05-18
In TimelineForwardPaginationTest, explicitly test for the events we expect

michaelkaye:
This permits a clear error when the events are missing / extra and while
not making the test invulnerable to future changes in events, should be
explicit on what's changed.

 
## Type of change

- [ ] Feature
- [ ] Bugfix
- [X] Technical
- [ ] Other :

## Content

The test was failing with a count of 58 events not the expected 57. 

To diagnose why, I split out the events by type, and asserted each count individually.

I believe we've gained one state event of some type, but I don't know which one (as we didn't have historical information).

From now, the test will error with a quite specific reason (missing an event or found an event)

A similar issue was fixed in another pagination test (different style of test so splitting out checks by message type was harder)

## Motivation and context

Fix integration test

## Checklist

### DIFF
--- a/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/session/room/timeline/TimelineForwardPaginationTest.kt
+++ b/matrix-sdk-android/src/androidTest/java/org/matrix/android/sdk/session/room/timeline/TimelineForwardPaginationTest.kt
@@ -140,9 +140,24 @@ class TimelineForwardPaginationTest : InstrumentedTest {
             aliceTimeline.hasMoreToLoad(Timeline.Direction.BACKWARDS).shouldBeFalse()
 
             assertEquals(EventType.STATE_ROOM_CREATE, snapshot.lastOrNull()?.root?.getClearType())
-            // 6 for room creation item (backward pagination), 1 for the context, and 50 for the forward pagination
-            // 6 + 1 + 50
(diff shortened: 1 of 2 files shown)


michaelkaye:
I believe the "context" in the original comment that was removed was a single message - paginating back 50 should add another 50 to it; but i might be wrong - someone with knowledge of the original test might help here.

michaelkaye:
I think something similar is happening in PreviousLastForwardTest (it is timing out waiting for there to be exactly X items in the snapshot, but there are x+1 or x-1)

michaelkaye:
Original test failure:
```
org.matrix.android.sdk.session.room.timeline.TimelineForwardPaginationTest > forwardPaginationTest[test(AVD) - 9] FAILED 
	org.amshove.kluent.internal.ComparisonFailedException: Expected: <57> but was: <58>
	at org.amshove.kluent.internal.AssertionsKt.assertEquals(Assertions.kt:129)
```

michaelkaye:
Looks like buildkite failure is unrelated

ericdecanini (approved):

michaelkaye:
> Does buildkite let us rerun tests? There's a good chance I think that rerunning it would pass

It does, but i don't have access to do so. I'm fairly sure this isn't something I can have affected as buildkite doesn't touch any of the tests themselves.

PR #6089, opened 2022-05-18
Ensure calling 'fail()' is not caught by the Exception handler

bmarty:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : test which should fail may not fail

## Content

Ensure tests which should fail are actually failing.

## Motivation and context

Make test fail if they have to.

A quick local local test confirm that the test was not failing when `fail()` is called in a try block and `Throwable` are caught.

## Screenshots / GIFs

NA

## Tests

- Run android tests

There are 2 failing tests, not related to this PR: `testUnwedging()` and `trustKeyBackupVersionWithRecoveryKeyTest()`

image

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

### DIFF
--- a/changelog.d/6089.misc
+++ b/changelog.d/6089.misc
@@ -1,0 +1,1 @@
+Test: Ensure calling 'fail()' is not caught by the catch block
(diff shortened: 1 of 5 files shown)


Florian14 (approved):

BillCarsonFr (approved):
LGTM

Issue #6093, opened 2022-05-18
[SDK] Avoid resetting internal state when calling `getLoginFlow`


ouchadam:
### Description

In order to support the new FTUE login flow we'll need to avoid resetting the internal login state upon calling `getLoginFlow` as this has the side effect of invalidating a previous successful result upon failure

PR #6094, opened 2022-05-18
SDK - Avoid resetting login state when calling `AuthenticationService.getLoginFlow`

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical - SDK change
- [ ] Other :

## Content

Fixes https://github.com/vector-im/element-android/issues/6093

Avoids resetting the internal authentication state when calling `getLoginFlow` until the flow is successfully fetched.  

## Motivation and context

To allow the UX flow of defaulting to a valid homeserver, editing with an invalid url and then returning to the previously selected homeserver.

## Screenshots / GIFs

| BEFORE | AFTER |
| --- | --- |
ezgif-4-26b6c4bc54|ezgif-4-d8f5749e9e

## Tests

- Start the sign in process
- Edit the homeserver url to an invalid url and press next
- After seeing the error quit the edit server screen
- Continue to sign in with the previously selected homeserver
- Notice a crash

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 31

### DIFF
--- a/changelog.d/6093.sdk
+++ b/changelog.d/6093.sdk
@@ -1,0 +1,1 @@
+Allowing AuthenticationService.getLoginFlow to fail without resetting state from previously successful calls


bmarty (approved):
Thanks!

onurays (approved):
Checked the commit history to see if we have an edge case here, nothing suspicious. LGTM!

PR #6095, opened 2022-05-18
Well known lookups should be to port 443 only.

michaelkaye:
## Type of change

- [ ] Feature
- [X] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Per https://spec.matrix.org/unstable/client-server-api/#well-known-uri, step two:

```
1. Extract the server name from the user’s Matrix ID by splitting the Matrix ID at the first colon.
2. Extract the hostname from the server name.
3. Make a GET request to https://hostname/.well-known/matrix/client. 
```

The hostname should not include a port.

## Motivation and context

There were originally issues with unit tests running against servers on non-standard ports (eg http://localhost:8080). These would then attempt to lookup well-known data by accessing https://localhost:8080/ (NB: https).

This attempts to fix those tests by following the well-known resolution steps and removing the port.

## Tests

Tests continue to work; login has become faster for hosts over http because the timeout attempting to connect has been removed.

However, we should be aware that this is a change in behaviour and may possibly cause some servers with an interesting configuration to have issues, eg if your server_name is `target-domain.com:8443` and you host your well-known at `https://target-domain.com:8443/` not `https://target-domain.com:443/`.

## Tested devices

- [ ] Physical
- [X] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixPatterns.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixPatterns.kt
@@ -177,7 +177,7 @@ object MatrixPatterns {
      * - "@alice:domain.org".getDomain() will return "domain.org"
      * - "@bob:domain.org:3455".getDomain() will return "domain.org:3455"
      */
-    fun String.getDomain(): String {
+    fun String.getServerName(): String {
         if (BuildConfig.DEBUG && !isUserId(this)) {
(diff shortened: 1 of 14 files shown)


michaelkaye:
Hydrogen is making a request to eg https://hydrogen.element.io/null/_matrix/client/r0/login when i give my homeserver as eg "matrix.org:8080", so disreguarding that as it doesn't appear to be implemented yet.

PR #6096, opened 2022-05-18
Fix usage of System.currentTimeMillis().

bmarty:
Should make the CI happy.

### DIFF
--- a/library/core-utils/src/main/java/im/vector/lib/core/utils/flow/TimingOperators.kt
+++ b/library/core-utils/src/main/java/im/vector/lib/core/utils/flow/TimingOperators.kt
@@ -16,6 +16,7 @@
 
 package im.vector.lib.core.utils.flow
 
+import android.os.SystemClock
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.channels.Channel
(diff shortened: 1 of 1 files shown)


library/core-utils/src/main/java/im/vector/lib/core/utils/flow/TimingOperators.kt
```diff
@@ -68,10 +69,10 @@ fun <T> Flow<T>.chunk(durationInMillis: Long): Flow<List<T>> {
 
 @ExperimentalCoroutinesApi
 fun <T> Flow<T>.throttleFirst(windowDuration: Long): Flow<T> = flow {
-    var windowStartTime = System.currentTimeMillis()
+    var windowStartTime = SystemClock.elapsedRealtime()
```
ouchadam, line 72: ideally we'd make use of the `Clock` here as directly interacting with system isn't too test friendly (and have the `Clock` as part of `core`), the but I understand this is a quick workaround to get the CI green again

ouchadam (approved):

onurays (changes requested):
Other usage in `DefaultClock` class is also failing. Can you fix it too?

ouchadam:
@onurays the usage in `DefaultClock` should be allowed via the 2 usages threshold https://github.com/vector-im/element-android/blob/develop/tools/check/forbidden_strings_in_code.txt#L179

unfortunately when the script fails it prints out the first usage it finds, which tends to be the allowed usage :cry:

onurays:
> @onurays the usage in `DefaultClock` should be allowed via the 2 usages threshold https://github.com/vector-im/element-android/blob/develop/tools/check/forbidden_strings_in_code.txt#L179
> 
> unfortunately when the script fails it prints out the first usage it finds, which tends to be the allowed usage 😢

Oh, right. So the purpose is to use Clock everywhere, but difficult to inject here.

onurays (approved):
Aggree on @ouchadam's comment. Let's make the CI happy again.

library/core-utils/src/main/java/im/vector/lib/core/utils/flow/TimingOperators.kt
bmarty: I agree, yes, quick fix for now.
Also will be a bit painful to pass a `Clock` to every `throttleFirst` call.

PR #6098, opened 2022-05-19
Labs flag for enabling live location sharing (PSF-799)

onurays:
Within this PR, we remove feature flag and add a labs flags for live location sharing.

| Labs |
|------|
||


ouchadam (approved):
LGTM :100:

ouchadam:
do we need a changelog entry?

vector/src/main/java/im/vector/app/features/VectorFeatures.kt
```diff
@@ -26,7 +26,6 @@ interface VectorFeatures {
     fun isOnboardingUseCaseEnabled(): Boolean
     fun isOnboardingPersonalizeEnabled(): Boolean
     fun isOnboardingCombinedRegisterEnabled(): Boolean
-    fun isLiveLocationEnabled(): Boolean
```
ouchadam, line 29: not for this PR, in the future I could imagine sharing the `VectorFeatures` with the lab features and having a single point of truth for toggling features

PR #6103, opened 2022-05-20
Glide - Use current drawable while loading new static map image

onurays:
Interestingly, Glide is clearing current image before the new image is loaded which causes a flicker on timeline item. This PR suggests to use current drawable as placeolder while loading the new static map image.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/AbsMessageLocationItem.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/AbsMessageLocationItem.kt
@@ -75,6 +75,7 @@ abstract class AbsMessageLocationItem<H : AbsMessageLocationItem.Holder> : AbsMe
         GlideApp.with(holder.staticMapImageView)
                 .load(location)
                 .apply(RequestOptions.centerCropTransform())
+                .placeholder(holder.staticMapImageView.drawable)
                 .listener(object : RequestListener<Drawable> {
                     override fun onLoadFailed(e: GlideException?,
                                               model: Any?,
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/AbsMessageLocationItem.kt
```diff
@@ -75,6 +75,7 @@ abstract class AbsMessageLocationItem<H : AbsMessageLocationItem.Holder> : AbsMe
         GlideApp.with(holder.staticMapImageView)
                 .load(location)
                 .apply(RequestOptions.centerCropTransform())
+                .placeholder(holder.staticMapImageView.drawable)
```
bmarty, line 78: This fix seems to come from this suggestion: https://github.com/bumptech/glide/issues/2505#issuecomment-532168345

It's not recommended to use that in RecyclerView, which is the case here (see the comments below).

I should work, but may lead to some issue.

Not a blocker anyway.

onurays: > This fix seems to come from this suggestion: https://github.com/bumptech/glide/issues/2505#issuecomment-532168345

Exactly, I couldn't find any other solutions. I could understand the below comment if we were setting the actual source but we are just setting the placeholder.

+ We are going to have a test session next week.

yostyle: I'm agree with @bmarty about this fix. Further tests are needed.

yostyle (approved):

PR #6107, opened 2022-05-20
Fix many invites cancelled after 1 or 2 are successful.

bmarty:
Draft PR to get other people thoughts.

I did not test the fix yet 🙈 

Compare files by ignoring spaces.

### DIFF
--- a/vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
@@ -28,8 +28,9 @@ import im.vector.app.core.resources.StringProvider
 import im.vector.app.features.userdirectory.PendingSelection
 import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.catch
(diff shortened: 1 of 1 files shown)


vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
```diff
@@ -55,39 +56,38 @@ class InviteUsersToRoomViewModel @AssistedInject constructor(
...
-                                    R.plurals.invitations_sent_to_one_and_more_users,
-                                    selections.size - 1,
-                                    selections.first().getBestName(),
-                                    selections.size - 1
-                            )
-                        }
-                        _viewEvents.post(InviteUsersToRoomViewEvents.Success(successMessage))
-                    }
-        }
+                    _viewEvents.post(InviteUsersToRoomViewEvents.Success(successMessage))
```
bmarty, line 88: The ViewEvent was posted too early and so the Activity is finished, the coroutine scope is cancelled, and the pending invitation request are cancelled too.

bmarty:
Was not intended to be merged.

PR #6110, opened 2022-05-20
Fix sending multiple invites to a room reaching only one or two people

### LINKED ISSUE #6109: Sending multiple invites to a room only works partially
### Steps to reproduce

1. Create a room or get into a room where you can invite people.
2. Invite several people (at least 3 or 4).

#### What did you expect?

Everyone should be invited to the room.

#### What happened instead?

Only the first 1 or 2 people actually get invited.

### Your phone model

Emulator

### Operating system version

Android 12

1.4.18

### Homeserver

matrix.org

### Will you send logs?

No

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes #6109. Based on the work of @bmarty on #6107.

Prevents early cancelation of the coroutine scope used to send the invites.

## Motivation and context

When inviting several people to a room, only the first couple of users were actually invited. The `InviteUsersToRoomActivity` was being finished ahead of time, cancelling the `viewModelScope` used to send the invite requests, and this prevented the rest of the users from being invited. This was caused by an `onEach` statement being used as `onCompletion`.

## Tests

- Create a room or open a room where you can invite other users.
- Invite several users, at least 3 or 4.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 12.

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
@@ -28,8 +28,9 @@ import im.vector.app.core.resources.StringProvider
 import im.vector.app.features.userdirectory.PendingSelection
 import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.flow.onCompletion
 import org.matrix.android.sdk.api.session.Session
 import org.matrix.android.sdk.api.session.getRoom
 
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
```diff
@@ -63,11 +65,9 @@ class InviteUsersToRoomViewModel @AssistedInject constructor(
                             is PendingSelection.UserPendingSelection     -> room.membershipService().invite(user.user.userId, null)
                             is PendingSelection.ThreePidPendingSelection -> room.membershipService().invite3pid(user.threePid)
                         }
-                    }
-                    .catch { cause ->
-                        _viewEvents.post(InviteUsersToRoomViewEvents.Failure(cause))
-                    }
-                    .collect {
+                    }.onCompletion { error ->
```
jmartinesp, line 68: It seems like ordering here is very important. If we use `catch` before `onCompletion`,  the possible error value that `onCompletion` can receive will always be null and we won't be able to distinguish between successful and unsuccessful results.

vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
```diff
@@ -87,6 +87,9 @@ class InviteUsersToRoomViewModel @AssistedInject constructor(
                         }
                         _viewEvents.post(InviteUsersToRoomViewEvents.Success(successMessage))
                     }
+                    .catch { cause ->
+                        _viewEvents.post(InviteUsersToRoomViewEvents.Failure(cause))
+                    }.collect()
```
bmarty, line 92: The other change in #6107 was to replace this by `.launchIn(viewModelScope)` and remove the first line `viewModelScope.launch {`. Did you see it? Maybe it was not working as expected?

vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
bmarty: Good to know, thanks!

vector/src/main/java/im/vector/app/features/invite/InviteUsersToRoomViewModel.kt
jmartinesp: I reverted that while trying some alternatives using `.collect()` and I forgot to set back the `.launchIn(...)` call. It should be fixed now.

bmarty (approved):

PR #6124, opened 2022-05-23
Add feature flag for live location sharing

onurays:
We were in rush to remove the feature flag for live location sharing. We need to keep it until updating the privacy.


Johennes:
As discussed separately, this PR can be discarded because live location sharing is supposed to ship in the next release and the privacy policy is expected to be updated by the end of the month.

onurays:
Releasing the whole feature in next release is still in discussion. Closing for now.

PR #6051, opened 2022-05-13
Start DM on first message (UI)

Florian14:
## Type of change

- [X] WIP Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

This PR displays a "fake" local timeline after selecting the users to create a DM. Sending a message or event is not handled yet, so the real room cannot be created for now.
The flow is hidden behind a debug feature flag accessible from the debug menu.

## Motivation and context

UI: https://github.com/vector-im/element-android/issues/5525

## Screenshots / GIFs

https://user-images.githubusercontent.com/11990514/177182407-61af8ca8-e78e-4bf7-b01b-3bbe8aef492f.mp4

## Tests

- Activate the start DM feature in the debug menu -> Features
- On the home screen, select the first tab (Direct Messages)
- Click on the fab button to start a DM, select several users and press "GO"
- A kind of timeline should now appear, there is no action in the toolbar (except the close button)
- It is not possible to access the room settings or details
- Composer actions are not working (an error appears when we try to send an event to the room)
- When we go back to the room list or relaunch the app, the "fake" room is deleted from the DB and disappears from the room list

## Tested devices

- [ ] Physical
- [X] Emulator
- OS version(s): API 32

## Checklist

- [ ] Changes has been tested on an Android device or Android emulator with API 21
- [X] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [ ] Pull request includes a new file under ./changelog.d. See https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#changelog
- [X] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [X] You've made a self review of your PR
- [ ] If you have modified the screen flow, or added new screens to the application, you have updated the test [UiAllScreensSanityTest.allScreensTest()](https://github.com/vector-im/element-android/blob/main/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt#L73)

Issue #5525, opened 2022-03-11
Create DM room only on first message


giomfo:
User Story

- As a user sending a DM to someone I do not have an active DM conversation with
- I want the room to be created only when I first send a message (text, image or voice)
- So that the recipient receives the invite only when there is a message in the room

Detail

- The user selects someone (Corey) to DM, the current implementation is : 
<img src="https://user-images.githubusercontent.com/8969772/157971649-4a62ab01-6380-4553-a130-7019f05a67f6.png" width="50%">

- [x] We request first to rename the option "CREATE" with "GO"
- [x] When the user presses 'GO': 
   - if there is already a DM with Corey : open the DM
   - if there isn't DM with Corey, instead of the room being created (and Corey being invited to an empty room), The user sees the following interface with the composer (the room has not been created).
<img width="243" alt="image" src="https://user-images.githubusercontent.com/8969772/157973680-87081c7b-f74b-4e2b-af6e-9673e892bba5.png">

- [x] If the user doesn’t send a message ( = Dismiss modal  (close button))  no room is created. The user has to start again.
- [x] If the user does send a message, then the DM room is created by inviting Corey and the message is sent.

Additional Requests

- [x] If several contacts have been selected in the picker before clicking on the GO option. The interface is adapted : 
<img width="239" alt="image" src="https://user-images.githubusercontent.com/8969772/157975054-f4ee2aac-33a9-4042-bbe1-8a639b0f03fb.png">

- [x] If the user selects at least one email address in the selected contact list, the resulting DM will be unencrypted. We have to hide the following tile in the interface
<img width="451" alt="image" src="https://user-images.githubusercontent.com/8969772/157976272-daa8eff2-5c7e-4d17-87d3-c9152fa1cadf.png">

gaelledel:
- [ ] More space is needed next to the GO button (on the right) - security padding is mandatory on mobile min 12px
- [ ] Title in the fake room should read " Direct message" in singular
- [ ] We need a loading Screen between the screen when user sends the first message and when the room is created. The sting should read: "We're creating your room"
- [ ] If the invitee has not joined the room yet, we should hide the calling options in the nav. --> Gaelle to create Jira ticket 
- [ ] When inviting email address only the subsequent screen displays: " Empty Room" Instead of showing the full email address string to the user.
[...]

PR #6127, opened 2022-05-23
Create the DM when sending an event

Florian14:
## Type of change

- [X] WIP Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Depends on #6051 

Create the DM room when an event is sent from the local room timeline (text, image, audio, poll, location sharing...)

## Motivation and context

Continue #5525 

## Screenshots / GIFs

||Sender|Receiver|
|-|-|-|
|By text|https://user-images.githubusercontent.com/11990514/178769378-f9348056-c725-418a-a830-f0afbad358bd.mp4|https://user-images.githubusercontent.com/11990514/178769798-f7f90ce8-d410-47c2-9a07-672bd4278e06.mp4|
|By image|https://user-images.githubusercontent.com/11990514/178770472-850fa810-aaa2-41b9-9a81-989a96da9d95.mp4|https://user-images.githubusercontent.com/11990514/178770509-30da044d-4b56-4f25-aa8e-fb78d29d5aa8.mp4|
|By location|https://user-images.githubusercontent.com/11990514/178770654-7bef2d98-33a0-4eef-bae9-5682bede97f1.mp4|https://user-images.githubusercontent.com/11990514/178770719-80cb39ec-231b-4d51-bedb-00282632f600.mp4|

## Tests

1. Open a local room timeline (by clicking on the fab button from the DM room list, selecting one or several users, and pressing the "GO" button)
2. Send an event from the composer view (message, image or anything else)
3. The DM room should be created, the selected users should be invited and we can continue our discussion in that room.
4. Press the back button should redirect to the home screen, the local room is not displayed anymore.

## Tested devices

- [ ] Physical
- [X] Emulator
- OS version(s): API 32

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/EventType.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/EventType.kt
@@ -70,6 +70,9 @@ object EventType {
     const val STATE_ROOM_ENCRYPTION = "m.room.encryption"
     const val STATE_ROOM_SERVER_ACL = "m.room.server_acl"
 
+    // This type is for local purposes, it should never be processed by the server
+    const val LOCAL_STATE_ROOM_THIRD_PARTY_INVITE = "local.room.third_party_invite"
+
     // Call Events
     const val CALL_INVITE = "m.call.invite"
     const val CALL_CANDIDATES = "m.call.candidates"
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
@@ -17,13 +17,16 @@
 package org.matrix.android.sdk.api.session.room.model.create
 
 import android.net.Uri
+import com.squareup.moshi.JsonClass
 import org.matrix.android.sdk.api.crypto.MXCRYPTO_ALGORITHM_MEGOLM
 import org.matrix.android.sdk.api.session.identity.ThreePid
 import org.matrix.android.sdk.api.session.room.model.GuestAccess
 import org.matrix.android.sdk.api.session.room.model.PowerLevelsContent
 import org.matrix.android.sdk.api.session.room.model.RoomDirectoryVisibility
 import org.matrix.android.sdk.api.session.room.model.RoomHistoryVisibility
+import org.matrix.android.sdk.internal.di.MoshiProvider
 
+@JsonClass(generateAdapter = true)
 open class CreateRoomParams {
     /**
      * A public visibility indicates that the room will be shown in the published room list.
@@ -61,12 +64,12 @@ open class CreateRoomParams {
      * A list of user IDs to invite to the room.
      * This will tell the server to invite everyone in the list to the newly created room.
      */
-    val invitedUserIds = mutableListOf<String>()
+    var invitedUserIds: MutableList<String> = mutableListOf()
 
     /**
      * A list of objects representing third party IDs to invite into the room.
      */
-    val invite3pids = mutableListOf<ThreePid>()
+    var invite3pids = mutableListOf<ThreePid>()
 
     /**
      * Initial Guest Access.
@@ -99,14 +102,14 @@ open class CreateRoomParams {
      * The server will clobber the following keys: creator.
      * Future versions of the specification may allow the server to clobber other keys.
      */
-    val creationContent = mutableMapOf<String, Any>()
+    var creationContent = mutableMapOf<String, Any>()
 
     /**
      * A list of state events to set in the new room. This allows the user to override the default state events
      * set in the new room. The expected format of the state events are an object with type, state_key and content keys set.
      * Takes precedence over events set by preset, but gets overridden by name and topic keys.
      */
-    val initialStates = mutableListOf<CreateRoomStateEvent>()
+    var initialStates = mutableListOf<CreateRoomStateEvent>()
 
     /**
      * Set to true to disable federation of this room.
@@ -151,7 +154,7 @@ open class CreateRoomParams {
      * Supported value: MXCRYPTO_ALGORITHM_MEGOLM.
      */
     var algorithm: String? = null
-        private set
+        internal set
 
     var historyVisibility: RoomHistoryVisibility? = null
 
@@ -161,10 +164,18 @@ open class CreateRoomParams {
 
     var roomVersion: String? = null
 
-    var featurePreset: RoomFeaturePreset? = null
+    @Transient var featurePreset: RoomFeaturePreset? = null
 
     companion object {
         private const val CREATION_CONTENT_KEY_M_FEDERATE = "m.federate"
         private const val CREATION_CONTENT_KEY_ROOM_TYPE = "type"
+
+        fun fromJson(json: String?): CreateRoomParams? {
+            return json?.let { MoshiProvider.providesMoshi().adapter(CreateRoomParams::class.java).fromJson(it) }
+        }
     }
+}
+
+internal fun CreateRoomParams.toJSONString(): String {
+    return MoshiProvider.providesMoshi().adapter(CreateRoomParams::class.java).toJson(this)
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
@@ -167,8 +167,8 @@ open class CreateRoomParams {
     @Transient var featurePreset: RoomFeaturePreset? = null
 
     companion object {
-        private const val CREATION_CONTENT_KEY_M_FEDERATE = "m.federate"
-        private const val CREATION_CONTENT_KEY_ROOM_TYPE = "type"
+        internal const val CREATION_CONTENT_KEY_M_FEDERATE = "m.federate"
+        internal const val CREATION_CONTENT_KEY_ROOM_TYPE = "type"
 
         fun fromJson(json: String?): CreateRoomParams? {
             return json?.let { MoshiProvider.providesMoshi().adapter(CreateRoomParams::class.java).fromJson(it) }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
@@ -64,7 +64,7 @@ open class CreateRoomParams {
      * A list of user IDs to invite to the room.
      * This will tell the server to invite everyone in the list to the newly created room.
      */
-    var invitedUserIds: MutableList<String> = mutableListOf()
+    var invitedUserIds = mutableListOf<String>()
 
     /**
      * A list of objects representing third party IDs to invite into the room.
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendEventTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendEventTask.kt
@@ -16,10 +16,12 @@
 package org.matrix.android.sdk.internal.crypto.tasks
 
 import org.matrix.android.sdk.api.session.events.model.Event
+import org.matrix.android.sdk.api.session.room.model.localecho.RoomLocalEcho
 import org.matrix.android.sdk.api.session.room.send.SendState
 import org.matrix.android.sdk.internal.network.GlobalErrorReceiver
 import org.matrix.android.sdk.internal.network.executeRequest
 import org.matrix.android.sdk.internal.session.room.RoomAPI
+import org.matrix.android.sdk.internal.session.room.create.CreateRoomFromLocalRoomTask
 import org.matrix.android.sdk.internal.session.room.membership.LoadRoomMembersTask
 import org.matrix.android.sdk.internal.session.room.send.LocalEchoRepository
 import org.matrix.android.sdk.internal.task.Task
@@ -37,12 +39,18 @@ internal class DefaultSendEventTask @Inject constructor(
         private val localEchoRepository: LocalEchoRepository,
         private val encryptEventTask: EncryptEventTask,
         private val loadRoomMembersTask: LoadRoomMembersTask,
+        private val createRoomFromLocalRoomTask: CreateRoomFromLocalRoomTask,
         private val roomAPI: RoomAPI,
         private val globalErrorReceiver: GlobalErrorReceiver
 ) : SendEventTask {
 
     override suspend fun execute(params: SendEventTask.Params): String {
         try {
+            if (RoomLocalEcho.isLocalEchoId(params.event.roomId.orEmpty())) {
+                // Room is local, so create a real one and send the event to this new room
+                return createRoomAndSendEvent(params)
+            }
+
             // Make sure to load all members in the room before sending the event.
             params.event.roomId
                     ?.takeIf { params.encrypt }
@@ -76,6 +84,12 @@ internal class DefaultSendEventTask @Inject constructor(
             Timber.w(e, "Unable to send the Event")
             throw e
         }
+    }
+
+    private suspend fun createRoomAndSendEvent(params: SendEventTask.Params): String {
+        val roomId = createRoomFromLocalRoomTask.execute(CreateRoomFromLocalRoomTask.Params(params.event.roomId.orEmpty()))
+        Timber.d("State event: convert local room (${params.event.roomId}) to existing room ($roomId) before sending the event.")
+        return execute(params.copy(event = params.event.copy(roomId = roomId)))
     }
 
     @Throws
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendEventTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendEventTask.kt
@@ -46,8 +46,7 @@ internal class DefaultSendEventTask @Inject constructor(
 
     override suspend fun execute(params: SendEventTask.Params): String {
         try {
-            if (RoomLocalEcho.isLocalEchoId(params.event.roomId.orEmpty())) {
-                // Room is local, so create a real one and send the event to this new room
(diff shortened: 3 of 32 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/tasks/SendEventTask.kt
```diff
@@ -78,6 +86,12 @@ internal class DefaultSendEventTask @Inject constructor(
         }
     }
 
+    private suspend fun createRoomAndSendEvent(params: SendEventTask.Params): String {
```
mnaturel, line 89: Maybe we could add a unit test for the case of local room id?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomTask.kt
```diff
@@ -77,6 +81,7 @@ internal class DefaultCreateLocalRoomTask @Inject constructor(
         @SessionDatabase private val realmConfiguration: RealmConfiguration,
         private val createRoomBodyBuilder: CreateRoomBodyBuilder,
         private val userService: UserService,
+        private val cryptoService: DefaultCryptoService,
```
mnaturel, line 84: I think we should inject the interface `CryptoService` instead, no?

mnaturel: Oh it is because the `onStateEvent` is not part of the service interface... Strange maybe we should move the method to the interface. @bmarty What do you think?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomTask.kt
```diff
@@ -228,11 +237,40 @@ internal class DefaultCreateLocalRoomTask @Inject constructor(
             )
         }
 
+        val localRoomThreePidEvents = invited3Pids.map { body ->
```
mnaturel, line 240: Maybe this could be extracted in 2 private methods? And I am struggling to understand the need for both `LocalRoomThirdPartyInviteContent` and `RoomThirdPartyInviteContent`. Since we are creating a local room, I thought we would only have `LocalRoomThirdPartyInviteContent`. Maybe this could be explained through methods names or code comments.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateRoomFromLocalRoomTask.kt
```diff
@@ -0,0 +1,303 @@
...
+            }
+        } catch (exception: TimeoutCancellationException) {
+            throw CreateRoomFailure.CreatedWithTimeout(roomId)
+        }
+
+        createTombstoneEvent(params, roomId)
+        return roomId
+    }
+
+    /* ==========================================================================================
```
mnaturel, line 132: I think we agreed to not use section bloc comments. They could be removed if we split this class into several components.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateRoomFromLocalRoomTask.kt
```diff
@@ -0,0 +1,303 @@
...
+     * Private methods
+     * ========================================================================================== */
+
+    /**
+     * Retrieve the room configuration by parsing the state events related to the local room.
+     */
+    private suspend fun getCreateRoomParams(params: CreateRoomFromLocalRoomTask.Params): CreateRoomParams {
```
mnaturel, line 139: To reduce the scope of the task, maybe we can move all code related to creating the `CreateRoomParams` into a dedicated class. I would be also easier to test. What do you think?

mnaturel (commented):
Cool feature! I added some suggestions mostly to ease reading of code. Let me know what do you think about it. I didn't test it but seeing the videos of the PR it looks good.

Florian14:
> Cool feature! I added some suggestions mostly to ease reading of code. Let me know what do you think about it. I didn't test it but seeing the videos of the PR it looks good.

@mnaturel Thanks for your review! I totally agree about adding tests, I didn't mention it in the PR description but I'll add some unit tests, I would do it in a dedicated PR but I'll see if I can add some in this PR. BTW it's good to have your feedback on which part of code you suggest adding tests!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomTask.kt
Florian14: I created a `LocalRoomThirdPartyInviteContent` because I had to persist the full `ThreePid` to propagate it to create the "real" room. When we invite someone by email to a room, the email address is obfuscated in the `RoomThirdPartyInviteContent` and there is no way to retrieve it.
I created a new object for this purpose because I would not modify the existing object behaviour.
I also need the "RoomThirdPartyInviteContent" for the display of the timeline of the local room which should be as close as possible to a real room.
That's why I have both objects, I'll see to clarify this part.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
```diff
@@ -151,7 +154,7 @@ open class CreateRoomParams {
      * Supported value: MXCRYPTO_ALGORITHM_MEGOLM.
      */
     var algorithm: String? = null
-        private set
+        internal set
```
Florian14, line 157: need to change the visibility for JSON conversion

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/room/create/DefaultCreateRoomFromLocalRoomTaskTest.kt
```diff
@@ -0,0 +1,148 @@
...
+            stateEventDataSource = fakeStateEventDataSource.instance,
+            clock = clock
+    )
+
+    @Before
+    fun setup() {
+        mockkStatic("org.matrix.android.sdk.internal.database.RealmQueryLatchKt")
+        coJustRun { awaitNotEmptyResult<Any>(realmConfiguration = any(), timeoutMillis = any(), builder = any()) }
+
+
```
ElementBot, line 75: - :no_entry_sign: Needless blank line(s)

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/session/room/create/DefaultCreateRoomFromLocalRoomTaskTest.kt
```diff
@@ -0,0 +1,147 @@
...
+            createRoomTask = createRoomTask,
+            stateEventDataSource = fakeStateEventDataSource.instance,
+            clock = clock
+    )
+
+    @Before
+    fun setup() {
+        mockkStatic("org.matrix.android.sdk.internal.database.RealmQueryLatchKt")
+        coJustRun { awaitNotEmptyResult<Any>(realmConfiguration = any(), timeoutMillis = any(), builder = any()) }
+
```
ElementBot, line 74: - :no_entry_sign: Trailing space(s)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateRoomFromLocalRoomTask.kt
Florian14: I changed the internal implementation and removed this part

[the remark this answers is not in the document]
Florian14: I changed the internal implementation and removed this part

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/helper/TimelineEventVisibilityHelper.kt
```diff
@@ -179,8 +179,9 @@ class TimelineEventVisibilityHelper @Inject constructor(private val userPreferen
 
         // Hide fake events for local rooms
         if (RoomLocalEcho.isLocalEchoId(roomId) &&
```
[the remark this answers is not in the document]
Florian14: https://github.com/vector-im/element-android/pull/6127/commits/71938a9ade9a3aa70cf573a10ed4a741084a2af7

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomTask.kt
Florian14: Update: I don't need it anymore the local event for the purpose of the room creation but it is still necessary for the correct display of the local timeline. I can remove it if we agree that the obfuscated three pid is enough.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomTask.kt
Florian14: Yes, I would not change it, I need it to register the local room in the crypto store in the same way as it is done for an existing room. I agree that it should be refactored, but I suppose that the intention was to not expose it in the interface.

bmarty: We should not inject service inside the SDK, and we should not inject implementations as well, but we have some exceptions...
The Service interface should contain only methods useful for the application, not the internal stuff.
So OK to keep it like this.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/EventType.kt
```diff
@@ -70,6 +70,9 @@ object EventType {
     const val STATE_ROOM_ENCRYPTION = "m.room.encryption"
     const val STATE_ROOM_SERVER_ACL = "m.room.server_acl"
 
+    // This type is for local purposes, it should never be processed by the server
+    const val LOCAL_STATE_ROOM_THIRD_PARTY_INVITE = "local.room.third_party_invite"
```
bmarty, line 74: Could be internal, as well as `LocalRoomThirdPartyInviteContent` (and maybe move this class to the internal package?)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
```diff
@@ -61,12 +64,12 @@ open class CreateRoomParams {
      * A list of user IDs to invite to the room.
      * This will tell the server to invite everyone in the list to the newly created room.
      */
-    val invitedUserIds = mutableListOf<String>()
+    var invitedUserIds: MutableList<String> = mutableListOf()
```
bmarty, line 67: Was it necessary to change from `val` to  `var`?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomStateEventsTask.kt
```diff
@@ -0,0 +1,298 @@
...
+/**
+ * Generate a list of local state events from the given [CreateRoomBody].
+ * The states events are generated according to the given configuration and following the matrix specification.
+ * This list reflects as much as possible a list of state events related to a real room configured and got from the server.
+ *
+ * Ref: https://spec.matrix.org/latest/client-server-api/#post_matrixclientv3createroom
+ */
+internal interface CreateLocalRoomStateEventsTask : Task<Params, List<Event>> {
+    data class Params(
+            val roomCreatorUserId: String,
```
bmarty, line 67: The value will always be the current userId, no? In this case, it can be injected in the implementation.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomStateEventsTask.kt
```diff
@@ -0,0 +1,298 @@
...
+                    ).toContent(),
+            )
+            val thirdPartyInviteEvent = createLocalStateEvent(
+                    type = EventType.STATE_ROOM_THIRD_PARTY_INVITE,
+                    content = RoomThirdPartyInviteContent(body.address, null, null, null).toContent(),
+            )
+            listOf(localThirdPartyInviteEvent, thirdPartyInviteEvent)
+        }.flatten()
+        addAll(threePidEvents)
+    }
```
bmarty, line 180: Maybe rewrite to:

```kotlin
    private fun MutableList<Event>.createRoomThreePidEvents() {
        createRoomBody.invite3pids.orEmpty().forEach { body ->
            val localThirdPartyInviteEvent = createLocalStateEvent(
                    type = EventType.LOCAL_STATE_ROOM_THIRD_PARTY_INVITE,
                    content = LocalRoomThirdPartyInviteContent(
                            isDirect = createRoomBody.isDirect.orFalse(),
                            membership = Membership.INVITE,
                            displayName = body.address,
                            thirdPartyInvite = body.toThreePid()
                    ).toContent(),
            )
            val thirdPartyInviteEvent = createLocalStateEvent(
                    type = EventType.STATE_ROOM_THIRD_PARTY_INVITE,
                    content = RoomThirdPartyInviteContent(
                            displayName = body.address,
                            keyValidityUrl = null,
                            publicKey = null,
                            publicKeys = null
                    ).toContent(),
            )
            add(localThirdPartyInviteEvent)
            add(thirdPartyInviteEvent)
        }
    }
```

for clarity

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomStateEventsTask.kt
```diff
@@ -0,0 +1,298 @@
...
+                ).toContent(),
+        )
+        add(roomCreateEvent)
+    }
+
+    /**
+     * Generate the create state event related to the power levels using the given overridden values or the default values according to the specification.
+     * Ref: https://spec.matrix.org/latest/client-server-api/#mroompower_levels
+     */
+    private fun MutableList<Event>.createRoomPowerLevelsEvent() = apply {
```
bmarty, line 120: Using `= apply` is not necessary, since the return value is not used.
Better to use `{`. (at all place in this class)

bmarty (commented):
Thanks for the PR! A few remarks on the implementation, will do some tests this afternoon.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
Florian14: Yes, otherwise Moshi will not be able to restore this field.
Btw I do not need to explicitly declare the type

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomStateEventsTask.kt
Florian14: I thought by mistake it was necessary to chain the calls but not for my usage. https://github.com/vector-im/element-android/pull/6127/commits/110cabaca17a3e195279a48ba28cf18bbba0264c

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomStateEventsTask.kt
Florian14: 💯 I don't know why I made it complicated! https://github.com/vector-im/element-android/pull/6127/commits/cac4df7d664b1e081f9c9bb36d217afbb410b5c9

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/create/CreateLocalRoomStateEventsTask.kt
Florian14: https://github.com/vector-im/element-android/pull/6127/commits/eab4ebc3b1dcbba8eedf2812e9427fd3e9d228ed

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/events/model/EventType.kt
Florian14: https://github.com/vector-im/element-android/pull/6127/commits/2be2a057959f2c9588d2f8b8cc3af91a6abd277a

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/create/CreateRoomParams.kt
Florian14: This is not a data class, when I originally tried by keeping the `val`, some fields were not persisted, I thought that it was for another reason but changing to `var` fixed the issue. I suppose that as these are not constructor fields, Moshi cannot do the same trick 😞

bmarty: > This is not a data class

My bad, OK then!

bmarty (approved):
Thanks for the update.
I have tested the feature, nomical case, i.e. inviting a matrixId and sending a text message, and this is working pretty well.
2 small glitches observed: first one on the timeline, when the room is created. 2nd one when going back to the room list, I can see twice the rooms for a few ms.
Nothing blocking though!

Approved!

Florian14:
> Thanks for the update. I have tested the feature, nomical case, i.e. inviting a matrixId and sending a text message, and this is working pretty well. 2 small glitches observed: first one on the timeline, when the room is created. 2nd one when going back to the room list, I can see twice the rooms for a few ms. Nothing blocking though!
> 
> Approved!

Thanks for your feedback!

The first glitch should be fixed by adding a loading wheel during the creation but I did not find a way to do it atm.
The second one is more annoying because the local rooms are deleted when displaying the room list, I can try to do the deletion as soon as the new room has been created, but not sure it will fix the glitch

Issue #6012, opened 2022-05-10
[Location sharing] - Maximized view for live location shares


mnaturel:
#### What would you like to do?
As a user in a room where two or more people currently have active live location shares, I can click on any of the location sharing messages in the timeline and see the locations of all people in the room with active location shares.

#### Why would you like to do it?
See the locations of everybody in the room who is sharing locations on a single map

#### How would you like to achieve it?
See: https://www.figma.com/file/uthEK6xuo9hSQRdWvPAre5/Location-Sharing?node-id=402%3A49959

Different steps:
- [x] Navigation from timeline to map screen
- [ ] Display of user pins
- [ ] Display of bottom sheet for the list details of users

PR #6129, opened 2022-05-23
[Location sharing] - Show user live location pins in map view (PSF-1019)

mnaturel:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Introducing `LocationSharingService` service interface on SDK side to group all location sharing related features.
On app side, display of all user pins currently sharing their live location in the map view opened when clicking on a message tile in the timeline.

## Motivation and context

Part of #6012 

## Tests

It is suggested to test with a physical device since there may be some issues with pin display in MapView with emulator.
- Ensure the location sharing is enabled in Settings -> Preferences
- Ensure the live location sharing is enable in Settings -> Labs
- Go to a room
- Start sharing your live location
- Press the message when it is in the running state
- Check the pin is correctly displayed on the map and it is moving on location update

Special case to test:
- Since there is a database migration, try update the app from an old version of the app
- Ensure there is no crash at startup

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Android 11

## Checklist


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/livelocation/LiveLocationAggregationProcessor.kt
```diff
@@ -35,6 +36,7 @@ import timber.log.Timber
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 
+// TODO add unit tests
```
BillCarsonFr, line 39: TODO?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/location/DefaultLocationSharingService.kt
```diff
@@ -0,0 +1,49 @@
...
+import dagger.assisted.AssistedFactory
+import dagger.assisted.AssistedInject
+import org.matrix.android.sdk.api.session.room.location.LocationSharingService
+import org.matrix.android.sdk.api.session.room.model.livelocation.LiveLocationShareAggregatedSummary
+import org.matrix.android.sdk.internal.database.mapper.LiveLocationShareAggregatedSummaryMapper
+import org.matrix.android.sdk.internal.database.model.livelocation.LiveLocationShareAggregatedSummaryEntity
+import org.matrix.android.sdk.internal.database.query.findRunningLiveInRoom
+import org.matrix.android.sdk.internal.di.SessionDatabase
+
+// TODO add unit tests
```
BillCarsonFr, line 31: TODO?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/livelocation/LiveLocationAggregationProcessor.kt
mnaturel: Yes it was intented due to lack of time so that it can be merged. It will be added into another dedicated PR.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/location/DefaultLocationSharingService.kt
mnaturel: Same explanation as the other TODO for unit tests: it was intented due to lack of time so that it can be merged. It will be added into another dedicated PR.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/livelocation/LiveLocationAggregationProcessor.kt
mnaturel: Issue created for that: https://github.com/vector-im/element-android/issues/6155

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/location/DefaultLocationSharingService.kt
mnaturel: Issue created for that: https://github.com/vector-im/element-android/issues/6156

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewFragment.kt
```diff
@@ -43,6 +59,14 @@ class LocationLiveMapViewFragment : VectorBaseFragment<FragmentSimpleContainerBi
 
     private val args: LocationLiveMapViewArgs by args()
```
onurays, line 60: This should be removed.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewFragment.kt
```diff
@@ -16,24 +16,40 @@
...
+import im.vector.app.features.location.zoomToLocation
+import timber.log.Timber
+import java.lang.ref.WeakReference
 import javax.inject.Inject
 
 /**
- * Screen showing a map with all the current users sharing their live location in room.
+ * Screen showing a map with all the current users sharing their live location in a room.
  */
 @AndroidEntryPoint
```
onurays, line 54: We should use constructor injection and also inject UrlMapProvider there.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewModel.kt
```diff
@@ -0,0 +1,55 @@
...
+    interface Factory : MavericksAssistedViewModelFactory<LocationLiveMapViewModel, LocationLiveMapViewState> {
+        override fun create(initialState: LocationLiveMapViewState): LocationLiveMapViewModel
+    }
+
+    companion object : MavericksViewModelFactory<LocationLiveMapViewModel, LocationLiveMapViewState> by hiltMavericksViewModelFactory()
+
+    /**
+     * Map to keep track of symbol ids associated to each user Id.
+     */
+    val mapSymbolIds = mutableMapOf<String, Long>()
```
onurays, line 44: We should keep this in view state.

onurays (changes requested):
I will fix my remarks + all markers are gone after rotating the device.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewFragment.kt
```diff
@@ -70,6 +101,104 @@ class LocationLiveMapViewFragment : VectorBaseFragment<FragmentSimpleContainerBi
...
+            symbolManager.update(it)
+        }
+    }
+
+    private fun removeOutdatedSymbols(userLiveLocations: List<UserLiveLocationViewState>, symbolManager: SymbolManager) {
+        val userIdsToRemove = viewModel.mapSymbolIds.keys.subtract(userLiveLocations.map { it.userId }.toSet())
+        userIdsToRemove
+                .mapNotNull { userId ->
+                    removeUserPinFromMapStyle(userId)
+                    viewModel.mapSymbolIds[userId]
```
onurays, line 157: This does nothing.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewFragment.kt
onurays: Done.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewFragment.kt
onurays: Done.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewModel.kt
onurays: Done.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewFragment.kt
onurays: Removed.

onurays (approved):
Thank you @mnaturel. Nice work and LGTM!

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/database/mapper/LiveLocationShareAggregatedSummaryMapperTest.kt
```diff
@@ -0,0 +1,72 @@
...
+                isActive = isActive,
+                endOfLiveTimestampMillis = timeout,
+                lastLocationContent = lastKnownLocationContent
+        )
+        val content = mockk<Content>()
+        every { ContentMapper.map(lastKnownLocationContent) } returns content
+
+        val summary = mapper.map(entity)
+
+        // note: unfortunately the implementation relies on an inline method to map the lastLocationDataContent
```
ouchadam, line 62: for this case I would lean towards calling into the real `ContentMapper`, especially as that's where the majority of the logic of the summary mapper sits, it also means the test can be simplified! 

```kotlin
private const val ANY_USER_ID = "a-user-id"
private const val ANY_ACTIVE_STATE = true
private const val ANY_TIMEOUT = 123L
private val A_LOCATION_INFO = LocationInfo("a-geo-uri")

class LiveLocationShareAggregatedSummaryMapperTest {

    private val mapper = LiveLocationShareAggregatedSummaryMapper()

    @Test
    fun `given an entity then result should be mapped correctly`() {
        val entity = anEntity(content = MessageBeaconLocationDataContent(locationInfo = A_LOCATION_INFO))

        val summary = mapper.map(entity)

        summary shouldBeEqualTo LiveLocationShareAggregatedSummary(
                userId = ANY_USER_ID,
                isActive = ANY_ACTIVE_STATE,
                endOfLiveTimestampMillis = ANY_TIMEOUT,
                lastLocationDataContent = MessageBeaconLocationDataContent(locationInfo = A_LOCATION_INFO)
        )
    }

    private fun anEntity(content: MessageBeaconLocationDataContent) = LiveLocationShareAggregatedSummaryEntity(
            userId = ANY_USER_ID,
            isActive = ANY_ACTIVE_STATE,
            endOfLiveTimestampMillis = ANY_TIMEOUT,
            lastLocationContent = Moshi.Builder().build().adapter(MessageBeaconLocationDataContent::class.java).toJson(content)
    )
}
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/livelocation/LiveLocationAggregationProcessor.kt
```diff
@@ -137,5 +142,16 @@ internal class LiveLocationAggregationProcessor @Inject constructor(
...
     }
 
+    private fun deactivateAllPreviousBeacons(realm: Realm, roomId: String, userId: String, currentEventId: String) {
+        LiveLocationShareAggregatedSummaryEntity
+                .findActiveLiveInRoomForUser(
+                        realm = realm,
+                        roomId = roomId,
+                        userId = userId
+                )
+                .filterNot { it.eventId == currentEventId }
```
ouchadam, line 152: do we expect to have many events per user/room? it might be safer to do the filtering within the database rather than on the query results (I'm imagining a user with years worth of history creating big in memory lists)

vector/src/main/java/im/vector/app/features/location/live/map/UserLiveLocationViewStateMapper.kt
```diff
@@ -0,0 +1,56 @@
...
+    suspend fun map(liveLocationShareAggregatedSummary: LiveLocationShareAggregatedSummary) =
+            suspendCancellableCoroutine<UserLiveLocationViewState?> { continuation ->
+                val userId = liveLocationShareAggregatedSummary.userId
+                val locationData = liveLocationShareAggregatedSummary.lastLocationDataContent
+                        ?.getBestLocationInfo()
+                        ?.geoUri
+                        .toLocationData()
+
+                when {
+                    userId.isNullOrEmpty() || locationData == null -> continuation.resume(null) {
```
ouchadam, line 38: do we need to use the `resume(value, callback)` overload of `resume`? assuming they're the same, it looks like we make use of `resume(value)` and avoid need the empty callback with comment

vector/src/test/java/im/vector/app/features/location/live/map/GetListOfUserLiveLocationUseCaseTest.kt
```diff
@@ -0,0 +1,110 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.matrix.android.sdk.api.session.room.model.livelocation.LiveLocationShareAggregatedSummary
+import org.matrix.android.sdk.api.session.room.model.message.MessageBeaconLocationDataContent
+
+class GetListOfUserLiveLocationUseCaseTest {
+
+    @get:Rule
+    val mvRxTestRule = MvRxTestRule()
```
ouchadam, line 42: we should be able to avoid this rule in a non MvRx viewmodel test :crossed_fingers:

ariskotsomitopoulos (approved):
PR seems fine to me, I observed the following issue (not sure if is related to this PR):

Timeline view:
Screenshot 2022-05-27 at 12 25 45

After click the location and opens a separate fragment, the pin icon colors are inverted?
Screenshot 2022-05-27 at 12 25 49

mnaturel:
> PR seems fine to me, I observed the following issue (not sure if is related to this PR):
> 
> Timeline view: Screenshot 2022-05-27 at 12 25 45
> 
> After click the location and opens a separate fragment, the pin icon colors are inverted? Screenshot 2022-05-27 at 12 25 49

I think you had tested it with emulator, am I right? If so, this is a known issue with MapBox SDK on emulator. It does not happen on a real device.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewFragment.kt
mnaturel: @onurays In a previous PR I had a [comment about using `@AndroidEntryPoint`](https://github.com/vector-im/element-android/pull/6092#discussion_r876857466) to inject fragment. And I think this is a better way to inject fragment.

mnaturel:
> Thank you @mnaturel. Nice work and LGTM!

Thanks for the fixes!

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/database/mapper/LiveLocationShareAggregatedSummaryMapperTest.kt
mnaturel: Thanks for the suggested code! I will update it.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/livelocation/LiveLocationAggregationProcessor.kt
mnaturel: I see you point. Here we only query the active lives, meaning we should not have a lot of items in the list. But I agree, it is safer to integrate the filter inside the query.
In the future, I guess we may allow multiple lives per user (one per device).

vector/src/main/java/im/vector/app/features/location/live/map/GetListOfUserLiveLocationUseCase.kt
```diff
@@ -0,0 +1,40 @@
...
+import androidx.lifecycle.asFlow
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
+import kotlinx.coroutines.flow.mapLatest
+import org.matrix.android.sdk.api.session.Session
+import org.matrix.android.sdk.api.session.getRoom
+import javax.inject.Inject
+
+class GetListOfUserLiveLocationUseCase @Inject constructor(
+        private val session: Session,
```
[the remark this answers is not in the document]
mnaturel: If I understood well, this is what is called when injecting the `Session`: https://github.com/vector-im/element-android/blob/955f366690def07fa90877cb49224ca7cbab0f99/vector/src/main/java/im/vector/app/core/di/SingletonModule.kt#L136 Am I right?

vector/src/main/java/im/vector/app/features/location/live/map/UserLiveLocationViewStateMapper.kt
mnaturel: I didn't found the `resume(value)` signature method, it seems the callback is mandatory...

vector/src/test/java/im/vector/app/features/location/live/map/GetListOfUserLiveLocationUseCaseTest.kt
mnaturel: In fact, you are right it is not needed in this particular useCase but I added it as extra security. If we decide to switch the Coroutine Dispatcher inside the implementation, it will not work. The rule sets the `TestCoroutineDispatcher` to avoid such problems. We could also define our own rule if we want to avoid using that `MvRx` rule.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/livelocation/LiveLocationAggregationProcessor.kt
mnaturel: See [066c540](https://github.com/vector-im/element-android/pull/6129/commits/066c540eb70bef15b3fb59fcc72ea283117e0934)

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/database/mapper/LiveLocationShareAggregatedSummaryMapperTest.kt
mnaturel: See [f707f17](https://github.com/vector-im/element-android/pull/6129/commits/f707f177d923623b7afa0f4201756614e29224e9)

vector/src/main/java/im/vector/app/features/location/live/map/GetListOfUserLiveLocationUseCase.kt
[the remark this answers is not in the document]
ouchadam: with a quick search I can see that we're injecting the `Session` into lots of places, not a blocking comment :+1: 

the main reason to avoid injecting the `Session` directly is that it means the class won't be able to support multi sessions in the future (although perhaps we should have a Session wrapper which delegates to the active session rather than a separate active session holder...)

vector/src/main/java/im/vector/app/features/location/live/map/UserLiveLocationViewStateMapper.kt
ouchadam: we're using the non callback version here  https://github.com/vector-im/element-android/blob/955f366690def07fa90877cb49224ca7cbab0f99/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/network/RetrofitExtensions.kt#L45

perhaps it's a different import :thinking:

vector/src/test/java/im/vector/app/features/location/live/map/GetListOfUserLiveLocationUseCaseTest.kt
ouchadam: I would argue for manually injecting dispatchers rather than relying on a rule to override the global dispatchers (it's not possible to pass a dispatcher to the android `ViewModel/LiveData` which is why it's needed in some cases) 

we have [injectable dispatchers](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/core/dispatchers/CoroutineDispatchers.kt) and [test dispatchers](https://github.com/vector-im/element-android/blob/develop/vector/src/test/java/im/vector/app/test/TestCoroutineDispatchers.kt) for this scenario

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewFragment.kt
onurays: @ganfra is the master of the Dagger. Feel free to revert :)

ouchadam (approved):
thanks for the updates :100:

PR #6136, opened 2022-05-24
Excluding the gms play-service-location from maplibre for fdroid builds 

ouchadam:
Draft whilst testing!


vector/build.gradle
```diff
@@ -507,9 +507,14 @@ dependencies {
     implementation 'commons-codec:commons-codec:1.15'
 
     // MapTiler
-    implementation 'org.maplibre.gl:android-sdk:9.5.2'
-    implementation 'org.maplibre.gl:android-plugin-annotation-v9:1.0.0'
-
+    fdroidImplementation(libs.maplibre.androidSdk) {
+        exclude group: 'com.google.android.gms', module: 'play-services-location'
```
ouchadam, line 511: we should be safe to exclude this due to maplibre doing a runtime check for `play services location` classes when attempting to create a location provider https://github.com/maplibre/maplibre-gl-native/blob/ea234edf67bb3aec75f077e15c1c30c99756b926/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/engine/LocationEngineProvider.java#L59

bmarty: Thanks for checking. For my understanding,  Is it mandatory to specify the module here?

vector/build.gradle
```diff
@@ -507,9 +507,14 @@ dependencies {
...
+    }
+    fdroidImplementation(libs.maplibre.pluginAnnotation) {
+        exclude group: 'com.google.android.gms', module: 'play-services-location'
```
bmarty, line 514: I am pretty sure you have checked that, but does `libs.maplibre.pluginAnnotation` also bring this module?

bmarty (approved):
Perfect, thanks!

vector/build.gradle
ouchadam: not mandatory, I can reduce the explicitness if wanted

Claire1817 (approved):

vector/build.gradle
bmarty: OK, also this seems to be the only imported module, see:
- https://github.com/maplibre/maplibre-gl-native/blob/ea234edf67bb3aec75f077e15c1c30c99756b926/platform/android/MapboxGLAndroidSDK/build.gradle#L16
- https://github.com/maplibre/maplibre-gl-native/blob/ea234edf67bb3aec75f077e15c1c30c99756b926/platform/android/gradle/dependencies.gradle#L68

No need to reduce the explicitness.

Thanks!

vector/build.gradle
bmarty: OK, thanks!

bmarty (approved):
Can be merged!

PR #6149, opened 2022-05-25
Make widget web view request system permissions for camera and microphone

Johennes:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Previously the widget web view prompted to grant the widget permissions but it didn't actually request those permissions from the system. So if the web view requested, e.g. the camera permission but the app hadn't previously been granted that permission, the web view wouldn't get camera access even when the widget permission request had been confirmed.

With this change, the app will also request camera and microphone permissions from the system when needed.

## Motivation and context

The concrete use case that led to this change is running Element Call in a widget web view after a fresh app install (where no system permissions have been granted before and so the widget would never actually get camera or microphone access).

## Screenshots / GIFs

Before this change, only the widget permission dialog was displayed:

Screenshot_20220525_123021

After this change, once the widget permission dialog is confirmed, the actual system permission dialogs appear (unless permission has already been granted before):

Screenshot_20220525_123041
Screenshot_20220525_123051

## Tests

- Perform a clean app install or reset the app's camera & microphone permissions
- Join #ecwidgettest:matrix.org
- Tap through into Element Call in the widget web view
- Confirm the widget permission dialog
- Notice the new system permission dialogs

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 12

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/webview/WebChromeEventListener.kt
+++ b/vector/src/main/java/im/vector/app/features/webview/WebChromeEventListener.kt
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2022 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.features.webview
+
+import android.webkit.PermissionRequest
+
+interface WebChromeEventListener {
+
+    /**
+     * Triggered when the web view requests permissions.
+     *
+     * @param request The permission request.
+     */
+    fun onPermissionRequest(request: PermissionRequest) {
+        // NO-OP
+    }
+
+}
--- a/vector/src/main/java/im/vector/app/features/widgets/WidgetFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/widgets/WidgetFragment.kt
@@ -26,6 +26,8 @@ import android.view.Menu
 import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
+import android.webkit.PermissionRequest
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.core.view.isInvisible
 import androidx.core.view.isVisible
 import com.airbnb.mvrx.Fail
@@ -42,7 +44,9 @@ import im.vector.app.core.platform.OnBackPressed
 import im.vector.app.core.platform.VectorBaseFragment
 import im.vector.app.core.utils.openUrlInExternalBrowser
 import im.vector.app.databinding.FragmentRoomWidgetBinding
+import im.vector.app.features.webview.WebChromeEventListener
 import im.vector.app.features.webview.WebViewEventListener
+import im.vector.app.features.widgets.webview.WebviewPermissionUtils
 import im.vector.app.features.widgets.webview.clearAfterWidget
 import im.vector.app.features.widgets.webview.setupForWidget
 import kotlinx.parcelize.Parcelize
@@ -63,6 +67,7 @@ data class WidgetArgs(
 class WidgetFragment @Inject constructor() :
         VectorBaseFragment<FragmentRoomWidgetBinding>(),
         WebViewEventListener,
+        WebChromeEventListener,
         OnBackPressed {
 
     private val fragmentArgs: WidgetArgs by args()
@@ -75,7 +80,7 @@ class WidgetFragment @Inject constructor() :
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         setHasOptionsMenu(true)
-        views.widgetWebView.setupForWidget(this)
+        views.widgetWebView.setupForWidget(this, this)
         if (fragmentArgs.kind.isAdmin()) {
             viewModel.getPostAPIMediator().setWebView(views.widgetWebView)
         }
@@ -269,6 +274,19 @@ class WidgetFragment @Inject constructor() :
 
     override fun onHttpError(url: String, errorCode: Int, description: String) {
         viewModel.handle(WidgetAction.OnWebViewLoadingError(url, true, errorCode, description))
+    }
+
+    private val permissionResultLauncher = registerForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) { result ->
+        WebviewPermissionUtils.onPermissionResult(result)
+    }
+
+    override fun onPermissionRequest(request: PermissionRequest) {
+        WebviewPermissionUtils.promptForPermissions(
+                title = R.string.room_widget_resource_permission_title,
+                request = request,
+                context = requireContext(),
+                activity = requireActivity(),
+                activityResultLauncher = permissionResultLauncher)
     }
 
     private fun displayTerms(displayTerms: WidgetViewEvents.DisplayTerms) {
--- a/vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
+++ b/vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
@@ -15,17 +15,30 @@
  */
 package im.vector.app.features.widgets.webview
 
+import android.Manifest
 import android.annotation.SuppressLint
 import android.content.Context
 import android.webkit.PermissionRequest
+import androidx.activity.result.ActivityResultLauncher
 import androidx.annotation.StringRes
+import androidx.fragment.app.FragmentActivity
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import im.vector.app.R
+import im.vector.app.core.utils.checkPermissions
 
 object WebviewPermissionUtils {
 
+    private var permissionRequest: PermissionRequest? = null
+    private var selectedPermissions = listOf<String>()
+
     @SuppressLint("NewApi")
-    fun promptForPermissions(@StringRes title: Int, request: PermissionRequest, context: Context) {
+    fun promptForPermissions(
+            @StringRes title: Int,
+            request: PermissionRequest,
+            context: Context,
+            activity: FragmentActivity,
+            activityResultLauncher: ActivityResultLauncher<Array<String>>
+    ) {
         val allowedPermissions = request.resources.map {
             it to false
         }.toMutableList()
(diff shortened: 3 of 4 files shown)


vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
```diff
@@ -15,17 +15,30 @@
...
+import androidx.activity.result.ActivityResultLauncher
 import androidx.annotation.StringRes
+import androidx.fragment.app.FragmentActivity
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import im.vector.app.R
+import im.vector.app.core.utils.checkPermissions
 
 object WebviewPermissionUtils {
 
+    private var permissionRequest: PermissionRequest? = null
```
ouchadam, line 31: I'm a little weary of including mutable static state in the `object` instance, would prefer for this class to become instantiable (and then injected where needed) or have the activity/fragment to hold on to the state instead

vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
```diff
@@ -37,16 +50,50 @@ object WebviewPermissionUtils {
...
+                    }
                 }
                 .setNegativeButton(R.string.room_widget_resource_decline_permission) { _, _ ->
                     request.deny()
                 }
                 .show()
     }
 
+    fun onPermissionResult(result: Map<String, Boolean>) {
+        permissionRequest?.let { request ->
```
ouchadam, line 76: is there a case where we would call `onPermissionResult` without having a valid `permissionRequest` available?

I would lean towards throwing instead of silently swallowing the missing `promptForPermissions` (if the ordering is implicitly required)

Johennes: No, that shouldn't happen, I believe. The order is required as you say. I've added an NPE instead of the `let`. Let me know if that's what you had in mind.

vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
Johennes: I had also thought of this, yeah. I had previously decided against putting the state into the fragment because that way state and logic are split between the fragment and `WebviewPermissionUtils` which I found harder to reason about. I made the utils into a normal class now and used a private instance variable on the fragment. If this needs to be injectable, I might need some help with that. 🙈

vector/src/main/java/im/vector/app/features/widgets/WidgetFragment.kt
```diff
@@ -72,6 +72,7 @@ class WidgetFragment @Inject constructor() :
 
     private val fragmentArgs: WidgetArgs by args()
     private val viewModel: WidgetViewModel by activityViewModel()
+    private val permissionUtils = WebviewPermissionUtils()
```
ouchadam, line 75: if we want to inject this via dagger (not really needed whilst the `WebviewPermissionUtils` has no dependencies of its own) we can include the instance via the injectable constructor of the fragment 

```kotlin
WidgetFragment @Inject constructor() 
```

becomes...

```kotlin
WidgetFragment @Inject constructor(
  private val permissionUtils: WebviewPermissionUtils
) 
```

and the `WebviewPermissionUtils` itself will need to be marked as injectable

```kotlin
class WebviewPermissionUtils @Inject constructor()
```

Fragments in the Element are injectable because of the [fragment module](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/core/di/FragmentModule.kt#L857)

vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
```diff
@@ -37,16 +50,50 @@ object WebviewPermissionUtils {
...
+            val grantedPermissions = selectedPermissions.filter { webPermission ->
+                val androidPermission = webPermissionToAndroidPermission(webPermission)
+                        ?: return@filter true // No corresponding Android permission exists
+                return@filter result[androidPermission]
+                        ?: return@filter true // Android permission already granted before
+            }
+            if (grantedPermissions.isNotEmpty()) {
```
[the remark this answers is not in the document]
Johennes: Hm, I think in this snippet we'd put the _Android_ permission strings into `request.grant(...)`, right? We have to put the _web_ permission strings into that one. That's essentially the reason why I mapped inside the filter.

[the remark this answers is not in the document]
Johennes: I agree that the `filter` clause isn't very easy to parse though. ☹️

vector/src/main/java/im/vector/app/features/widgets/WidgetFragment.kt
Johennes: Ah nice! Thanks a lot for explaining. Made the change to be a good citizen and help remember for next time. 🙂

ouchadam: future us will appreciate the injectable boilerplate being done :smile:

vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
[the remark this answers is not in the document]
Johennes: This kept nagging me so I added a unit test for this part to be able to safely change the implementation. I found that your updated snippet still doesn't work, sadly.

First, `onlySupportedAndroidPermissions` is actually not needed / desired because there are web permissions such as `PermissionRequest.RESOURCE_PROTECTED_MEDIA_ID` for which (I think) no Android system permission exists. In those cases we still want to grant those permissions to the web view (based on the fact that the user confirmed them in the _widget_ permission dialog) so we can't filter them out.

Secondly, the chained call of `mapWithResult` uses the web permission string as key in the _Android_ permission result map which will always result in `null`.

I've tried to resolve this but in the end it comes down to having the current complicated filter/map logic inside `mapWithResult` which seemed more or less equivalent to what I have now. 😕

At least we have the tests now though so this endeavor hasn't been in vein. 😀

vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
```diff
@@ -15,17 +15,33 @@
...
+import java.lang.NullPointerException
+import javax.inject.Inject
 
-object WebviewPermissionUtils {
+class WebviewPermissionUtils @Inject constructor() {
+
+    private var permissionRequest: PermissionRequest? = null
+    private var selectedPermissions = listOf<String>()
 
     @SuppressLint("NewApi")
```
mnaturel, line 37: I am not sure we still need this annotation. I have run the lint task and I had no error...

vector/src/androidTest/java/im/vector/app/features/widgets/WebviewPermissionUtilsTest.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.amshove.kluent.shouldBeEqualTo
+import org.junit.FixMethodOrder
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import org.junit.runners.MethodSorters
+
+@RunWith(JUnit4::class)
+@FixMethodOrder(MethodSorters.JVM)
+class WebviewPermissionUtilsTest : InstrumentedTest {
```
mnaturel, line 32: I am not sure it should be intrumented test here. It could be only unit tests.

mnaturel (commented):
I have added small comments. I need to test it.

changelog.d/6149.bugfix
```diff
@@ -0,0 +1 @@
+Make widget web view request system permissions for camera and microphone (PSF-1061)
```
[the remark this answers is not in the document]
Johennes: Happy to remove it but having it in the changelog / release notes would allow us to directly cross reference with Jira. I think it's essentially also not better or worse than including it in commits and pull requests – which we already do today.

vector/src/main/java/im/vector/app/features/widgets/WidgetFragment.kt
```diff
@@ -271,6 +278,20 @@ class WidgetFragment @Inject constructor() :
         viewModel.handle(WidgetAction.OnWebViewLoadingError(url, true, errorCode, description))
     }
 
+    private val permissionResultLauncher = registerForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) { result ->
```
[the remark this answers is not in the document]
Johennes: The registration methods available there, sadly don't expose the map of permission results: https://github.com/vector-im/element-android/blob/4094a66f3ce65d5fa92ac31b7749c731b3f33cf5/vector/src/main/java/im/vector/app/core/utils/PermissionsTools.kt#L67

vector/src/main/java/im/vector/app/features/widgets/webview/WebviewPermissionUtils.kt
Johennes: Removed it

vector/src/androidTest/java/im/vector/app/features/widgets/WebviewPermissionUtilsTest.kt
Johennes: Good call, made the change

vector/src/main/java/im/vector/app/features/webview/WebChromeEventListener.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.webkit.PermissionRequest
+
+interface WebChromeEventListener {
+
+    /**
+     * Triggered when the web view requests permissions.
+     *
+     * @param request The permission request.
+     */
+    fun onPermissionRequest(request: PermissionRequest) {
```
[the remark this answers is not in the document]
Johennes: I actually mimicked this from `WebViewEventListener` which is a sibling interface. I'm not sure what the specific reason was there though.

vector/src/androidTest/java/im/vector/app/features/widgets/WebviewPermissionUtilsTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.amshove.kluent.shouldBeEqualTo
+import org.junit.FixMethodOrder
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+import org.junit.runners.MethodSorters
+
+@RunWith(JUnit4::class)
+@FixMethodOrder(MethodSorters.JVM)
+class WebviewPermissionUtilsTest {
```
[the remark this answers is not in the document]
Johennes: Good catch, thanks

mnaturel:
After some tests, I have a remark concerning the UX:
Even if I granted the microphone permission once, the dialog with checkbox to request permission appears again. I am wondering if we could directly request Android permissions without the extra dialog with checkboxes? It will be less painful for users.

changelog.d/6149.bugfix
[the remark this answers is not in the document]
ganfra: This will be visible from the release page, so would be better to have a corresponding github issue...

vector/src/main/java/im/vector/app/features/webview/WebChromeEventListener.kt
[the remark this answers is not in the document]
ganfra: Yes please remove the //NO-OP for this as you have only one method. Also you could use a  fun interface, like that: 
```
fun interface ShieldComputer {
      fun onPermissionRequest(request: PermissionRequest)
}
```

vector/src/main/java/im/vector/app/features/widgets/WidgetFragment.kt
```diff
@@ -75,7 +82,7 @@ class WidgetFragment @Inject constructor() :
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         setHasOptionsMenu(true)
-        views.widgetWebView.setupForWidget(this)
+        views.widgetWebView.setupForWidget(this, this)
```
ganfra, line 85: We should avoid having 2 parameters referencing as `this`

ganfra (approved):
Just small remarks otherwise LGTM

Johennes:
> After some tests, I have a remark concerning the UX: Even if I granted the microphone permission once, the dialog with checkboxes to request permission appears again. I am wondering if we could directly request Android permissions without the extra dialog with checkboxes? It will be less painful for users.

The reason the intermediate dialog is needed is that the app currently doesn't keep track of which widget runs in the web view. So if we wouldn't always show the first prompt, then the second prompt would effectively grant the permission to _all_ widgets – which we likely don't want.

I agree that the UX isn't particularly great. This PR was only meant to fix the bug where the web view wasn't actually granted the permission though. We already have https://element-io.atlassian.net/browse/PSF-1033 (Web view permissions are not persisted - Android) to persist the permissions per widget which can serve as the basis for generally improving the flow.

changelog.d/6149.bugfix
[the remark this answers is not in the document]
Johennes: I assumed with the filename it'll link to this PR and, therefore, had intentionally explained the issue that is fixed in detail up in the description to avoid having to create a separate issue just to close it again.

vector/src/main/java/im/vector/app/features/webview/WebChromeEventListener.kt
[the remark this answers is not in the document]
Johennes: Ok, made the change. I decided not to make it a fun interface though since we might in future want to extend it by adding further `WebChromeClient` methods.

vector/src/main/java/im/vector/app/features/widgets/WidgetFragment.kt
Johennes: Good call. I've introduced a compound interface to avoid this now.

mnaturel (approved):

PR #6111, opened 2022-05-20
Sonarqube including coverage on each PR.

michaelkaye:
Changes:

 - remove a lot of post-pr.yml tests (integration and sonarqube) as these run on each PR now
 - Leave behind the UI test as that is not part of 'code coverage'
 - Ignore a number of tests that were failing during the testing process - sonar coverage only works when tests run fine; so ensure we only have fine tests.
    - NB: some of these tests may run fine locally, or may be able to be reintroduced with some tweaks; however at this point I'm explicitly ignoring any test that has failed, so we can start having a reliable test run - i know this is removing some probably good tests, and that is an explicit goal.
  - Update configuration to use a new sonarcloud.io project for uploads; this enables the PR notifications to be set correctly.

Once merged, we'll want to update the list of required checks before merging is OK; at the moment some of the checks are not being run or have been combined; this is global across the project and needs to occur after merging to develop.

After merge I suggest everyone updates their branches against latest develop before attempting a PR merge.

A concern might be usage of the macos runner to get the integration tests running each time. There's not much we can do about that at present.

PR #6150, opened 2022-05-25
DO NOT MERGE: What happens without macos

michaelkaye:
Test what happens to builds if we run on ubuntu not macos

Code changes from #6111 + running on linux runner which might fail. If this works reliably then we can migrate away from macos runner and make ios team happier ^^


michaelkaye:
So maybe option five; which is building on GHA, then running on a off-GHA android emulator.

Issue #6153, opened 2022-05-25
Room upgrade command not doing anything


ericdecanini:
### Steps to reproduce

1. Enable developer mode in advanced settings
2. Enter the `/upgraderoom [newVersion]` command

#### What did you expect?

Upgrade room bottom sheet opens

#### What happened instead?

Nothing

### Your phone model

Emulator

### Operating system version

Android 12

### Application version and app store

1.4.16-dev

### Homeserver

matrix.org

### Will you send logs?

No

PR #6154, opened 2022-05-25
Fixes /upgraderoom command not doing anything

ericdecanini:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes the `/upgraderoom` command not doing anything due to a posted live data event getting overridden before it's observed

## Motivation and context

Closes https://github.com/vector-im/element-android/issues/6153

## Screenshots / GIFs

| Before | After |
|-|-|
| https://user-images.githubusercontent.com/20701752/170278411-d41ca380-5176-484d-92f2-f682fc0d6d3b.mp4 | https://user-images.githubusercontent.com/20701752/170278455-82228685-0104-4572-aa77-63ed7dd93214.mp4 |

## Tests

- Enable developer mode in advanced settings
- Go to a room and enter `/upgraderoom [anyNumber]`
- See upgrade bottom sheet now pops up

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 12

## Checklist


vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -472,14 +472,16 @@ class MessageComposerViewModel @AssistedInject constructor(
...
-                            _viewEvents.post(MessageComposerViewEvents.SlashCommandResultOk(parsedCommand))
-                            popDraft()
+                            viewModelScope.launch {
+                                _viewEvents.set(
+                                        MessageComposerViewEvents.ShowRoomUpgradeDialog(
+                                                parsedCommand.newVersion,
+                                                room.roomSummary()?.isPublic ?: false
+                                        )
+                                )
+                                _viewEvents.set(MessageComposerViewEvents.SlashCommandResultOk(parsedCommand))
```
bmarty, line 482: The pb was we were posting 2 ViewEvents in a row?
Maybe keep only the first one and assume the receiver will do the same treatment than when the second one is sent? This will avoid us to add new API to the DataSource.

ericdecanini: I tested this but the receiver doesn't accomplish the same. Without this ok command, future messages won't send. Even if we handle things on the receiver side, it's only confusing the intent on both sides.

I do also think that having access to a non async `emit` in the data source is very important, otherwise we'd always have to find workarounds like this whenever we want to post two events in succession

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
```diff
@@ -25,6 +25,9 @@ interface DataSource<T> {
 }
 
 interface MutableDataSource<T> : DataSource<T> {
+
+    suspend fun set(value: T)
```
fedrunov, line 29: I understand where `set` comes from, but it looks weird here for me. Maybe `emit` or `blockingPost` could fit better. What do you think?

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
```diff
@@ -58,6 +65,10 @@ open class PublishDataSource<T> : MutableDataSource<T> {
         return mutableFlow
     }
 
+    override suspend fun set(value: T) {
+        mutableFlow.emit(value)
```
fedrunov, line 69: PublishDataSource has `DROP_OLDEST` policy for buffer overflow and emit will only suspend when policy is set to `SUSPEND`, so it won't work here 🤔

fedrunov (changes requested):

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
ericdecanini: This only applies for `tryEmit`. As `emit` is already a suspend function, it doesn't make use of a buffer
https://github.com/Kotlin/kotlinx.coroutines/issues/2387#issuecomment-727220851

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
ericdecanini: I don't have very strong feelings towards using a name that's not `set`, but `blockingPost` sounds like something that shouldn't suspend (sort of like blocking the thread its on like `blockingGet`), and `emit` confuses the below `post` methods because ultimately both are `emit` methods

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
fedrunov: Read the doc for `emit`: 
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-shared-flow/emit.html

`This call can suspend only when the [BufferOverflow] strategy is [SUSPEND] and there are subscribers collecting this shared flow.`

It's suspend, but you need a proper policy to be set, to make it work

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
fedrunov: well `blockingGet` was just to provide any alternative to `emit` :) I agree that having `emit` and `post` is confusing, but `set` is even more confusing. You don't actually set anything, you just post event and ensure that next event will be posted only after first one was consumed. 

Maybe `postSafely`?

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
ericdecanini: Interesting 🤔 but at the same time, I wonder if it's an issue though. The buffer doesn't suspend, but instead (in this case) would simply drop the oldest and work synchronously (within the coroutine). What's important is we have that functionality so that events "set" at the same time don't get lost as they do with the current `post` method

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
ericdecanini: `postSafely` assumes that `post` has a chance of throwing which it doesn't. Main reason for opting for `set` is to have those parallels with LiveData's `set` and `post` as the behaviour we ultimately achieve here are the same

ericdecanini: I'm starting to think `emit` is a good idea know 👀😂

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
fedrunov: I'm not sure what do you mean. If `emit` is not suspended on buffer overflow, emitting new events after buffer is overflown will cause older events to be dropped. Which mean if you emit 10 things sequentially it wont' wait until first is fully consumed before emitting second. Instead it will emit all of them as they are not suspend functions and will drop older when buffer capacity is reached. So you will lost events

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
fedrunov: and I'm starting to think that param is much better one :) 

@bmarty could you join this conversation? I'd like to hear you thoughts

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
```diff
@@ -472,14 +472,16 @@ class MessageComposerViewModel @AssistedInject constructor(
...
                         is ParsedCommand.UpgradeRoom                       -> {
-                            _viewEvents.post(
-                                    MessageComposerViewEvents.ShowRoomUpgradeDialog(
-                                            parsedCommand.newVersion,
-                                            room.roomSummary()?.isPublic ?: false
-                                    )
-                            )
-                            _viewEvents.post(MessageComposerViewEvents.SlashCommandResultOk(parsedCommand))
-                            popDraft()
+                            viewModelScope.launch {
```
bmarty, line 475: Why do we need `viewModelScope.launch` here and not at line 482?

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
bmarty: In our case that the PR is aiming to fix, there are only 2 Events, and we have a subscriber, so I think we will not have any overflow of the buffer.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
ericdecanini: 482 is still within this same scope

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerViewModel.kt
ganfra: Technically, we should't be able to have multiple Events sent in a row before the previous one is processed, otherwise this should more be part of ViewState. 
That being said, I'm not against changing the API to only use `emit` instead of `tryEmit`

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
bmarty: @ganfra will have a look on it.

fedrunov (approved):

vector/src/main/java/im/vector/app/core/utils/DataSource.kt
ganfra: Ok so, we hit this issue because the `post` method is called on the main thread, while collecting events is also made on the main thread. Just launching a new coroutine will fix this.
But we might still want to increase the buffer so it won't happen again. I have added some small tests on this behavior. 
Let me know what you think before we remove your new methods.

ericdecanini:
Thanks for the commit @ganfra, it looks good to me and the issue is still fixed with the changes

ganfra:
> Thanks for the commit @ganfra, it looks good to me and the issue is still fixed with the changes (still requesting for your approval for the pr)

We should then remove the `suspend fun emit ` from DataSource. Can you do it? then I'll approve the PR.

ericdecanini:
> > Thanks for the commit @ganfra, it looks good to me and the issue is still fixed with the changes (still requesting for your approval for the pr)
> 
> We should then remove the `suspend fun emit ` from DataSource. Can you do it? then I'll approve the PR.

Whoops, missed that it was still there. All done

ganfra (approved):
LGTM

vector/src/test/java/im/vector/app/core/utils/DataSourceTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        publishDataSource.post(1)
+
+        publishDataSource.stream()
+                .test(this)
+                .assertNoValues()
+                .finish()
+    }
+
+    @Test
+    fun `given PublishDataSource with a large enough buffer size, when posting values after observing, then only the latest values are observed`() = runTest {
```
bmarty, line 47:
```suggestion
    fun `given PublishDataSource with a large enough buffer size, when posting values after observing, then all values are observed`() = runTest {
```

(IIUC)

bmarty (approved):
Just a small suggestion about a test name, else LGTM!

PR #6164, opened 2022-05-26
Ask the user to confirm urls which contain unicode direction overrides

### LINKED ISSUE #6163: Ask for confirmation when a url with unicode directional characters are present
Inform the user when they've tapped a potentially misleading URL which contains unicode directional override characters

To stop potential URL masquerading

By reusing the URL confirmation check when a unicode directional override is present in the URL

_No response_

### Additional context

_No response_

ouchadam:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Checks for the presence of Right to Left unicode directional overrides when tapping on a URL, if one is found we'll show the user the URL with the overrides disabled and ask them to confirm they want to continue.

## Motivation and context

Fixes #6163 

Newer versions of Android (31+) seemingly handle the oddness out of the box by avoiding marking the URLs as clickable links, however previous versions will still attempt to link out to them, although the formatting is often broken.

## Screenshots / GIFs

| BEFORE | AFTER | 
| --- | --- |
2022-05-26T10:26:35,157524612+01:00|Screenshot_20220526_113846

## Tests

- Using Element web /devtools construct and post a link containing a RtL override `\u202E` 
- Using an Android 25 device view and tap the link
- Notice a popup informing the user that the URL may not be what they expect

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 21, 25, 28, 31

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
@@ -1927,10 +1927,14 @@ class TimelineFragment @Inject constructor(
             if (!isManaged) {
                 when {
                     url.containsRtLOverride()                                                  -> {
-                        displayUrlConfirmationDialog(title.ensureEndsLeftToRight(), url.filterDirectionOverrides())
+                        displayUrlConfirmationDialog(
+                                seenUrl = title.ensureEndsLeftToRight(),
+                                actualUrl = url.filterDirectionOverrides(),
+                                continueTo = url
+                        )
                     }
                     title.isValidUrl() && url.isValidUrl() && URL(title).host != URL(url).host -> {
-                        displayUrlConfirmationDialog(title.ensureEndsLeftToRight(), url)
+                        displayUrlConfirmationDialog(title, url)
                     }
                     else                                                                       -> {
                         openUrlInExternalBrowser(requireContext(), url)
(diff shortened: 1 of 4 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -1922,30 +1925,36 @@ class TimelineFragment @Inject constructor(
                         }
                     })
             if (!isManaged) {
-                if (title.isValidUrl() && url.isValidUrl() && URL(title).host != URL(url).host) {
-                    MaterialAlertDialogBuilder(requireActivity(), R.style.ThemeOverlay_Vector_MaterialAlertDialog_NegativeDestructive)
```
ouchadam, line 1926: the dialog displaying has been extracted to its own function

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -1922,30 +1925,36 @@ class TimelineFragment @Inject constructor(
...
-                                openUrlInExternalBrowser(requireContext(), url)
-                            }
-                            .setNegativeButton(R.string.action_cancel, null)
-                            .show()
-                } else {
-                    // Open in external browser, in a new Tab
-                    openUrlInExternalBrowser(requireContext(), url)
+                when {
+                    url.containsRtLOverride() || (title.isValidUrl() && url.isValidUrl() && URL(title).host != URL(url).host) -> {
+                        displayUrlConfirmationDialog(title.ensureEndsLeftToRight(), url.filterDirectionOverrides())
```
ouchadam, line 1930: we have to ensure the text returns to LtR at the end with `title.ensureEndsLeftToRight()`, otherwise the override character will cause the rest of the string resource to also become inverted 

| WITHOUT LTR OVERRIDE |
| --- |
|Screenshot_20220526_124304

bmarty: `url.filterDirectionOverrides()` will also be used to open the url, if the user accepts. Will it work fine? Or considerer adding `url` as a new param.

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -1922,30 +1925,36 @@ class TimelineFragment @Inject constructor(
...
-                            .setPositiveButton(R.string._continue) { _, _ ->
-                                openUrlInExternalBrowser(requireContext(), url)
-                            }
-                            .setNegativeButton(R.string.action_cancel, null)
-                            .show()
-                } else {
-                    // Open in external browser, in a new Tab
-                    openUrlInExternalBrowser(requireContext(), url)
+                when {
+                    url.containsRtLOverride() || (title.isValidUrl() && url.isValidUrl() && URL(title).host != URL(url).host) -> {
```
bmarty, line 1929: Maybe split into 2 cases rather than using `||` for clarity?

bmarty (commented):
Thanks.
The existing case was working finally? (you told me it was not the case)

ouchadam:
> Thanks. The existing case was working finally? (you told me it was not the case)

kind of~ after looking into the logic, the url detection only takes into account urls with a protocol, for example `[google.com](https://youtube.com)` will skip the check

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ouchadam: ah! no it should be a separate parameter, will update :+1:

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ouchadam: e6198d7bf601aa73d955d42e3248b34f93af6075

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
ouchadam: 67f1929784a0554fdd90ce099d160fd4b9a6d98a

bmarty (approved):
Thanks for the update!

Issue #5974, opened 2022-05-09
Remove Login2


ouchadam:
Once the FTUE is complete we can remove the unused proof of concept Login2 package

PR #6171, opened 2022-05-27
[SDK] Allow passwords to be set at the point of reset confirmation

### LINKED ISSUE #6169: [SDK] Allow supplying a password to the `LoginWizard.resetPasswordMailConfirmation` step
### Description

Currently the SDK `LoginWizard` expects both the email and password to be supplied when resetting an account password. This is specific to Element's current reset flow rather than dictated by the spec/api. 

The new FTUE reset password flow separates the email confirmation and new password confirmation steps, matching the reset password endpoints.

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Fixes #6169 

Updates the `LoginWizard` reset api to allow passing the new password at the point of confirmation
- Makes the `newPassword` parameter in `LoginWizard.resetPassword` optional
- A missing password at the point of `resetPasswordMailConfirmed` will throw an `IllegalStateException` 
- Also introduces a `ResetState` to the `OnboardingViewModel` in preparation for the flow changes

## Motivation and context

To enable the FTUE reset password journey which involves supplying a password after the reset email has been confirmed instead of the current flow which involves passing the email and password upfront.

## Screenshots / GIFs

No UI changes

## Tests

No production changes

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 31

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/login/LoginWizard.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/login/LoginWizard.kt
@@ -63,14 +63,15 @@ interface LoginWizard {
      * [resetPasswordMailConfirmed] is successfully called.
      *
      * @param email an email previously associated to the account the user wants the password to be reset.
-     * @param newPassword the desired new password
+     * @param newPassword the desired new password, can be optionally set here or as part of [resetPasswordMailConfirmed]
      */
-    suspend fun resetPassword(email: String,
-                              newPassword: String)
+    suspend fun resetPassword(email: String, newPassword: String? = null)
 
     /**
      * Confirm the new password, once the user has checked their email
      * When this method succeed, tha account password will be effectively modified.
+     *
+     * @param newPassword the desired new password, required if a password was not supplied to [resetPassword]
      */
-    suspend fun resetPasswordMailConfirmed()
+    suspend fun resetPasswordMailConfirmed(newPassword: String? = null)
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/login/LoginWizard.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/login/LoginWizard.kt
@@ -63,15 +63,14 @@ interface LoginWizard {
      * [resetPasswordMailConfirmed] is successfully called.
      *
      * @param email an email previously associated to the account the user wants the password to be reset.
-     * @param newPassword the desired new password, can be optionally set here or as part of [resetPasswordMailConfirmed]
      */
-    suspend fun resetPassword(email: String, newPassword: String? = null)
+    suspend fun resetPassword(email: String)
 
     /**
      * Confirm the new password, once the user has checked their email
      * When this method succeed, tha account password will be effectively modified.
      *
-     * @param newPassword the desired new password, required if a password was not supplied to [resetPassword]
+     * @param newPassword the desired new password
      */
-    suspend fun resetPasswordMailConfirmed(newPassword: String? = null)
+    suspend fun resetPasswordMailConfirmed(newPassword: String)
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DefaultLoginWizard.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DefaultLoginWizard.kt
@@ -30,6 +30,7 @@ import org.matrix.android.sdk.internal.auth.data.ThreePidMedium
 import org.matrix.android.sdk.internal.auth.data.TokenLoginParams
 import org.matrix.android.sdk.internal.auth.db.PendingSessionData
 import org.matrix.android.sdk.internal.auth.registration.AddThreePidRegistrationParams
+import org.matrix.android.sdk.internal.auth.registration.AddThreePidRegistrationResponse
 import org.matrix.android.sdk.internal.auth.registration.RegisterAddThreePidTask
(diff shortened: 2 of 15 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/login/LoginWizard.kt
```diff
@@ -63,14 +63,15 @@ interface LoginWizard {
      * [resetPasswordMailConfirmed] is successfully called.
      *
      * @param email an email previously associated to the account the user wants the password to be reset.
-     * @param newPassword the desired new password
+     * @param newPassword the desired new password, can be optionally set here or as part of [resetPasswordMailConfirmed]
      */
-    suspend fun resetPassword(email: String,
-                              newPassword: String)
+    suspend fun resetPassword(email: String, newPassword: String? = null)
```
bmarty, line 68: When I created this API, I was following the UI, when the form was asking for the email and the new password. I have checked and it's fine to provide the Pwd later.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/login/DefaultLoginWizard.kt
```diff
@@ -136,4 +136,14 @@ internal class DefaultLoginWizard(
         // Set to null?
         // resetPasswordData = null
     }
+
+    private fun PendingSessionData.readResetPasswordDataOrThrow(newPassword: String?): Pair<AddThreePidRegistrationResponse, String> {
+        return when (resetPasswordData) {
+            null -> throw IllegalStateException("developer error, no reset password in progress")
+            else -> {
+                val password = newPassword ?: resetPasswordData.newPassword ?: throw IllegalStateException("developer error, no new password set")
```
bmarty, line 144: Do we want to enforce that `newPassword` and `resetPasswordData.newPassword` are not not null simultaneously? Or are the same?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/auth/login/LoginWizard.kt
```diff
@@ -63,14 +63,15 @@ interface LoginWizard {
      * [resetPasswordMailConfirmed] is successfully called.
      *
      * @param email an email previously associated to the account the user wants the password to be reset.
-     * @param newPassword the desired new password
+     * @param newPassword the desired new password, can be optionally set here or as part of [resetPasswordMailConfirmed]
```
bmarty, line 66: Maybe add the point that if the pwd is provided here, it will be temporarily stored in the SDK encrypted DB.
To avoid that consider passing the pwd in [resetPasswordMailConfirmed]

bmarty (commented):
LGTM;
But having two ways to do the same thing is maybe not ideal from a SDK user POV.

Have you considered fully remove `newPassword` parameter in `resetPassword`?

I would vote to do so, no matter if there is an API break.

ouchadam:
> Have you considered fully remove `newPassword` parameter in `resetPassword`?

I originally had this but wasn't sure of the consequences of no longer persisting the password within the `pendingState` in the database

I'm happy to update and remove!

ouchadam:
will close and rework to provide a single way to reset the password

vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
```diff
@@ -247,21 +247,15 @@ class OnboardingViewModel @AssistedInject constructor(
 
     private fun handleLoginWithToken(action: OnboardingAction.LoginWithToken) {
         val safeLoginWizard = loginWizard
+        setState { copy(isLoading = true) }
 
-        if (safeLoginWizard == null) {
-            setState { copy(isLoading = false) }
```
ouchadam, line 252: by making the `loginWizard` non null (which the compiler is warning is impossible) we can simplify the branching

bmarty: I have noticed this warning. I think this is relatively new, probably since a Kotlin version upgrade. I am also wondering why the compiler is not failing, since warnings are errors on our project. So looks more like a IDE inspection warning.

bmarty (approved):
LGTM!

vector/src/main/java/im/vector/app/features/login2/LoginViewModel2.kt
```diff
@@ -470,7 +472,8 @@ class LoginViewModel2 @AssistedInject constructor(
 
             currentJob = viewModelScope.launch {
                 try {
-                    safeLoginWizard.resetPasswordMailConfirmed()
+                    val state = awaitState()
+                    safeLoginWizard.resetPasswordMailConfirmed(state.resetPasswordNewPassword!!)
```
Claire1817, line 476: is it possible to avoid the not-null assertion operator ?

ouchadam: it's possible (we're avoiding it in the [OnboardingViewModel](https://github.com/vector-im/element-android/pull/6171/files#diff-31d1cb8d12c34d852a52405908bc8f1a427a1f08485a662527a6ff5c11e7053eR459) but I've taken a shortcut here in the LoginViewModel2 as this entire package will be [removed](https://github.com/vector-im/element-android/issues/5974) once the FTUE is finished 

for context, `Login2` is the proof of concept which has lead to the FTUE project

I'd prefer to avoid making tooo many changes to this package, but would be happy to make the change if you feel it's worth doing :+1:

Claire1817 (approved):
Tested: Ok :)

Issue #4133, opened 2021-10-01
Better handle room upgrades within Spaces 


BillCarsonFr:
### Steps to reproduce

Reflect web changes https://github.com/vector-im/element-web/issues/19246

### Your use case

The client should resolve upgrades locally where possible to deduplicate space listings

ericdecanini:
@BillCarsonFr I've done some investigating and testing on this issue and my question is do we still need to action this?

On testing, there's no situation where a listing is duplicated. When a room is upgraded and an old member is not invited to the upgraded room, they may see the old room in their room list that they can go into and see that it is upgraded. This seems to be normal behaviour. This happens thanks to receiving a tombstone state event from the server on upgrade.

The moment they get an invite to the new room, the old room's listing disappears and gets replaced by the invite.

ericdecanini:
https://user-images.githubusercontent.com/20701752/170678438-5baac528-c21d-4e46-afe4-9654c4724d69.mp4

The above shows me being in a room, then another user upgrading the room, looking for any duplicates in room list and explore rooms, then being invited into the room and looking for any duplicates.

Fortunately we don't have any situations where there are duplicates, but there are situations where the same room is either in Home or in the Space but never both. I'll investigate this

PR #6200, opened 2022-05-31
Fixes room not being in space after upgrade

ericdecanini:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes room not being in space after upgrade and possible duplications in Explore Rooms

## Motivation and context

Closes https://github.com/vector-im/element-android/issues/4133

## Screenshots / GIFs

https://user-images.githubusercontent.com/20701752/171157326-36fe8661-9f7a-4351-aa2d-729d658353f6.mp4

## Tests

- In a space, go to a room with two of your test accounts and use the `/upgraderoom 7` to upgrade the room from 9 to 7 (yes this still counts as an upgrade). Don't automatically reinvite users
- On your other test account, ensure that you still see the room in the same space but going into it shows you that it's been upgraded
- Invite your account and see that the room disappears and gets replaced by the invite
- After accepting the invite see that in explore rooms you no longer see the pre-upgrade room

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Android 12 (both)

## Checklist


vector/src/main/java/im/vector/app/features/home/room/detail/upgrade/UpgradeRoomViewModelTask.kt
```diff
@@ -84,8 +84,6 @@ class UpgradeRoomViewModelTask @Inject constructor(
 //                                autoJoin = currentInfo.autoJoin ?: false,
                                 suggested = currentInfo.suggested
                         )
-
-                        parentSpace.removeChildren(params.roomId)
```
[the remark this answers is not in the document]
ericdecanini: I would guess this was the first attempt at removing the duplication, and though it worked for that purpose, you would no longer see the pre-upgraded room in the space but you'd see it in Home

ouchadam (approved):
I'm not tooo familiar with `Spaces`, codewise this looks good to me! :100:

vector/src/main/java/im/vector/app/features/spaces/explore/SpaceDirectoryController.kt
```diff
@@ -209,4 +211,7 @@ class SpaceDirectoryController @Inject constructor(
             }
         }
     }
+
+    private fun SpaceChildInfo.isUpgradedRoom(data: SpaceDirectoryState) =
+            data.knownRoomSummaries.any { it.roomId == childRoomId && it.versioningState != VersioningState.NONE }
```
ouchadam, line 216: super nit: could be a little easier on the brain to lift the `isUpgraded = it.versioningState != VersioningState.NONE` to an extension on the `VersioningState` 

has the benefit of keeping the conditions here positive and future proofs this logic from any possible new _versioning states_ which aren't upgrades

ericdecanini: I agree, good shout!

vector/src/main/java/im/vector/app/features/spaces/explore/SpaceDirectoryController.kt
```diff
@@ -110,6 +110,7 @@ class SpaceDirectoryController @Inject constructor(
                     ?.filter {
                         it.parentRoomId == (data.hierarchyStack.lastOrNull() ?: data.spaceId)
                     }
+                    ?.filterNot { it.isUpgradedRoom(data) }
```
fedrunov, line 113: it will filter out rooms with `UPGRADED_ROOM_NOT_JOINED` VersioningState, is it intentional? If room was upgraded, but user haven't joined upgraded room yet, what user will see? We change state from `NONE` to `UPGRADED_ROOM_NOT_JOINED` on TombStone event and so it could happen while user is on a background and doesn't interact with the app.

vector/src/main/java/im/vector/app/features/home/room/detail/upgrade/UpgradeRoomViewModelTask.kt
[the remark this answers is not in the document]
fedrunov: This could be valuable for other cases, some counters could use spaces children number or some check (if space is empty or not), so absence of removal could cause some side effects.

@BillCarsonFr you added this line, maybe you could verify if removing it could cause any problems?

fedrunov (changes requested):

vector/src/main/java/im/vector/app/features/spaces/explore/SpaceDirectoryController.kt
ericdecanini: This is intentional yeah, this only applies to the old version of upgraded rooms, so them not being active there's no use having them shown in Explore Rooms

BillCarsonFr: Just to clarify, we should filter out if the room is an old version of a room that is also in the space (we want to avoid apparant duplicates in the list).

vector/src/main/java/im/vector/app/features/home/room/detail/upgrade/UpgradeRoomViewModelTask.kt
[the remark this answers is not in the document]
fedrunov: I was thinking about some places, where we could get rooms without applying this filter, like getting space''s room count or something like that. 🤔

vector/src/main/java/im/vector/app/features/spaces/explore/SpaceDirectoryController.kt
fedrunov: So, when room is upgraded, everyone in the room gets tombstone event and state is changed to `UPGRADED_ROOM_NOT_JOINED` until they will join new room. `isUpgradedRoom` will return true and so such users won't see room in a list. Shouldn't we hide it only after users will join new Room?

fedrunov: But you don't keep it, you filter out any room with `UPGRADED_ROOM_NOT_JOINED`, so it's different from web, no?

ericdecanini: I mean keep in the sense that it's not deleted from the space in the db, it's just filtered out in the UI

fedrunov (approved):

BillCarsonFr (changes requested):
I noticed that this change is not consistent with web (did some tests).
Could you please sync with web and see why they are not doing it?
https://github.com/vector-im/element-meta/issues/460
We should be consistent.

ericdecanini:
> I noticed that this change is not consistent with web (did some tests). Could you please sync with web and see why they are not doing it? [vector-im/element-meta#460](https://github.com/vector-im/element-meta/issues/460) We should be consistent.

We have this issue to decide and fix what should happen with regards to the room showing after an upgrade. This PR focuses on fixing the orphaning (issue you linked) and deduplication of the room after an upgrade. I think we should keep the focus of this PR on only that and any other changes, we address it in this issue below. This will also give us enough time to fix it for both web and android and gain parity again

https://github.com/vector-im/element-android/issues/6411

Issue #6191, opened 2022-05-30
[SDK]  Add support for signing out all devices when resetting password


ouchadam:
### Description

Update the SDK to align with the js-sdk https://github.com/matrix-org/matrix-react-sdk/pull/8259 by adding a parameter to `sign out all devices` when resetting the password [spec](https://spec.matrix.org/v1.2/client-server-api/#post_matrixclientv3accountpassword)

This will allow consumers of the SDK to opt out of signing out of all devices if wanted

PR #6207, opened 2022-05-31
SDK - Logout all devices

ouchadam:
Draft as this depends on #6171 

## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

#6191 Adding SDK support for MSC2457 - opting in/out of logging out of all devices when changing password

- Updates the database version and provides a migration for the new `HomeserverCapabilities.canControlLogoutDevices` field 

## Motivation and context

To enable the next steps of the _FTUE forgotten password flow_ to introduce a new sign out all devices check box.

## Screenshots / GIFs

no UI changes here, is made use of in https://github.com/vector-im/element-android/pull/6263

## Tests

No logic changes

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):


Florian14 (approved):
LGTM

fedrunov:
Just to understand - when we logout user from all sessions we do that only server-side? There is no event coming from server to clients, which we need to handle? Like push notification to let user know that they was logged out and won't receive notifications anymore?

ouchadam:
> Just to understand - when we logout user from all sessions we do that only server-side? There is no event coming from server to clients, which we need to handle? Like push notification to let user know that they was logged out and won't receive notifications anymore?

Correct, it's server controlled, there's no real-time push event but the users token is invalidated and they'll receive a 401 the next time they attempt to make a request with their locally stored token (which the clients already handle)

for context, logging out from all sessions is the current default behaviour for Android, the change in this PR enables the SDK to opt out

fedrunov (approved):

PR #6214, opened 2022-05-31
Homeserver version without patch number parsing

### LINKED ISSUE #6017: Homeserver version parser should support version without patch
For instance `v1.4` should be correctly parsed, currently only `v1.4.0` will work.

The code is here: https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt#L41

Also discussed in #5997 

Also required to close this issue: add some unit tests on the parser (could candidate for TDD).

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Fixes #6017 Supporting Homeserver versions without a patch number.

Versions without a patch are considered as having a patch of `0` eg 1.5 is parsed as `major: 1 minor: 5 patch: 0`

- Makes the Homeserver version regex support optional patch numbers 
- Adds unit tests around the parsing

## Motivation and context

To support version formats defined by the spec 

## Screenshots / GIFs

No UI changes

## Tests

Handled by unit tests 

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/extensions/Strings.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/extensions/Strings.kt
@@ -27,3 +27,8 @@ fun CharSequence.ensurePrefix(prefix: CharSequence): CharSequence {
  * Append a new line and then the provided string.
  */
 fun StringBuilder.appendNl(str: String) = append("\n").append(str)
+
+/**
+ * Returns null if the string is empty.
+ */
+fun String.ensureNotEmpty() = ifEmpty { null }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
@@ -38,14 +38,14 @@ internal data class HomeServerVersion(
     }
 
     companion object {
-        internal val pattern = Regex("""[r|v](\d+)\.(\d+)\.(\d+)""")
+        internal val pattern = Regex("""[r|v](\d+)\.(\d+)(?:\.(\d+))?""")
 
         internal fun parse(value: String): HomeServerVersion? {
             val result = pattern.matchEntire(value) ?: return null
             return HomeServerVersion(
                     major = result.groupValues[1].toInt(),
                     minor = result.groupValues[2].toInt(),
-                    patch = result.groupValues[3].toInt()
+                    patch = result.groupValues.getOptional(index = 3, default = "0").toInt()
             )
         }
 
@@ -59,3 +59,5 @@ internal data class HomeServerVersion(
         val v1_3_0 = HomeServerVersion(major = 1, minor = 3, patch = 0)
     }
 }
+
+private fun List<String>.getOptional(index: Int, default: String) = getOrNull(index)?.ifEmpty { default } ?: default
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
@@ -16,6 +16,8 @@
 
 package org.matrix.android.sdk.internal.auth.version
 
+import org.matrix.android.sdk.api.extensions.ensureNotEmpty
+
 /**
  * Values will take the form "rX.Y.Z".
  * Ref: https://matrix.org/docs/spec/client_server/latest#get-matrix-client-versions
@@ -45,7 +47,7 @@ internal data class HomeServerVersion(
             return HomeServerVersion(
                     major = result.groupValues[1].toInt(),
                     minor = result.groupValues[2].toInt(),
-                    patch = result.groupValues.getOptional(index = 3, default = "0").toInt()
+                    patch = result.groupValues.getOrNull(index = 3)?.ensureNotEmpty()?.toInt() ?: 0
             )
         }
 
@@ -59,5 +61,3 @@ internal data class HomeServerVersion(
         val v1_3_0 = HomeServerVersion(major = 1, minor = 3, patch = 0)
     }
 }
-
-private fun List<String>.getOptional(index: Int, default: String) = getOrNull(index)?.ifEmpty { default } ?: default
--- a/matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersionTest.kt
+++ b/matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersionTest.kt
@@ -22,27 +22,23 @@ import org.junit.Test
 class HomeServerVersionTest {
 
     @Test
-    fun `given a semantic version, when parsing then converts to home server version`() {
-        val cases = buildList {
-            addAll(
-                    listOf(
-                            case("0.5.1", expected = aVersion(0, 5, 1)),
-                            case("1.0.0", expected = aVersion(1, 0, 0)),
-                            case("1.10.3", expected = aVersion(1, 10, 3))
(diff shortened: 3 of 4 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
```diff
@@ -38,14 +38,14 @@ internal data class HomeServerVersion(
     }
 
     companion object {
-        internal val pattern = Regex("""[r|v](\d+)\.(\d+)\.(\d+)""")
+        internal val pattern = Regex("""[r|v](\d+)\.(\d+)(?:\.(\d+))?""")
```
ouchadam, line 41: this is the fix, to make the last capture group optional

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersionTest.kt
```diff
@@ -0,0 +1,56 @@
...
+                case("r1", expected = null),
+                case("a", expected = null),
+                case("1a.2b.3c", expected = null),
+                case("r", expected = null)
+        )
+
+        (supportedVersions + unsupportedVersions).forEach { (input, expected) ->
+            val result = HomeServerVersion.parse(input)
+
+            assertEquals(expected, result, "Expected $input to be $expected but got $result")
```
ouchadam, line 44: ended up using the `assertEquals` directly instead of `shouldBeEqual` in order to provide a custom message

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
bmarty: It should work, but strange that it's not just `[r|v](\d+)\.(\d+)(\.(\d+))?`. Regex syntax I guess.

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersionTest.kt
bmarty: I am always bitten by the parameter ordering. With JUnit the message is first.

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersionTest.kt
```diff
@@ -0,0 +1,56 @@
...
+}
+
+private fun aVersion(major: Int, minor: Int, patch: Int) = HomeServerVersion(major, minor, patch)
+private fun case(input: String, expected: HomeServerVersion?) = Case(input, expected)
+
+private fun List<Case>.withPrefixes(vararg prefixes: String) = map { case ->
+    prefixes.map { prefix -> case.copy(input = "$prefix${case.input}") }
+}.flatten()
+
+private data class Case(val input: String, val expected: HomeServerVersion?)
```
bmarty, line 56: OOI why all that private stuff is not inside `class HomeServerVersionTest`?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
```diff
@@ -59,3 +59,5 @@ internal data class HomeServerVersion(
         val v1_3_0 = HomeServerVersion(major = 1, minor = 3, patch = 0)
     }
 }
+
+private fun List<String>.getOptional(index: Int, default: String) = getOrNull(index)?.ifEmpty { default } ?: default
```
bmarty, line 63: `ifEmpty` is maybe not necessary. I do not think it can be empty, since we have `\d+`. Maybe add a test with "v1.4.", expecting `null`.

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersionTest.kt
```diff
@@ -0,0 +1,56 @@
...
+
+class HomeServerVersionTest {
+
+    @Test
+    fun `given a semantic version, when parsing, then converts to home server version`() {
+        val supportedVersions = listOf(
+                case("1.5", expected = aVersion(1, 5, 0)),
+                case("0.5.1", expected = aVersion(0, 5, 1)),
+                case("1.0.0", expected = aVersion(1, 0, 0)),
+                case("1.10.3", expected = aVersion(1, 10, 3))
```
bmarty, line 30: Could have a trailing comma.

bmarty (commented):
Thanks. Some remarks.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
ouchadam: the optional block returns an empty string, with `ifEmpty` removed

```
java.lang.NumberFormatException: For input string: ""
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:662)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at org.matrix.android.sdk.internal.auth.version.HomeServerVersion$Companion.parse$matrix_sdk_android_debug(HomeServerVersion.kt:48)
```

have updated the test cases to include trailing `.` https://github.com/vector-im/element-android/pull/6214/commits/4501c7cf45ebd37bd5fdc190a0ea5cff2bc68341

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersionTest.kt
ouchadam: it could be inside the class, although there's nothing instance related in the helpers which is why I opted for having the helpers at the module level (I tend to lift pure functions out of classes)

bmarty: Very good point, thanks!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/auth/version/HomeServerVersion.kt
bmarty: I see, thanks. I would avoid creating this extension and would have written at line 48:
```kotlin
patch = result.groupValues.getOrNull(3)?.ensureNotEmpty()?.toInt() ?: 0
```

Since, this is not necessary to do `"0".toInt()`.

The extension:
```kotlin
fun String.ensureNotEmpty() = takeIf { it.isNotEmpty() }
```
could be added to the core <del>module</del> package. I am surprised we do not have create it yet, we have 76 occurrences of `.takeIf { it.isNotEmpty() }` in the codebase :)

This is obviously not a blocker :).

bmarty (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/extensions/Strings.kt
```diff
@@ -27,3 +27,8 @@ fun CharSequence.ensurePrefix(prefix: CharSequence): CharSequence {
  * Append a new line and then the provided string.
  */
 fun StringBuilder.appendNl(str: String) = append("\n").append(str)
+
+/**
+ * Returns null if the string is empty.
+ */
+fun String.ensureNotEmpty() = ifEmpty { null }
```
bmarty, line 34: thanks, better than my proposal!

PR #6218, opened 2022-06-01
Replace legacy lock screen implementation

### LINKED ISSUE #6217: Replace PIN code screen implementation

Replace PF-LockScreen library with our own implementation.

While this library is really handy, the internal APIs used have been deprecated and modifying it can be difficult. It also lacks some features as key invalidation on new biometric method enrollment, which is needed by some of our customers.

Re-implement this feature following our code guidelines, using the new APIs available and simplifying how it works internally.

Modifying our PF-LockScreen fork was a possibility, but in the long term not depending on a library that's no longer maintained and uses deprecated APIs is probably the best way to go.

### Additional context

_No response_

Yes

jmartinesp:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Our PF-LockScreen fork is replaced by our own implementation of this feature. This should be easier to customise and maintain, it allows us to use the new biometric APIs from Android, removing the deprecated ones, and adds missing features like biometric key invalidation on new method enrolled.

In the beginning this was a separate module, but due to dependencies with some classes in the vector module it had to be included as part of this module. It's in its own `im.vector.app.features.pin.lockscreen` package so it can be split in the future. Sadly, the resources needed are mostly in the `library/ui-styles` module and are not as easy to split as they could be, but most of them are prefixed using `lockscreen_` to be easily distinguished.

## Motivation and context

Fixes #6217, depends on #6270 .

## Screenshots / GIFs

No UI changes, we're using the same resources and styles.

## Tests

- Have a logged in user.
- Go to Settings -> Security & Privacy -> Protect access.
- Enable PIN, set up a PIN code.
- Enable biometrics too if possible, confirm with a fingerprint (this can also be done on emulators from the settings dialog, but emulator with API 29 had some issues with fingerprints).
- To test this feature more easily, disable "Require PIN after 2 minutes" so the PIN code screen is displayed every time you return to the app.

Which things could be useful to test:
- Enabling / disabling biometrics.
- Failing the biometric prompt too many times, it should automatically disable biometric auth.
- Failing PIN code auth too many times should force a log out.
- Changing PIN.

Note: if the user updates the app and they already had both PIN code and biometric authentication enabled, _key invalidation will only start working after the first successful biometric auth_. A system/biometric key will be created and maintained. Also, key invalidation is only available since Android 6.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 9, 10, 12 (both emulator and physical devices).

## Checklist


vector/src/androidTest/java/im/vector/app/features/pin/lockscreen/biometrics/BiometricHelperTests.kt
```diff
@@ -0,0 +1,224 @@
...
+    private val biometricManager = mockk<BiometricManager>(relaxed = true)
+    private val lockScreenKeyRepository = mockk<LockScreenKeyRepository>(relaxed = true)
+
+    @Before
+    fun setup() {
+        clearAllMocks()
+    }
+
+    @Test
+    fun canUseWeakBiometricAuthReturnsTrueIfIsFaceUnlockEnabledAndCanAuthenticate() {
```
ganfra, line 63: Would be more readable to use back-ticks like this:

```
fun `i am a test method`(){
  //
}
```

vector/src/main/java/im/vector/app/core/di/LockScreenModule.kt
```diff
@@ -0,0 +1,84 @@
...
+            pinCodeLength = 4,
+            isWeakBiometricsEnabled = false,
+            isDeviceCredentialUnlockEnabled = false,
+            isStrongBiometricsEnabled = true,
+            needsNewCodeValidation = true,
+    )
+
+    @Provides
+    @Singleton
+    fun provideKeyHelper(@ApplicationContext context: Context) = LockScreenKeyRepository(context, "vector-lockscreen")
```
ganfra, line 55: You should annotate your constructor of LockScreenKeyRepository directly with `@Inject` so you don't need to do it in module.
Also if it doesn't hold a state or init is not heavy it shouldn't be scoped

ganfra (changes requested):
Nice work! 
Some remarks, mostly on the DI.
Also, could you use svg instead of xxxhdpi pngs?

vector/src/main/java/im/vector/app/core/di/LockScreenModule.kt
jmartinesp: About the repository being declared in the module, I did it like this to configure the `baseName` "vector-lockscreen" parameter. I could do it by using a default value for baseName or by declaring this value with some qualifier and then injecting it automatically, but it seems too complex for this.

As for state, it _could happen_ that 2 different badly used repositories created some key at the same time, that's why I used singleton here.

vector/src/main/java/im/vector/app/features/pin/lockscreen/ui/LockScreenAction.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package im.vector.app.features.pin.lockscreen.ui
+
+import androidx.fragment.app.FragmentActivity
+import im.vector.app.core.platform.VectorViewModelAction
+
+sealed class LockScreenAction : VectorViewModelAction {
+    data class PinCodeEntered(val value: String) : LockScreenAction()
+    data class ShowBiometricPrompt(val callingActivity: FragmentActivity) : LockScreenAction()
```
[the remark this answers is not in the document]
jmartinesp: We need to pass it to `BiometricHelper`,  so I don't think we can avoid doing this without changing the whole structure 😞.

vector/src/main/java/im/vector/app/features/pin/lockscreen/views/LockScreenCodeView.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Number of digits entered.
+     */
+    val enteredDigits: Int get() = code.size
+
+    /**
+     * Callback called when the PIN code has been completely entered.
+     */
+    var onCodeCompleted: ((String) -> Unit)? = null
```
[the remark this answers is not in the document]
jmartinesp: Isn't it a bit redundant? It will get automatically translated to a lambda anyway. In any case, I'll change it if that's what we prefer.

vector/src/main/java/im/vector/app/core/di/LockScreenModule.kt
ganfra: Ok for this class then

vector/src/main/java/im/vector/app/features/pin/lockscreen/views/LockScreenCodeView.kt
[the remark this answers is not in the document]
ganfra: As you want, I don't have preference for this

vector/src/main/java/im/vector/app/features/pin/lockscreen/ui/LockScreenListener.kt
```diff
@@ -19,45 +19,45 @@ package im.vector.app.features.pin.lockscreen.ui
 /**
  * Listener class to be notified of any event that could happen in the lock screen UI.
  */
-open class LockScreenListener {
+interface LockScreenListener {
     /**
      * In PIN creation mode, called when the first PIN code has been entered.
      */
-    open fun onFirstCodeEntered() {}
+    fun onFirstCodeEntered() {}
```
ganfra, line 26: The default impl could just return Unit but it's a detail
`fun onFirstCodeEntered() = Unit`

vector/src/main/java/im/vector/app/features/pin/lockscreen/di/LockScreenModule.kt
```diff
@@ -52,19 +54,17 @@ object LockScreenModule {
...
+    fun provideKeyRepository(@ApplicationContext context: Context) = LockScreenKeyRepository(context, "vector-lockscreen")
 
     @Provides
-    @Singleton
-    fun provideBiometricUtils(
+    fun provideBiometricHelper(
             @ApplicationContext context: Context,
             lockScreenKeyRepository: LockScreenKeyRepository,
             configurationProvider: LockScreenConfiguratorProvider,
     ) = BiometricHelper(context, lockScreenKeyRepository, configurationProvider, BiometricManager.from(context))
```
ganfra, line 64: Use Inject annotation for BiometricHelper and PinCodeHelper, no?

vector/src/main/java/im/vector/app/features/pin/lockscreen/ui/LockScreenFragment.kt
```diff
@@ -164,7 +164,7 @@ class LockScreenFragment : VectorBaseFragment<FragmentLockScreenBinding>() {
 
     private fun setupCodeView(lockScreenCodeView: LockScreenCodeView, configuration: LockScreenConfiguration) = with(lockScreenCodeView) {
         codeLength = configuration.pinCodeLength
-        onCodeCompleted = { code ->
+        onCodeCompleted = LockScreenCodeView.CodeCompletedListener { code ->
```
ganfra, line 167: Use fonction reference?

ganfra (commented):
Just small remaining questions

vector/src/main/java/im/vector/app/features/pin/lockscreen/ui/LockScreenFragment.kt
jmartinesp: Sorry, I'm not following. How should I use a function reference here? By extracting its contents to private function first?

jmartinesp: Oh, ok. Honestly, I don't think its worth it, we'd just be moving the `viewModel.handle(LockScreenAction.PinCodeEntered(code))` code into a `handleCodeCompleted(code: String)` which probably doesn't add much value. If this was used in several places it could make sense.

jmartinesp:
> Thanks for the changes, all good for me. Note that I didn't test it, do you prefer I do or are you confident enough?

I tested everything that came to mind, but honestly another round of testing for edge cases wouldn't hurt if you have the time.

library/ui-styles/src/main/res/values/styles_pin_code.xml
```diff
@@ -22,13 +22,13 @@
     </style>
 
     <style name="PinCodeDeleteButtonStyle">
-        <item name="android:src">@drawable/delete_lockscreen_pf</item>
+        <item name="android:src">@drawable/lockscreen_delete</item>
```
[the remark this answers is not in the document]
jmartinesp: Oops, true. @ganfra also mentioned this, but I forgot to re-check those. These drawables are taken from the old lib, but I'll try to replace them with svg if possible.

Issue #3830, opened 2021-08-13
Switching spaces is slow


robintown:
#### Describe the bug

On my modestly sized account with ~8 spaces containing ~10 rooms each, switching spaces is generally slow. It often takes 1 to 2 seconds from the moment when I tap on a space in the side panel, to when I actually see the room list populated with that space's rooms. In my opinion, the performance of this common interaction should be improved.

#### Smartphone (please complete the following information):
 - Device: Google Pixel 3a
 - OS: LineageOS 15.1

#### Additional context
 - App version and store: 1.1.14 - F-Droid
 - Homeserver: townsendandsmith.ml

SimonBrandner:
Similar experience on Motorola G7 plus

SimonBrandner:
This seems to go away for me after I clear my cache but quickly comes back which I find quite odd, are there any guesses at what could be the cause of this? Are we the only people experiencing this?

zaibon:
I'm also experiencing this issue but even worst. It can rakes tens of seconds for the switch to a new space actually happen.

rda0:
> I am experiencing this too. Switching spaces is extremely slow, sometimes takes minutes or does not work at all. Clearing the cache does not help. This is especially annoying, since somehow the active space gets randomly switch around (or by opening via notifications?) and I always need to switch back to home.
> 
> Version: 1.4.2 Phone: Samsung Galaxy S9

Deleting cache, storage, completely reinstalling the app and logging in a new session resolved the problem. It is now taking me mostly less than 1s to switch spaces.

SimonBrandner:
The issue has been resolved for me as well - I had to switch to a different account which seems to have done the trick.

Is anyone else still seeing this?

robintown:
I am currently seeing general slowness across the entire app, which would suggest memory issues of some sort. Not sure whether that counts as this issue.

robintown:
By disabling threads, I've now definitely gotten the app to a place where it is responsive overall, but still slow (1~10 seconds) to switch spaces.

SimonBrandner:
> By disabling threads, I've now definitely gotten the app to a place where it is responsive overall, but still slow (1~10 seconds) to switch spaces.

For me, space switching has dramatically improved after my account migration from `matrix.envs.net` to `element.io` though now I am experiencing very slow syncs...

timokoesters:
I think sync might actually be the reason. On my device I have these logs on every sync:

[...]

PR #5860, opened 2022-04-28
Replaces subtitle in Search Rooms with room context rather than last event

ericdecanini:
Drafting for early discussion purposes

Assignees assigned for taking a look at my query below, didn't want to overfill the reviewers section
 
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Instead of showing the last event (last message sent, etc.), room list items in the search screen will show room context (parent space, user id, canonical alias, etc)

## Motivation and context

Closes https://github.com/vector-im/element-android/issues/3943

This brings Android into parity with desktop clients.

Getting the space that the room belonged to was not as straightforward as initially expected. Spaces that rooms belonged to often weren't being saved properly into Realm which meant that rooms that belonged to spaces had an empty list of parents. This is why this PR contains updates to `RoomSummaryUpdater` and related classes

Other changes in this PR include: Centering of the room title when the subtitle is empty

I also took this opportunity to cleanup some commented code

## Screenshots / GIFs

| Before | After |
|-|-|
| image | image |

As you can see above, for space rooms, we show the space or subspace they belong to. For non-space rooms, we show their alias if its available. For dms, we show the user id of the other recipient. Otherwise, we show nothing

## Tests

- From the home screen, navigate to the search screen (by clicking on the filter icon in the action bar)
- Search for rooms in spaces, standalone rooms, dm rooms, dms, and rooms with aliases. See that the subtitle is appropriate for each of them and matches what you find on the desktop clients

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 10

## Checklist

- [ ] Changes has been tested on an Android device or Android emulator with API 21
- [ ] UI change has been tested on both light and dark themes
- [ ] Accessibility has been taken into account. See https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#accessibility
- [x] Pull request is based on the develop branch
- [x] Pull request includes a new file under ./changelog.d. See https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#changelog
- [ ] Pull request includes screenshots or videos if containing UI changes
- [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
- [x] You've made a self review of your PR
[...]

PR #6233, opened 2022-06-03
Fix slow paged room list regression

BillCarsonFr:
## Type of change

- [x] Bugfix

## Content

A new parameter was added to resolve the list of parent summaries when loading the rooms paged list.
This param is affecting performances a lot when you have some decent space hierarchy. And it was called unproperly when displaying the regular room list (should only be needed for search/filter rooms).

Also the new added `flattenParents` field (that was slowing the queries) was never used. Instead the space parent was used. I think improperly because it only reflect the presence of a `m.space.parent`relation, that is not used to decide hierarchy (`m.space.child`) are used for that.

@ericdecanini  can you confirm?

See https://github.com/vector-im/element-android/issues/3830 regarding recent degradation of performances.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
@@ -226,8 +226,12 @@ interface RoomService {
     ): LiveData<PagedList<RoomSummary>>
 
     /**
-     * TODO Doc.
-     * @param getFlattenParents When true, the list of known parent and grand parent summaries will be resolved.
+     * Get's a live paged list from a filter that can be dynamically updated.
+     *
+     * @param queryParams The filter to use
+     * @param pagedListConfig The paged list configuration (page size, initial load, prefetch distance...)
+     * @param sortOrder defines how to sort the results
+     * @param getFlattenParents When true, the list of known parents and grand parents summaries will be resolved.
      * This can have significant impact on performance, better be used only on manageable list (filtered by displayName, ..).
      */
     fun getFilteredPagedRoomSummariesLive(
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
@@ -227,11 +227,14 @@ interface RoomService {
 
     /**
      * TODO Doc.
+     * @param getFlattenParents When true, the list of known parent and grand parent summaries will be resolved.
+     * This can have significant impact on performance, better be used only on manageable list (filtered by displayName, ..).
      */
     fun getFilteredPagedRoomSummariesLive(
             queryParams: RoomSummaryQueryParams,
             pagedListConfig: PagedList.Config = defaultPagedListConfig,
-            sortOrder: RoomSortOrder = RoomSortOrder.ACTIVITY
+            sortOrder: RoomSortOrder = RoomSortOrder.ACTIVITY,
+            getFlattenParents: Boolean = false,
     ): UpdatableLivePageResult
 
     /**
--- a/vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryItemFactory.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryItemFactory.kt
@@ -207,7 +207,9 @@ class RoomSummaryItemFactory @Inject constructor(
 
     private fun getSearchResultSubtitle(roomSummary: RoomSummary): String {
         val userId = roomSummary.directUserId
-        val spaceName = roomSummary.spaceParents?.firstOrNull()?.roomSummary?.name
+        val spaceName = roomSummary.flattenParents
+                .takeIf { it.isNotEmpty() }
+                ?.joinToString(", ") { it.name }
         val canonicalAlias = roomSummary.canonicalAlias
 
         return (userId ?: spaceName ?: canonicalAlias).orEmpty()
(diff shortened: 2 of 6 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
```diff
@@ -227,11 +227,14 @@ interface RoomService {
...
      * TODO Doc.
+     * @param getFlattenParents When true, the list of known parent and grand parent summaries will be resolved.
+     * This can have significant impact on performance, better be used only on manageable list (filtered by displayName, ..).
      */
     fun getFilteredPagedRoomSummariesLive(
             queryParams: RoomSummaryQueryParams,
             pagedListConfig: PagedList.Config = defaultPagedListConfig,
-            sortOrder: RoomSortOrder = RoomSortOrder.ACTIVITY
+            sortOrder: RoomSortOrder = RoomSortOrder.ACTIVITY,
+            getFlattenParents: Boolean = false,
```
ericdecanini, line 237:
```suggestion
            getFlattenParents: Boolean = false, // Heavily impacts performance when true. Only use when space parent is absolutely required
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
```diff
@@ -227,11 +227,14 @@ interface RoomService {
 
     /**
      * TODO Doc.
+     * @param getFlattenParents When true, the list of known parent and grand parent summaries will be resolved.
+     * This can have significant impact on performance, better be used only on manageable list (filtered by displayName, ..).
```
ericdecanini, line 231: Feels weird to have a partial java doc covering only one parameter. I would resolve this TODO and complete the javadoc here (and preferably for the method above too, to avoid confusion later)
```suggestion
     * Gets filtered page of room summaries so that [insert use of filtered here]
     *
     * @param queryParams - tba
     * @param pagedListConfig - tba
     * @param sortOrder - tba
     * @param getFlattenParents When true, the list of known parent and grand parent summaries will be resolved.
```

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryItemFactory.kt
```diff
@@ -207,7 +207,9 @@ class RoomSummaryItemFactory @Inject constructor(
 
     private fun getSearchResultSubtitle(roomSummary: RoomSummary): String {
         val userId = roomSummary.directUserId
-        val spaceName = roomSummary.spaceParents?.firstOrNull()?.roomSummary?.name
+        val spaceName = roomSummary.flattenParents
+                .takeIf { it.isNotEmpty() }
+                ?.joinToString(", ") { it.name }
```
ericdecanini, line 212: Although this is nice, I'm thinking now to maintain parity with web, we only take the last element in `flattenParents` for the subtitle. If design would deem it's better to have all parents separated by comma, we can have a meta ticket to change that across all platforms

ericdecanini (changes requested):
I think the doc change is very much needed. There's a similar method in `RoomService` called `getPagedRoomSummariesLive` that feels like the default method used while `getFilteredPagedRoomSummariesLive` sounds like it's only used in `FilteredRoomsActivity`. This issue highlights that this isn't the case, and it would be good to make that clearer in the code through these java docs

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryItemFactory.kt
BillCarsonFr: Doesn't look to me what web is taking the last one only. What are the requirement exactly?
Can you take over it as I believe you are the one that initialy developed it?

fedrunov:
@BillCarsonFr were you able to reproduce or you just assume that flattened parents is the cause? I've tested on many spaces and there were no problems with db read, so if you can reproduce, could you share your setup?

Also, issue was created before flatten parents flag was added, so I'm not sure this actually fixes problem

BillCarsonFr:
> @BillCarsonFr were you able to reproduce or you just assume that flattened parents is the cause? I've tested with many spaces and there were no problems with db read, so if you can reproduce, could you share your setup?
> 
> Also, issue was created before flatten parents flag was added ([PR](https://github.com/vector-im/element-android/pull/5860)). So I'm not sure this actually fixes problem

I can reproduce 100% with my personal account.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
bmarty: Good point, but please add this mention in the Kdoc and not here, or it will not be exported.

bmarty (commented):
Please add a file for the changelog

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/RoomService.kt
bmarty: If you use `tba` you should keep the `TODO`

bmarty (approved):
LGTM after a static review.
Please do not merge on develop, I will create a hotfix, as it's blocking the release 1.4.18.

bmarty:
All the 4 commits have been cherry-picked to create the hotfix 1.4.19. The changes are now on `develop` and on `main`.

waclaw66:
Another regression https://github.com/vector-im/element-android/issues/6272 that didn't make it to 1.4.19...

PR #5909, opened 2022-05-03
Fixes crash when accepting VOIP calls

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes a crash when receiving and consuming call events #5421 (tentatively fixes #5591 and possibly other VOIP quirks)

- The event insertion logic is designed to be single threaded however due to coroutine continuation, there's unintended multiple thread access when the _processing_ and _post processing_ overlap
- Fixed by converting the launching to a flow which will sequentially process the launch logic on the single threaded scope

## Motivation and context

To fix a VOIP crash

## Screenshots / GIFs

- Adds a small delay to the postProcessing step
- Send 5 messages 

#### Before 

- Notice multiple concurrent launches    

```
Launch: LIVE_ENTITY_BACKGROUND
Process start: pool-7-thread-1
Process end: LIVE_ENTITY_BACKGROUND
Post start: LIVE_ENTITY_BACKGROUND

Launch: LIVE_ENTITY_BACKGROUND
Process start: pool-7-thread-1
Process end: LIVE_ENTITY_BACKGROUND
Post start: LIVE_ENTITY_BACKGROUND

Launch: LIVE_ENTITY_BACKGROUND
Process start: pool-7-thread-1
Process end: LIVE_ENTITY_BACKGROUND
Post start: LIVE_ENTITY_BACKGROUND

Launch: LIVE_ENTITY_BACKGROUND
Process start: pool-7-thread-1
Process end: LIVE_ENTITY_BACKGROUND
Post start: LIVE_ENTITY_BACKGROUND

Launch: LIVE_ENTITY_BACKGROUND
Process start: pool-7-thread-1
Process end: LIVE_ENTITY_BACKGROUND
Post start: LIVE_ENTITY_BACKGROUND
Post end: LIVE_ENTITY_BACKGROUND
Post end: LIVE_ENTITY_BACKGROUND
Post end: LIVE_ENTITY_BACKGROUND
Post end: LIVE_ENTITY_BACKGROUND
Post end: LIVE_ENTITY_BACKGROUND
```

#### After

- Notice sequential launches, each waiting for the post processing to complete    

```
Launch: LIVE_ENTITY_BACKGROUND
Process start: pool-7-thread-1
Process end: LIVE_ENTITY_BACKGROUND
Post start: LIVE_ENTITY_BACKGROUND
Post end: LIVE_ENTITY_BACKGROUND

[...]

PR #6279, opened 2022-06-10
Fix EventInsertLiveObserver gets blocked by reverting and adding lock…

### LINKED ISSUE #6278: [SDK]  EventInsertLiveObserver processing is not triggered
### Description

Sometime EventInsertLiveObserver gets blocked so there is no event being processed anymore.
Probably introduced by this change: https://github.com/vector-im/element-android/commit/c09a93c171ec4e2cc1942bbe4ea5613e1ddb5768

ganfra:
Fix #6278 
Introduce a Lock instead of relying on Flow.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/EventInsertLiveObserver.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/EventInsertLiveObserver.kt
@@ -19,10 +19,9 @@ package org.matrix.android.sdk.internal.database
 import com.zhuinden.monarchy.Monarchy
 import io.realm.RealmConfiguration
 import io.realm.RealmResults
-import kotlinx.coroutines.flow.MutableSharedFlow
-import kotlinx.coroutines.flow.launchIn
-import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import org.matrix.android.sdk.internal.database.mapper.asDomain
 import org.matrix.android.sdk.internal.database.model.EventEntity
 import org.matrix.android.sdk.internal.database.model.EventInsertEntity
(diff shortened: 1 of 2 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/EventInsertLiveObserver.kt
```diff
@@ -39,65 +38,58 @@ internal class EventInsertLiveObserver @Inject constructor(
...
-                // don't use copy from realm over there
-                val copiedEvent = EventInsertEntity(
-                        eventId = it.eventId,
-                        eventType = it.eventType
-                ).apply {
-                    insertType = it.insertType
+        observerScope.launch {
+            lock.withLock {
+                if (!results.isLoaded || results.isEmpty()) {
+                    return@launch
```
bmarty, line 51: Probably equivalent, but is it safer to use `return@withLock` instead of `return@launch`?

onurays (approved):
LGTM, do you think there will be a performance effect?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/EventInsertLiveObserver.kt
ganfra: Honestly I don't know, but I can change if you prefer

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/EventInsertLiveObserver.kt
```diff
@@ -39,65 +38,58 @@ internal class EventInsertLiveObserver @Inject constructor(
...
-        onResultsChangedFlow
-                .onEach { handleChange(it) }
-                .launchIn(observerScope)
-    }
-
     override fun onChange(results: RealmResults<EventInsertEntity>) {
-        if (!results.isLoaded || results.isEmpty()) {
-            return
-        }
-        observerScope.launch { onResultsChangedFlow.emit(results) }
```
ouchadam, line 58: do we know the cause of the issue/reproduction steps?

ouchadam:
for context, the PR that introduced the change https://github.com/vector-im/element-android/pull/5909

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/EventInsertLiveObserver.kt
ganfra: I reproduce a lot. Just sent some reactions, if you go back on room list and open again the timeline, there is a high chance you don't see the reactions as the event is not aggregated.
For the exact cause I don't know, but looks like the inner channel gets blocked by the the looper thread...

ouchadam (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/EventInsertLiveObserver.kt
ouchadam: very strange :thinking: 

I haven't been able to reproduce on my side, the flow always seems to emit

| GIF | 
| --- |
reactions

bmarty: (same for me, I cannot repro)

bmarty (changes requested):
Thanks for the update.
"Request changes" state to prevent merging this PR, I will cherry-pick commit on the hotfix branch.

bmarty:
Commits are included in 1.4.20

Issue #5782, opened 2022-04-19
FTUE - Link connect to server to new Server Selection


ouchadam:
- When tapping Connect to server the user should be presented the sign up page with the Edit server screen already open

daniellekirkwood:
We will test with this case and see if we need to change or pivot what we're doing... Assigning to the 3 of us to ensure we don't miss it.

daniellekirkwood:
## Is this a V1 or a V2 issue? 
It's not strictly needed, it's not a blocker for the release but it would improve the user's experience. 

@ouchadam to hide the button in the meantime

daniellekirkwood:
No longer exists

PR #6293, opened 2022-06-13
FTUE - Minor tweaks

ouchadam:
## Type of change

- [x] WIP Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Making the confetti celebration during the FTUE onboarding always appearing during the account creation breaker. Previously we would show the confetti after the optionally personalisation steps.

- Hiding the connect to server button as we'll revisit for the V2 implementation #5782 

## Motivation and context

- We want all users to see the confetti during account creation 
- We want to avoid showing previous flows

##### Confetti move

| BEFORE | AFTER |
| --- | --- |
|before-account-created|after-account-created

##### Hidden connect to server

| BEFORE | AFTER |
| --- | --- |
|Screenshot_20220613_111152|Screenshot_20220613_111400

## Tests

- With the combined register flag enabled 
- Complete the account registration 
- Notice the confetti when asked to personalise or go home

- With the combined register flag enabled 
- When selecting the app use case
- Notice there's no connect to server button

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 31


bmarty (approved):
LGTM, thanks!

ganfra (approved):
LGTM

PR #6296, opened 2022-06-13
Allow sharing text mimetype content via android's share menu

### LINKED ISSUE #6285: Allow 'Share' function to work on .ics/calendar files
### Your use case

Dear Element Team: I love this app already.
My request/suggestion would be to allow 'Share' function to send .ics/calendar files thru Element. (Broadly, if possible, to allow the function of even unknown file types to be sent through.
Pics, vids, etc send effortlessly. With calendar entries, with the file type .ics will not send. Clicking share from my calendar to send to Element only returns the message 'Couldn't handle share data'.
Lastly, a secondary request/suggestion: If it were possible to send larger file sizes,  that would be convenient, for sending other already supported file types.
Thank you very much for the consideration. I appreciate your time and am very grateful that you for helping to preserve my privacy. Thanks again. Sincerely,   

I am able to send the .ics file type through standard email.
I am also able to change the file extension to a .txt, sucessfully send the file via Element, then the receiver is able to use the file by changing it back to the .ics.

### Additional context

N/A

Yes

ouchadam:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Allows `text/*` content types to be shared via android share menu as file attachments. 

## Motivation and context

Fixes #6285 

## Screenshots / GIFs

|Before|After|
|-|-|
|share-ics|after-share-calendar|

## Tests

- Attempt to share an .ics file to element from another app, like google chrome

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/vector/src/main/java/im/vector/app/features/attachments/AttachmentsHelper.kt
+++ b/vector/src/main/java/im/vector/app/features/attachments/AttachmentsHelper.kt
@@ -214,7 +214,7 @@ class AttachmentsHelper(val context: Context, val callback: Callback) : Restorab
                         it.toContentAttachmentData()
                     }
             )
-        } else if (type.startsWith("application") || type.startsWith("file") || type.startsWith("*")) {
+        } else if (type.startsWith("application") || type.startsWith("file") || type.startsWith("text") || type.startsWith("*")) {
             callback.onContentAttachmentsReady(
                     MultiPicker.get(MultiPicker.FILE).getIncomingFiles(context, intent).map {
                         it.toContentAttachmentData()
--- a/changelog.d/6285.feature
+++ b/changelog.d/6285.feature
@@ -1,0 +1,1 @@
+Allow sharing text based content via android's share menu (eg .ics files)


vector/src/main/java/im/vector/app/features/attachments/AttachmentsHelper.kt
```diff
@@ -214,7 +214,7 @@ class AttachmentsHelper(val context: Context, val callback: Callback) : Restorab
                         it.toContentAttachmentData()
                     }
             )
-        } else if (type.startsWith("application") || type.startsWith("file") || type.startsWith("*")) {
+        } else if (type.startsWith("application") || type.startsWith("file") || type.startsWith("text") || type.startsWith("*")) {
```
ouchadam, line 217: do we want to explicitly handle all the possible cases? or should our else case allow handling any type as a file?

Claire1817: For me i prefer to explicitly handle all the possible cases to keep the control of what we send. It's easy to add a new type if a user need it.

Claire1817 (approved):

fedrunov (approved):
LGTM

PR #6297, opened 2022-06-13
Making bug report screenshots opt in instead of opt out

### LINKED ISSUE #6261: Rageshake submission dialog should default screenshots to off
Not send screenshots with a rageshake by default

Because it's too easy to miss that a screenshot is being attached, so too easy to send one by default and at least in my case the screenshots are rarely relevant to the rageshake or often contain private data

By defaulting the screenshot option to off

_No response_

### Additional context

_No response_

No

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : Bug report defaults

## Content

Makes screenshots opt-in rather than opt-out as screenshots contain messages which the logs do not

## Motivation and context

Fixes #6261 

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot_20220613_180354|Screenshot_20220613_181111|
 

## Tests

- Open a bug report
- Notice the send screenshot toggle is disabled

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/vector/src/main/res/layout/activity_bug_report.xml
+++ b/vector/src/main/res/layout/activity_bug_report.xml
@@ -154,7 +154,7 @@
                     android:layout_height="wrap_content"
                     android:layout_marginStart="10dp"
                     android:layout_marginEnd="10dp"
-                    android:checked="true"
+                    android:checked="false"
                     android:text="@string/send_bug_report_include_screenshot" />
 
                 <ImageView
--- a/changelog.d/6261.misc
+++ b/changelog.d/6261.misc
@@ -1,0 +1,1 @@
+Making screenshots in bug reports opt in instead of opt out
--- a/vector/src/main/java/im/vector/app/features/rageshake/BugReportActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/rageshake/BugReportActivity.kt
@@ -259,7 +259,7 @@ class BugReportActivity : VectorBaseActivity<ActivityBugReportBinding>() {
     }
 
     private fun onSendScreenshotChanged() {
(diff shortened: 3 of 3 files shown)


vector/src/main/res/layout/activity_bug_report.xml
```diff
@@ -154,7 +154,7 @@
                     android:layout_height="wrap_content"
                     android:layout_marginStart="10dp"
                     android:layout_marginEnd="10dp"
-                    android:checked="true"
+                    android:checked="false"
```
bmarty, line 157: This toggle also affects the visibility of the screenshot on this screen (`bug_report_screenshot_preview`).
Maybe hide the screenshot by default, or make it always visible? But maybe stay like that, so the user can see the screenshot that will be sent if they check the box, and when they check and uncheck, the screenshot will be GONE.
Also, if the checkbox is not checked, and the user sends the report, maybe add a dialog like "Do you want to include a screenshot to your report?". I am afraid that we will get less screenshots with this change, even if generally there are quite useless.

bmarty (approved):
LGTM, just some thoughts that you may already have considered yourself.

yostyle (approved):
LGTM

vector/src/main/res/layout/activity_bug_report.xml
ouchadam: :+1:  for making the screenshot preview always visible  https://github.com/vector-im/element-android/pull/6297/commits/d8814974c56550ca96cb77c44b8bd368adecb2bb

for the UX change this feels like something that would need product input, most bug reports which include screenshots often don't contain the reported bug (eg crashes)

PR #6307, opened 2022-06-14
Missing unit test coverage 

### LINKED ISSUE #6299: Sonarqube missing coverage
We're seeing 0% coverage reports on the sonar task which is causes PRs to be unmergeable

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Fixes #6299 missing unit test coverage  

- The current android gradle plugin we're using (7.1.3) has a bug which causes unit test coverage generated with `testCoverageEnabled` to be invalid, we can generate the coverage via jacoco instead.
- The fix is to split the unit tests and instrumentation tasks so that we can change the configuration and collect all the results
- Also disables instrumenting the code for debug builds which should speed up compilation a little~

## Motivation and context

The include unit test coverage in the sonar reports

## Screenshots / GIFs

https://github.com/vector-im/element-android/pull/6323

2022-06-16T15:36:09,854306238+01:00

## Tests

- with the gradle build option `testCoverageEnabled = true` run the unit tests `./gradlew testGplayDebugUnitTest`
- attempt to generate a jacoco report from the resulting `unit_test_coverage/*.exec` file
- notice no coverage is picked up

## Tested devices
N/A CI related

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,7 +51,7 @@ jobs:
           emulator-build: 7425822
           script: | 
             ./gradlew unitTestsWithCoverage --stacktrace $CI_GRADLE_ARG_PROPERTIES
-            ./gradlew instrumentationTestsWithCoverage -Pandroid.testInstrumentationRunnerArguments.notPackage=im.vector.app.ui --stacktrace $CI_GRADLE_ARG_PROPERTIES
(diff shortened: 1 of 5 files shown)


bmarty (approved):
LGTM, but see my question on the linked PR.

onurays (approved):
LGTM. There are failing actions but not related to this PR.

PR #6320, opened 2022-06-15
CreatePollViewModel unit tests [PSF-1122]

onurays:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Motivation and context

Increase CreatePollViewModel test coverage to 100% (83/83)
--

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewModelTest.kt
```diff
@@ -0,0 +1,202 @@
...
+
+    private fun createPollViewModel(pollMode: PollMode): CreatePollViewModel {
+        return if (pollMode == PollMode.EDIT) {
+            CreatePollViewModel(CreatePollViewState(editPollArgs), fakeSession)
+        } else {
+            CreatePollViewModel(CreatePollViewState(createPollArgs), fakeSession)
+        }
+    }
+
+    @Before
```
mnaturel, line 66: We should unmock mocked objects after each test in general. We can add:

```
@After
fun tearDown() {
    unmockkAll()
}
```

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewModelTest.kt
```diff
@@ -0,0 +1,202 @@
...
+import org.matrix.android.sdk.api.session.room.model.message.PollType
+import org.matrix.android.sdk.api.session.room.send.SendService
+import org.matrix.android.sdk.api.util.Cancelable
+
+class CreatePollViewModelTest {
+
+    @get:Rule
+    val mvrxTestRule = MvRxTestRule()
+
+    private val fakeSession = mockk<Session>()
```
mnaturel, line 53: We should use/create FakeObject classes to be reusable for other unit tests. Here you can use `FakeSession` and `FakeRoom`. You can add `givenXXX` methods in them to mock some calls (see existing examples). For the `SendService`, a `FakeSendService` can be created and instantiated inside the `FakeRoom` class. We may only use `mockk<>` inside tests class when it is very specific to the tested class or simple object.

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewModelTest.kt
```diff
@@ -0,0 +1,202 @@
...
+                .finish()
+    }
+
+    @Test
+    fun `given there is not any options when the question is added then poll cannot be created and more options can be added`() = runTest {
+        val createPollViewModel = createPollViewModel(PollMode.CREATE)
+        createPollViewModel.handle(CreatePollAction.OnQuestionChanged(fakeQuestion))
+
+        // We need to wait for createPollViewModel.onChange is triggered
+        delay(10)
```
mnaturel, line 88: You can use `assertStatesChanges` instead of `assertState` to avoid having to add a delay which may lead to test result inconsistencies.

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewStates.kt
```diff
@@ -0,0 +1,71 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.app.features.poll.create
+
+import im.vector.app.features.poll.PollMode
+import kotlin.random.Random
+
+object CreatePollViewStates {
```
mnaturel, line 22: I think we should move this class to the test package as it is only used as testing purpose. We could also rename it into `FakeCreatePollViewStates` or we could also declare all the fake data at the top of the related test file: `CreatePollViewModelTest` before the class declaration. I have a preference for the latter as we can see tha fake data and the tests at the same time.

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewStates.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    const val fakeRoomId = "fakeRoomId"
```
mnaturel, line 24: For all `const val`, it is preferable to use upper case format like: 

`const val A_FAKE_ROOM_ID = "fakeRoomId"`

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewModelTest.kt
ouchadam: different cpu/computers will be able to execute faster/slower, we should avoid relying on delays whenever possible :crossed_fingers: 

`// We need to wait for createPollViewModel.onChange is triggered` in the test enviroment we should control all the dispatching and block asynchronous calls (which is the purpose of `runTest` and `MvRxTestRule`)

ouchadam: different cpu/computers will be able to execute faster/slower, we should avoid relying on delays whenever possible :crossed_fingers: 

`// We need to wait for createPollViewModel.onChange is triggered` in the test enviroment we should control all the dispatching and block asynchronous calls (which is the purpose of `runTest` and `MvRxTestRule`)

onurays: I tried `assertStatesChanges` and verify with a timeout but no luck.

ouchadam: the issue here looks like your `handle` is not blocking correctly :thinking: 

the `ViewModel.test()`extension is how we're capturing emissions from the `ViewModel`, if we interact with the `ViewModel` before `test()` then we'll miss state events

```kotlin
val test = createPollViewModel.test() // register state and event listeners

viewModel.handle()  // blocking - causes a state update

test.assert() // asserts against the captured states
```

however here I can see the `handle` emissions happening after `.test()`


```
println("start")
val createPollViewModel = createPollViewModel(PollMode.CREATE)
createPollViewModel.handle(CreatePollAction.OnQuestionChanged(fakeQuestion))
println("handle")

// We need to wait for createPollViewModel.onChange is triggered
delay(10)
createPollViewModel
        .also { println("test") }
        .test() // prints captures emissions
        .assertState(pollViewStateWithOnlyQuestion)
        .finish()
```

```
start
handle
test
CreatePollViewState(roomId=fakeRoomId, editedEventId=null, mode=CREATE, question=What is your favourite coffee?, options=[, ], canCreatePoll=false, canAddMoreOptions=true, pollType=DISCLOSED_UNSTABLE)
```
notice how the emission happens after `test()`, which shows the logic and updates inside `handle` are non blocking (a test's worst enemy!)

ouchadam: separating the `test()` resolves the issue for me 

```
println("start")
val createPollViewModel = createPollViewModel(PollMode.CREATE)
val test = createPollViewModel.test()

createPollViewModel.handle(CreatePollAction.OnQuestionChanged(fakeQuestion))
println("handled")

test
        .also { println("test") }
        .assertStates(initialCreatePollViewState, pollViewStateWithOnlyQuestion)
        .finish()
```

```
start
CreatePollViewState(roomId=fakeRoomId, editedEventId=null, mode=CREATE, question=, options=[, ], canCreatePoll=false, canAddMoreOptions=true, pollType=DISCLOSED_UNSTABLE)
CreatePollViewState(roomId=fakeRoomId, editedEventId=null, mode=CREATE, question=What is your favourite coffee?, options=[, ], canCreatePoll=false, canAddMoreOptions=true, pollType=DISCLOSED_UNSTABLE)
handled
test
```

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewModelTest.kt
onurays: Right, done.

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewModelTest.kt
onurays: I tried many approaches. The problem is not able to sync the `canCreatePoll` state field. To be clear: Could you please try to run `given there is a question when enough options are added then poll can be created and more options can be added` with your suggestion?

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewModelTest.kt
onurays: Exactly, done.

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewStates.kt
onurays: Exactly, done.

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewStates.kt
onurays: Right, done.

vector/src/test/java/im/vector/app/features/poll/create/CreatePollViewModelTest.kt
mnaturel: I understood the problem: in the `test()` extension method, we use a different dispatcher to assert values on the flows than the one used in the `MvRxTestRule`. @onurays I can apply a fix if you want?

ouchadam: I might be missing something here, there's no issue in the `test()` extension, the problems are 

- the usage of `.test()` should be before any interactions with the `ViewModel`
- `ViewModels` can emit multiple states, especially when multiple handles are called


```kotlin
@Test
fun `given there is not a question when enough options are added then poll cannot be created and more options can be added`() = runTest {
    val createPollViewModel = createPollViewModel(PollMode.CREATE)
    val test = createPollViewModel.test()

    repeat(CreatePollViewModel.MIN_OPTIONS_COUNT) {
        createPollViewModel.handle(CreatePollAction.OnOptionChanged(it, A_FAKE_OPTIONS[it]))
    }

    test.assertStatesChanges(
            initialCreatePollViewState,
            { copy(options = listOf(A_FAKE_OPTIONS[0], "")) },
            { copy(options = listOf(A_FAKE_OPTIONS[0], A_FAKE_OPTIONS[1])) },
    ).finish()
}

@Test
fun `given there is a question when enough options are added then poll can be created and more options can be added`() = runTest {
    val createPollViewModel = createPollViewModel(PollMode.CREATE)
    val test = createPollViewModel.test()

    createPollViewModel.handle(CreatePollAction.OnQuestionChanged(A_FAKE_QUESTION))
    repeat(CreatePollViewModel.MIN_OPTIONS_COUNT) {
        createPollViewModel.handle(CreatePollAction.OnOptionChanged(it, A_FAKE_OPTIONS[it]))
    }

    test.assertStatesChanges(
            initialCreatePollViewState,
            { copy(question = A_FAKE_QUESTION) },
            { copy(options = listOf(A_FAKE_OPTIONS[0], "")) },
            { copy(options = listOf(A_FAKE_OPTIONS[0], A_FAKE_OPTIONS[1])) },
    ).finish()
}
```

ouchadam: the ordering is important because we're subscribing to the `ViewModel` flows in order to capture their emissions

if all the helper logic was inlined it would look like...

```kotlin
// .test()
val captures = mutableListOf()
viewModel.stateFlow
  .onEach { captures.add(it}
  .launchIn(nonBlockingContext)

// handle
viewModel.setState {
    copy("foo")
}

// assert
captures shouldBeEqualTo expectedEmissions
```

ouchadam: turns out there's a bug/coroutine-test incompatibility  https://github.com/airbnb/mavericks/issues/599

```kotlin
@get:Rule
val mvrxTestRule = MvRxTestRule(testDispatcher = UnconfinedTestDispatcher())

@Test
fun `can create poll`() = runTest {
    val createPollViewModel = createPollViewModel(PollMode.CREATE)
    val test = createPollViewModel.test()

    createPollViewModel.handle(CreatePollAction.OnQuestionChanged(A_FAKE_QUESTION))
    repeat(CreatePollViewModel.MIN_OPTIONS_COUNT) {
        createPollViewModel.handle(CreatePollAction.OnOptionChanged(it, A_FAKE_OPTIONS[it]))
    }

    test.assertStatesChanges(
            initialCreatePollViewState,
            { copy(question = A_FAKE_QUESTION) },
            { copy(options = listOf(A_FAKE_OPTIONS[0], "")) },
            { copy(options = listOf(A_FAKE_OPTIONS[0], A_FAKE_OPTIONS[1])) },
            { copy(canCreatePoll = true) },
    ).finish()
}
```

onurays: Thank you both for the great help! Please see my last commit (and approve :).

ouchadam (approved):
great job! LGTM :100:

vector/src/test/java/im/vector/app/test/fakes/FakeCreatePollViewStates.kt
```diff
@@ -0,0 +1,131 @@
...
+import org.matrix.android.sdk.api.session.events.model.toContent
+import org.matrix.android.sdk.api.session.room.model.message.MessagePollContent
+import org.matrix.android.sdk.api.session.room.model.message.PollAnswer
+import org.matrix.android.sdk.api.session.room.model.message.PollCreationInfo
+import org.matrix.android.sdk.api.session.room.model.message.PollQuestion
+import org.matrix.android.sdk.api.session.room.sender.SenderInfo
+import org.matrix.android.sdk.api.session.room.timeline.TimelineEvent
+import kotlin.random.Random
+
+object FakeCreatePollViewStates {
```
ouchadam, line 34: minor naming comment for fakes vs fixtures 

Fakes are objects where behaviour has been changed/overridden or they're extended with test helpers for capturing information, whereas a fixture is a reusable set of defaults, 

I would argue the content of the file fits more into the description of a fixture 

for example a `FakeRoom` could return a `FakeTimeline` which returns a list of `PollEvent fixtures`, usually it's case of `class = fake`, `data class = fixture`

mnaturel (approved):

PR #6323, opened 2022-06-16
Ignore! - Test coverage changes

ouchadam:
Test


bmarty:
@ouchadam why it's not 100% of code coverage here? It looks like the added `fun` is fully tested.

ouchadam:
it's a missing `null` test case 

2022-06-16T16:25:10,570424601+01:00

Issue #5114, opened 2022-02-01
Update design and behaviour on widget bottom sheet


bmarty:
- Create a room
- Invite User 2, User 2 joins the room
- Add a widget to the room, ex: Etherpad
- User 2 open the widget. This bottom sheet is displayed:

<img width="484" alt="image" src="https://user-images.githubusercontent.com/3940906/151936699-6686303b-9dec-441a-a399-6fac0ae13166.png">

TODO:
- [ ] At least the textColor of the 2 buttons should be white. The style of the button are maybe not correct.
- [ ] If the bottom sheet is dismissed, the screen should be closed, like when the user click on "DECLINE"
- [x] @amshakal I do not know if you want to iterate on other things?

daniellekirkwood:
Let's go ahead and leave the requirements for this issue to be:
- [ ] Change color of text in darkmode to white
- [ ] Bottom sheet behaviour updates

amshakal:
Nothing besides what's been stated above already. Thanks!

ahmed-radhouane:
Hello @daniellekirkwood , @amshakal 

 - About this issue I'd like to discuss the UI requirement. 
> At least the textColor of the 2 buttons should be white. The style of the button are maybe not correct.

I do not think that making the textColor as white/ black for both dark/ white modes is a UI friendly for two reasons : 
 - it could make confusion to the user to distinguish that these two labels are actions needed, specially seeing the screen, it has a text content so it could make user inconfortable with that. 
 - I didn't found in the application similar UI for actions (white text actions) 
 
 So I would like to suggest this solution with keep both the action aspect of the two labels, and which I found more adapted to the application UI style used in the application : 
 
 Dark Mode | Light Mode 
--- | --- 
<img src="https://user-images.githubusercontent.com/73594434/153059101-9ce3fd65-6d7c-42af-85c1-dd889c816025.png" width="300"> | <img src="https://user-images.githubusercontent.com/73594434/153059259-c0594b14-e089-4df6-906f-9476c129d112.png" width="300">

Please let me know if the result fits what needed. 
Thank you.

amshakal:
Hi there, thanks for the feedback! I think the comment above has been misinterpreted, and I can see how that can happen. :D By changing the text colour, the red/green background would stay as is, so that it looks like a proper button, like in our design system. Here is an example of how it will look. This style is also consistent throughout our platforms as buttons. 
[...]

PR #6326, opened 2022-06-16
Cleanup of #5187

bmarty:
See #5187 for more details.

Will squash and merge to cleanup git history.

After my change:

image

Fixes https://github.com/vector-im/element-android/issues/5114


vector/src/main/java/im/vector/app/features/widgets/permissions/RoomWidgetPermissionBottomSheet.kt
```diff
@@ -102,8 +102,9 @@ class RoomWidgetPermissionBottomSheet :
     }
 
     override fun onCancel(dialog: DialogInterface) {
+        viewModel.handle(RoomWidgetPermissionActions.BlockWidget)
```
mnaturel, line 105: Maybe should we reuse the existing `doDecline` method?

mnaturel (approved):
I have just added a small remark to mutualize some code.

vector/src/main/java/im/vector/app/features/widgets/permissions/RoomWidgetPermissionBottomSheet.kt
bmarty: https://github.com/vector-im/element-android/pull/5187#discussion_r813882777 😄

mnaturel: I see. It is up to you but we could create a sub method without including call to `dismiss()` as you suggested?

Claire1817 (approved):

PR #6341, opened 2022-06-17
Fixed issues with reporting sync state events from different threads

artkoenig:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Replaced `LiveData` with `Flow` for sync state reporting.

## Motivation and context

`SyncRequestStateTracker` instance is reused in different threads, because of the `Scope` annotation. If `LiveData::postValue` is called fast enough, some events can get lost, because only the last one is dispatched. 
So in my case I could reproduce loosing of `InitialSyncRequestState.Idle` due to sending `IncrementalSyncIdle` from another thread.

Signed-off-by: Artjom König artkoenig@gmail.com

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/sync/SyncService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/sync/SyncService.kt
@@ -60,9 +60,9 @@ interface SyncService {
     fun getSyncStateLive(): LiveData<SyncState>
 
     /**
-     * Get the [SyncRequestState] as a LiveData.
+     * Get the [SyncRequestState] as a SharedFlow.
      */
-    fun getSyncRequestStateLive(): LiveData<SyncRequestState>
+    fun getSyncRequestStateFlow(): SharedFlow<SyncRequestState>
 
     /**
      * This method returns a flow of SyncResponse. New value will be pushed through the sync thread.
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncRequestStateTracker.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncRequestStateTracker.kt
@@ -16,23 +16,26 @@
 
 package org.matrix.android.sdk.internal.session.sync
 
-import androidx.lifecycle.MutableLiveData
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.launch
 import org.matrix.android.sdk.api.session.sync.InitialSyncStep
 import org.matrix.android.sdk.api.session.sync.SyncRequestState
 import org.matrix.android.sdk.internal.session.SessionScope
 import javax.inject.Inject
 
 @SessionScope
-internal class SyncRequestStateTracker @Inject constructor() :
-        ProgressReporter {
+internal class SyncRequestStateTracker @Inject constructor(
+        private val coroutineScope: CoroutineScope
+) : ProgressReporter {
 
-    val syncRequestState = MutableLiveData<SyncRequestState>()
+    val syncRequestState = MutableSharedFlow<SyncRequestState>()
 
     private var rootTask: TaskInfo? = null
 
     // Only to be used for incremental sync
     fun setSyncRequestState(newSyncRequestState: SyncRequestState.IncrementalSyncRequestState) {
-        syncRequestState.postValue(newSyncRequestState)
+        emitSyncState(newSyncRequestState)
     }
 
     /**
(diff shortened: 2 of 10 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/GroupSyncHandler.kt
```diff
@@ -37,7 +37,7 @@ internal class GroupSyncHandler @Inject constructor() {
         data class LEFT(val data: Map<String, Any>) : HandlingStrategy()
     }
 
-    fun handle(
+    suspend fun handle(
```
bmarty, line 40: We cannot have suspendable here since we are caling this inside a Realm transaction which does not support thread change.

Inside maybe you can replace the `MutableLiveData` here https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncRequestStateTracker.kt#L29 by a `Flow` instead?

bmarty (commented):
Thanks for the PR. One first remark.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/handler/GroupSyncHandler.kt
artkoenig: Sure, I'm curious - what about running the code inside the `handleResponse` function (https://github.com/vector-im/element-android/blob/3b2f9d9404129aafd477f1ad91f88a0c7d21a411/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncResponseHandler.kt#L69)  on a single-thread executor (`Executors.newSingleThreadExecutor()`)? Would it not fix the thread switching related problems?

artkoenig: Btw. I see that `Realm.WRITE_EXECUTOR` which is used inside the transaction is already single threaded...

artkoenig: The `Flow`-refactoring is done

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/sync/SyncService.kt
```diff
@@ -62,7 +62,7 @@ interface SyncService {
     /**
      * Get the [SyncRequestState] as a LiveData.
      */
-    fun getSyncRequestStateLive(): LiveData<SyncRequestState>
+    fun getSyncRequestStateFlow(): SharedFlow<SyncRequestState>
```
bmarty, line 65: Can I ask you to update the comment above please? (just replace `LiveData` by `SharedFlow`).

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncRequestStateTracker.kt
```diff
@@ -16,23 +16,26 @@
...
 import org.matrix.android.sdk.api.session.sync.InitialSyncStep
 import org.matrix.android.sdk.api.session.sync.SyncRequestState
 import org.matrix.android.sdk.internal.session.SessionScope
 import javax.inject.Inject
 
 @SessionScope
-internal class SyncRequestStateTracker @Inject constructor() :
-        ProgressReporter {
+internal class SyncRequestStateTracker @Inject constructor(
+        private val coroutineScope: CoroutineScope
```
bmarty, line 29: Note that the CoroutineScope [from the crypto module](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/crypto/CryptoModule.kt#L135) will be injected here.

This is not a big deal, I think, but this will have to be re-worked (by us)

bmarty (approved):
Thanks for the update (and sorry for the delay to review it). Just one remark about the Kdoc, then this can be merged. For the remark about the CoroutineScope we will handle it.

Note that I will squash and merge your commits to cleanup the Git history. Feel free to squash your commits and force push if you want to.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/SyncRequestStateTracker.kt
bmarty: (and this is tracked by #6432)

artkoenig:
Hi @bmarty, are there still any issues preventing this to be merged?

bmarty (approved):
Thanks for the update.
I have triggered the CI on this PR.

PR #6356, opened 2022-06-21
[Location sharing] - Make stop of a live from another device possible (PSF-1060)

### LINKED ISSUE #6349: [Location sharing] - Stop a live from another device
Make the stop of a live be possible from another device.

When starting a live from device A, we should be able to stop the live from another device B using the same account.

Move the stop logic from the service which sends the location events.

_No response_

### Additional context

_No response_

Yes

mnaturel:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Moving the sending of the stopping state event outside the Android service used to start and send location updates.
Observing the live status in DB from the same Android service to stop sending locations when live has ended and to stop the service if there is no more active lives.

## Motivation and context

Closes #6349

## Tests

2 devices are needed

- Ensure lab flag for live location sharing is enabled
- Start a live on a device
- Stop the live from another device (logged in with the same account)
- Check the live is correctly stopped and the location sharing service is closed in the first device

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 11, 10

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
@@ -60,4 +60,10 @@ interface LocationSharingService {
      * Returns a LiveData on the list of current running live location shares.
      */
     fun getRunningLiveLocationShareSummaries(): LiveData<List<LiveLocationShareAggregatedSummary>>
+
+    /**
+     * Returns a LiveData on the list of live location shares with the given eventIds.
+     * @param eventIds the list of event ids
+     */
+    fun getLiveLocationShareSummaries(eventIds: List<String>): LiveData<List<LiveLocationShareAggregatedSummary>>
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
@@ -16,6 +16,7 @@
 
 package org.matrix.android.sdk.api.session.room.location
 
+import androidx.annotation.MainThread
 import androidx.lifecycle.LiveData
 import org.matrix.android.sdk.api.session.room.model.livelocation.LiveLocationShareAggregatedSummary
 import org.matrix.android.sdk.api.util.Cancelable
@@ -60,11 +61,13 @@ interface LocationSharingService {
     /**
      * Returns a LiveData on the list of current running live location shares.
      */
+    @MainThread
     fun getRunningLiveLocationShareSummaries(): LiveData<List<LiveLocationShareAggregatedSummary>>
 
     /**
      * Returns a LiveData on the live location share summary with the given eventId.
      * @param beaconInfoEventId event id of the initial beacon info state event
      */
+    @MainThread
     fun getLiveLocationShareSummary(beaconInfoEventId: String): LiveData<Optional<LiveLocationShareAggregatedSummary>>
 }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
@@ -19,6 +19,7 @@ package org.matrix.android.sdk.api.session.room.location
 import androidx.lifecycle.LiveData
 import org.matrix.android.sdk.api.session.room.model.livelocation.LiveLocationShareAggregatedSummary
 import org.matrix.android.sdk.api.util.Cancelable
+import org.matrix.android.sdk.api.util.Optional
 
 /**
  * Manage all location sharing related features.
@@ -62,8 +63,8 @@ interface LocationSharingService {
     fun getRunningLiveLocationShareSummaries(): LiveData<List<LiveLocationShareAggregatedSummary>>
 
     /**
-     * Returns a LiveData on the list of live location shares with the given eventIds.
-     * @param eventIds the list of event ids
+     * Returns a LiveData on the live location share summary with the given eventId.
+     * @param beaconInfoEventId event id of the initial beacon info state event
      */
-    fun getLiveLocationShareSummaries(eventIds: List<String>): LiveData<List<LiveLocationShareAggregatedSummary>>
+    fun getLiveLocationShareSummary(beaconInfoEventId: String): LiveData<Optional<LiveLocationShareAggregatedSummary>>
 }
--- a/vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
+++ b/vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
@@ -100,7 +100,7 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
                 ?.let { result ->
                     when (result) {
                         is UpdateLiveLocationShareResult.Success -> {
-                            roomArgsMap[result.beaconEventId] = roomArgs
+                            addRoomArgs(result.beaconEventId, roomArgs)
                             locationTracker.requestLastKnownLocation()
                         }
                         is UpdateLiveLocationShareResult.Failure -> {
@@ -132,21 +132,16 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
 
     fun stopSharingLocation(roomId: String) {
         Timber.i("### LocationSharingService.stopSharingLocation for $roomId")
-        synchronized(roomArgsMap) {
-            val beaconIds = roomArgsMap
-                    .filter { it.value.roomId == roomId }
-                    .map { it.key }
-            beaconIds.forEach { roomArgsMap.remove(it) }
-
-            tryToDestroyMe()
-        }
+        removeRoomArgs(roomId)
+        tryToDestroyMe()
     }
 
+    @Synchronized
     override fun onLocationUpdate(locationData: LocationData) {
         Timber.i("### LocationSharingService.onLocationUpdate. Uncertainty: ${locationData.uncertainty}")
 
         // Emit location update to all rooms in which live location sharing is active
-        roomArgsMap.toMap().forEach { item ->
+        roomArgsMap.forEach { item ->
             sendLiveLocation(item.value.roomId, item.key, locationData)
         }
     }
@@ -173,6 +168,7 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
         stopSelf()
     }
 
+    @Synchronized
     private fun tryToDestroyMe() {
         if (roomArgsMap.isEmpty()) {
             Timber.i("### LocationSharingService. Destroying self, time is up for all rooms")
@@ -191,6 +187,19 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
         super.onDestroy()
         Timber.i("### LocationSharingService.onDestroy")
         destroyMe()
+    }
+
+    @Synchronized
+    private fun addRoomArgs(beaconEventId: String, roomArgs: RoomArgs) {
+        roomArgsMap[beaconEventId] = roomArgs
+    }
+
+    @Synchronized
+    private fun removeRoomArgs(roomId: String) {
+        val beaconIds = roomArgsMap
+                .filter { it.value.roomId == roomId }
+                .map { it.key }
+        beaconIds.forEach { roomArgsMap.remove(it) }
     }
 
     private fun launchInIO(block: suspend CoroutineScope.(Session) -> Unit) =
--- a/vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
+++ b/vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
@@ -18,22 +18,27 @@ package im.vector.app.features.location
 
 import android.content.Intent
 import android.os.Binder
+import android.os.Handler
 import android.os.IBinder
 import android.os.Parcelable
 import dagger.hilt.android.AndroidEntryPoint
 import im.vector.app.core.di.ActiveSessionHolder
 import im.vector.app.core.services.VectorService
+import im.vector.app.features.location.live.GetLiveLocationShareSummaryUseCase
 import im.vector.app.features.notifications.NotificationUtils
 import im.vector.app.features.session.coroutineScope
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.flow.distinctUntilChangedBy
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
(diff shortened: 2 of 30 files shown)


vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
```diff
@@ -208,6 +179,34 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
...
+    private fun addRoomArgs(beaconEventId: String, roomArgs: RoomArgs) {
+        roomArgsMap[beaconEventId] = roomArgs
+    }
+
+    @Synchronized
+    private fun removeRoomArgs(roomId: String) {
+        val beaconIds = roomArgsMap
+                .filter { it.value.roomId == roomId }
+                .map { it.key }
+        beaconIds.forEach { roomArgsMap.remove(it) }
```
onurays, line 192: Nit: I would simplify this like:

```
roomArgsMap
                .filter { it.value.roomId == roomId }
                .forEach { roomArgsMap.remove(it.key) }
```

vector/src/main/java/im/vector/app/features/location/live/StopLiveLocationShareUseCase.kt
```diff
@@ -0,0 +1,37 @@
...
+
+package im.vector.app.features.location.live
+
+import org.matrix.android.sdk.api.session.Session
+import org.matrix.android.sdk.api.session.getRoom
+import org.matrix.android.sdk.api.session.room.location.UpdateLiveLocationShareResult
+import javax.inject.Inject
+
+class StopLiveLocationShareUseCase @Inject constructor(
+        private val session: Session
```
ouchadam, line 25: I can't remember the outcome of this, if the `StopLiveLocationShareUseCase` instance can exist across sessions, the inject session will be invalid

might be safer to inject and lazily fetch from the `ActiveSessionHolder`, or scope the `UseCase` to session?

vector/src/main/java/im/vector/app/features/location/LocationSharingServiceConnection.kt
```diff
@@ -49,30 +51,46 @@ class LocationSharingServiceConnection @Inject constructor(
...
-        callback?.onLocationServiceStopped()
+        onCallbackActionNoArg(Callback::onLocationServiceStopped)
     }
 
     override fun onServiceError(error: Throwable) {
-        callback?.onLocationServiceError(error)
+        forwardErrorToCallbacks(error)
+    }
+
+    @Synchronized
```
ouchadam, line 77: do these need to be synchronised? Ideally we would only add/remove the callbacks from 1 thread (such as the main thread)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
```diff
@@ -59,5 +61,13 @@ interface LocationSharingService {
     /**
      * Returns a LiveData on the list of current running live location shares.
      */
+    @MainThread
```
ouchadam, line 64: can these only be called by the main thread? I was under the impression `LiveData's` observing mechanism would handle that for us

vector/src/main/java/im/vector/app/features/location/LocationSharingServiceConnection.kt
```diff
@@ -49,30 +51,46 @@ class LocationSharingServiceConnection @Inject constructor(
...
+    private fun onCallbackActionNoArg(action: Callback.() -> Unit) {
+        callbacks.forEach(action)
+    }
+
+    @Synchronized
+    private fun forwardErrorToCallbacks(error: Throwable) {
+        callbacks.forEach { it.onLocationServiceError(error) }
```
ouchadam, line 94: if the entry points and service callbacks are all on the same thread, this doesn't need to be synchronised, (could argue for introducing a main thread handler to enforce a single thread if multiple threads are used!) 

also worth highlighting that `synchronised` doesn't guard against concurrent modifications on the same thread eg a `onLocationServiceError` that calls `unbind`

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewModel.kt
```diff
@@ -70,7 +79,12 @@ class LocationLiveMapViewModel @AssistedInject constructor(
     }
 
     private fun handleStopSharing() {
-        locationSharingServiceConnection.stopLiveLocationSharing(initialState.roomId)
+        viewModelScope.launch {
+            val result = stopLiveLocationShareUseCase.execute(initialState.roomId)
```
ouchadam, line 83: could the live location end up in a stuck state if the viewmodel was to close/cancel before the `stopLiveLocationShareUseCase` completed?

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
```diff
@@ -208,6 +179,34 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
...
+    }
+
+    private fun listenForLiveSummaryChanges(roomId: String, eventId: String) {
+        activeSessionHolder
+                .getSafeActiveSession()
+                ?.let { session ->
+                    mainHandler.post {
```
onurays, line 199: Do we need `mainHandler` here? We generally use it while triggering a listener function.

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
```diff
@@ -115,53 +120,18 @@ class LocationSharingService : VectorService(), LocationTracker.Callback {
...
-
-    private suspend fun sendStoppedBeaconInfo(session: Session, roomId: String): UpdateLiveLocationShareResult? {
-        return session.getRoom(roomId)
-                ?.locationSharingService()
-                ?.stopLiveLocationShare()
+        removeRoomArgs(roomId)
+        tryToDestroyMe()
     }
 
+    @Synchronized
```
ouchadam, line 129: in the same context as https://github.com/vector-im/element-android/pull/6356/files#r905917814

it would be great to avoid synchronising and make the entry/exit points thread safe, helps reduce the risk of deadlocks or unnecessary thread locking (especially when the main thread is involved :crossed_fingers: )

onurays (approved):
LGTM, please see my small remarks.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
mnaturel: In fact, I added this annotation because the implementation relies on `Monarchy.findAllMappedWithChanges` method. In this method, there is an `assertMainThread()` to ensure we are in the main thread. If it is not the case, it throws an exception.
Ideally, we should not rely on `LiveData` but only on `Flow`. However, last time I tried using the `toFlow` extension available in `Realm`, it has lead to problems about Frozen objects...

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
mnaturel: See [this comment](https://github.com/vector-im/element-android/pull/6356#discussion_r906047460) for explanation. Since we are in Android Service, this is the only way I found to access main thread.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/location/LocationSharingService.kt
ouchadam: I see, thanks for the explanation :+1:

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewModel.kt
mnaturel: I guess, it may happen the live is not stopped if `ViewModel` is cleared before the new state event is sent. When we are in this case, there will be no change on the UI and the service to send locations will still be running. So the user will hopefully see that it was not stopped. 

I see there is a `Session.coroutineScope` but not sure it is a good idea to use it here since we post a result ViewEvent in the coroutine.

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
ouchadam: this is where I'm a little confused, the android service should already be calling back on the main thread, we're jumping to the IO thread here https://github.com/vector-im/element-android/blob/81e14c7c3b6529f08d2cf0b0cacc5a548c836c12/vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt#L89 :thinking:

mnaturel: Yes you are right. Maybe we whould not switch the `Dispatcher` and rely on the fact the SDK is handling correctly the Network calls. I will try to remove it to see.

mnaturel: It seems to work. I will remove the `Handler` I have introduced and the switch to the `Dispatcher.io`.

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
mnaturel: In the `LocationSharingService`, there is the `onLocationUpdate()` which is not called on the Main Thread since we introduced a [`Deboucer` in the `LocationTracker` which runs on a background thread](https://github.com/vector-im/element-android/blob/3dcec85dea47feb823cc8dcc03cf7871af91be45/vector/src/main/java/im/vector/app/features/location/LocationTracker.kt#L67). Do you think I should try to switch back to main thread when notifying callbacks in `LocationTracker` to be able to remove the synchronization ?

vector/src/main/java/im/vector/app/features/location/LocationSharingServiceConnection.kt
ouchadam: :raised_hands:  

this is a tricky one.... I would lean towards not guarding against it and trying to enforce that adding/removing should always happen outside of the callbacks, but it's only enforceable at the code review point :thinking: 

we could use a `CopyOnWriteArray` or copy the list each time we iterate `callbacks.toList()`, both have their own set of drawbacks~

what would be your preference?

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
mnaturel: Okay, thanks. I think I tried the `Flow` approach but I don't remember why I dropped it... I will try to give it another chance.

vector/src/main/java/im/vector/app/features/location/live/map/LocationLiveMapViewModel.kt
ouchadam: ahh does this mean the user will see a foreground notification showing that sharing is still a happening? 

if so it could be handy to support cancelling via the notification (as a separate feature)

mnaturel: Yes, there will be the remaining foreground notification. I think we cannot really add a stop action on the notification since the user may be sharing its live location in several rooms. Or maybe we can decide via the Notification, it stops all current running lives. I will suggest the idea of such action.

vector/src/main/java/im/vector/app/features/location/LocationSharingServiceConnection.kt
mnaturel: Ah I see, thanks! I will go for `toList()`, we won't have large lists so I think it is okay.

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
mnaturel: Thanks! Fix here: [519d43c](https://github.com/vector-im/element-android/pull/6356/commits/519d43ceb78fc23c7f899cdf0033c2eb4c1055d8)

vector/src/main/java/im/vector/app/features/location/live/StopLiveLocationShareUseCase.kt
mnaturel: You are right. Find a fix here: [9450267](https://github.com/vector-im/element-android/pull/6356/commits/945026730c8cb29d26f86e35b5d260ea03b70168)

vector/src/main/java/im/vector/app/features/location/LocationSharingService.kt
mnaturel: I removed all the synchronization and introduced a `SharedFlow` for the location updates: [353a8a7](https://github.com/vector-im/element-android/pull/6356/commits/353a8a70eba70635c010047912600acfcb7beaf8)

ouchadam (approved):
thanks for the updates :100: 

thinking out loud, `asFlow` and `LiveData` are a little tricky to test, do you think it's worth some sort of abstraction or test helpers?

vector/src/main/java/im/vector/app/features/location/live/GetLiveLocationShareSummaryUseCase.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.matrix.android.sdk.api.session.room.model.livelocation.LiveLocationShareAggregatedSummary
+import timber.log.Timber
+import javax.inject.Inject
+
+class GetLiveLocationShareSummaryUseCase @Inject constructor(
+        private val session: Session,
+) {
+
+    @MainThread
+    fun execute(roomId: String, eventId: String): Flow<LiveLocationShareAggregatedSummary> {
```
ouchadam, line 35: not a blocker, I'm wondering if we could/should stick with coroutines here and rely on a `withContext(Main)` (injecting the dispatchers) :thinking:

mnaturel: You are right. When calling a usecase dealing with coroutine, we should not have to worry about the `Dispatcher` in which it should run. I have done some changes here: [7099665](https://github.com/vector-im/element-android/pull/6356/commits/70996655e150dc5f3cb246e0d06e0b7e7fdb2a6d). I also fixed an issue with the `FakeSession` and its dispatchers (see [8476fba](https://github.com/vector-im/element-android/pull/6356/commits/8476fba5ff1d5140e755b9085c3f4b5ad0a1f6c1)).

vector/src/test/java/im/vector/app/test/fakes/FakeFlowLiveDataConversions.kt
```diff
@@ -0,0 +1,33 @@
...
+
+package im.vector.app.test.fakes
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.asFlow
+import io.mockk.every
+import io.mockk.mockkStatic
+import kotlinx.coroutines.flow.flowOf
+
+class FakeFlowLiveDataConversions {
```
ouchadam, line 25: :100:  makes sense! 

I'm not sure how we'll handle multiple calls to `asFlow` yet but we can figure that out when needed

ouchadam (approved):

PR #6358, opened 2022-06-21
App launch crash when migrating database from 29 -> 30 

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Adds missing field check when migrating the database to v30

## Motivation and context

To fix a migration crash

## Screenshots / GIFs

No UI changes

## Tests

- Install v1.4.20 and sign in
- Install v1.4.23 on top of 1.4.20

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

PR #6361, opened 2022-06-22
Fix migration to 30. Seems that `isNull` with long path is not working :/

bmarty:
App launch crash when migrating database from 29 -> 30

Replacement for #6358 

Tested with Debugger, this is fine.


ouchadam (approved):

ouchadam:
unsigned task failed with `
* What went wrong:
> Task :vector:optimizeGplayReleaseResources
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)`

going to merge as the other build tasks passed

Issue #6080, opened 2022-05-17
Crash with dispatchTo


bmarty:
### Steps to reproduce

Reported by rageshakes:
https://github.com/matrix-org/element-android-rageshakes/issues?q=is%3Aopen+dispatchTo+

```
Thread: main, Exception: java.util.ConcurrentModificationException
at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:757)
at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:780)
at im.vector.app.core.utils.JsonViewerStylerKt.dispatchTo(JsonViewerStyler.kt:4)
at org.matrix.android.sdk.internal.session.DefaultSession$$ExternalSyntheticLambda1.run(R8$$SyntheticClass:5)
```

### Outcome

Crash

### Will you send logs?

No

bmarty:
Related to #5821 probably.

ouchadam:
Fixed in 1.4.26

PR #6362, opened 2022-06-22
Fixing concurrent `dispatchTo` crash

### LINKED ISSUE #5821: Signing out causes a crash
### Steps to reproduce

 -I'm logged into the app
2. What do you click?
 -I try to sign out
3. More steps…
 -About 3 out of 10 times the app crashes when I confirm signing out
 -The next time I log in, I'm greated with the crash report screen

#### What did you expect?
I expected to be signed out and taken to the welcome screen.
#### What happened instead?
The app crashed instead.

Submitted 2 rageshakes:
https://github.com/matrix-org/element-android-rageshakes/issues/37244
https://github.com/matrix-org/element-android-rageshakes/issues/37247

### Your phone model

Samsung Galaxy A40

### Operating system version

Android 11

1.4.12

### Homeserver

matrix.org

### Will you send logs?

Yes

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes #5821 & #6080

- Uses a _safe to concurrently modify_ structure for the `SessionListeners` list

## Motivation and context

We have a few implementations of the `SessionListener` which remove themselves from the `SessionListeners` in `onSessionStopped` ([SessionScopedProperty](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/session/SessionScopedProperty.kt#L39), [InvitesAcceptor](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/invite/InvitesAcceptor.kt#L146)) , this is a problem because we iterate through all the listeners in order to call `onSessionStopped`, resulting in a concurrent modification exception.

The error itself seems to be race condition presenting itself intermittently due to `synchronised` usages allowing the next main looper to be used, I'm only able to reproduce locally by directly including a listener which removes itself ASAP but the crash is quite common in the Play console

```kotlin
// Session.close()
sessionListeners.addListener(object : Session.Listener {
    override fun onSessionStopped(session: Session) {
        sessionListeners.removeListener(this)
    }
})
```
 
The fix is to use a structure which allows concurrent modification, in this case a `CopyOnWriteArraySet` which does have an extra performance cost _but_ at the point of modification (rather than iteration) which we don't do very often.

## Screenshots / GIFs

|Before|After|
|-|-|
|before-dispatchto|after-dispatchto|

## Tests

- In the code, add a new session listener which removes itself `onSessionStopped`
- Notice a crash when signing out

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/changelog.d/5821.bugfix
+++ b/changelog.d/5821.bugfix
@@ -1,0 +1,1 @@
+Fixes concurrent modification crash when signing out or launching the app
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionListeners.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/SessionListeners.kt
@@ -19,12 +19,13 @@ package org.matrix.android.sdk.internal.session
 import org.matrix.android.sdk.api.extensions.tryOrNull
 import org.matrix.android.sdk.api.session.Session
 import timber.log.Timber
+import java.util.concurrent.CopyOnWriteArraySet
 import javax.inject.Inject
 
 @SessionScope
 internal class SessionListeners @Inject constructor() {
 
-    private val listeners = mutableSetOf<Session.Listener>()
+    private val listeners = CopyOnWriteArraySet<Session.Listener>()
 
     fun addListener(listener: Session.Listener) {
         synchronized(listeners) {


bmarty (approved):
LGTM, thanks!

PR #6366, opened 2022-06-22
Poll view state unit tests [PSF-1130]

onurays:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Determining `PollState` and `PollOptionViewState` require business login. This PR aims to clean up `MessageItemFactory`, refactor this by creating a factory class and increase the test coverage.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactory.kt
```diff
@@ -0,0 +1,135 @@
...
+    ): PollState = when {
+        !informationData.sendState.isSent() -> PollState.Sending
+        pollResponseSummary?.isClosed.orFalse() -> PollState.Ended
+        pollContent.getBestPollCreationInfo()?.isUndisclosed().orFalse() -> PollState.Undisclosed
+        pollResponseSummary?.myVote?.isNotEmpty().orFalse() -> PollState.Voted(pollResponseSummary?.totalVotes ?: 0)
+        else -> PollState.Ready
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun List<PollAnswer>.mapToOptions(
```
ouchadam, line 89: rather than exposing the internals for testing I would take a step back and consider what we can and can't easily unit test, in this case it's the `AndroidViews/Epoxy Items`

which means instead of returning an epoxy `PollItem`, we could create a dedicated proxy `ViewState` data class for the `PolItem` 

``` kotlin
// MessageItemFactory

createPollItem() {
  val viewState = pollItemViewStateFactory.create()
  return PollItem_()
    .callback()
    .highlight()
    .state(viewState)
}
```

If `Views` have no/minimal business logic, then it means our `ViewStates` could be considered UI proxies (which means less need for slow/flaky instrumentation tests)

whichhhh would turn out class api into...

```kotlin
val viewState = PollItemViewStateFactory().create(
  pollContent: MessagePollContent,
  informationData: MessageInformationData,
)
```

as the factory returns immutable data we can make simple asserts against the result 

```kotlin
@Test
fun `given a certain input, when creating state, then has specific result`() {
  val specific: MessagePollContent

  val result = factory.create(specificInput) 

  result shouldBeEqualTo expectedOutput
} 
```

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactoryTest.kt
```diff
@@ -0,0 +1,282 @@
...
+            testDispatcher = testDispatcher // See https://github.com/airbnb/mavericks/issues/599
+    )
+
+    private lateinit var pollItemFactory: PollItemFactory
+
+    @Before
+    fun setup() {
+        // We are not going to test any UI related code
+        pollItemFactory = PollItemFactory(
+                stringProvider = mockk(),
```
ouchadam, line 98: these should be test fakes rather than direct mocks and as they're unused by the tests it also suggests we aren't asserting any of the data they provide

ouchadam: for example we have a [FakeStringProvider](https://github.com/vector-im/element-android/blob/develop/vector/src/test/java/im/vector/app/test/fakes/FakeStringProvider.kt#L23) which allows us to assert against the correct resource ids 


```kotlin
val fake = FakeStringProvider()

val result = fake.getString(R.string.ok)

result shouldBeEqualTo R.string.ok.toTestString()
```

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactoryTest.kt
```diff
@@ -0,0 +1,282 @@
...
+                                    isSelected = A_MESSAGE_INFORMATION_DATA.pollResponseAggregatedSummary?.myVote == option.id,
+                            )
+                        }
+                    }
+        }
+    }
+
+    @Test
+    fun `given an ended poll then all option view states is Ended`() = runTest {
+        with(pollItemFactory) {
```
ouchadam, line 247: I would argue that the `optionViewStates` can exist on a parent `PollItemViewState`, if we want to avoid too many conditions under a single class, we can extract out another factory which the `PollItemViewStateFactory` could depend on


```kotlin
// Our PollItemViewState test doesn't need to know how the option view state is generated
// that class can have its own tests 
fakeOptionViewStateFactory.given(A_OPTION_VIEW_STATE)

val result = factory.create()

result shouldBeEqualTo Result(
  optionViewStates = A_OPTION_VIEW_STATE
)
```

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactoryHelper.kt
```diff
@@ -0,0 +1,82 @@
...
+import android.text.style.ForegroundColorSpan
+import android.view.View
+import im.vector.app.R
+import im.vector.app.core.resources.ColorProvider
+import im.vector.app.core.resources.StringProvider
+import im.vector.app.core.utils.DimensionConverter
+import im.vector.app.features.home.room.detail.timeline.TimelineEventController
+import im.vector.app.features.home.room.detail.timeline.item.MessageInformationData
+
+object MessageItemFactoryHelper {
```
mnaturel, line 34: I think in general it is better to create classic injectable classes and avoid `object`. Objects are not scoped to only where they are needed and are not suitable for unit tests.

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactoryTest.kt
```diff
@@ -0,0 +1,282 @@
...
+                )
+        )
+)
+
+class PollItemFactoryTest {
+
+    private val testDispatcher = UnconfinedTestDispatcher()
+
+    @get:Rule
+    val mvRxTestRule = MvRxTestRule(
```
mnaturel, line 88: Here this rule is not needed since the `PollItemFactory` does not rely on coroutines. And it is only useful in `ViewModel` and when switching the context `Dispatcher`.

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactoryTest.kt
```diff
@@ -0,0 +1,282 @@
...
+        )
+    }
+
+    @After
+    fun tearDown() {
+        unmockkAll()
+    }
+
+    @Test
+    fun `given a sending poll state then PollState is Sending`() = runTest {
```
mnaturel, line 111: `runTest` can be removed from all tests. It is only useful when testing `suspend` functions.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactory.kt
onurays: Very nice idea, done!

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactoryTest.kt
```diff
@@ -0,0 +1,282 @@
...
+                        A_POLL_CONTENT.getBestPollCreationInfo()?.answers?.get(index)?.let { option ->
+                            pollOptionViewState shouldBeEqualTo PollOptionViewState.PollSending(option.id ?: "", option.getBestAnswer() ?: "")
+                        }
+                    }
+        }
+    }
+
+    @Test
+    fun `given a sent poll then all option view states is PollReady`() = runTest {
+        with(pollItemFactory) {
```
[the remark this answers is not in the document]
onurays: Done (hopefully).

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactoryTest.kt
onurays: Done.

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactoryTest.kt
```diff
@@ -0,0 +1,282 @@
...
+                stringProvider = mockk(),
+                avatarSizeProvider = mockk(),
+                colorProvider = mockk(),
+                dimensionConverter = mockk(),
+        )
+    }
+
+    @After
+    fun tearDown() {
+        unmockkAll()
```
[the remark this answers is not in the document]
onurays: Right, removed.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactoryHelper.kt
onurays: No need to a helper class after refactoring.

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemFactoryTest.kt
onurays: Done.

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemViewStateFactoryTest.kt
```diff
@@ -0,0 +1,221 @@
...
+class PollItemViewStateFactoryTest {
+
+    private lateinit var pollItemViewStateFactory: PollItemViewStateFactory
+
+    private val stringProvider = FakeStringProvider()
+
+    @Before
+    fun setup() {
+        // We are not going to test any UI related code
+        pollItemViewStateFactory = PollItemViewStateFactory(
```
ouchadam, line 89: we can inline the creation as the junit test runner will create a new instance of the test class for each test

```
private val stringProvider = FakeStringProvider()
private val pollItemViewStateFactory = PollItemViewStateFactory(stringProvider.instance)
```

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemViewStateFactoryTest.kt
```diff
@@ -0,0 +1,223 @@
...
+
+        val sendingPollInformationData = A_MESSAGE_INFORMATION_DATA.copy(sendState = SendState.SENDING)
+        val pollViewState = pollItemViewStateFactory.create(
+                pollContent = A_POLL_CONTENT,
+                informationData = sendingPollInformationData,
+        )
+
+        pollViewState shouldBeEqualTo PollViewState(
+                question = A_POLL_CONTENT.getBestPollCreationInfo()?.question?.getBestQuestion() ?: "",
+                totalVotes = stringProvider.instance.getString(R.string.poll_no_votes_cast),
```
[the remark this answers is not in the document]
bmarty: So the test is not passing?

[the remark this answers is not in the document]
bmarty: (oh yes, please ignore me)

bmarty (approved):
Only one small question, will force push

vector/src/test/java/im/vector/app/features/home/room/detail/timeline/factory/PollItemViewStateFactoryTest.kt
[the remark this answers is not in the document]
ouchadam: ideally our expected data creation would avoid calling any real functions or mocks as it makes them more susceptible to unrelated changes

eg if the `StringProvider.getString` function parameters change, this unrelated individual test will also need to be updated 

In this case we know the expected value is the computed value of `R.string.poll_no_votes_cast` which we can create with the `.toTestString()` extension, fetching the value from the fake we've already setup is unnecessary

PR #6368, opened 2022-06-22
[WIP] Modularisation

ouchadam:
Work in progress! Turning the `vector` module into an android library in order to start extracting feature modules out on top.

- Creates a new `vector-app` which is our Android Application entry point
- All epoxy layout references must use a generated R2 from butterknife
- The Hilt entry point needs to be in the Application module


ouchadam:
as this is just a proof of concept we can reference as a closed PR

vector-app/build.gradle
```diff
@@ -0,0 +1,554 @@
+import com.android.build.OutputFile
```
bmarty, line 1: A bit annoying that it's seen by Git as a new file.

vector-app/build.gradle
```diff
@@ -0,0 +1,554 @@
...
+
+apply plugin: 'com.android.application'
```
bmarty, line 3: `com.android.library` ?
**EDIT** OK, I get the direction of the dependency now.

vector/build.gradle
```diff
@@ -130,87 +21,28 @@ android {
...
-
-        // The following argument makes the Android Test Orchestrator run its
-        // "pm clear" command after each test invocation. This command ensures
-        // that the app's state is completely cleared between tests.
-        testInstrumentationRunnerArguments clearPackageData: 'true'
+        buildConfigField "String", "GIT_REVISION", "\"1\""
+        buildConfigField "String", "GIT_REVISION_DATE", "\"2\""
+        buildConfigField "String", "GIT_BRANCH_NAME", "\"3\""
+        buildConfigField "String", "BUILD_NUMBER", "\"4\""
+        buildConfigField "String", "VERSION_NAME", "\"5\""
```
bmarty, line 45: This would need to be updated I guess.

vector-app/build.gradle
```diff
@@ -0,0 +1,554 @@
...
+        ]
+    }
+
+    buildFeatures {
+        viewBinding true
+    }
+}
+
+dependencies {
+    implementation project(':vector')
```
bmarty, line 325: When I first read the PR, I thought the dependency would be the opposite. So maybe in the file `settings.xml` put this module above `:vector`.

bmarty (commented):
Thanks, good idea to do it this way first.
I have made some comments, just to open some discussions.

vector/build.gradle
ouchadam: yep, the versioning and build config will need to be lifted out of the vector module and supplied via the dagger graph, it's similar for the variant specific logic, ideally only the application module would know about the different variants and the _vector_ module would receive different configuration/implementations

PR #6372, opened 2022-06-23
Fixing wrong error message when signing in with wrong credentials

### LINKED ISSUE #6371: Wrong error message when signing in with wrong credentials
### Steps to reproduce

1. In the login with matrix screen
2. When entering wrong credentials
3. Notice the error "Sorry, the server isn't accepting new accounts"

#### What did you expect?
An error mentioning wrong credentials

#### What happened instead?
The wrong error was displayed

### Your phone model

_No response_

### Operating system version

_No response_

1.4.24

### Homeserver

_No response_

### Will you send logs?

No

Yes

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes #6371 Wrong error message when signing in with wrong credentials

- Takes into account if we're in the sign up / in mode when inferring the registration disabled errors. Disabled registration uses the same error code as invalid credentials. 

## Motivation and context

To avoid confusing the user!

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot_20220623_181959|Screenshot_20220623_181805|

## Tests

- Attempt to log in with invalid credentials
- Notice an error message about registration being disabled

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthLoginFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthLoginFragment.kt
@@ -286,7 +286,7 @@ class FtueAuthLoginFragment @Inject constructor() : AbstractSSOFtueAuthFragment<
             throwable.isWeakPassword() || throwable.isInvalidPassword() -> {
                 views.passwordFieldTil.error = errorFormatter.toHumanReadable(throwable)
             }
-            throwable.isRegistrationDisabled() -> {
+            isSignupMode && throwable.isRegistrationDisabled() -> {
                 MaterialAlertDialogBuilder(requireActivity())
                         .setTitle(R.string.dialog_title_error)
                         .setMessage(getString(R.string.login_registration_disabled))
--- a/changelog.d/6371.bugfix
+++ b/changelog.d/6371.bugfix
@@ -1,0 +1,1 @@
+Fixes wrong error message when signing in with wrong credentials


vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthLoginFragment.kt
```diff
@@ -286,7 +286,7 @@ class FtueAuthLoginFragment @Inject constructor() : AbstractSSOFtueAuthFragment<
             throwable.isWeakPassword() || throwable.isInvalidPassword() -> {
                 views.passwordFieldTil.error = errorFormatter.toHumanReadable(throwable)
             }
-            throwable.isRegistrationDisabled() -> {
+            isSignupMode && throwable.isRegistrationDisabled() -> {
```
ouchadam, line 289: registration disabled is only a valid error for the sign up flow

mnaturel (approved):

onurays (approved):
Oh, nice catch! LGTM.

bmarty:
Sorry to comment on a closed PR, but I think we do not need a dialog in case of wrong password. the error is already displayed inlined in the form. For disabled registration, there is (was?) no form IIRC, so we need to keep it.
Also, I think this issue is sort of a regression, since the error should be handled in the Fragment (to update the form) and so not handled by the default error handler which is displaying a dialog. The issue is fixed so this is fine, I'm just sharing my thoughts.

PR #6382, opened 2022-06-27
Hotfix/1.4.25

bmarty:
Not supposed to be merged from GH.

`clearWith` will end only when the list is empty, so in this particular case, we do not empty the list, so it ends up with an infinite migration. After that occurs, ANR and at newt launch the DB is deleted, because considered corrupted. So impacted user will performs an init sync, so this is better than nothing.

Impacted users: testers who get the update 1.4.24 from the PlayStore. This version was not pushed to production.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/migration/MigrateSessionTo030.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/migration/MigrateSessionTo030.kt
@@ -35,6 +35,7 @@ internal class MigrateSessionTo030(realm: DynamicRealm) : RealmMigrator(realm, 3
                 .equalTo(ChunkEntityFields.IS_LAST_FORWARD, false)
                 .findAll()
 
+        val nbOfDeletedChunks = chunks.size
         var nbOfDeletedTimelineEvents = 0
         var nbOfDeletedEvents = 0
         chunks.forEach { chunk ->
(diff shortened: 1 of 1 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/migration/MigrateSessionTo030.kt
```diff
@@ -35,16 +36,23 @@ internal class MigrateSessionTo030(realm: DynamicRealm) : RealmMigrator(realm, 3
...
-            chunk.getList(ChunkEntityFields.TIMELINE_EVENTS.`$`).clearWith { timelineEvent ->
+            val realmObjectToDelete = mutableListOf<DynamicRealmObject>()
+            chunk.getList(ChunkEntityFields.TIMELINE_EVENTS.`$`).forEach { timelineEvent ->
                 // Don't delete state events
                 val event = timelineEvent.getObject(TimelineEventEntityFields.ROOT.`$`)
                 if (event?.isNull(EventEntityFields.STATE_KEY) == true) {
-                    event.deleteFromRealm()
-                    timelineEvent.deleteFromRealm()
+                    realmObjectToDelete.add(event)
+                    realmObjectToDelete.add(timelineEvent)
```
bmarty, line 47: Cannot delete while iterating, so just store in a list and delete later.

ouchadam (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/migration/MigrateSessionTo030.kt
ganfra, line 47: TimelineEvent can always be delete, it's just the root Event which should not be kept if state event.

bmarty: Good to know, but this is not new? So the `clearWith` may work if we always delete the TimelineEvent.

bmarty: OK, I will test that

bmarty:
With the latest commit, I get the log `MigrateSessionTo030: 190 deleted TimelineEvent(s) and 149 deleted Event(s).`

ganfra (approved):

PR #6392, opened 2022-06-28
Ensure clearWith lambda is deleting all the list item

bmarty:
Else we will get an infinite loop. This specific error will help to figure out what is happening.

Tested locally with a Realm schema upgrade:

```kotlin
internal class MigrateSessionTo032(realm: DynamicRealm) : RealmMigrator(realm, 32) {

    override fun doMigrate(realm: DynamicRealm) {
        val chunks = realm.where("ChunkEntity")
                .findAll()

        chunks.forEach { chunk ->
            chunk.getList(ChunkEntityFields.TIMELINE_EVENTS.`$`).clearWith {
                // NO OP, let the app crash!
            }
        }
    }
}
```

And got this crash with stacktrace in log:

```shell
java.lang.RuntimeException: Unable to create application im.vector.app.VectorApplication: java.lang.IllegalStateException: `clearWith` MUST delete all elements of the RealmList
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6465)
        at android.app.ActivityThread.access$1300(ActivityThread.java:219)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1859)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
     Caused by: java.lang.IllegalStateException: `clearWith` MUST delete all elements of the RealmList
        at org.matrix.android.sdk.internal.extensions.RealmExtensionsKt.clearWith(RealmExtensions.kt:36)
        at org.matrix.android.sdk.internal.database.migration.MigrateSessionTo032.doMigrate(MigrateSessionTo032.kt:31)
        ...
```

With this code:

```kotlin
            chunk.getList(ChunkEntityFields.TIMELINE_EVENTS.`$`).clearWith {
                it.deleteFromRealm()
            }
```

No crash is observed and the DB is properly migrated.

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/extensions/RealmExtensions.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/extensions/RealmExtensions.kt
@@ -20,6 +20,7 @@ import io.realm.RealmList
 import io.realm.RealmObject
 import io.realm.RealmObjectSchema
 import org.matrix.android.sdk.internal.database.model.HomeServerCapabilitiesEntityFields
+import org.matrix.android.sdk.internal.util.fatalError
 
 internal fun RealmObject.assertIsManaged() {
     check(isManaged) { "${javaClass.simpleName} entity should be managed to use this function" }
@@ -27,14 +28,19 @@ internal fun RealmObject.assertIsManaged() {
 
 /**
  * Clear a RealmList by deleting all its items calling the provided lambda.
+ * The lambda is supposed to delete the item, which means that after this operation, the list will be empty.
  */
 internal fun <T> RealmList<T>.clearWith(delete: (T) -> Unit) {
-    while (!isEmpty()) {
-        val previousSize = size
-        first()?.let { delete.invoke(it) }
-        if (previousSize != size + 1) {
-            error("`clearWith` MUST delete all elements of the RealmList")
-        }
+    map { item ->
+        // Create a lambda for all items of the list
+        { delete(item) }
+    }.forEach { lambda ->
+        // Then invoke all the lambda
+        lambda.invoke()
+    }
+
+    if (!isEmpty()) {
+        fatalError("`clearWith` MUST delete all elements of the RealmList")
     }
 }
 
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/extensions/RealmExtensions.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/extensions/RealmExtensions.kt
@@ -39,7 +39,7 @@ internal fun <T> RealmList<T>.clearWith(delete: (T) -> Unit) {
         lambda.invoke()
     }
 
-    if (!isEmpty()) {
+    if (isNotEmpty()) {
         fatalError("`clearWith` MUST delete all elements of the RealmList")
     }
 }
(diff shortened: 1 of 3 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/extensions/RealmExtensions.kt
```diff
@@ -30,7 +30,11 @@ internal fun RealmObject.assertIsManaged() {
  */
 internal fun <T> RealmList<T>.clearWith(delete: (T) -> Unit) {
     while (!isEmpty()) {
+        val previousSize = size
         first()?.let { delete.invoke(it) }
+        if (previousSize != size + 1) {
```
ouchadam, line 35: should this be `previousSize == size`? (to highlight no deletion has occurred)

bmarty: I see your point, but I expect `previousSize == size + 1` to be the only valid condition, so I think it's better like that.

We may have lambda which add object(s) to the list, or delete multiple objects, we live in a crazy world!

This will be possible to delete another item than the provide one in the lambda, for instance the first one, but at the end this will have the same result.

Another solution to avoid that would have to change the Extension to:

```kotlin
internal fun <T : RealmObject /* <- Needed contract to be able to use `isValid` */> RealmList<T>.clearWith(delete: (T) -> Unit) {
    while (!isEmpty()) {
        first()?.let {
            delete.invoke(it)
            if (it.isValid) {
                // Object has not been deleted
                error("`clearWith` MUST delete all elements of the RealmList")
            }
        }
    }
}
```

or even just delete it with a warning:

```kotlin
internal fun <T : RealmObject /* <- Needed contract to be able to use `isValid` */> RealmList<T>.clearWith(delete: (T) -> Unit) {
    while (!isEmpty()) {
        first()?.let {
            delete.invoke(it)
            if (it.isValid) {
                // Object has not been deleted
                Timber.w("`clearWith` MUST delete all elements of the RealmList")
                it.deleteFromRealm()
            }
        }
    }
}
```

bmarty: I do not think we can use `map`, because the RealmList is getting empty while we call `deleteFromRealm()` on its items. This is not really Kotlin logic.
I will add some Kdoc on the fun.

bmarty: Oh i understand your code with `map` better now, have read too fast. It may work. Let me give it a try.

bmarty: @ouchadam @mnaturel , I have updated the PR, and tested again dummy migration, it behaves at expected.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/extensions/RealmExtensions.kt
```diff
@@ -20,17 +20,27 @@ import io.realm.RealmList
...
-        first()?.let { delete.invoke(it) }
+    map { item ->
+        // Create a lambda for all items of the list
+        { delete(item) }
+    }.forEach { lambda ->
+        // Then invoke all the lambda
+        lambda.invoke()
+    }
+
+    if (!isEmpty()) {
```
mnaturel, line 42: Could we use `isNotEmpty()` to improve readability?

ouchadam (approved):
thanks for the update!

mnaturel (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/extensions/RealmExtensions.kt
mnaturel: Thanks!

PR #6415, opened 2022-06-29
Replacing epoxy item annotation layout references with getDefaultLayoutId

### LINKED ISSUE #6389: Replace Epoxy layout annotation references
AGP 8 is planning to make R references non final which will cause our epoxy items to become non-compiling due to their usage of a layout reference in the class annotation, this is also a problem for libraries.

We can avoid this by switching to `override fun getDefaultLayoutId(): Int`

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Replaces the R.layout references from the epoxy model annotations as library R files are not marked as final meaning their values can't be used in annotations. The android gradle plugin also warns that with version 8.0 this will be the default behaviour for application modules going forwards. 

## Motivation and context

Closes #6389

## Screenshots / GIFs
No UI changes

## Tests

- Login in and navigation around the app

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/vector/src/debug/java/im/vector/app/features/debug/features/BooleanFeatureItem.kt
+++ b/vector/src/debug/java/im/vector/app/features/debug/features/BooleanFeatureItem.kt
@@ -23,11 +23,12 @@ import android.widget.Spinner
 import android.widget.TextView
 import com.airbnb.epoxy.EpoxyAttribute
 import com.airbnb.epoxy.EpoxyModelClass
+import im.vector.app.R
 import im.vector.app.core.epoxy.VectorEpoxyHolder
 import im.vector.app.core.epoxy.VectorEpoxyModel
 
-@EpoxyModelClass(layout = im.vector.app.R.layout.item_feature)
-abstract class BooleanFeatureItem : VectorEpoxyModel<BooleanFeatureItem.Holder>() {
+@EpoxyModelClass
+abstract class BooleanFeatureItem : VectorEpoxyModel<BooleanFeatureItem.Holder>(R.layout.item_feature) {
 
     @EpoxyAttribute
     lateinit var feature: Feature.BooleanFeature
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageAudioItem.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageAudioItem.kt
@@ -38,7 +38,7 @@ import im.vector.app.features.home.room.detail.timeline.style.TimelineMessageLay
 import im.vector.app.features.themes.ThemeUtils
 
 @EpoxyModelClass
-abstract class MessageAudioItem : AbsMessageItem<MessageAudioItem.Holder>(R.layout.item_timeline_event_base) {
+abstract class MessageAudioItem : AbsMessageItem<MessageAudioItem.Holder>() {
 
     @EpoxyAttribute
     var filename: String = ""
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageAudioItem.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageAudioItem.kt
@@ -37,8 +37,8 @@ import im.vector.app.features.home.room.detail.timeline.helper.ContentUploadStat
 import im.vector.app.features.home.room.detail.timeline.style.TimelineMessageLayout
 import im.vector.app.features.themes.ThemeUtils
 
-@EpoxyModelClass(layout = R.layout.item_timeline_event_base)
-abstract class MessageAudioItem : AbsMessageItem<MessageAudioItem.Holder>() {
+@EpoxyModelClass
+abstract class MessageAudioItem : AbsMessageItem<MessageAudioItem.Holder>(R.layout.item_timeline_event_base) {
 
     @EpoxyAttribute
     var filename: String = ""
--- a/vector/src/main/java/im/vector/app/features/location/live/map/LiveLocationBottomSheetController.kt
+++ b/vector/src/main/java/im/vector/app/features/location/live/map/LiveLocationBottomSheetController.kt
@@ -26,8 +26,6 @@ import im.vector.app.core.resources.StringProvider
 import im.vector.app.core.resources.toTimestamp
 import im.vector.app.core.time.Clock
 import im.vector.app.features.home.AvatarRenderer
-import im.vector.app.features.location.live.map.bottomsheet.LiveLocationUserItem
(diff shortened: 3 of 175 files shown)


vector/src/debug/java/im/vector/app/features/debug/features/BooleanFeatureItem.kt
```diff
@@ -23,11 +23,12 @@ import android.widget.Spinner
...
 import com.airbnb.epoxy.EpoxyAttribute
 import com.airbnb.epoxy.EpoxyModelClass
+import im.vector.app.R
 import im.vector.app.core.epoxy.VectorEpoxyHolder
 import im.vector.app.core.epoxy.VectorEpoxyModel
 
-@EpoxyModelClass(layout = im.vector.app.R.layout.item_feature)
-abstract class BooleanFeatureItem : VectorEpoxyModel<BooleanFeatureItem.Holder>() {
+@EpoxyModelClass
+abstract class BooleanFeatureItem : VectorEpoxyModel<BooleanFeatureItem.Holder>(R.layout.item_feature) {
```
bmarty, line 31: Good idea to provide it to the parent constructor!

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageAudioItem.kt
```diff
@@ -37,8 +37,8 @@ import im.vector.app.features.home.room.detail.timeline.helper.ContentUploadStat
 import im.vector.app.features.home.room.detail.timeline.style.TimelineMessageLayout
 import im.vector.app.features.themes.ThemeUtils
 
-@EpoxyModelClass(layout = R.layout.item_timeline_event_base)
-abstract class MessageAudioItem : AbsMessageItem<MessageAudioItem.Holder>() {
+@EpoxyModelClass
+abstract class MessageAudioItem : AbsMessageItem<MessageAudioItem.Holder>(R.layout.item_timeline_event_base) {
```
bmarty, line 41: Maybe not a good idea, but `AbsMessageItem` may have `R.layout.item_timeline_event_base` as a default value for `layoutId`, WDYT?

vector/src/main/java/im/vector/app/features/location/live/map/LiveLocationBottomSheetController.kt
```diff
@@ -26,8 +26,6 @@ import im.vector.app.core.resources.StringProvider
 import im.vector.app.core.resources.toTimestamp
 import im.vector.app.core.time.Clock
 import im.vector.app.features.home.AvatarRenderer
-import im.vector.app.features.location.live.map.bottomsheet.LiveLocationUserItem
-import im.vector.app.features.location.live.map.bottomsheet.liveLocationUserItem
```
bmarty, line 30: Was it unused import? Wondering why ktlint did not complain about it.

vector/src/main/java/im/vector/app/features/login/terms/PolicyItem.kt
```diff
@@ -28,7 +28,7 @@ import im.vector.app.core.epoxy.VectorEpoxyHolder
 import im.vector.app.core.epoxy.onClick
 import im.vector.app.core.extensions.setHorizontalPadding
 
-@EpoxyModelClass(layout = R.layout.item_policy)
+@EpoxyModelClass
```
bmarty, line 31: Strange that this class does not extend `VectorEpoxyModel` like the other. Maybe a small mistake from me. If there is no good reason not to do it, maybe an opportunity to fix that...

vector/src/main/java/im/vector/app/features/reactions/EmojiSearchResultItem.kt
```diff
@@ -19,16 +19,16 @@ import android.graphics.Typeface
...
 import im.vector.app.core.epoxy.VectorEpoxyHolder
+import im.vector.app.core.epoxy.VectorEpoxyModel
 import im.vector.app.core.epoxy.onClick
 import im.vector.app.core.extensions.setTextOrHide
 import im.vector.app.features.reactions.data.EmojiItem
 
-@EpoxyModelClass(layout = R.layout.item_emoji_result)
-abstract class EmojiSearchResultItem : EpoxyModelWithHolder<EmojiSearchResultItem.Holder>() {
+@EpoxyModelClass
+abstract class EmojiSearchResultItem : VectorEpoxyModel<EmojiSearchResultItem.Holder>(R.layout.item_emoji_result) {
```
bmarty, line 31: ... and this has been changed here (and a few times elsewhere), thanks about that!

bmarty (approved):
Nice (and probably a bit painful) work!

Will be happy to see all those warnings vanish:

image

vector/src/main/java/im/vector/app/features/location/live/map/LiveLocationBottomSheetController.kt
bmarty: (The code does not compile on my side if I just remove these 2 lines on my current branch, but the CI seems to be happy here. Please double check!)

ouchadam: [LiveLocationUserItem](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/location/live/map/LiveLocationUserItem.kt#L17) has the wrong package declared for the file location

vector/src/main/java/im/vector/app/features/login/terms/PolicyItem.kt
ouchadam: ah great catch, will fix! this must have been one of the first items I updated before deciding it would be better to provide the id via the constructor

vector/src/main/java/im/vector/app/features/location/live/map/LiveLocationBottomSheetController.kt
bmarty: https://github.com/vector-im/element-android/pull/6415#discussion_r910737101

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageAudioItem.kt
ouchadam: make sense, I didn't realise they were all using the exact same base layout https://github.com/vector-im/element-android/pull/6415/commits/d0e5b3eb21232abcc87d10100cc9a9e90c14bf5a

bmarty (approved):
Thanks for the update. Maybe merge the PR just after the release branch's been created.

ouchadam:
@bmarty sounds good to me!

PR #6434, opened 2022-07-01
Simple check to prevent frozen class modificaiton

BillCarsonFr:
## Type of change
- [x] Technical

Quick technical change to add a check to prevent the modification of some serializable classes.
Historically megolm session where serialized using java serialization mechanism but without proper version. Any change to those class will make it impossible for the app to load saved session.

This change adds a check to fail when such frozen class is modified.

### DIFF
--- a/tools/check/check_code_quality.sh
+++ b/tools/check/check_code_quality.sh
@@ -17,6 +17,21 @@
 #
 
 #######################################################################################################################
+# Check frozen class modification
+#######################################################################################################################
+
+echo "Check if frozen class modified"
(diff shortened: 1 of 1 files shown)


BillCarsonFr:
@bmarty Adding you as reviewer, not sure if it's the best place/way to do that.

bmarty (approved):
LGTM, I assume that you have checked the error. I will create a 2 PRs to test again.

This check could be a good candidate for Danger, but the tool is not yet setup in the project. We will migrate this check later.

Issue #5687, opened 2022-04-04
Element doesn't respect system font scaling


thefirstofthe300:
### Steps to reproduce

1) Set system font size to any value (smallest for greatest contrast)
2) Open Element and observe current font size
3) Set system font size to another value (largest for greatest contrast
4) Open Element and observe that the font sizes have stayed the same (with the notable exception of the chips used to denote @'ing someone

### Outcome

Font sizes don't change 

### Will you send logs?

No

fedrunov:
We have currently app-wide setting to define font scale (Settings->Preferences->Font size) and the problem is that on app start we always set font scale for the app from that setting. So if system setting is the "largest" text scale but in-app setting is "normal", the resulting applying scale will be "normal", and if user change system setting we ignore it, applying only whatever set in the app settings. The only moment when we sync with system property is first start of the app.-  we get system property and apply it to the app setting. 

I think we should have "system default" setting as default one, until user manually change it to something specific in the app settings. If "system default" is set, app will use system settings, reflecting any changes in them. @daniellekirkwood what do you think?

This is how in-app font scale setting dialog look like now:
![Screenshot 2022-05-24 at 12 18 43](https://user-images.githubusercontent.com/66663241/170010671-5f09fb5e-d376-47e4-a880-dd4c637800b3.png)

daniellekirkwood:
Thanks, @fedrunov !!

Let's go ahead and add "Use system default" to the top of this list with a radio button. We'll use font size "Normal" for the label.

fedrunov:
Here how it looks like:

![image](https://user-images.githubusercontent.com/66663241/170479509-47161c65-260f-4a9b-b4f6-0482f594f184.png)

I think it would be better if list will contain only scale factor items and "use system default" will be on top as a checkbox. When it's checked list will be "readonly" and item, corresponding to system setting will be selected. When it's unchecked list will be interactive and user could select an option to override system setting for element.

fedrunov:
@niquewoodhouse, @amshakal what do you think?

amshakal:
Hey! I agree with your solution, the only thing I'll add is a line seperating system default from the rest of the options. Happy to create a visual representation of it if needed.

PR #6453, opened 2022-07-04
font scale setting screen

fedrunov:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Screen, where user can select font scaling specifically for element app, or choose to use system-wide setting 

## Motivation and context

https://github.com/vector-im/element-android/issues/5687

## Screenshots / GIFs

|Light theme|DarkTheme|
|-|-|
|Screenshot 2022-07-07 at 11 11 52|Screenshot 2022-07-07 at 11 12 01|

## Tests

- Navigate to Settings->Prefences
- Click on Font Size
- Choose settings

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Android 12

## Checklist

- [] Accessibility has been taken into account. See https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#accessibility


vector/src/main/java/im/vector/app/core/utils/SystemSettingsProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+interface SystemSettingsProvider {
+
+    /**
+     * @return system setting for font scale
+     */
+    fun getSystemFontScale(): Float
+}
+
+class SystemSettingsProviderImpl @Inject constructor(
```
ouchadam, line 34: rather than `Impl` we could provide more context here by renaming the implementation `AndroidSystemSettingsProvider`, for other cases where the interface and implementation are the same (extracted for testing or design reasons) the convention is to use `Default{Interface}`

vector/src/test/java/im/vector/app/features/settings/font/FontScaleSettingViewModelTest.kt
```diff
@@ -0,0 +1,103 @@
...
+                fakeConfiguration.instance,
+                fakeFontScalePreferences
+        ).also {
+            viewModel = it
+            initialState = state
+        }
+    }
+
+    @Test
+    fun `when handling FontScaleChangedAction, then changes state and emits RestartActivity event`() = runTest {
```
ouchadam, line 56: a convention commonly used by gherkin style unit tests is to use line breaks as a way to signify the different parts of the test

```
// given
setup parameters and dependencies

// when
call method under test

// then
assert result
```

vector/src/main/java/im/vector/app/core/utils/SystemSettingsProvider.kt
fedrunov: I don't really think there is any other system rather than Android we could have, so `System` and `AndroidSystem` are quite same for me. And since so, I'll rename to `AndroidSystemSettingsProvider` :)

vector/src/main/java/im/vector/app/core/utils/SystemSettingsProvider.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package im.vector.app.core.utils
+
+import android.content.Context
+import android.provider.Settings
+import javax.inject.Inject
+
+/**
+ * A helper to get system settings.
```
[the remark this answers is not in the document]
fedrunov: > does the consumer of getSystemFontScale benefit from knowing it's a System setting

Yes, the main purpose of this class is to provide system settings to be compared with app settings or used instead of it. So there is a clear and significant difference between system setting and app setting for consumer of this class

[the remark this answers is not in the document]
fedrunov: > this might be more flexible if it was coupled to the font domain rather than SystemSettings, eg FontSettingsProvider

yes, it would be more flexible, but that's also a level of abstraction which we don't need and most likely we won't need in a future. There are no other system settings we provide for now and adding another class for font domain will just increase number of classes involved in resolving this setting, without giving us any real benefit from it's flexibility

vector/src/test/java/im/vector/app/features/settings/font/FontScaleSettingViewModelTest.kt
```diff
@@ -0,0 +1,103 @@
...
+        val scaleOptions = fakeFontScalePreferences.getAvailableScales()
```
ouchadam, line 57: ideally our tests would only make real calls to the method currently under test, which for this test case is `ViewModel.handle`

by calling methods on other classes we're coupling our tests to them, if `FontScalePreferences.getAvailableScales` were to change, this test would also need to be updated

something else to bear in mind is which information is needed by the test, for this test case we want to know if given a list of options, when selecting one, is the state updated, restart emitted and the selection persisted, which means we can simplify to...

```kotlin
// A reusable fixture for the given data
private val A_SELECTION = aFontScaleValue(index = 1)

// the ViewModel tests don't need to know about the real font scale options
private val A_SCALE_OPTIONS_WITH_SELECTION = listOf(
        aFontScaleValue(index = 0),
        A_SELECTION,
)

@Test
fun `given scale options containing selection, when handling FontScaleChangedAction with selection, then persists selection, updates selected index and emits RestartActivity`() {
    // we setup the fake with our static test data related to the test name
    fakeFontScalePreferences.givenAvailableScaleOptions(A_SCALE_OPTIONS_WITH_SELECTION)
    viewModelWith(initialState)
    val test = viewModel.test()

    viewModel.handle(FontScaleSettingAction.FontScaleChangedAction(A_SELECTION))

    test
            .assertStatesChanges(
                    // the ViewModel mutates the initial state in its `init`
                    initialState.copy(availableScaleOptions = A_SCALE_OPTIONS_WITH_SELECTION),
                    { copy(persistedSettingIndex = A_SELECTION.index) }
            )
            .assertEvents(FontScaleSettingViewEvents.RestartActivity)
            .finish()
    fakeFontScalePreferences.verifyAppScaleFontValue(A_SELECTION)
}

// our tests only make use of the index 
fun aFontScaleValue(index: Int) = FontScaleValue(index, "foo", -1f, 0)
```

vector/src/main/java/im/vector/app/features/settings/FontScalePreferences.kt
```diff
@@ -0,0 +1,135 @@
...
+    /** Returns list of all available font scale values.
+     * @return list of values
+     */
+    fun getAvailableScales(): List<FontScaleValue>
+}
+
+/**
+ * Object to manage the Font Scale choice of the user.
+ */
+@Singleton
```
ouchadam, line 66: what's the reason for using `@Singleton`? I would lean against the usage as the class isn't heavy to instantiate and doesn't directly hold any mutable state

once the classloader encounters a singleton it'll be in memory for the lifetime of the application (this class wouldn't be needed once the app is in the background for example)

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
```diff
@@ -200,4 +211,11 @@ object VectorStaticModule {
     @Provides
     @Singleton
     fun providesBuildMeta() = BuildMeta()
+
+    @Provides
+    @Singleton
+    @DefaultPreferences
+    fun providesDefaultSharedPreferences(context: Context): SharedPreferences {
```
ouchadam, line 218: :100:

ouchadam (approved):
thanks for the updates, looks great! :100:

PR #6457, opened 2022-07-04
Implement the current spec for event match conditions

### LINKED ISSUE #2541: at-room parsing is wrongly identifying messages
The push rules don't trigger for at-room notifications put together with other sentences (`at-room1` for example), but Android's parsing appears to do a simple contains check instead. This can cause irrelevant notifications.

deepbluev7:
This fixes that people randomly get pinged on every reply to a user
names @roomba:server.tld.

fixes #2541

Signed-off-by: Nicolas Werner <nicolas.werner@hotmail.de>

 
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Implements the spec from https://github.com/matrix-org/matrix-spec-proposals/pull/3690 so that replying to users, whose name starts with @room, shouldn't ping you anymore.

## Motivation and context

I complained about this issue a few years ago and it still isn't fixed and now exactly what I predicted happened: A user named @roomba joined my rooms and this is causing quite a few complaints from Element Android users. Since not replying to people is a hard thing to remember, I thought I would try fixing it instead!

## Tests

I actually didn't, since I don't have any Android device. I'll see if I can test it in the emulator on a different device later. But if you want to test it, just join #gentoo:matrix.org and see if you have thousands of room mentions, where someone replied to roomba.

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


deepbluev7:
(Still need to install java to be able to lint and the android SDK to be able to test this)

deepbluev7:
Actually, thinking about it, not sure if I have time to set up a complete Android dev env soon, some feedback would be nice and maybe it can be merged without me manually testing it, if all the tests and lints pass?

deepbluev7:
I think the sonarqube stuff failing is not my fault, is it?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushrules/EventMatchCondition.kt
```diff
@@ -31,40 +29,36 @@ class EventMatchCondition(
...
-                    pattern.removePrefix("*").removeSuffix("*").simpleGlobToRegExp()
-                } else {
-                    pattern.simpleGlobToRegExp()
-                }
-                val regex = Regex(modPattern, RegexOption.DOT_MATCHES_ALL)
+            if (key == "content.body") {
+                val regex = Regex("(\\W|^)" + pattern.simpleGlobToRegExp() + "(\\W|$)", setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.IGNORE_CASE))
                 regex.containsMatchIn(value)
+            } else {
+                val regex = Regex(pattern.simpleGlobToRegExp(), setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.IGNORE_CASE))
```
BillCarsonFr, line 60: Shouldn't you keep the existing optimisation to remove * prefix/suffix? as per https://github.com/vector-im/element-android/pull/5008

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushrules/EventMatchCondition.kt
```diff
@@ -31,40 +29,36 @@ class EventMatchCondition(
...
-                val modPattern = if (pattern.hasSpecialGlobChar()) {
-                    // Regex.containsMatchIn() is way faster without leading and trailing
-                    // stars, that don't make any difference for the evaluation result
-                    pattern.removePrefix("*").removeSuffix("*").simpleGlobToRegExp()
-                } else {
-                    pattern.simpleGlobToRegExp()
-                }
-                val regex = Regex(modPattern, RegexOption.DOT_MATCHES_ALL)
+            if (key == "content.body") {
+                val regex = Regex("(\\W|^)" + pattern.simpleGlobToRegExp() + "(\\W|$)", setOf(RegexOption.DOT_MATCHES_ALL, RegexOption.IGNORE_CASE))
```
BillCarsonFr, line 57: How different from caseInsensitiveFind is this? can it reuse it too?

deepbluev7: caseInsesitiveFind doesn't evaluate globs (since it escapes the string, from what I can tell), otherwise it is the same.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushrules/EventMatchCondition.kt
deepbluev7: So in theory we could, but in this branch we are not matching the content.body field, but other fields of the event, which usually don't contain a long text. If someone explicitly writes a rule to match `*a*` on the sender field, we could in theory match 255 characters. It would also need to be extended to not do a containsMatchIn, since if there was no `*` at the start of the string, we need to match the start of the string.

The important part here is in the following line, which does `matches` instead of `containsMatchIn`. This would need to change to:

```
if (pattern.startsWith("*") && pattern.endsWith("*")) {
    val regex = Regex(pattern.removePrefix("*").removeSuffix("*").simpleGlobToRegExp(), RegexOption.DOT_MATCHES_ALL)
    regex.containsMatchIn(value)
} else if (pattern.startsWith("*")) {
  // match if value endsWith pattern
} else if (pattern.endsWith("*")) {
  // match if value startsWith pattern
} else {
  // match if complete value matches pattern
}
```

The last 3 branches could probably be collapsed just into one, but considering that fields apart from `content.body` are usually short, patterns usually never start AND end with a `*` and this would mostly affect the user themselves if they find a way to write such a rule that actually is slow to evaluate, I would say it is not worth it in this case.

Now on the other hand in the half that matches on `content.body`, we need to check for word boundaries at the start and end for most matches. Now I don't think this matters if we have a glob on both sides, so I think keeping the optimization can actually be useful there. But for context, the old optimization was because Element Android did `*@room*` to match the body, which this PR explicitly removes, since that matches the wrong stuff. So this only optimizes some badly written user rules and tbh probably should be implemented upstream in the regex engine... Anyway, I'll add the optimization back to the content.body branch.

deepbluev7: Added the optimization back in the body branch, not in the other branch since I don't really see a need for it, but if you think I should, I can do that as well.

deepbluev7:
Jo

Florian14 (approved):
Changes lgtm, tested ok, CI is happy after merging develop into the branch.
Thanks for your contribution and fixing this uncommon but annoying issue

deepbluev7:
Thank you! <3

PR #6458, opened 2022-07-04
Rename Android Service to use `AndroidService` suffix

bmarty:
To limit confusion with our Matrix SDK Services

### DIFF
--- a/changelog.d/6458.misc
+++ b/changelog.d/6458.misc
@@ -1,0 +1,1 @@
+Rename Android Service to use `AndroidService` suffix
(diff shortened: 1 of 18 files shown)


bmarty:
> now that we're consistent we can add these classes to the coverage exclusions https://github.com/vector-im/element-android/blob/develop/coverage.gradle#L9 (we should favour having testable collaborators instead of logic directly in the services)

Good point! Done in [4f19034](https://github.com/vector-im/element-android/pull/6458/commits/4f19034a94dc13b4021f42063e7c22abc1f600ac)

ouchadam (approved):
thanks for the update!

PR #6462, opened 2022-07-05
Timeline: fix validation of timeline event changes

### LINKED ISSUE #6461: Crashes on Timeline [Thread] due to range validation
### Steps to reproduce

1. Open a Thread
2. Put in airplane mode to have a gap when coming back online
3. Send 11 messages from other session
4. Get out of airplane
5. Crashes

#### What did you expect?
Doesn't crash
#### What happened instead?
Crashes

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

Yes

ganfra:
Fix #6461

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineChunk.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineChunk.kt
@@ -490,38 +490,11 @@ internal class TimelineChunk(
     private fun handleDatabaseChangeSet(results: RealmResults<TimelineEventEntity>, changeSet: OrderedCollectionChangeSet) {
         val insertions = changeSet.insertionRanges
         for (range in insertions) {
-            // Check if the insertion's displayIndices match our expectations - or skip this insertion.
-            // Inconsistencies (missing messages) can happen otherwise if we get insertions before having loaded all timeline events of the chunk.
-            if (builtEvents.isNotEmpty()) {
-                // Check consistency to item before insertions
-                if (range.startIndex > 0) {
-                    val firstInsertion = results[range.startIndex]!!
-                    val lastBeforeInsertion = builtEvents[range.startIndex - 1]
-                    if (firstInsertion.displayIndex + 1 != lastBeforeInsertion.displayIndex) {
-                        Timber.i(
-                                "handleDatabaseChangeSet: skip insertion at ${range.startIndex}/${builtEvents.size}, " +
-                                        "displayIndex mismatch at ${range.startIndex}: ${firstInsertion.displayIndex} -> ${lastBeforeInsertion.displayIndex}"
-                        )
-                        continue
-                    }
-                }
-                // Check consistency to item after insertions
-                if (range.startIndex < builtEvents.size) {
-                    val lastInsertion = results[range.startIndex + range.length - 1]!!
-                    val firstAfterInsertion = builtEvents[range.startIndex]
-                    if (firstAfterInsertion.displayIndex + 1 != lastInsertion.displayIndex) {
-                        Timber.i(
-                                "handleDatabaseChangeSet: skip insertion at ${range.startIndex}/${builtEvents.size}, " +
-                                        "displayIndex mismatch at ${range.startIndex + range.length}: " +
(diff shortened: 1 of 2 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineChunk.kt
```diff
@@ -558,6 +528,21 @@ internal class TimelineChunk(
         }
     }
 
+    private fun validateInsertion(range: OrderedCollectionChangeSet.Range, results: RealmResults<TimelineEventEntity>): Boolean {
```
ouchadam, line 531: :100: for pulling out a function

bmarty (commented):
I can confirm that this fixes the crash from #6461, but I have another crash when closing the thread and going back to the main timeline.

```
Thread: DefaultTimeline_Thread, Exception: java.lang.IllegalStateException: Cannot create asynchronous query while in a write transaction in /tmp/realm-java/realm/realm-library/src/main/cpp/io_realm_internal_OsResults.cpp line 243
at io.realm.internal.OsResults.nativeStartListening(Native Method)
at io.realm.internal.OsResults.addListener(OsResults.java:673)
at io.realm.RealmResults.addChangeListener(RealmResults.java:734)
at org.matrix.android.sdk.internal.session.room.timeline.TimelineChunk.(TimelineChunk.kt:117)
at org.matrix.android.sdk.internal.session.room.timeline.TimelineChunk.createTimelineChunk(TimelineChunk.kt:567)
at org.matrix.android.sdk.internal.session.room.timeline.TimelineChunk.chunkObjectListener$lambda-2(TimelineChunk.kt:96)
at org.matrix.android.sdk.internal.session.room.timeline.TimelineChunk.$r8$lambda$pCFJ5imq9fIZIR0Q2XoxgvbuKjU(Unknown Source:0)
at org.matrix.android.sdk.internal.session.room.timeline.TimelineChunk$$ExternalSyntheticLambda1.onChange(Unknown Source:4)
```

See https://github.com/matrix-org/element-android-rageshakes/issues/40144

ganfra:
> 

Yes, this is one of the other issues I was talking about. But it happens randomly and already exists before my fix.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineChunk.kt
```diff
@@ -558,6 +528,21 @@ internal class TimelineChunk(
...
+        // Insertion can only happen from LastForward chunk after a sync.
+        if (isLastForward.get()) {
+            val firstBuiltEvent = builtEvents.firstOrNull()
+            if (firstBuiltEvent != null) {
+                val lastInsertion = results[range.startIndex + range.length - 1] ?: return false
+                if (firstBuiltEvent.displayIndex + 1 != lastInsertion.displayIndex) {
```
ouchadam, line 537: for my understanding, is the only false case when the `firstBuiltEvent` exists and it doesn't have the same display index as the `results[range.startIndex + range.length - 1]`

it might be a little easier to read as a when rather than nested ifs 

```kotlin
private fun validateInsertion(range: OrderedCollectionChangeSet.Range, results: RealmResults<TimelineEventEntity>): Boolean {
    // Insertion can only happen from LastForward chunk after a sync.
    val firstForwardEvent = isLastForward.get().takeIf { it }?.let { builtEvents.firstOrNull()  }
    return when {
        firstForwardEvent == null -> true
        firstForwardEvent.nextDisplayIndexDoesNotMatch(results.last(range)) -> {
            Timber.v("There is no continuation in the chunk, chunk is not fully loaded yet, skip insert.")
            false
        }
        else -> true
    }
}

private fun RealmResults<TimelineEventEntity>.last(range: OrderedCollectionChangeSet.Range): TimelineEventEntity? {
    return this[range.startIndex + range.length - 1]
}

private fun TimelineEvent.nextDisplayIndexDoesNotMatch(event: TimelineEventEntity?): Boolean {
    return (displayIndex + 1) != event?.displayIndex
}
```

ouchadam (approved):

PR #6467, opened 2022-07-05
Crash - Sharing text 

### LINKED ISSUE #6451: Beta 1.4.26: Sharing URLs crashes Element
### Steps to reproduce

1. In any web browser (or feed reader or YouTube) choose "Share with" and select Element from the menu
2. Element crashes

https://user-images.githubusercontent.com/478228/177135083-403fdea5-8cee-4b5a-ba9b-adf47a9469e1.mp4

#### What did you expect?

Element opens and lets me select where to share the URL to

#### What happened instead?

Element instantly crashes, bouncing me back to the app I was sharing from.
This happens with every URL sharing (tested with Vivaldi, Chrome, YouTube, Inoreader).

On next startup, Element asked me to send logs (which I did).

Side note: Sharing images works however.

### Your phone model

Google Pixel 6

### Operating system version

Android 12

1.4.26 [40104262] (G-b9381) from Play Store Beta

### Homeserver

private; matrix-synapse-py3 1.61.1+bullseye1

### Will you send logs?

Yes

No

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes a crash when `text/plain` is shared due to a regression introduced by #6296 which meant plain text was attempting to be read as a file

- Checks for the presence of plain text before `text/*` files (the fix!) 
- Separates the `Share Intent Handling` from the `AttachmentsHelper`, means the callbacks can be separated (and will make introducing tests easier)

## Motivation and context

To fix #6451 

## Screenshots / GIFs

TIL incognito chrome tabs aren't captured by screen recorders :sweat_smile: 

|Before|After|
|-|-|
|before-share-crash|after-sharing-text|
 

## Tests

- With Element installed and signed in
- Open a website in Google Chrome
- Attempt to share the url link via the menu item `Share...`

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/vector/src/main/java/im/vector/app/features/attachments/AttachmentsHelper.kt
+++ b/vector/src/main/java/im/vector/app/features/attachments/AttachmentsHelper.kt
@@ -196,32 +196,38 @@ class AttachmentsHelper(val context: Context, val callback: Callback) : Restorab
      */
     fun handleShareIntent(context: Context, intent: Intent): Boolean {
         val type = intent.resolveType(context) ?: return false
-        if (type.startsWith("image")) {
-            callback.onContentAttachmentsReady(
-                    MultiPicker.get(MultiPicker.IMAGE).getIncomingFiles(context, intent).map {
-                        it.toContentAttachmentData()
-                    }
-            )
-        } else if (type.startsWith("video")) {
-            callback.onContentAttachmentsReady(
-                    MultiPicker.get(MultiPicker.VIDEO).getIncomingFiles(context, intent).map {
-                        it.toContentAttachmentData()
-                    }
-            )
-        } else if (type.startsWith("audio")) {
-            callback.onContentAttachmentsReady(
-                    MultiPicker.get(MultiPicker.AUDIO).getIncomingFiles(context, intent).map {
-                        it.toContentAttachmentData()
-                    }
-            )
-        } else if (type.startsWith("application") || type.startsWith("file") || type.startsWith("text") || type.startsWith("*")) {
-            callback.onContentAttachmentsReady(
-                    MultiPicker.get(MultiPicker.FILE).getIncomingFiles(context, intent).map {
-                        it.toContentAttachmentData()
-                    }
-            )
-        } else {
-            return false
+        when {
+            type.startsWith("image") -> {
+                callback.onContentAttachmentsReady(
+                        MultiPicker.get(MultiPicker.IMAGE).getIncomingFiles(context, intent).map {
+                            it.toContentAttachmentData()
+                        }
+                )
+            }
+            type.startsWith("video") -> {
+                callback.onContentAttachmentsReady(
+                        MultiPicker.get(MultiPicker.VIDEO).getIncomingFiles(context, intent).map {
+                            it.toContentAttachmentData()
+                        }
+                )
+            }
+            type.startsWith("audio") -> {
+                callback.onContentAttachmentsReady(
+                        MultiPicker.get(MultiPicker.AUDIO).getIncomingFiles(context, intent).map {
+                            it.toContentAttachmentData()
+                        }
+                )
+            }
+            type.startsWith("application") || type.startsWith("file") || type.startsWith("text") || type.startsWith("*") -> {
+                callback.onContentAttachmentsReady(
+                        MultiPicker.get(MultiPicker.FILE).getIncomingFiles(context, intent).map {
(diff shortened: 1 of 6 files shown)


vector/src/main/java/im/vector/app/features/attachments/ShareIntentHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * This methods aims to handle incoming share intents.
+     *
+     * @return true if it can handle the intent data, false otherwise
+     */
+    fun handleIncomingShareIntent(context: Context, intent: Intent, onFile: (List<ContentAttachmentData>) -> Unit, onPlainText: (String) -> Unit): Boolean {
+        val type = intent.resolveType(context) ?: return false
+        return when {
+            type == "text/plain" -> handlePlainText(intent, onPlainText)
```
ouchadam, line 35: having the plain text handling earlier in the check and as part of the share intent handler is the fix, previously we were relying on the handler to skip on plain text

vector/src/main/java/im/vector/app/features/attachments/AttachmentsHelper.kt
```diff
@@ -45,7 +45,6 @@ class AttachmentsHelper(val context: Context, val callback: Callback) : Restorab
         }
 
         fun onContentAttachmentsReady(attachments: List<ContentAttachmentData>)
-        fun onAttachmentsProcessFailed()
```
ouchadam, line 48: it turns out this was never used

ganfra (approved):
LGTM!

vector/src/main/java/im/vector/app/features/attachments/ShareIntentHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+import javax.inject.Inject
+
+class ShareIntentHandler @Inject constructor() {
+
+    /**
+     * This methods aims to handle incoming share intents.
+     *
+     * @return true if it can handle the intent data, false otherwise
+     */
+    fun handleIncomingShareIntent(context: Context, intent: Intent, onFile: (List<ContentAttachmentData>) -> Unit, onPlainText: (String) -> Unit): Boolean {
```
bmarty, line 32: Is it possible to inject the context in the constructor?

vector/src/main/java/im/vector/app/features/attachments/ShareIntentHandler.kt
bmarty, line 35: Could be nice to use constant defined in https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/MimeTypes.kt#L22 (or define new one there)

vector/src/main/java/im/vector/app/features/attachments/ShareIntentHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+        val type = intent.resolveType(context) ?: return false
+        return when {
+            type == "text/plain" -> handlePlainText(intent, onPlainText)
+            type.startsWith("image") -> {
```
bmarty, line 36: See also https://github.com/vector-im/element-android/blob/main/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/util/MimeTypes.kt#L38

bmarty (commented):
Nice work!
Just added some non-blocking remarks

vector/src/main/java/im/vector/app/features/attachments/ShareIntentHandler.kt
ouchadam: will do! the context is only being used to pull out the `contentResolver` :+1:

vector/src/main/java/im/vector/app/features/attachments/ShareIntentHandler.kt
ouchadam: updated 169ac9d0a08da3ad0acddaf74c013eb2733279f8

vector/src/main/java/im/vector/app/features/attachments/ShareIntentHandler.kt
ouchadam: 59ef8e10c8e649a8a2b9226cead1291e68416358

Issue #4777, opened 2021-12-20
Forceful conversion to numbered lists in messages causes messages to change in meaning


Avamander:
### Steps to reproduce

1. Someone asks me a question:
```
We have three choices here:
1. Pizza with pineapple
2. Sushi
3. Ramen

What do you want to eat?
```
2. I answer: `2. sounds good`
3. I'm brought pizza with pineapple because Element Android changed my `2.` to `1.` due to its internal conversion to numbered list (with a single item)

This is **incredibly** awful.

#### What did you expect?
That my message is displayed how it was sent.

#### What happened instead?
The meaning of my message was changed.

### Operating system version

Android 10

### Application version and app store

1.3.9

### Homeserver

matrix.org

ouchadam:
this looks like a missing feature with the markdown/html library we use, have raised an issue https://github.com/noties/Markwon/issues/381

notramo:
This is standard Markdown feature (automatically numbering ordered list).
Maybe disable ordered lists in Element?

basxto:
Biggest issue is that Element Desktop and Element Android handle this differently, so the answer will look different with different clients. Element Desktop also makes it into a list, but it does not do automatical numbering.

List decorations are also dropped in the _Direct Messages_, there the answer would be just rendered as `sounds good`

HarHarLinks:
it is really horrible when you try to make an appointment and e.g. suggest "4. April" and depending whether the other person(s) use(s) element web, android, or iOS, they see "1. April". Then suddenly on 1. April you get a call why you're late.

daniellekirkwood:
Have triaged on the WTFs board, will remove myself now 👍

Seil0:
How is this only a minor issue? This makes element unusable on Android.

notramo:
Because if sponsor reports issue, then it's a serious issue despite if it doesn't even affects the UX (or very minimally - see confetti animation or typing indicator).
If somebody other reports it, who doesn't pay, then it's a minor issue, despite it can be so serious that users migrate back to proprietary chat platforms. See the #2504 call glare bug, which repeatedly makes users migrate back to Facebook™ Messenger™, but the Element Android don't give a s*** for since 2 years, despite it's probably 30 lines of code and probably can be fixed in 20 minute by somebody who is familiar with the codebase (I unfortunately don't know Kotlin, and I'm not familiar with Android development).
So don't expect a fix for this bug either.
[...]

PR #6473, opened 2022-07-05
Fixing numbered lists always starting from 1

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Includes the html tag start position when parsing html Lists. 

- Copies the [original ListHandler](https://github.com/noties/Markwon/blob/master/markwon-html/src/main/java/io/noties/markwon/html/tag/ListHandler.java#L44) and takes into account the initial starting value.
- Providing our own list tag handler overrides the default

## Motivation and context

To fix the confusing behaviour of numbered lists losing their initial value #4777 

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot_20220705_143148|Screenshot_20220705_142811|

## Tests

- From Element web, send a message starting with `#.` eg `5. July`
- Notice the 5 has been converted to a 1

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/vector/src/main/java/im/vector/app/features/html/ListHandlerWithInitialStart.java
+++ b/vector/src/main/java/im/vector/app/features/html/ListHandlerWithInitialStart.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2022 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


opusforlife2:
OHGODTHANKYOU

vector/src/main/java/im/vector/app/features/html/ListHandlerWithInitialStart.java
```diff
@@ -0,0 +1,111 @@
...
+        if (!ol && !ul) {
+            return;
+        }
+
+        final MarkwonConfiguration configuration = visitor.configuration();
+        final RenderProps renderProps = visitor.renderProps();
+        final SpanFactory spanFactory = configuration.spansFactory().get(ListItem.class);
+
+        // Modified line
+        int number = Integer.parseInt(block.attributes().containsKey(START_KEY) ? block.attributes().get(START_KEY) : "1");
```
ouchadam, line 67: this line is the fix for taking into account the starting position

fedrunov (approved):
LGTM

Issue #5803, opened 2022-04-20
Set up Nightly builds


manuroe:
We want to automate the release of develop builds so that we can have feedbacks sooner a la develop.element.io.

### Requirements

1. We want the nightly build to be as close as possible to the release version of the app
2. We want the nightly build to be installed along with the release version of the app
3. We want the nightly build to have its own signature so that we can share some secrets with GitHub, without taking the risk to compromise the signature of the main app
4. The nightly build is set up for the GPlay variant of the app (nightly builds for F-Droid users already exist at https://fdroid.krombel.de/)
5. We want the nightly builds to be built once a day, during the night, when the CI is not too busy by building PRs
6. We want the nightly build to be built and released automatically
7. Nightly build are built from the `develop` branch of the app. So it includes all merged PR since the previous build
8. We want nightly build to upgrade installed nightly build app automatically
9. We want to be able to detect nightly build and the version on rageshakes.
10. We want nightly build to be available to anyone
11. Analytics?

### Actions

- [x] Create a new `buildTypes` named `nightly`, base on `release`. Use `initWith release`
- [x] Create a new signature
- [x] add the signature to the project (we already have the debug signature to the project).
- [x] Store the secrets into Passbolt
- [x] Share the secrets to GitHub Secret
- [x] Nightly build will have `.nightly` suffix for the appId and ` nightly` suffix for the app name (**EDIT** We do not update the app name to not alter strings with template using the app name). We could also change the background of the icon to something dark (code 07007E) (**EDIT** This is done).
- [x] `versionName` will be the current version of develop with a `-nightly` suffix (for instance "1.4.28-nightly")
- [x] `versionCode` will be the versionCode from develop, i.e. based on timestamp (no change to do here)
- [x] Create a dedicated file `google-services.json` with the package name of the nightly
- [x] Configure a GitHub action to build the nightly build and publish them
- [x] Rageshake will include a new "Nightly" label
- [x] Use Firebase app distribution to release the nightly build. With gradle https://firebase.google.com/docs/app-distribution/android/distribute-gradle, with client: https://firebase.google.com/docs/app-distribution/android/distribute-cli. Not sure what is the best option so far (?).
- [x] Document how to subscribe to the nightly build

### Once complete

- [ ] Advertise that the nightly build is available, providing the documentation about it

genofire:
[...]

PR #6478, opened 2022-07-05
Nightly build publish on Firebase

bmarty:
Related to #5803 

Description WIP


vector/src/nightly/java/im/vector/app/config/AnalyticsConfig.kt
```diff
@@ -0,0 +1,27 @@
+/*
```
bmarty, line 1: It's a pain I had to copy paste all those files from `/release`. Still looking for a better solution.
But there are diff in this file and in file `shortcuts.xm` :/

vector/build.gradle
```diff
@@ -1,6 +1,7 @@
 import com.android.build.OutputFile
 
 apply plugin: 'com.android.application'
+apply plugin: 'com.google.firebase.appdistribution'
```
bmarty, line 4: I hope this will not be a problem for F-Droid build, but since we already have non free `com.google.android.gms.oss-licenses-plugin` just below, I think this will be OK.

ouchadam: if we find out it is a problem we can use a similar mechanism to the coverage task and [conditionally apply the plugin](https://github.com/vector-im/element-android/pull/6478/files#diff-f3c10f5faa6b2c6a9a08eb7516581fb5afea4f1e5382c9b561666b8a6f4b6826L12)

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,48 @@
...
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: Install towncrier
+        run: |
+          python3 -m pip install towncrier
+      - name: Prepare changelog file
+        run: |
+          mv towncrier.toml towncrier.toml.bak
+          sed 's/CHANGES\.md/CHANGES_NIGHTLY\.md/' towncrier.toml.bak > towncrier.toml
```
ouchadam, line 38: should we commit the _changes_ file after successfully deploying the nightly?

vector/src/main/res/drawable/ic_launcher_background.xml
```diff
@@ -5,6 +5,6 @@
     android:viewportHeight="108">
   <path
       android:pathData="m0,0h108v108h-108z"
-      android:fillColor="#0DBD8B"
+      android:fillColor="@color/launcher_background"
```
ouchadam, line 8: :100:

vector/build.gradle
```diff
@@ -268,6 +278,40 @@ android {
             }
             // signingConfig signingConfigs.release
         }
+
+        nightly {
+            applicationIdSuffix ".nightly"
```
ouchadam, line 283: we'll also need to update the sygnal config to include the new package/sha

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,48 @@
...
+          rm towncrier.toml.bak
+          yes n | towncrier --version nightly
+      - name: Build and upload Gplay Nightly APK
+        run: |
+          ./gradlew assembleGplayNightly appDistributionUploadGplayNightly $CI_GRADLE_ARG_PROPERTIES --stacktrace
```
ouchadam, line 43: in the future we may want to also run the tests beforehand to ensure we don't release a potentially broken build~

docs/nightly_build.md
```diff
@@ -0,0 +1,54 @@
...
+
+Nightly builds are built and released to Firebase every days, and automatically.
+
+This is recommended to exclusively use this app, with your main account, instead of Element Android, and fallback to Element Android just in case of regression, to discover as soon as possible any regression, and report it to the team. To avoid double notification, you may want to disable the notification from the Element Android production version. Just open Element Android, navigate to `Settings/Notifications` and uncheck `Enable notifications for this session`.
+
+*Note:* Due to a limitation of Firebase, the nightly build is the universal build, which means that the size of the APK is a bit bigger, but this should not have any other side effect.
+
+## How to register to get nightly build
```
ouchadam, line 21: it's probably worth calling out the tester limit of 500 https://firebase.google.com/docs/app-distribution/add-remove-testers?platform=android#add-remove-testers-group , we'll need to request more slots (if needed) https://console.firebase.google.com/project/_/appdistribution/limits

docs/nightly_build.md
```diff
@@ -0,0 +1,54 @@
...
+
+Provide your email to the Android team, who will add it to the list "External testers" on Firebase. You will then receive an invite on the provided email.
```
ouchadam, line 23: should we mention the user will need to have a google account to access the nightly build via firebase?

vector/build.gradle
bmarty: Are you sure about that? (I will double check that I can receive Push with the nightly)

docs/nightly_build.md
bmarty: Are you sure about that? There are other email type in the group `external-testers`

vector/build.gradle
ganfra: For me it should be working as it's a suffix

ganfra (approved):
Ok for me if pushs are working

.github/workflows/nightly.yml
bmarty: I do not think this is necessary, this is just a temporary file which will contain the changelog sent to firebase.

vector/build.gradle
ouchadam: ah you're right! it's working locally for me, I thought we'd need to introduce the package into https://github.com/matrix-org/sygnal/blob/main/sygnal.yaml.sample#L203

I guess we're okay because the GCM api key is the same for nightly and debug (I thought firebase would validate package names, shas and api keys :thinking: )

ouchadam (approved):

PR #6482, opened 2022-07-06
Add GitHub action to check for a towncrier file

### LINKED ISSUE #6481: Let a GitHub action check if there is a file for towncrier on every PR.
Using this action:

```sh
      - name: Check for presence of a Change Log fragment (only pull requests)
        run: |
          # Fetch the pull request' base branch so towncrier will be able to
          # compare the current branch with the base branch.
          # Source: https://github.com/actions/checkout/#fetch-all-branches.
          git fetch --no-tags origin +refs/heads/${BASE_BRANCH}:refs/remotes/origin/${BASE_BRANCH}
          towncrier check --compare-with origin/${BASE_BRANCH}
        env:
          BASE_BRANCH: ${{ github.base_ref }}
        if: github.event_name == 'pull_request'
```

Inspired from https://github.com/oasisprotocol/oasis-core/blob/66d5a6292f9014ad5208fce2e34afd0ceff359af/.github/workflows/ci-lint.yml#L51-L60 (maybe outdated)

This may not be a required action, since for instance Dependabot are not creating file for towncrier (yet? if it is possible it could be great)

bmarty:
This PR adds a GitHub action to check if a file has been added to the directory `changelog.d`.

This PR does not include such file (yet), because I want the new check to detect it.

This GHA will not be mandatory for all PRs, but can be used as a reminder.

The plan is also to check that `towncrier` can be setup on GHA, because it will be used in #6478 

Closes #6481

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -207,7 +207,7 @@ jobs:
   towncrier:
     name: Towncrier check
     runs-on: ubuntu-latest
-    if: github.head_ref == 'develop'
+    if: github.event_name == 'pull_request' && github.head_ref == 'develop'
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python 3.8
@@ -226,4 +226,3 @@ jobs:
           towncrier check --compare-with origin/${BASE_BRANCH}
       env:
         BASE_BRANCH: ${{ github.base_ref }}
-        if: github.event_name == 'pull_request'
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -203,3 +203,26 @@ jobs:
           name: detekt-report
           path: |
             */build/reports/detekt/detekt.html
+
+  towncrier:
+    name: Check that there is at least one file for the changelog
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Install towncrier
+        run: |
+          python3 -m pip install towncrier
+      - name: Run towncrier
+        run: |
+        # Fetch the pull request' base branch so towncrier will be able to
+        # compare the current branch with the base branch.
+        # Source: https://github.com/actions/checkout/#fetch-all-branches.
+        git fetch --no-tags origin +refs/heads/${BASE_BRANCH}:refs/remotes/origin/${BASE_BRANCH}
+        towncrier check --compare-with origin/${BASE_BRANCH}
+      env:
+        BASE_BRANCH: ${{ github.base_ref }}
+        if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


.github/workflows/quality.yml
```diff
@@ -203,3 +203,26 @@ jobs:
           name: detekt-report
           path: |
             */build/reports/detekt/detekt.html
+
+  towncrier:
+    name: Towncrier check
+    runs-on: ubuntu-latest
+    if: always() && github.event_name == 'pull_request'
```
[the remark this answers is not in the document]
michaelkaye: Other thought is whether you want this to apply if the target is not `develop` - eg PRs against a release branch?

.github/workflows/quality.yml
```diff
@@ -203,3 +203,26 @@ jobs:
...
+      - name: Install towncrier
+        run: |
+          python3 -m pip install towncrier
+      - name: Run towncrier
+        run: |
+        # Fetch the pull request' base branch so towncrier will be able to
+        # compare the current branch with the base branch.
+        # Source: https://github.com/actions/checkout/#fetch-all-branches.
+        git fetch --no-tags origin +refs/heads/${BASE_BRANCH}:refs/remotes/origin/${BASE_BRANCH}
+        towncrier check --compare-with origin/${BASE_BRANCH}
```
michaelkaye, line 226:
```suggestion
          # Fetch the pull request' base branch so towncrier will be able to
          # compare the current branch with the base branch.
          # Source: https://github.com/actions/checkout/#fetch-all-branches.
          git fetch --no-tags origin +refs/heads/${BASE_BRANCH}:refs/remotes/origin/${BASE_BRANCH}
          towncrier check --compare-with origin/${BASE_BRANCH}
```

michaelkaye: (for indentation to match earlier indentation)

.github/workflows/quality.yml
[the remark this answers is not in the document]
michaelkaye: the always() in other examples is if the step depends on other steps; it might fail to run because those others failed; but this is entirely parallel.

(though, it might be good to have this check at the start of a tree of checks - only bother running the slow tests once the fast tests are passing, as otherwise you will need to re-run the slow checks again when the changelog file has been committed)

michaelkaye (approved):
Looks like it'll do the right thing, as I see; few commments/thoughts but none of them block merging.

The suggestion is just indentation.

.github/workflows/quality.yml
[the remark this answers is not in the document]
bmarty: OK, thanks for explanation. Adding `always()` was an attempt (9335886a5af70f07deb61799dffd5120a4e52079) to trigger this action for this particular PR. I will drop this commit.

> if the target is not develop

In this case I think we do not want to run this check. But I have no idea how to do that. Will do some research.

.github/workflows/quality.yml
bmarty: If I do that, I am loosing the syntactic coloration, which is strange:

Without indentation:
image

With indentation:
image

this is maybe a bug in the editor, but it makes me not confident to do this change...

bmarty: OK, done in 6feca4badf939353fab736bcf173cb684e28e854

PR #6494, opened 2022-07-07
Suppress webview / checkbox permission dialog

Johennes:
This suppresses the web view / checkbox permission dialog and redirects straight to the system permission dialog (in case it's needed).


onurays (approved):
Nice improvement, thank you!

PR #6501, opened 2022-07-08
[Timeline] - Collapse redacted events (PSG-523)

### LINKED ISSUE #6487: [Timeline] - Collapse redacted events
In the timeline, when we have more than or exactly 3 successive redacted events, we should collapse the events.

To avoid taking non necessary spaces in the timeline

Collapse the successive events as other collapsable events.

_No response_

### Additional context

_No response_

Yes

mnaturel:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Collapsing multiple successive redacted events to save space on timeline. Refactoring a little the code to ease the readability.

## Motivation and context

Closes #6487 

## Tests

- Go to a room
- Send multiple different types of messages
- Delete successive sent messages
- Check successive redacted messages are collapsed
- Check you can expand to see redacted messages individually

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 11

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MergedHeaderItemFactory.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MergedHeaderItemFactory.kt
@@ -24,7 +24,6 @@ import im.vector.app.features.home.room.detail.timeline.TimelineEventController
 import im.vector.app.features.home.room.detail.timeline.helper.AvatarSizeProvider
 import im.vector.app.features.home.room.detail.timeline.helper.MergedTimelineEventVisibilityStateChangedListener
 import im.vector.app.features.home.room.detail.timeline.helper.TimelineEventVisibilityHelper
-import im.vector.app.features.home.room.detail.timeline.helper.canBeMerged
 import im.vector.app.features.home.room.detail.timeline.helper.isRoomConfiguration
 import im.vector.app.features.home.room.detail.timeline.item.BasedMergedItem
 import im.vector.app.features.home.room.detail.timeline.item.MergedRoomCreationItem
@@ -35,6 +34,7 @@ import im.vector.app.features.home.room.detail.timeline.tools.createLinkMovement
 import org.matrix.android.sdk.api.crypto.MXCRYPTO_ALGORITHM_MEGOLM
 import org.matrix.android.sdk.api.extensions.orFalse
 import org.matrix.android.sdk.api.query.QueryStringValue
+import org.matrix.android.sdk.api.session.events.model.Event
 import org.matrix.android.sdk.api.session.events.model.EventType
 import org.matrix.android.sdk.api.session.events.model.content.EncryptionEventContent
 import org.matrix.android.sdk.api.session.events.model.toModel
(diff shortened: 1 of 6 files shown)


fedrunov (approved):
Looks good, but added some NITs

vector/src/main/res/values/strings.xml
```diff
@@ -3094,4 +3094,8 @@
     <string name="live_location_labs_promotion_description">Please note: this is a labs feature using a temporary implementation. This means you will not be able to delete your location history, and advanced users will be able to see your location history even after you stop sharing your live location with this room.</string>
     <string name="live_location_labs_promotion_switch_title">Enable location sharing</string>
 
+    <plurals name="room_removed_messages">
+        <item quantity="one">%d message removed</item>
+        <item quantity="other">%d messages removed</item>
```
bmarty, line 3099: When expanded, we can see the wording "Message deleted". So this is not coherent with this new wording (deleted/removed).
If we want to keep "%d messages removed", maybe we could change the other wording to "Message removed"?
The action to redact a message is "Remove..." so I think we should change the other wording.

Also, FTR (but I do not call for a change in this PR) the message on Element Web is:

image

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MergedHeaderItemFactory.kt
```diff
@@ -78,19 +79,60 @@ class MergedHeaderItemFactory @Inject constructor(
...
+     * @param nextEvent is an older event than event
+     * @param addDaySeparator true to add a day separator
+     */
+    private fun isStartOfRedactedEventsSummary(
+            event: TimelineEvent,
+            nextEvent: TimelineEvent?,
+            addDaySeparator: Boolean,
+    ): Boolean {
+        return event.root.isRedacted() &&
+                ((nextEvent?.root?.getClearType() != EventType.REDACTION && !nextEvent?.root?.isRedacted().orFalse()) || addDaySeparator)
```
mnaturel, line 132: This condition does not work for all cases: some redacted events may not be collapsed since the `nextEvent` (older than the current `event`) is of type `EventType.REDACTION`. We should go further in the events until we have a non redacted event which is not of type `EventType.REDACTION`.

mnaturel: Fixed in [59632bf](https://github.com/vector-im/element-android/pull/6501/commits/59632bf00ffca58033a5bb2ef7fa4373ccd4ee96)

vector/src/main/res/values/strings.xml
mnaturel: I agree we should change to the same syntax "remove". But I guess we should create a new String for that and we cannot update the english translation of the existing one? And I am concerned about the synchronization of this change with the other platforms if we do such a change.

Also, the difference with the merged message with Web is explained by the fact we may want message to be shorter in mobile so I am guessing that is why we do not repeat the author name of the message. The authors are already visible through avatar icons.

bmarty: We can update the English string if the meaning is close. This is the case here. Translators will see the the string has been change and may update the translations. Synchronization is important, but wording coherency in a single client is more important IMHO. We may discuss that with other teams.

mnaturel: Agree. Updated in [6eb05f8](https://github.com/vector-im/element-android/pull/6501/commits/6eb05f8eb51b7c6afa3505a1ab736d9b4be15353)

PR #6517, opened 2022-07-11
Avoiding direct `BuildConfig` usage

### LINKED ISSUE #6406: Avoid direct BuildConfig usages
In order to migrate to a modular world we'll need to avoid directly using the `BuildConfig` as we'll want to have a single entry point rather than multiple build files with their own configuration.

The abstraction should be injectable via the dagger graph and provided by the main `Android Application` module

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Moves the `AnalyticsConfig` to the `Config`
- Introduces config based `LocationConfig`, `VoipConfig`, `CryptoConfig`
- Reintroduces the live location sharing feature flag so that clients can opt out at runtime 
- Moves `debug` and `application id` to the injectable `BuildMeta`

## Motivation and context

Fixes #6406 

Avoids interacting directly with the `BuildConfig` outside the application level dagger module, this allows us to extract logic to modules which are relying on the build configuration

## Screenshots / GIFs
No UI changes

## Tests

- General app smoke test

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 29


vector/build.gradle
```diff
@@ -284,7 +264,6 @@ android {
             versionName "${versionMajor}.${versionMinor}.${versionPatch}${getGplayVersionSuffix()}"
 
             resValue "bool", "isGplay", "true"
-            buildConfigField "boolean", "ALLOW_FCM_USE", "true"
```
ouchadam, line 287: `ALLOW_FCM_USE` is not used

vector/src/main/java/im/vector/app/core/di/ConfigurationModule.kt
```diff
@@ -0,0 +1,78 @@
...
+import im.vector.app.features.analytics.AnalyticsConfig
+import im.vector.app.features.call.webrtc.VoipConfig
+import im.vector.app.features.crypto.keysrequest.OutboundSessionKeySharingStrategy
+import im.vector.app.features.home.room.detail.composer.voice.VoiceMessageConfig
+import im.vector.app.features.location.LocationSharingConfig
+import im.vector.app.features.raw.wellknown.CryptoConfig
+
+@InstallIn(SingletonComponent::class)
+@Module
+object ConfigurationModule {
```
ouchadam, line 36: a dedicated module to provide the abstractions for `Config/BuildConfig`

vector/src/main/java/im/vector/app/features/notifications/NotificationBroadcastReceiver.kt
```diff
@@ -48,31 +48,32 @@ class NotificationBroadcastReceiver : BroadcastReceiver() {
     @Inject lateinit var activeSessionHolder: ActiveSessionHolder
     @Inject lateinit var analyticsTracker: AnalyticsTracker
     @Inject lateinit var clock: Clock
+    @Inject lateinit var actionIds: NotificationActionIds
```
ouchadam, line 51: the notification ids include the application package name which is now injected

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
```diff
@@ -189,15 +189,22 @@ object VectorStaticModule {
...
-
     @Provides
     fun providesPhoneNumberUtil(): PhoneNumberUtil = PhoneNumberUtil.getInstance()
 
     @Provides
     @Singleton
-    fun providesBuildMeta() = BuildMeta()
+    fun providesBuildMeta() = BuildMeta(
+            isDebug = BuildConfig.DEBUG,
+            sdkInt = Build.VERSION.SDK_INT,
```
bmarty, line 199: `sdkInt` is a bit out of scope for `BuildMeta` object (I just see that `BuildMeta` is not new, was not aware of this class before), since `Build.VERSION.SDK_INT` have different values at runtime (and not at compile time). We have `BuildVersionSdkIntProvider` on the SDK, and also `AndroidVersionTestOverrider` (that looks a bit hacky, I would be happier to have a similar `BuildVersionSdkIntProvider`) app side, which allow to do some test and overriding.

vector/src/main/java/im/vector/app/core/resources/BuildMeta.kt
```diff
@@ -16,8 +16,16 @@
 
 package im.vector.app.core.resources
 
-import android.os.Build
-
 data class BuildMeta(
-        val sdkInt: Int = Build.VERSION.SDK_INT
+        val isDebug: Boolean,
```
bmarty, line 20: Do we really need to propagate `isDebug` to submodules? I think if we chose the `Debug` (resp `Release`) `buildType`, all the submodules will use the same `buildType`, and so all `BuildConfig.DEBUG` will have the same values for all modules.

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -81,7 +81,8 @@ class HomeActivityViewModel @AssistedInject constructor(
         private val analyticsStore: AnalyticsStore,
         private val lightweightSettingsStorage: LightweightSettingsStorage,
         private val vectorPreferences: VectorPreferences,
-        private val analyticsTracker: AnalyticsTracker
+        private val analyticsTracker: AnalyticsTracker,
+        private val analyticsConfig: AnalyticsConfig,
```
bmarty, line 85: `analyticsConfig` appear to be unused, at least there is no other change form this file in GH interface.

vector/src/main/java/im/vector/app/features/notifications/NotificationActionIds.kt
```diff
@@ -0,0 +1,41 @@
...
+    val JOIN_ACTION = "${buildMeta.applicationId}.NotificationActions.JOIN_ACTION"
+    val REJECT_ACTION = "${buildMeta.applicationId}.NotificationActions.REJECT_ACTION"
+    val QUICK_LAUNCH_ACTION = "${buildMeta.applicationId}.NotificationActions.QUICK_LAUNCH_ACTION"
+    val MARK_ROOM_READ_ACTION = "${buildMeta.applicationId}.NotificationActions.MARK_ROOM_READ_ACTION"
+    val SMART_REPLY_ACTION = "${buildMeta.applicationId}.NotificationActions.SMART_REPLY_ACTION"
+    val DISMISS_SUMMARY_ACTION = "${buildMeta.applicationId}.NotificationActions.DISMISS_SUMMARY_ACTION"
+    val DISMISS_ROOM_NOTIF_ACTION = "${buildMeta.applicationId}.NotificationActions.DISMISS_ROOM_NOTIF_ACTION"
+    val TAP_TO_VIEW_ACTION = "${buildMeta.applicationId}.NotificationActions.TAP_TO_VIEW_ACTION"
+    val DIAGNOSTIC_ACTION = "${buildMeta.applicationId}.NotificationActions.DIAGNOSTIC"
+    val PUSH_ACTION = "${buildMeta.applicationId}.PUSH"
```
bmarty, line 40: Since these are not `const val` anymore, I think the naming convention is not followed anymore here.

vector/src/main/java/im/vector/app/features/notifications/NotificationUtils.kt
```diff
@@ -68,16 +67,13 @@ import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.random.Random
 
-/**
- * Util class for creating notifications.
- * Note: Cannot inject ColorProvider in the constructor, because it requires an Activity
- */
```
bmarty, line 74: Always a bit sad to see comment blocks vanish.

vector/src/main/java/im/vector/app/core/di/ConfigurationModule.kt
```diff
@@ -0,0 +1,78 @@
...
+            }
+    )
+
+    @Provides
+    fun providesLocationSharingConfig() = LocationSharingConfig(
+            mapTilerKey = Config.LOCATION_MAP_TILER_KEY,
+    )
+
+    @Provides
+    fun providesVoipConfig() = VoipConfig(
```
bmarty, line 75: 💯  for splitting between several small config classes!

bmarty (commented):
This is a nice improvement. My 2 cents

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
ouchadam: it's because the previous import has been replaced with an instance property of the same name `import im.vector.app.config.analyticsConfig` -> `private val analyticsConfig: AnalyticsConfig`
 

https://github.com/vector-im/element-android/pull/6517/files#diff-7153f5b6cf4c510293d53c73474e854a3a84e6003c88341472c0808377bf3462R118

vector/src/main/java/im/vector/app/core/di/SingletonModule.kt
ouchadam: used the existing provider https://github.com/vector-im/element-android/pull/6517/commits/3bfe25a49f8ae397ae7d45d3ac10c8db230f3427

vector/src/main/java/im/vector/app/core/resources/BuildMeta.kt
bmarty: Good points. Thanks!

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
bmarty: OK, sure.

vector/src/main/java/im/vector/app/features/notifications/NotificationUtils.kt
bmarty: OK, this is old comment I think, you are right. Also I do not understand why this class has `@Singleton`. And yes this would definitely needs a renaming. This class is maybe 6 or 7 years old :).

bmarty:
More remarks:

- There is some cleanup to do at https://github.com/vector-im/element-android/blob/dd397b9a48d8a00d042050ad7f0b59122e4a31b5/vector/src/main/java/im/vector/app/features/attachments/AttachmentsHelper.kt#L42
This is using BuildConfig from the SDK, this is not correct. We should use `BuildMeta.lowPrivacyLoggingEnabled` now. (the log could be moved to the implementation in TimelineFragment)
- There are still some occurences of `BuildConfig.DEBUG` in the SDK, is the plan to replace all with `BuildMeta`?
- Same remark for `Build.VERSION.SDK_INT`, we should use `BuildVersionSdkIntProvider` everywhere. Could be in a dedicated PR though.

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -2306,81 +2362,99 @@ class TimelineFragment @Inject constructor(
...
             }
+
             is EventSharedAction.IgnoreUser -> {
                 action.senderId?.let { askConfirmationToIgnoreUser(it) }
             }
+
             is EventSharedAction.OnUrlClicked -> {
                 onUrlClicked(action.url, action.title)
             }
+
```
bmarty, line 2439: Strange to see all those new lines. Is there a reason?

bmarty (approved):
thanks for the update. one final remark

PR #6535, opened 2022-07-13
fix olm session proliferation

### LINKED ISSUE #6534: Proliferation of olm session
According to some android logs, we can see as much as 300 olm session with the same device.

BillCarsonFr:
## Type of change

- [x] Bugfix

## Content

When a to_device olm message cannot be decrypted (BAD_ENCRYPTED_MESSAGE), the sending devices was marked as wedged (ratchet got desync). 
In order to fix it, the SDK will automatically try to refresh the session with this device by creating a new session and sending a new dummy prekey message.
This unwedging mechanism can only be done once per hour.

The issue is that when a device was marked as wedged it was never removed later. Meaning that if the app stayed on for long, a new olm session will be created every hour. This could explain the proliferation of olm sessions noticed in some logs:
Fixes #6534

## Tests

Can be tested by commenting out persisting olm sessions in the code, then force discard session and send messages. Kill/Restart the sending device will then restore an old ratchet version causing a desync.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/changelog.d/6534.bugfix
+++ b/changelog.d/6534.bugfix
@@ -1,0 +1,1 @@
+Unwedging could cause the SDK to force creating a new olm session every hour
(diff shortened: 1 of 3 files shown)


yostyle (approved):
LGTM

PR #6545, opened 2022-07-13
FTUE - Combined register copy review

### LINKED ISSUE #6546: FTUE - Register screen copy review
As a result of the copy review we have made the following changes to the register screen

- Removed the subtitle
- Removes the selected homeserver description 
- Introduced a footer showing the full matrix id
- Noticed an inconsistency between Android and iOS where we aren't checking for user name availability when deselecting the user name entry 

[Figma](https://www.figma.com/file/xdV4PuI3DlzA1EiBvbrggz/Login-Flow-v2?node-id=5261%3A49831)
2022-07-13T16:29:05,703928530+01:00

ouchadam:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Matches iOS and checks for user name availability on the loss of focus on the user name field
- Updates the register page design to match the latest iteration of the designs (removes the subtitle and selected homeserver description)
- Adds a footer to the user name reinforcing the full matrix id as a way of discoverability 
- Adds a check for minimum 8 characters password length

## Motivation and context
Fixes #6546

To simplify the registration screen copy and reinforce the matrix id concept

## Screenshots / GIFs

| BEFORE | AFTER |
| --- | --- |
|before-register|after-register

| PASSWORD MINIMUM | USERNAME TAKEN | FULL MATRIX ID |
| --- | --- | --- |
|after-password-min|after-username-taken|after-full-matrix-id

## Tests

- With the combined register feature flag enabled and start the registration process

- Attempt to register with an account that already exists
- Attempt to register with a password shorter than 8 characters

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

### DIFF
--- a/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
@@ -144,7 +144,7 @@ class OnboardingViewModel @AssistedInject constructor(
             is OnboardingAction.UpdateSignMode -> handleUpdateSignMode(action)
             is OnboardingAction.InitWith -> handleInitWith(action)
             is OnboardingAction.HomeServerChange -> withAction(action) { handleHomeserverChange(action) }
-            is OnboardingAction.MaybeUpdateHomeserverFromMatrixId -> handleMaybeUpdateHomeserver(action)
+            is OnboardingAction.UserNameEnteredAction -> handleUserNameEntered(action)
             is AuthenticateAction -> withAction(action) { handleAuthenticateAction(action) }
             is OnboardingAction.LoginWithToken -> handleLoginWithToken(action)
             is OnboardingAction.WebLoginSuccess -> handleWebLoginSuccess(action)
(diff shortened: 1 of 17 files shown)


vector/src/main/java/im/vector/app/features/onboarding/ftueauth/FtueAuthCombinedRegisterFragment.kt
```diff
@@ -55,8 +59,11 @@ import org.matrix.android.sdk.api.failure.isLoginEmailUnknown
 import org.matrix.android.sdk.api.failure.isRegistrationDisabled
 import org.matrix.android.sdk.api.failure.isUsernameInUse
 import org.matrix.android.sdk.api.failure.isWeakPassword
+import reactivecircus.flowbinding.android.widget.textChanges
 import javax.inject.Inject
 
+private const val MINIMUM_PASSWORD_LENGTH = 8
```
bmarty, line 65: My 2 cents: some homeservers may have other rules for password (length, char type usage, etc.). We may have a first error with "password must be 8 chars minimum", then later "password must be 10 chars minimum", which is quite annoying. I would avoid clients to do local check on this field and let the config server side.

vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
```diff
@@ -167,13 +170,47 @@ class OnboardingViewModel @AssistedInject constructor(
...
+    private fun handleUserNameEntered(action: OnboardingAction.UserNameEnteredAction) {
+        when (action) {
+            is OnboardingAction.UserNameEnteredAction.Login -> maybeUpdateHomeserver(action.userId)
+            is OnboardingAction.UserNameEnteredAction.Registration -> maybeUpdateHomeserver(action.userId, continuation = { userName ->
+                checkUserNameAvailability(userName)
+            })
+        }
+    }
+
+    private fun maybeUpdateHomeserver(userNameOrMatrixId: String, continuation: suspend (String) -> Unit = {}) {
```
Florian14, line 182: Not sure that the lint will be ok but I suggest to name the string parameter inside the lambda for clarity
```suggestion
    private fun maybeUpdateHomeserver(userNameOrMatrixId: String, continuation: suspend (userName: String) -> Unit = {}) {
```

bmarty: I agree with this remark, also having an optional lambda for `continuation` is a bit strange to me.
Adam may iterate later, we need to merge the PR and review the next one before tomorrow, and this is not a blocker.

bmarty (approved):

PR #6548, opened 2022-07-13
Realm migration bg

bmarty:
This PR tries to move the Realm migration DB into background.

The first commit ensure there is always a migration to do and the migration will last 5 minutes. To be dropped of course.

|Before - migration in `VectorApplication.onCreate()`|After - migration in `MainActivity` view model|
|-|-|
|DbMigrationInFg|DbMigrationInBg|

New design with a wording (will iterate with @ganfra ):

image

Before we can have ANR if the migration takes time. After this should not happen anymore.

I create the PR to make this work public.

There are still some limitations:
- [ ] the code is quite dirty.
- [ ] we should consider using a mono Activity pattern to avoid some UI side effect when opening the app from a notification, after the app has been killed.
- [x] deep links management are not handled yet
- [x] ShareActivity management
- [x] Notification intent management  
- [x] Handle the todo : `TODO This will not work, handle this case.`
- [ ] probably some remaining crashes...

### DIFF
--- a/vector/src/main/java/im/vector/app/features/MainActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/MainActivity.kt
@@ -22,6 +22,7 @@ import android.os.Bundle
 import android.os.Parcelable
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
+import com.airbnb.mvrx.viewModel
 import com.bumptech.glide.Glide
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import dagger.hilt.android.AndroidEntryPoint
@@ -47,6 +48,8 @@ import im.vector.app.features.signout.hard.SignedOutActivity
 import im.vector.app.features.themes.ActivityOtherThemes
 import im.vector.app.features.ui.UiStateRepository
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import kotlinx.parcelize.Parcelize
@@ -84,6 +87,8 @@ class MainActivity : VectorBaseActivity<ActivityMainBinding>(), UnlockedActivity
         }
     }
 
+    private val mainViewModel: MainViewModel by viewModel()
+
     override fun getBinding() = ActivityMainBinding.inflate(layoutInflater)
 
     override fun getOtherThemes() = ActivityOtherThemes.Launcher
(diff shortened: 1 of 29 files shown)


vector/src/main/java/im/vector/app/features/MainActivity.kt
```diff
@@ -103,15 +116,36 @@ class MainActivity : VectorBaseActivity<ActivityMainBinding>(), UnlockedActivity
...
-        // Handle some wanted cleanup
-        if (args.clearCache || args.clearCredentials) {
-            doCleanUp()
+    }
+
+    private fun handleAppStarted() {
+        if (intent.hasExtra(EXTRA_NEXT_INTENT)) {
+            // Start the next Activity
+            val nextIntent = intent.getParcelableExtra<Intent>(EXTRA_NEXT_INTENT)
+            startIntentAndFinish(nextIntent)
```
bmarty, line 137: this is working but when going back, we can see the MainActivity screen for a few ms...

vector/src/main/java/im/vector/app/features/start/StartAppAndroidService.kt
```diff
@@ -0,0 +1,63 @@
...
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        showStickyNotification()
+        startPollingActiveSession()
+        return START_STICKY
+    }
+
+    private fun startPollingActiveSession() {
+        globalScope.launch {
+            do {
+                delay(1.seconds.inWholeMilliseconds)
```
ganfra, line 52: Maybe add a listener to tell when the session is ready instead of polling?

ganfra (commented):
2 small remarks, otherwise LGTM. Not tested.

vector/src/main/java/im/vector/app/features/start/StartAppAndroidService.kt
bmarty: I thought about it, but I think this is not necessary because it will make the `ActiveSessionHolder` more complex for not real added value

vector/src/main/java/im/vector/app/features/MainActivity.kt
```diff
@@ -103,15 +130,54 @@ class MainActivity : VectorBaseActivity<ActivityMainBinding>(), UnlockedActivity
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        args = parseArgs()
-        if (args.clearCredentials || args.isUserLoggedOut || args.clearCache) {
-            clearNotifications()
+
+        startAppViewModel.onEach {
+            renderState(it)
```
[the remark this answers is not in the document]
bmarty: I have done it in [912d71b](https://github.com/vector-im/element-android/pull/6548/commits/912d71b0761c26a6c472e4bc88096f7215801981)

ganfra (approved):

bmarty:
OK. I have dropped the tmp commit.
PR can be merged if CI is happy.

PR #6618, opened 2022-07-21
new room list implementation for EditLayout

### LINKED ISSUE #6611: App Layout: New room list screen
For EditLayout project there is no need to change anything in `HomeDetailFragment` but `RoomListFragment` which is used to show rooms lists in the app doesn't fit new requirements. To support enabling/disabling by FeatureFlags we decided to put new home screen implementation to a separate class, so this class needs to be created

- [ ] Implementation will only be used for Home screen
- [ ] needs to support different types of data to be presented in list 
- [ ] needs to support carousels
- [ ] needs to provide a way to configure types of data, being presented

fedrunov:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : prep work for EditLayout project

There is already `RoomListFragment` and some classes around it, which are dedicated to showing different list of rooms with multiple sections, but they all are tied together and have a hardcoded sections per use case. In EditLayout we'll need to provide a list which can be configured by user and this require different approach + this list will not have sections as collapsible list parts and also it needs to be put under Feature flag. As a result I've created a new implementation, but in this PR is more like copy-past dummy with excessive things taken away. I'll expand this implementation and probably even change it's design, while I'll complete tasks with sections, necessary for EditLayout and in the end I would like to align `RoomListFragment` and `HomeRoomListFragment` to avoid duplication of code, introduced in this PR. In the best scenario there will be only on Fragment to show them all, but at this moment not everything is certain about EditLayout and also changes are too big for me to create a proper design for both things at once.

closes #6611

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/HomeDetailFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/HomeDetailFragment.kt
@@ -24,6 +24,7 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.core.view.isVisible
 import androidx.fragment.app.Fragment
+import com.airbnb.mvrx.Mavericks
 import com.airbnb.mvrx.activityViewModel
 import com.airbnb.mvrx.fragmentViewModel
 import com.airbnb.mvrx.withState
@@ -41,6 +42,7 @@ import im.vector.app.core.ui.views.CurrentCallsView
 import im.vector.app.core.ui.views.CurrentCallsViewPresenter
 import im.vector.app.core.ui.views.KeysBackupBanner
 import im.vector.app.databinding.FragmentHomeDetailBinding
+import im.vector.app.features.VectorFeatures
 import im.vector.app.features.call.SharedKnownCallsViewModel
 import im.vector.app.features.call.VectorCallActivity
 import im.vector.app.features.call.dialpad.DialPadFragment
@@ -48,8 +50,10 @@ import im.vector.app.features.call.webrtc.WebRtcCallManager
 import im.vector.app.features.home.room.list.RoomListFragment
 import im.vector.app.features.home.room.list.RoomListParams
 import im.vector.app.features.home.room.list.UnreadCounterBadgeView
+import im.vector.app.features.home.room.list.home.HomeRoomListFragment
 import im.vector.app.features.popup.PopupAlertManager
 import im.vector.app.features.popup.VerificationVectorAlert
+import im.vector.app.features.roomprofile.uploads.media.RoomUploadsMediaFragment
 import im.vector.app.features.settings.VectorLocale
 import im.vector.app.features.settings.VectorPreferences
 import im.vector.app.features.settings.VectorSettingsActivity.Companion.EXTRA_DIRECT_ACCESS_SECURITY_PRIVACY_MANAGE_SESSIONS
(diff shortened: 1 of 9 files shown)


vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
```diff
@@ -0,0 +1,226 @@
...
+                .setMessage(message)
+                .setPositiveButton(R.string.action_leave) { _, _ ->
+                    roomListViewModel.handle(HomeRoomListAction.LeaveRoom(roomId))
+                }
+                .setNegativeButton(R.string.action_cancel, null)
+                .show()
+    }
+
+    private fun getAdapterForData(data: HomeRoomSection): EpoxyControllerAdapter {
+        return when (data) {
```
ericdecanini, line 171: Optional: Is there any purpose for this `when` when we only have two cases? A single statement (maybe with `takeIf { it is RoomSummaryData }` might be simpler

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
```diff
@@ -0,0 +1,226 @@
...
+                context = requireActivity(),
+                roomId = event.roomSummary.roomId,
+                isInviteAlreadyAccepted = isInviteAlreadyAccepted,
+                trigger = ViewRoom.Trigger.RoomList
+        )
+    }
+
+    // region RoomListListener
+
+    override fun onRoomClicked(room: RoomSummary) {
```
ericdecanini, line 197: Not to be actioned, but I always wondered how feasible it is to lift interaction events like these ones into a separate class. What are your thoughts?

ericdecanini (approved):
Looking awesome!

vector/src/main/java/im/vector/app/features/home/HomeDetailFragment.kt
```diff
@@ -352,8 +357,12 @@ class HomeDetailFragment @Inject constructor(
             if (fragmentToShow == null) {
                 when (tab) {
                     is HomeTab.RoomList -> {
-                        val params = RoomListParams(tab.displayMode)
-                        add(R.id.roomListContainer, RoomListFragment::class.java, params.toMvRxBundle(), fragmentTag)
+                        if (vectorFeatures.isNewAppLayoutEnabled()) {
+                            add(R.id.roomListContainer, createHomeRoomListFragment(), fragmentTag)
```
bmarty, line 361: Is it necessary to invoke the `fragmentFactory` here? Just wondering if we could just to 

```suggestion
                            add(R.id.roomListContainer, HomeRoomListFragment::class.java, null, fragmentTag)
```

bmarty (approved):
LGTM, just one small question.

Also as a minor remark, changing the setting will not update the Home. The app will have to be restarted. Since the setting is tmp, I think this is fine.

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
fedrunov: Was thinking about this also, but for now I decided to keep it as it is. Will take a look when everything else will be complete

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
fedrunov: That's more for a next iterations. There will be more sense in this `when`, when more sections will be added

vector/src/main/java/im/vector/app/features/home/HomeDetailFragment.kt
fedrunov: I thought to follow existing pattern, taking in account that there could be params for HomeRoomListFragment later. But I think it's better to simplify for now and then extract to method if it will be necessary

PR #6643, opened 2022-07-26
Feature/bma/fix firebase popup

bmarty:
Ensure that the Firebase popup is displayed at most once a day.

There is no way to intercept the "Not now" action, this is Firebase internal.

Also the previous fix was not working.

With this implementation, we save the current day in shared pref to ensure that the popup is not displayed twice.

### DIFF
--- a/vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
+++ b/vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
@@ -20,6 +20,7 @@ import android.content.Context
 import androidx.core.content.edit
 import com.google.firebase.appdistribution.FirebaseAppDistribution
 import com.google.firebase.appdistribution.FirebaseAppDistributionException
+import im.vector.app.BuildConfig
 import im.vector.app.core.di.DefaultSharedPreferences
 import im.vector.app.core.time.Clock
 import timber.log.Timber
@@ -54,6 +55,7 @@ class NightlyProxy @Inject constructor(
     }
 
     private fun canDisplayPopup(): Boolean {
+        if (BuildConfig.APPLICATION_ID != "im.vector.app.nightly") return false
         val sharedPreferences = DefaultSharedPreferences.getInstance(context)
         val today = clock.epochMillis() / A_DAY_IN_MILLIS
         val lastDisplayPopupDay = sharedPreferences.getLong(SHARED_PREF_KEY, 0)
--- a/vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
+++ b/vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
@@ -29,7 +29,7 @@ class NightlyProxy @Inject constructor(
         private val clock: Clock,
         private val context: Context,
 ) {
-    fun updateIfNewReleaseAvailable() {
+    fun mayDisplayFirebasePopup() {
         if (!canDisplayPopup()) return
         val firebaseAppDistribution = FirebaseAppDistribution.getInstance()
         firebaseAppDistribution.updateIfNewReleaseAvailable()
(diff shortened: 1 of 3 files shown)


witchent:
Can we also have a setting (put it in labs or something) to completely disable this behaviour? I (and I am probably not alone in that) am using FDroid with the GPlay repo to keep Element up to date, so this Firebase thing is not needed.
Once a day is a lot better than before, but it would still be annoying.

vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
```diff
@@ -44,4 +53,24 @@ class NightlyProxy @Inject constructor() {
                     }
                 }
     }
+
+    private fun canDisplayPopup(): Boolean {
+        if (BuildConfig.APPLICATION_ID != "im.vector.app.nightly") return false
+        val sharedPreferences = DefaultSharedPreferences.getInstance(context)
```
ouchadam, line 59: we've recently added a provider for the default shared prefs via the annotation `@DefaultPreferences`
https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/core/di/SingletonModule.kt#L217

vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
```diff
@@ -44,4 +53,24 @@ class NightlyProxy @Inject constructor() {
                     }
                 }
     }
+
+    private fun canDisplayPopup(): Boolean {
+        if (BuildConfig.APPLICATION_ID != "im.vector.app.nightly") return false
```
ouchadam, line 58: do we need the application id guard if the logic lives within the nightly sourceset?

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
```diff
@@ -537,9 +537,7 @@ class HomeActivity :
         serverBackupStatusViewModel.refreshRemoteStateIfNeeded()
 
         // Check nightly
-        if (isFirstCreation()) {
-            nightlyProxy.updateIfNewReleaseAvailable()
-        }
+        nightlyProxy.mayDisplayFirebasePopup()
```
ouchadam, line 540: maybe we could avoid mentioning `firebase` here and instead rely on the entry point, something like `NightlyProxy.onHomeResumed`, what do you think?

vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
bmarty: ah yes, I will use it, thanks!

bmarty: 41d35e68a05a6c8cbbb2e006ee947b35f6f83428

vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
bmarty: It's in the `gplay` sourceset. I cannot add to `nightly` sourceset because of `initWith release`. So I cannot have a version for `release`, one for `debug` and one for `nightly` as I would like to, else I got duplicated class declaration .

Normally this popup would not affect GplayRelease version, but better safe than sorry.

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
bmarty: Yes, it's better, thanks. I was wondering if we want to call this also on LoginActivity, but I guess this is not necessary.

e5773203010f88fa2c05297d0d1cfd51d9448e91

vector/src/gplay/java/im/vector/app/nightly/NightlyProxy.kt
ouchadam: ah! thanks for explaining :+1:

ouchadam (approved):

PR #6656, opened 2022-07-27
Revert #6404

ganfra:
Revert #6404 as it breaks Thread and implementation is not working properly (timeline is not refreshed in live).
@Claire1817


ouchadam (approved):
looks like it'll be quite easy to reapply after the release :crossed_fingers:

Issue #6502, opened 2022-07-08
App Layout: Edit Layout [Android]


fedrunov:
Android specific [https://github.com/vector-im/element-meta/issues/220](https://github.com/vector-im/element-meta/issues/220)

- [x] new fragment for room list #6611 
- [x] Recents list #6504
- [x] List filters #6505
- [x] Layout settings: #6506
- [x] move dialpad from bottom tabs #6787
- [x] remove bottom tabs at Home #6565
- [x] invites #6889
- [x] invites margin #6947
- [x] Invites Empty States #6876 
- [x] Home screen empty states #6835
- [x] change app settings #6646
- [x] Accessibility #6856 
- [ ] Tests #6509
- [x] Release experience #6909

Nice-to-have
- [x] scroll to start when new item added/moved #6776
- [x] Analytics #6508
- [x] #6770

daniellekirkwood:
@fedrunov there's one remaining task in this epic - does it need to be done or are you happy that I close this out?

PR #6707, opened 2022-08-01
recents carousel for new home screen layout

### LINKED ISSUE #6504: App Layout: Recent chats list
Horizontal list with recent chats

[Figma](https://www.figma.com/file/Hw4gjP3pknMZUir8jQqmWH/%5BMobile%5D-IA-early-proposals?node-id=2593%3A140034)

image

Acceptance criteria:

- [x] Implement horizontal scrolling: [Carousel](https://github.com/airbnb/epoxy/wiki/Carousel) 
- [x] Use Breadcrumbs to form list
- [x] Display notification badge 
- [x] Click opens timeline 
- [x] Long click opens room menu (same as for general room list)
- [x] Doesn't depend on selected space

fedrunov:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

The carousel with recent chats, based on Breadcrumbs API

## Motivation and context

part of #6502 
closes #6504

## Screenshots / GIFs

|Light|Dark|Black|
|-|-|-|
|light|dark|black|

output-optimised


vector/src/main/java/im/vector/app/features/VectorFeatures.kt
```diff
@@ -53,5 +53,5 @@ class DefaultVectorFeatures : VectorFeatures {
     override fun isScreenSharingEnabled(): Boolean = true
     override fun forceUsageOfOpusEncoder(): Boolean = false
     override fun shouldStartDmOnFirstMessage(): Boolean = false
-    override fun isNewAppLayoutEnabled(): Boolean = false
+    override fun isNewAppLayoutEnabled(): Boolean = true
```
ericdecanini, line 56: Please keep this false by default

ericdecanini (changes requested):
Implementation looks good. Please just revert the VectorFeature

vector/src/main/java/im/vector/app/VectorApplication.kt
```diff
@@ -140,6 +144,11 @@ class VectorApplication :
         Mavericks.initialize(debugMode = false)
         EpoxyController.defaultDiffingHandler = EpoxyAsyncUtil.getAsyncBackgroundHandler()
         EpoxyController.defaultModelBuildingHandler = EpoxyAsyncUtil.getAsyncBackgroundHandler()
+        Carousel.setDefaultGlobalSnapHelperFactory(object : Carousel.SnapHelperFactory() {
```
mnaturel, line 147: Maybe we should extract this into a dedicated private method (e.g. `initDefaultSnapHelper`) to reduce the size of the `onCreate` method and ease the readability.

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
```diff
@@ -180,6 +183,14 @@ class HomeRoomListFragment @Inject constructor(
                     }
                 }.adapter
             }
+            is HomeRoomSection.RecentRoomsData -> RecentRoomCarouselController(
```
mnaturel, line 186: I see `RecentRoomCarouselController` is injectable. I think we should inject this controller in this fragment instead of `AvatarRenderer`.

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
```diff
@@ -180,6 +183,14 @@ class HomeRoomListFragment @Inject constructor(
...
+                    avatarRenderer = avatarRenderer
+            ).also { controller ->
+                controller.listener = this
```
mnaturel, line 189: I don't see any clean up of `RecyclerViews` in `onDestroyView()` of the fragment. Could we override it and clean the adapters associated to the lists (there is a `RecyclerView.cleanUp()` extension) + removing the listener from the controllers as we are passing the Fragment instance here.

vector/src/main/java/im/vector/app/features/home/room/list/home/recent/RecentRoomCarouselController.kt
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    override fun buildModels() {
+        val host = this
+        data?.let { data ->
+            carousel {
+                id("recents_carousel")
+                withModelsFrom(data) { originalRoomSummary ->
+
+                    var roomSummary = originalRoomSummary
```
mnaturel, line 48: I think this variable is not necessary. And if it is, it should be a `val` I guess.

mnaturel (commented):
Tested. LGTM. I have just added some comments to improve code.

vector/src/main/java/im/vector/app/VectorApplication.kt
fedrunov: I've moved it to `configureEpoxy` method along with other epoxy related things

mnaturel (commented):
Thanks for the updates! I have tested the app again but when enabling the new app layout, I don't see the carousel of recent rooms like before. Do I need to do something else? Also, since we no more have the left side menu in this new layout, I cannot find the hidden developer menu to access feature flags for example. Where can I find it when I am logged in?

fedrunov:
@mnaturel I blindly assumed that merging few PRs from dev won't break anything, while actually they broke whole thing. I've fixed problem in a separate pr and also update some margins for recents.

vector/src/main/res/layout/item_recent_room.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/recentRoot"
+    android:layout_width="60dp"
+    android:layout_height="wrap_content"
+    android:background="?android:colorBackground"
```
ElementBot, line 8: - :warning: Possible overdraw: Root element paints background `?android:colorBackground` with a theme that also paints a background (inferred theme is `@style/Theme.Vector.Light`)
- :warning: Possible overdraw: Root element paints background `?android:colorBackground` with a theme that also paints a background (inferred theme is `@style/Theme.Vector.Light`)

vector/src/main/res/layout/item_recent_room.xml
```diff
@@ -0,0 +1,61 @@
...
+    android:clickable="true"
+    android:focusable="true"
+    android:foreground="?attr/selectableItemBackground"
```
ElementBot, line 11: - :warning: Attribute `android:foreground` has no effect on API levels lower than 23 (current min is 21)
- :warning: Attribute `android:foreground` has no effect on API levels lower than 23 (current min is 21)

mnaturel:
> @mnaturel I blindly assumed that merging few PRs from dev won't break anything, while actually they broke whole thing. I've fixed problem in a separate pr and also update some margins for recents.

Thanks for the update. Now it works but I have a UX question: when I enter a room by pressing one of the item of the recent chats list and pressing back to come back home, it changes the ordering of the recent chats. I would expect to only change when posting messages on it. Was it done on purpose? And if the order changes would be possible to reload the position of the list at the start?

fedrunov:
> I would expect to only change when posting messages on it. 

This is how breadcrumbs work now, I haven't changed behaviour of that. @daniellekirkwood what do you think?

> And if the order changes would be possible to reload the position of the list at the start? Or maybe we could simply load the recent chat once when entering the home (not observing them).

I think this is only an issue when you have you list scrolled to the left (first item) and so when something moves to first position you expect list to stay at same position (0) not at same item (position = 1 after change). Do you know a way how to implement this? I know that we can react on inserting/changing item, but we'd also need to check scroll position (or first completely visible item position) before that, to not scroll when it's not necessary

daniellekirkwood:
> when I enter a room by pressing one of the item of the recent chats list and pressing back to come back home, it changes the ordering of the recent chats. I would expect to only change when posting messages on it. Was it done on purpose? 
Yes! This is how it's supposed to work - its the same on all the platforms. Recents act as Breadcrumbs so users can quickly jump in and out of the chats that they find most interesting. 

> And if the order changes would be possible to reload the position of the list at the start? Or maybe we could simply load the recent chat once when entering the home (not observing them).
It would make sense for the list to start at 0 again when hitting the home screen but happy to mark that as a "nice-to-have" for launch if we think it will take too much time to implement @fedrunov

mnaturel:
In this case, I guess we could try adding a call to `RecyclerView.scrollToPosition(0)` when arriving on the home screen?

amshakal:
Looking good. Just a few UI questions:
1. What font size are we using to display room names in recents? It seems way too big
2. When filters are deactivated and recents are activated, is it possible to have a line between them? It might imprive the hierarchy. Attaching image for reference. 
Screenshot 2022-08-08 at 1 09 18 pm

Otherwise, all is good.

fedrunov:
> In this case, I guess we could try adding a call to `RecyclerView.scrollToPosition(0)` when arriving on the home screen?

this will reset your scroll every time you go back, this could be very annoying. I've filed a separate issue for this, let's unblock this PR and move discussion there: https://github.com/vector-im/element-android/issues/6776

mnaturel (approved):

vector/src/main/res/layout/item_recent_room.xml
ElementBot, line 8: - :warning: Possible overdraw: Root element paints background `?android:colorBackground` with a theme that also paints a background (inferred theme is `@style/Theme.Vector.Light`)
- :warning: Possible overdraw: Root element paints background `?android:colorBackground` with a theme that also paints a background (inferred theme is `@style/Theme.Vector.Light`)

vector/src/main/res/layout/item_recent_room.xml
ElementBot, line 11: - :warning: Attribute `android:foreground` has no effect on API levels lower than 23 (current min is 21)
- :warning: Attribute `android:foreground` has no effect on API levels lower than 23 (current min is 21)

amshakal (approved):

ericdecanini (approved):
LGTM

PR #6710, opened 2022-08-01
Fixing `IllegalStateException: You should authenticate before using this` when restoring app

### LINKED ISSUE #6709: Crash when launching the app
```
Thread: main, Exception: java.lang.RuntimeException: Unable to start activity ComponentInfo{im.vector.app/im.vector.app.features.home.HomeActivity}: java.lang.IllegalStateException: You should authenticate before using this
```

```
Caused by: java.lang.IllegalStateException: 
  at im.vector.app.core.di.ActiveSessionHolder.getActiveSession (ActiveSessionHolder.kt:2)
  at im.vector.app.DaggerVectorApplication_HiltComponents_SingletonC$SingletonCImpl.-$$Nest$msession (DaggerVectorApplication_HiltComponents_SingletonC.java:3)
  at im.vector.app.DaggerVectorApplication_HiltComponents_SingletonC$ActivityCImpl$SwitchingProvider.get (DaggerVectorApplication_HiltComponents_SingletonC.java:166)
  at im.vector.app.core.di.VectorFragmentFactory.instantiate (VectorFragmentFactory.kt:11)
  at androidx.fragment.app.FragmentStateManager.<init> (FragmentStateManager.java:12)
  at androidx.fragment.app.FragmentManager.restoreSaveStateInternal (FragmentManager.java:35)
  at androidx.fragment.app.FragmentManager.attachController (FragmentManager.java:40)
  at androidx.fragment.app.FragmentActivity.lambda$init$3 (FragmentActivity.java:2)
  at androidx.fragment.app.FragmentActivity.$r8$lambda$euPNEtWNfVdMY89Jt5kWt_WEHqw (Unknown Source)
  at androidx.fragment.app.FragmentActivity$$ExternalSyntheticLambda0.onContextAvailable (Unknown Source:2)
  at androidx.activity.ComponentActivity.onCreate (ComponentActivity.java:5)
  at androidx.fragment.app.FragmentActivity.onCreate (FragmentActivity.java:1)
  at im.vector.app.core.platform.VectorBaseActivity.onCreate (VectorBaseActivity.kt:9)
  at im.vector.app.features.home.HomeActivity.onCreate (HomeActivi
```

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Always attempts to reinitialise the `Session` instance if we have a cached authenticated session.

## Motivation and context

Fixes #6709 - crash when launching the app 

- #6548 allowed the realm database migration to occur off the main thread but had the condition that the session must be initialised via the `MainActivity`, however we can't guarantee this screen will appear first as the system may attempt to restore other screens directly along with the Application at the same time

## Screenshots / GIFs

*With a manual process restart after 5 seconds~

|Before|After|
|-|-|
|before-restart-process|after-recreate-process|
 

## Tests

Add the following block to `HomeActivity` to force the process to be restarted with the current launch intent

```kotlin
fun onCreate() {
  Handler(Looper.getMainLooper()).postDelayed(Runnable {
      triggerRebirth(this@HomeActivity)
  }, 10000)
}

fun triggerRebirth() {
    val intent = Intent(this, HomeActivity::class.java)
            .apply {
                addFlags(Intent.FLAG_ACTIVITY_NEW_TASK and Intent.FLAG_ACTIVITY_CLEAR_TASK)
                putExtra(Mavericks.KEY_ARG, HomeActivityArgs(
                        clearNotification = true,
                        authenticationDescription = null,
                ))
            }
    startActivity(intent)
    Runtime.getRuntime().exit(0)
}
```

- Notice the app crashes

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):


vector/src/main/java/im/vector/app/core/di/ActiveSessionHolder.kt
```diff
@@ -80,18 +87,27 @@ class ActiveSessionHolder @Inject constructor(
     }
 
     fun hasActiveSession(): Boolean {
-        return activeSessionReference.get() != null
+        return activeSessionReference.get() != null || authenticationService.hasAuthenticatedSessions()
     }
 
     fun getSafeActiveSession(): Session? {
-        return activeSessionReference.get()
+        return runBlocking { getOrInitializeSession(startSync = true) }
```
ouchadam, line 94: this is the crash fix, we'll always attempt to initialise the session (including perform a migration if needed)

the hope is that most users will only perform a migration via the `MainActivity` flow (tapping the launcher icon/notification) which will give them a loading bar, however it's possible the ROMs/OEMs can restore the activity and application directly

vector/src/main/java/im/vector/app/core/di/SessionInitializer.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * A thread safe way to initialize the last authenticated Session instance.
+     *
+     * @param readCurrentSession expects an in-memory Session to be provided or null if not yet set.
+     * @param initializer callback to allow additional initialization on the Session, such as setting the in-memory Session instance.
+     * @return the initialized Session or null when no authenticated sessions are available.
+     */
+    suspend fun tryInitialize(readCurrentSession: () -> Session?, initializer: (Session) -> Unit): Session? {
+        return withContext(INITIALIZER_CONTEXT) {
```
ouchadam, line 41: the initialisation is thread safe and will block or wait for any in progress initialisers to complete, this means that ideally all access to `ActiveSessionHolder` should be off the main thread to avoid an possible ANRs, however this is quite a large refactor

the most common flow will be via the `MainActivity` which is off the main thread

vector/src/main/java/im/vector/app/core/pushers/VectorMessagingReceiver.kt
```diff
@@ -204,7 +198,7 @@ class VectorMessagingReceiver : MessagingReceiver() {
         }
     }
 
-    private fun getEventFastLane(session: Session, pushData: PushData) {
+    private suspend fun getEventFastLane(session: Session, pushData: PushData) {
```
ouchadam, line 201: reuses the suspension

vector/src/main/java/im/vector/app/core/di/ActiveSessionHolder.kt
```diff
@@ -80,18 +87,27 @@ class ActiveSessionHolder @Inject constructor(
     }
 
     fun hasActiveSession(): Boolean {
-        return activeSessionReference.get() != null
+        return activeSessionReference.get() != null || authenticationService.hasAuthenticatedSessions()
```
ouchadam, line 90: an active session is now a _in memory instance_ or _cached authenticated session_

this _should_ become async with the  `hasAuthenticatedSessions` being on the IO dispatcher, however as the active session is cached in memory almost straight away in most cases I've opt'd to avoid the breaking change (will hopefully be included in the realm-kotlin changes! :crossed_fingers: )

vector/src/main/java/im/vector/app/core/di/ActiveSessionHolder.kt
```diff
@@ -80,18 +87,27 @@ class ActiveSessionHolder @Inject constructor(
...
     }
 
     fun getActiveSession(): Session {
-        return activeSessionReference.get()
+        return getSafeActiveSession()
                 ?: throw IllegalStateException("You should authenticate before using this")
     }
 
+    suspend fun getOrInitializeSession(startSync: Boolean): Session? {
+        return activeSessionReference.get() ?: sessionInitializer.tryInitialize(readCurrentSession = { activeSessionReference.get() }) { session ->
```
ouchadam, line 103: we're eagerly reading from the in-memory cache to avoid a potential thread switch but it has the side effect of needing to query the in-memory value again within the initializer thread

vector/src/main/java/im/vector/app/core/di/ActiveSessionHolder.kt
ouchadam, line 94: ideally `getActiveSession` and `getSafeActiveSession` would be suspending instead of using `runBlocking` however the change needed is much bigger than the scope of this PR

vector/src/main/java/im/vector/app/features/start/StartAppViewModel.kt
```diff
@@ -55,11 +55,15 @@ class StartAppViewModel @AssistedInject constructor(
         handleLongProcessing()
         viewModelScope.launch(Dispatchers.IO) {
```
ganfra, line 56: Can you take care of inject Dispatchers instead of using directly IO?

vector/src/main/java/im/vector/app/core/di/SessionInitializer.kt
ganfra: Maybe safer to use a semaphore instead of a single thread dispatcher?

ganfra (commented):
Some small remarks otherwise LGTM

vector/src/main/java/im/vector/app/core/di/SessionInitializer.kt
ouchadam: will give it a try :+1:

vector/src/main/java/im/vector/app/features/start/StartAppViewModel.kt
ouchadam: will do :+1:

vector/src/main/java/im/vector/app/core/di/SessionInitializer.kt
ouchadam: works well! :100: https://github.com/vector-im/element-android/pull/6710/commits/9114630bbaec9020de897d970d7da1e2e739ae8f

vector/src/main/java/im/vector/app/features/start/StartAppViewModel.kt
ouchadam: https://github.com/vector-im/element-android/pull/6710/commits/3725921400631977c5fac483b18d25d10da84347

ganfra (approved):
All good, thanks!

PR #6714, opened 2022-08-02
Disable 'Enable biometrics' option if there are not biometric authenticators enrolled.

### LINKED ISSUE #6713: Users with no biometric features enrolled can enable biometric authentication
### Steps to reproduce

1. Make sure you have no biometric authenticators enrolled in your device (no fingerprint, no face unlock, etc.).
2. Open the app, go to Settings > Security & Privacy > Protect Access.
3. Enable PIN code.
4. See 'Enable biometrics' can be toggled.

#### What did you expect?

'Enable biometrics' should be disabled so it can't be toggled, as there are no biometric authenticators enrolled.

#### What happened instead?

It is enabled.

This can lead to crashes when starting the app, since we're asking for biometric auth but there is no biometric feature to unlock the prompt, and a `UserNotAuthenticatedException` will be thrown.

This can also be reproduced as a [weird bug in some emulators](https://github.com/vector-im/element-android/pull/6523#issuecomment-1187588542) *even if there are biometric features enrolled*.

### Your phone model

Emulator, Mi 9T Pro

### Operating system version

Android 12

1.4.32-dev [212085012] (G-1411fdd9) develop

### Homeserver

matrix.org

### Will you send logs?

No

Yes

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes #6713 .

## Motivation and context

Enabling biometric auth in the app is not restricted to users who have biometric authenticators enrolled in the device. This caused `UserNotAuthenticatedException` crashes for users who enabled this option without having those authenticators.

## Tests

- Make sure you have no biometric authenticators enrolled in your device (no fingerprint, no face unlock, etc.).
- Open the app, go to Settings > Security & Privacy > Protect Access.
- Enable PIN code.
- Check 'Enable biometrics' can't be toggled because it's disabled.
- Now add a fingerprint to your device.
- Go back to the app, in 'Protect Access'.
- Check that 'Enable biometrics' can be accessed now.

Note: I had some issues while testing this on emulators, it seems like an emulator bug. Having several fingerprints enrolled and enabling biometrics in the app caused the same `UserNotAuthenticatedException` crash mentioned above. Apparently, `BiometricManager` detects the valid authenticators but `BiometricPrompt` doesn't.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): 12

## Checklist


vector/src/main/java/im/vector/app/features/pin/lockscreen/ui/LockScreenFragment.kt
```diff
@@ -59,7 +59,7 @@ class LockScreenFragment : VectorBaseFragment<FragmentLockScreenBinding>() {
             if (state.lockScreenConfiguration.mode == LockScreenMode.CREATE) return@withState
 
             viewLifecycleOwner.lifecycleScope.launchWhenResumed {
-                if (state.isBiometricKeyInvalidated) {
+                if (state.canUseBiometricAuth && state.isBiometricKeyInvalidated) {
                     lockScreenListener?.onBiometricKeyInvalidated()
```
ouchadam, line 63: out of interest, what happens in this `onBiometricKeyInvalidated` flow?

vector/src/androidTest/java/im/vector/app/features/pin/lockscreen/crypto/KeyStoreCryptoTests.kt
```diff
@@ -69,10 +70,12 @@ class KeyStoreCryptoTests {
...
         keyStoreCrypto.hasValidKey() shouldBe true
 
-        val exception = KeyPermanentlyInvalidatedException()
-        every { secretStoringUtils.getEncryptCipher(any()) } throws exception
+        val keyInvalidatedException = KeyPermanentlyInvalidatedException()
+        every { secretStoringUtils.getEncryptCipher(any()) } throws keyInvalidatedException
+        keyStoreCrypto.hasValidKey() shouldBe false
 
-        runCatching { keyStoreCrypto.ensureKey() }
+        val userNotAuthenticatedException = UserNotAuthenticatedException()
```
ouchadam, line 77: thanks for adding the test case :100:

vector/src/main/java/im/vector/app/features/settings/VectorSettingsPinFragment.kt
```diff
@@ -66,20 +66,24 @@ class VectorSettingsPinFragment @Inject constructor(
...
+    private fun shouldEnableBiometricPref(isPinCodeChecked: Boolean): Boolean {
+        return isPinCodeChecked && biometricHelper.canUseAnySystemAuth
+    }
+
     override fun onResume() {
         super.onResume()
 
-        useBiometricPref.isEnabled = usePinCodePref.isChecked
-        useBiometricPref.isChecked = shouldCheckBiometricPref(usePinCodePref.isChecked)
+        useBiometricPref.isEnabled = shouldEnableBiometricPref(isPinCodeChecked = usePinCodePref.isChecked)
```
ouchadam, line 76: do you think it's worth extracting the `useBiometricPref.isEnabled  and useBiometricPref.isChecked` setting to a reusable function?

ouchadam (approved):
LGTM :+1: minor extraction comment, will leave up to you!

vector/src/main/java/im/vector/app/features/pin/lockscreen/ui/LockScreenFragment.kt
jmartinesp: There are 2 points where the `state. isBiometricKeyInvalidated ` can change and trigger this code:

1. `LockScreenViewModel.showBiometricPrompt`, this will call `BiometricHelper.authenticate` that will fail before showing any UI if the system key is not valid and instead it:
* Catches any exceptions.
* Posts a `LockScreenViewEvent.AuthError` to handle it in the UI.
* If the error was an instance of `KeyPermanentlyInvalidatedException`, it will call `removeBiometricAuthentication()` which will disable the 'biometric enabled' option, delete the broken system key and update the current state to display an alert saying that the key has just been invalidated.

2. `LockScreenViewModel.updateStateWithBiometricInfo` is automatically called when the  `LockScreenConfiguration` changes:
* It calls `biometricHelper.isSystemKeyValid`.
* This will reach `KeyStoreCrypto.hasValidKey()`, which tries to retrieve the current system key and either return true or catch both `KeyPermanentlyInvalidatedException` and `UserNotAuthenticatedException`, which are thrown when the internal `Cipher.init` fails, and return false.
* `state.isBiometricKeyInvalidated` will be updated with this returned value and used to display the same alert as in the case above.

ouchadam: thanks for explaining :+1:

PR #6728, opened 2022-08-03
Fixing CI crashing during compilation/linting

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Reduces the JVM memory allowance, we have multiple toolchains competing for memory which is causing the upper limit (currently 4g) to be reached however the VM doesn't have this much available, causing the compilation to crash.

- Makes use of the `GRADLE_OPTS` environmental var to avoid needing to pass the additional flags
- Re-enables parallel compilation, disables predexing and enforces a max worker count

## Motivation and context

We've been seeing the android lint task fail due to process crashes (out of memory) and the introduction of top level module exaggerated the issue https://github.com/vector-im/element-android/pull/6720

The fix is to limit the upper bound of the jvm during compilation to avoid over allocating available VM memory.

## Screenshots / GIFs
N/A

## Tests

- Run the CI!

## Tested devices
N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
 
 # Enrich gradle.properties for CI/CD
 env:
-  GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx2560m -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.daemon.jvm.options="-Xmx2560m" -Dkotlin.incremental=false
+  GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx3072m -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.daemon.jvm.options="-Xmx2560m" -Dkotlin.incremental=false
   CI_GRADLE_ARG_PROPERTIES: --stacktrace -PpreDexEnable=false --max-workers 2 --no-daemon
 
 jobs:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,10 +7,8 @@ on:
 
 # Enrich gradle.properties for CI/CD
 env:
-  CI_GRADLE_ARG_PROPERTIES: >
-    -Porg.gradle.jvmargs=-Xmx4g
-    -Porg.gradle.parallel=false
-    --no-daemon
+  GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx2560m -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.daemon.jvm.options="-Xmx2560m" -Dkotlin.incremental=false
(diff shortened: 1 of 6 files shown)


.github/workflows/nightly.yml
```diff
@@ -6,10 +6,8 @@ on:
     - cron: "0 4 * * *"
 
 env:
-  CI_GRADLE_ARG_PROPERTIES: >
-    -Porg.gradle.jvmargs=-Xmx4g
-    -Porg.gradle.parallel=false
-    --no-daemon
+  GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx3072m -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.daemon.jvm.options="-Xmx2560m" -Dkotlin.incremental=false
+  CI_GRADLE_ARG_PROPERTIES: --stacktrace -PpreDexEnable=false --max-workers 2 --no-daemon
```
mnaturel, line 10: I think we can remove the `--stacktrace` option on the line we call `CI_GRADLE_ARG_PROPERTIES` like you did on other config files. The `nightly_build.md` doc could be updated as well.

mnaturel:
A general question: how many times all jobs have passed? Maybe we should try at least another time to confirm?

ouchadam:
> A general question: how many times all jobs have passed? Maybe we should try at least another time to confirm?

2/3 times on this PR and a handful of times on https://github.com/vector-im/element-android/pull/6720

.github/workflows/build.yml
```diff
@@ -7,10 +7,8 @@ on:
 
 # Enrich gradle.properties for CI/CD
 env:
-  CI_GRADLE_ARG_PROPERTIES: >
-    -Porg.gradle.jvmargs=-Xmx4g
-    -Porg.gradle.parallel=false
-    --no-daemon
+  GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx3072m -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.daemon.jvm.options="-Xmx2560m" -Dkotlin.incremental=false
+  CI_GRADLE_ARG_PROPERTIES: --stacktrace -PpreDexEnable=false --max-workers 2 --no-daemon
```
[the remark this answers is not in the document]
ouchadam: I haven't tried, I've had lots of issues with CI and shared daemon builds in the past (mainly circleci / jenkins) 

didn't want to rock the boat tooo much by removing it yet :sweat_smile:

.github/workflows/nightly.yml
ouchadam: great catch, will do! :+1:

ouchadam: https://github.com/vector-im/element-android/pull/6728/commits/8470ac9731c18bfd12945d3afbc2722ab61d3b44

.github/workflows/build.yml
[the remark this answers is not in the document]
mnaturel: Okay I see, no problem.

PR #6736, opened 2022-08-04
Fixes crash when entering non ascii username during account creation

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Handles the exception throwing of the username availability check 

## Motivation and context

To avoid a crash when non ascii characters are used in the username during account creation

## Screenshots / GIFs

|Before|After|
|-|-|
|before-non-ascii-account|after-ascii|

## Tests

- In the account creation screen
- Enter non ascii characters in the username field

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
@@ -193,25 +193,32 @@ class OnboardingViewModel @AssistedInject constructor(
     }
 
     private suspend fun checkUserNameAvailability(userName: String) {
-        when (val result = registrationWizard.registrationAvailable(userName)) {
-            RegistrationAvailability.Available -> {
-                setState {
-                    copy(
-                            registrationState = RegistrationState(
-                                    isUserNameAvailable = true,
-                                    selectedMatrixId = when {
-                                        userName.isMatrixId() -> userName
-                                        else -> "@$userName:${selectedHomeserver.userFacingUrl.toReducedUrl()}"
-                                    },
-                            )
-                    )
-                }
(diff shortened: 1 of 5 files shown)


vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
```diff
@@ -193,25 +193,32 @@ class OnboardingViewModel @AssistedInject constructor(
...
-                                    isUserNameAvailable = true,
-                                    selectedMatrixId = when {
-                                        userName.isMatrixId() -> userName
-                                        else -> "@$userName:${selectedHomeserver.userFacingUrl.toReducedUrl()}"
-                                    },
-                            )
-                    )
-                }
-            }
+        runCatching { registrationWizard.registrationAvailable(userName) }.fold(
```
ouchadam, line 196: adding the `runCatching` is the fix

jmartinesp (changes requested):
There is one weird behaviour when trying to go back:

[emoji_back.webm](https://user-images.githubusercontent.com/480955/183628156-f2af103e-66d6-4c37-bf2f-4abe77dc33ac.webm)

It seems like `checkUserNameAvailability` is getting called on that case too.

jmartinesp (approved):
It seems like the issue above was fixed in #6741. After rebasing develop and testing again, it's working fine.

ouchadam:
(We've spoke in the room) The backflow is handled by https://github.com/vector-im/element-android/pull/6741 this PR is slightly behind develop

PR #6745, opened 2022-08-05
Why two tabs?

hawk-m42:


gregmiret:
Do you mean why separating direct messages and rooms ?
I also see it as a drawback and think it should be at least an option.

bmarty:
> Do you mean why separating direct messages and rooms ?

@hawk-m42 is talking about an indentation issue (formatting).

On my side Android Studio does not reformat those lines, so I prefer to keep the code like that for now.

PR #6748, opened 2022-08-05
Fixing flaky `CantVerifyTest`

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Adds a test rule to reset an active session in order to force the app to go through the authentication flow

## Motivation and context

- Fixes the `CantVerifyTest` failing if the test is executed after a session has been cached. The test can sometimes be triggered with a cached `Session` which means the app is already logged in. 

## Screenshots / GIFs
No production changes

## Tests

- Run the `CantVerifyTest` multiple times without clearing data between tests

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 31

### DIFF
--- a/vector/src/androidTest/java/im/vector/app/ClearCurrentSessionRule.kt
+++ b/vector/src/androidTest/java/im/vector/app/ClearCurrentSessionRule.kt
@@ -16,25 +16,39 @@
 
 package im.vector.app
 
-import androidx.datastore.preferences.preferencesDataStoreFile
+import android.content.Context
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.edit
 import androidx.test.platform.app.InstrumentationRegistry
+import im.vector.app.features.analytics.store.AnalyticsStore
 import kotlinx.coroutines.runBlocking
 import org.junit.rules.TestWatcher
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
+import kotlin.reflect.KClass
 
 class ClearCurrentSessionRule : TestWatcher() {
     override fun apply(base: Statement, description: Description): Statement {
         val context = InstrumentationRegistry.getInstrumentation().targetContext
         runBlocking {
+            reflectAnalyticDatastore(context).edit { it.clear() }
             runCatching {
                 val holder = (context.applicationContext as VectorApplication).activeSessionHolder
                 holder.getSafeActiveSession()?.signOutService()?.signOut(true)
-                context.preferencesDataStoreFile(name = "vector_analytics").delete()
                 (context.applicationContext as VectorApplication).vectorPreferences.clearPreferences()
                 holder.clearActiveSession()
             }
         }
         return super.apply(base, description)
     }
+}
+
+private fun KClass<*>.asTopLevel() = Class.forName("${qualifiedName}Kt")
+
+@Suppress("UNCHECKED_CAST")
+private fun reflectAnalyticDatastore(context: Context): DataStore<Preferences> {
+    val klass = AnalyticsStore::class.asTopLevel()
+    val method = klass.getMethod("access\$getDataStore", Context::class.java)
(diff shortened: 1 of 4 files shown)


vector/src/androidTest/java/im/vector/app/ClearCurrentSessionRule.kt
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+        return super.apply(base, description)
+    }
+}
+
+private fun KClass<*>.asTopLevel() = Class.forName("${qualifiedName}Kt")
+
+@Suppress("UNCHECKED_CAST")
+private fun reflectAnalyticDatastore(context: Context): DataStore<Preferences> {
```
ouchadam, line 50: uses reflection to avoid making the datastore within the `AnalyticsDatastore` public

vector/src/androidTest/java/im/vector/app/ClearCurrentSessionRule.kt
```diff
@@ -0,0 +1,54 @@
...
+
+class ClearCurrentSessionRule : TestWatcher() {
+    override fun apply(base: Statement, description: Description): Statement {
+        val context = InstrumentationRegistry.getInstrumentation().targetContext
+        runBlocking {
+            reflectAnalyticDatastore(context).edit { it.clear() }
+            runCatching {
+                val holder = (context.applicationContext as VectorApplication).activeSessionHolder
+                holder.getSafeActiveSession()?.signOutService()?.signOut(true)
+                (context.applicationContext as VectorApplication).vectorPreferences.clearPreferences()
```
fedrunov, line 39: `VectorPreferences.clearPreferences`  doesn't do a full clean up. There are plenty of things which are kept after logout and could affect tests actually. Is it ok?

vector/src/androidTest/java/im/vector/app/ClearCurrentSessionRule.kt
fedrunov, line 50: A small typo here, should be `AnalyticS`

fedrunov: I think it's worth to put this comment to the code. And, maybe, even put comment to the accessed method of the `AnalyticsDatastore`

vector/src/androidTest/java/im/vector/app/ClearCurrentSessionRule.kt
```diff
@@ -0,0 +1,54 @@
...
+import androidx.datastore.preferences.core.edit
+import androidx.test.platform.app.InstrumentationRegistry
+import im.vector.app.features.analytics.store.AnalyticsStore
+import kotlinx.coroutines.runBlocking
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+import kotlin.reflect.KClass
+
+class ClearCurrentSessionRule : TestWatcher() {
```
fedrunov, line 31: Could you please put some KDoc to clarify purpose of the class?

fedrunov (changes requested):

vector/src/androidTest/java/im/vector/app/ClearCurrentSessionRule.kt
ouchadam: I tried deleting via the `DefaultSharedPrefs` and it actually caused more trouble as it resets other legacy flags like the migration from Riot to Element (which causes a popup to be shown)

vector/src/androidTest/java/im/vector/app/ClearCurrentSessionRule.kt
ouchadam: updated! https://github.com/vector-im/element-android/pull/6748/commits/20b3dbc6e6e239104fd5bf5931f4ee663155f776 added doc to the rule, reflection helper and the extension property

fedrunov (approved):

PR #6774, opened 2022-08-08
Dedicated module for `strings`

### LINKED ISSUE #3955: Create a dedicated module for translations to be able to use strings from sub modules
Weblate configuration will have to be updated.

It will also probably break the forks which have changes in the translation, so please also add a documentation to help migrating the forks.

It could also be good for the iterative compilation time.

Be careful with this:
> apply plugin: 'placeholder-resolver'

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Creates a dedicated module `library/ui-strings` for holding all `strings` used by the project. 

## Motivation and context

Fixes #3955 - Sub module for strings

- This will allow other modules to include the `strings` without all of the kotlin code from `:vector`

## Screenshots / GIFs

No UI changes

## Tests

- Smoke test around the app to check for `$app_name`, such as the `Settings -> Security` page

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):


vector/build.gradle
```diff
@@ -139,6 +134,7 @@ dependencies {
     api project(":matrix-sdk-android")
     implementation project(":matrix-sdk-android-flow")
     implementation project(":library:jsonviewer")
+    implementation project(":library:strings")
```
bmarty, line 137: Maybe rename the module to `ui-strings` as per `ui-styles`. I am not sure from where this convention comes from. Any idea @ganfra ?

library/strings/build.gradle
```diff
@@ -0,0 +1,22 @@
+plugins {
+    id 'com.android.library'
+    id 'kotlin-android'
+    id 'com.likethesalad.stem-library'
```
bmarty, line 4: This is what was blocking from doing that work before. Are the string with template (such as ${app_name}) still working as expected?

bmarty: (OK covered by your test, sorry)

bmarty:
As a reminder, Weblate configuration will have to be updated too, once this is merged. (See #3955)

vector/build.gradle
ouchadam: will rename :+1:

bmarty:
@ouchadam I think the PR is the next one in the modularisation plan.

This PR could be merged, but since it requires some change on Weblate config, maybe better to have a plan to avoid mistakes.

I propose this steps:
- [ ] a Weblate sync (push then when the created PR is merged pull, but do not unlock)
- [ ] fix the conflict on this PR
- [ ] merge this PR
- [ ] update the Weblate config
- [ ] unlock Weblate

WDYT?

ouchadam:
sounds great to me! for this PR I mainly delete the branch and cherry pick the file move commit so the conflicts are easy to address

PR #6775, opened 2022-08-08
Feature/fga/realm kotlin crypto

ganfra:
This PR handles the crypto store migration to realm kotlin.
It also adds some handy methods to the RealmInstance and rework a bit coroutine scopes.

Not sure if we should merge in develop or wait for the whole codebase to be migrated to realm-kotlin.
Currently we are missing session db some features not available like flipper integration.


bmarty:
(for the TODO issue reported by Detekt, please just replace  `TODO:` by `TODO`)

PR #6834, opened 2022-08-15
Allow empty SSO `identity_providers`

### LINKED ISSUE #6827: SSO Login has disappeared from the app, both Android and iOS
### Steps to reproduce

1. I load the app and give our company Matrix URL
2. There used to be a "login with SSO" in the app. In the browser it is still there, but the app doesn't have it.
3. As a result, I can no longer login to Matrix/Element

#### What did you expect?

A button "Login with SSO".

#### What happened instead?

The button is not there. There's only the option to login with username and password, but that doesn't work for us.

### Your phone model

Any Android phone

### Operating system version

Any android OS

1.4.32

### Homeserver

Synapse 1.4.0

### Will you send logs?

No

No

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Treats null `identity_providers` as a valid  SSO case
- Models the `fallback` through the `ViewModel/Auth` chain to make the behaviour more obvious (and adds tests to the creation)

## Motivation and context

Fixes #6827 

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot_20220815_093113|Screenshot_20220815_125843|
 
## Tests

- Select a homeserver which provides SSO via the fallback mechanism (eg without using `identity_providers`) 
- Attempt to register/sign in

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 31

### DIFF
--- a/changelog.d/6827.bugfix
+++ b/changelog.d/6827.bugfix
@@ -1,0 +1,1 @@
+Fixing sign in/up for homeservers that rely on the SSO fallback url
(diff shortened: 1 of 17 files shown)


fedrunov (approved):
LGTM

PR #6837, opened 2022-08-15
Missing notifications due to stuck background workers

### LINKED ISSUE #6836: `SessionSafeCoroutineWorker` becomes stuck until a cache clear
### Steps to reproduce

1. Pragmatically introduce an uncaught error in the `doSafeWork` within a child of SessionSafeCoroutineWorker
2. Trigger the error
3. Remove the error and reinstall the app
4. Notice all future workers use the error flow instead of attempting to execute the work

#### What did you expect?
The work to trigger 

#### What happened instead?
It becomes stuck in the failure state until a cache clear (deleting the work manager database) 

### Your phone model

Emulator

### Operating system version

Android 12

1.4.32

### Homeserver

_No response_

### Will you send logs?

No

Yes

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Consumes the `lastFailureMessage` when handling the `SessionSafeCoroutineWorker.doWork()`, by not consuming, the worker is stuck in the failure state until the `WorkManager` database is cleared with a cache clean.

## Motivation and context

Fixes #6836 - implementations of `SessionSafeCoroutineWorker` suffer from uncaught exceptions causing the worker to become stuck in the failure state. 

We update the work params/input data on error but never reset the state back to the default arguments 

## Screenshots / GIFs
No UI changes

## Tests

- Add a forced uncaught crash to the `SyncWorker` 
- Install the app and attempt to receive messages whilst the app is in the background
- Remove the forced crash and attempt receive messages whilst the app is in the background
- Notice the background jobs are still failing with the original error
- `Settings -> General -> Clear cache` fixes the stuck state 

```
MatrixWorkerFactory.createWorker for org.matrix.android.sdk.internal.session.sync.job.SyncWorker
SyncWorker Params(sessionId=8742ee0c2fc165532212ce2480e7bbca, timeout=0, delay=0, periodic=false, forceImmediate=false, lastFailureMessage=java.lang.RuntimeException: manual throw)
Work cancelled due to input error from parent: java.lang.RuntimeException: manual throw
```

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 31


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionSafeCoroutineWorker.kt
```diff
@@ -89,10 +88,10 @@ internal abstract class SessionSafeCoroutineWorker<PARAM : SessionWorkerParams>(
      * This is called when the input parameters are correct, but contain an error from the previous worker.
      */
     @CallSuper
-    open fun doOnError(params: PARAM): Result {
+    open fun doOnError(params: PARAM, failureMessage: String): Result {
         // Forward the error
-        return Result.success(inputData)
```
ouchadam, line 94: previously we were always using the `inputData` which contained the `lastFailureMessage` which mean the next time the work ran the data would be restore with the pre-existing failure, causing the worker to never execute `doSafeWork`

the fix is to _consume_ the message and use the updated state for the result (so that it can be persisted without the `lastFailureMessage`)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionSafeCoroutineWorker.kt
```diff
@@ -55,14 +55,13 @@ internal abstract class SessionSafeCoroutineWorker<PARAM : SessionWorkerParams>(
...
-            } else {
-                doSafeWork(params)
+
+            when (val failureMessage = params.consumeFailure()) {
+                null -> doSafeWork(params)
+                else -> doOnError(params, failureMessage)
             }
         } catch (throwable: Throwable) {
-            buildErrorResult(params, throwable.localizedMessage ?: "error")
+            buildErrorResult(params, "${throwable::class.java.name}: ${throwable.localizedMessage ?: "N/A error message"}")
```
ouchadam, line 64: added extra details about the error to help debug in the future

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionSafeCoroutineWorker.kt
```diff
@@ -89,10 +88,10 @@ internal abstract class SessionSafeCoroutineWorker<PARAM : SessionWorkerParams>(
      * This is called when the input parameters are correct, but contain an error from the previous worker.
      */
     @CallSuper
-    open fun doOnError(params: PARAM): Result {
+    open fun doOnError(params: PARAM, failureMessage: String): Result {
         // Forward the error
-        return Result.success(inputData)
-                .also { Timber.e("Work cancelled due to input error from parent") }
+        return Result.success(WorkerParamsFactory.toData(paramClass, params))
+                .also { Timber.e("Work cancelled due to input error from parent: $failureMessage") }
```
ouchadam, line 94: includes the failure message in the error log

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionSafeCoroutineWorker.kt
bmarty: OK, so we do not forward the error anymore to the next worker, since it has been removed from the `params` parameter, right? (maybe update the comment above). The original idea to transmit the error was to ensure that the last worker of the chain will handle it, without breaking the whole worker chain forever.

It's still strange to me that the `lastFailureMessage` get restored when we start a new work.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionWorkerParams.kt
```diff
@@ -27,5 +27,9 @@ internal interface SessionWorkerParams {
      * Null when no error occurs. When chaining Workers, first step is to check that there is no lastFailureMessage from the previous workers
      * If it is the case, the worker should just transmit the error and shouldn't do anything else.
      */
-    val lastFailureMessage: String?
+    var lastFailureMessage: String?
+}
+
+internal fun SessionWorkerParams.consumeFailure(): String? {
+    return lastFailureMessage.also { lastFailureMessage = null }
```
bmarty, line 34: Maybe we could avoid making `lastFailureMessage` a `var` by using the `data.copy` fun. It could be painful since we have an interface here, so every impl would have to do the copy.

ouchadam: I was trying to avoid that as we have quite a few worker params, happy to make the change if preferred!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionSafeCoroutineWorker.kt
ouchadam: I was also surprised by this, I think it's caused by the `lastFailureMessage` causing a merge conflict in the input data

```kotlin
// requireBackgroundSync
val data = WorkerParamsFactory.toData(
        Params(
                sessionId = sessionId,
                timeout = serverTimeoutInSeconds,
                delay = 0L,
                periodic = false,
                random = UUID.randomUUID().toString()  // create a unique id for each work request
        )
)
```

Logging the work request `data` and the `inputData` on job execution 

```
// No forced crash - with random

requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"7675"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"7675"}, }
requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"107a"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"107a"}, }
requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"c502"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"c502"}, }
requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"a22b"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"a22b"}, }


// Force crash

requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"49b1"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"49b1"}, }
requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"fca1"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"49b1","lastFailureMessage":"error"}, }
requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"b7e6"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"49b1","lastFailureMessage":"error"}, }


// Force crash removed

requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"97c9"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"49b1","lastFailureMessage":"error"}, }
requireBackground sync params: Data {WORKER_PARAMS_JSON : {"random":"a2b6deb5-a33b-4361-ae3d-3dcb22913e36"}, }
doWork input: Data {WORKER_PARAMS_JSON : {"random":"49b1","lastFailureMessage":"error"}, }
```

notice that as soon as the `lastFailureMessage` is modified the request data becomes ignored, which I assume is caused by the _unspecified ordering_ of the default `OverwritingInputMerger` https://developer.android.com/reference/androidx/work/InputMerger

bmarty: > another solution is to set an InputMerger which always favours the latest input specifically for the SyncWorker

This should be far less confusing IMHO. The API of the work manager is so strange...

bmarty: (thanks for the explanations!)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionWorkerParams.kt
bmarty: Let's keep the `var` like this for now. Thanks!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/worker/SessionSafeCoroutineWorker.kt
ouchadam: comment updated 1a21ec5f36b333ab4a02d6ed04ed49bafc931b3d

ganfra: So, do we want to use an `InputMerger` instead of consuming the error?

ouchadam: I would lean towards consuming the error as having the workers stuck in the failure state sounds unintentional. Most of the workers try/catch their logic in `doSafeWork`, it only seems to be the uncaught exceptions that fall into the chain (stuck error) flow.

We have an instance of providing a merger which avoids merging for the [TimelineSendEventWorkCommon](https://github.com/vector-im/element-android/blob/develop/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/timeline/TimelineSendEventWorkCommon.kt#L51), although the name `startChain` is a little misleading as the `NoMerger` does the opposite and disables the result chaining.

ganfra: Yes if there are chained workers we need to cancel using this mechanism, otherwise workers will be stuck if we return Failed state. This worker API is so bad...

ganfra: I think it will break some flow if we consume the failure, as we always return Success, if there is a chain of workers, the next one will be running when it shouldn't...

ouchadam: in those cases, should we be using `Result.Failure` to cancel the chain? My understanding is that `APPEND_OR_REPLACE` would allow the chain to be recreated the next time a request comes in

ouchadam: I'll rework the PR to make use of the `chain` extension (`NoMerge`), we can think about the worker flows as a separate ticket (as other workers may be getting stuck)

ouchadam: updated 1fd1a4e82436f502b33bda2d3be8f642898689ce

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/sync/job/SyncWorker.kt
```diff
@@ -136,6 +137,7 @@ internal class SyncWorker(context: Context, workerParameters: WorkerParameters,
                     .setConstraints(WorkManagerProvider.workConstraints)
                     .setBackoffCriteria(BackoffPolicy.LINEAR, WorkManagerProvider.BACKOFF_DELAY_MILLIS, TimeUnit.MILLISECONDS)
                     .setInputData(data)
+                    .startChain(true)
```
ouchadam, line 140: this line is the fix, matches the multiple event worker by ignoring previous results when appending new workers

bmarty (approved):
LGTM, thanks for the update. I am more comfortable with this version of the fix.

PR #6840, opened 2022-08-16
added dialog to change app layout settings

### LINKED ISSUE #6506: App Layout: layout settings screen
Screenshot 2022-08-01 at 15 13 39
 
FIgma: https://www.figma.com/file/Hw4gjP3pknMZUir8jQqmWH/%5BMobile%5D-IA-early-proposals?node-id=3272%3A159882

Acceptance criteria

- [x] Bottom sheet dialog
- [x] Access from Home Screen context menu
- [x] Settings are stored in shared preferences
- [x] Enable recents
- [x] Enable filter tabs
- [ ] Allow different sorting for filtered room list

fedrunov:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

adds a bottom sheet dialog with app layout settings 

## Motivation and context

part of #6502 
closes #6506 

## Screenshots / GIFs

|Light|Dark|Black|
|-|-|-|
|light|dark|black|

## Tests

- open home
- open context menu (3 dots in top right corner)
- select "layout preferences"
- change preferences

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 11


changelog.d/6506.wip
```diff
@@ -0,0 +1 @@
+[App Layout] added dialog to configure app layout
```
ericdecanini, line 1: Nice idea using `[App Layout]`. I'll start doing this too

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeLayoutPreferences.kt
```diff
@@ -0,0 +1,107 @@
...
+
+class HomeLayoutPreferences @Inject constructor(
+        @DefaultPreferences private val preferences: SharedPreferences
+) {
+
+    companion object {
+        const val SETTINGS_PREFERENCES_HOME_RECENTS = "SETTINGS_PREFERENCES_HOME_RECENTS"
+        const val SETTINGS_PREFERENCES_HOME_FILTERS = "SETTINGS_PREFERENCES_HOME_FILTERS"
+        const val SETTINGS_PREFERENCES_USE_AZ_ORDER = "SETTINGS_PREFERENCES_USE_AZ_ORDER"
+    }
```
ericdecanini, line 32: Optional: Kotlin code style guidelines suggest that companion objects should be at the bottom of the class rather than the top

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeLayoutPreferences.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    fun setRecentsEnabled(isEnabled: Boolean) {
+        preferences.edit {
+            putBoolean(SETTINGS_PREFERENCES_HOME_RECENTS, isEnabled)
+        }
+    }
+
+    fun areFiltersEnabled(): Boolean {
+        return preferences.getBoolean(SETTINGS_PREFERENCES_HOME_FILTERS, false)
+    }
```
ericdecanini, line 51: Optional: Many of these functions are single liners, maybe it's better to use this syntax:
```suggestion
    fun areFiltersEnabled() = preferences.getBoolean(SETTINGS_PREFERENCES_HOME_FILTERS, false)
```

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListViewModel.kt
```diff
@@ -168,26 +202,32 @@ class HomeRoomListViewModel @AssistedInject constructor(
...
-                )
-            }
-
-            flow.emit(filtersData)
-        }.launchIn(viewModelScope)
+        favouritesFlow
+                .combine(dmsFLow) { hasFavourite, hasDm ->
+                    hasFavourite to hasDm
+                }.combine(filtersPreferencesFlow) { (hasFavourite, hasDm), areFiltersEnabled ->
+                    Triple(hasFavourite, hasDm, areFiltersEnabled)
```
ericdecanini, line 209: You can use `combine` without any prefix to combine three flows in one go
https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/combine.html

```suggestion
        combine(favouritesFlow, dmsFlow, filteresPreferencesFlow) { hasFavourite, hasDm, areFiltersEnabled ->
            Triple(hasFavourite, hasDm, areFiltersEnabled)
```

Also note capitalisation on dmsFLow

ericdecanini (approved):
Implementation looks good, but there are some code improvements I would strongly suggest

vector/src/main/java/im/vector/app/features/home/room/list/home/layout/HomeLayoutSettingBottomDialogFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+        views.homeLayoutSettingsFilters.isChecked = preferences.areFiltersEnabled()
+
+        if (preferences.isAZOrderingEnabled()) {
+            views.homeLayoutSettingsSortName.isChecked = true
+        } else {
+            views.homeLayoutSettingsSortActivity.isChecked = true
+        }
+
+        views.homeLayoutSettingsRecents.setOnCheckedChangeListener { _, isChecked ->
+            preferences.setRecentsEnabled(isChecked)
```
onurays, line 52: I think we shouldn't change preferences before the user clicks to "Done" button. What do you think?

vector/src/main/java/im/vector/app/features/home/room/list/home/filter/HomeFilteredRoomsController.kt
```diff
@@ -27,8 +27,7 @@ import org.matrix.android.sdk.api.session.room.members.ChangeMembershipState
 import org.matrix.android.sdk.api.session.room.model.RoomSummary
 
 class HomeFilteredRoomsController(
-        private val roomSummaryItemFactory: RoomSummaryItemFactory,
-        private val showFilters: Boolean,
+        private val roomSummaryItemFactory: RoomSummaryItemFactory
```
onurays, line 30: Please add comma (,) at the end.

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomSection.kt
```diff
@@ -21,12 +21,12 @@ import androidx.paging.PagedList
...
 import kotlinx.coroutines.flow.SharedFlow
 import org.matrix.android.sdk.api.session.room.model.RoomSummary
+import org.matrix.android.sdk.api.util.Optional
 
 sealed class HomeRoomSection {
     data class RoomSummaryData(
             val list: LiveData<PagedList<RoomSummary>>,
-            val showFilters: Boolean,
-            val filtersData: SharedFlow<List<HomeRoomFilter>>
+            val filtersData: SharedFlow<Optional<List<HomeRoomFilter>>>
```
onurays, line 29: Please add comma (,) at the end.

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListViewModel.kt
```diff
@@ -78,16 +80,45 @@ class HomeRoomListViewModel @AssistedInject constructor(
...
 
     init {
         configureSections()
+        observePreferences()
+    }
+
+    private fun observePreferences() {
+        preferences.registerFiltersListener { areFiltersEnabled ->
+            viewModelScope.launch {
+                filtersPreferencesFlow.emit(areFiltersEnabled)
```
onurays, line 95: Is this triggering immediately right after you register the listener? If so, we don't need the below lines.

onurays (commented):
Nice work! Can you please check my small remarks?

vector/src/main/java/im/vector/app/features/home/room/list/home/layout/HomeLayoutSettingBottomDialogFragment.kt
fedrunov: @amshakal could you please verify?

amshakal: Yes, I believe no changes should be made before the user hits 'DONE'

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListViewModel.kt
fedrunov: For some reason it doesn't. If I remove that emit, it will not show filters until you change setting

amshakal (approved):

vector/src/main/res/layout/bottom_sheet_home_layout_settings.xml
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/rootLayout"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:background="?colorSurface"
```
ElementBot, line 6: - :warning: Possible overdraw: Root element paints background `?colorSurface` with a theme that also paints a background (inferred theme is `@style/Theme.Vector.Light`)
- :warning: Possible overdraw: Root element paints background `?colorSurface` with a theme that also paints a background (inferred theme is `@style/Theme.Vector.Light`)

onurays (approved):
Thanks for the update!

vector/src/main/java/im/vector/app/features/home/room/list/home/layout/HomeLayoutSettingBottomDialogFragment.kt
bmarty: This is strange to me to have a "DONE" button on a BottomSheet. I would prefer to have the change effect live and visible to the background. For instance when disabling "Show recents", the user will have a direct feedback by seeing the breadcrumbs row vanish in the background. They can even play with the toggle several times. Since it's easy to revert the change, I would simply remove the DONE button. Else what would be done if the user dismiss the bottom sheet without clicking on DONE? Cancel the pending change? Not a good UX to me.

On Android settings and in app settings (preference screen), there is no DONE button, any changes are immediately taking effect. Better to do the same here IMHO.

CC @amshakal

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeLayoutPreferences.kt
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+package im.vector.app.features.home.room.list.home
+
+import android.content.SharedPreferences
+import androidx.core.content.edit
+import im.vector.app.core.di.DefaultPreferences
+import javax.inject.Inject
+
+class HomeLayoutPreferences @Inject constructor(
```
bmarty, line 24: I think it would be easier to use a `DataStore` rather than `SharedPreferences`. See `VectorSessionStore` or `AnalyticsStore` for some example.

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListViewModel.kt
bmarty: -> `DataStore` :)

vector/src/main/java/im/vector/app/features/home/room/list/home/layout/HomeLayoutSettingBottomDialogFragment.kt
```diff
@@ -0,0 +1,58 @@
...
+import im.vector.app.R
+import im.vector.app.core.platform.VectorBaseBottomSheetDialogFragment
+import im.vector.app.databinding.BottomSheetHomeLayoutSettingsBinding
+import im.vector.app.features.home.room.list.home.HomeLayoutPreferences
+import javax.inject.Inject
+
+@AndroidEntryPoint
+class HomeLayoutSettingBottomDialogFragment : VectorBaseBottomSheetDialogFragment<BottomSheetHomeLayoutSettingsBinding>() {
+
+    @Inject lateinit var preferences: HomeLayoutPreferences
```
bmarty, line 33: This should be moved to a `ViewModel`, and the UI should be updated in the `invalidate` method.

bmarty (commented):
I have made some more remarks, on the UX and on the impl.

vector/src/main/java/im/vector/app/features/home/room/list/home/layout/HomeLayoutSettingBottomDialogFragment.kt
fedrunov: Not sure we'll benefit from VM here. This dialog is too small and too specific to make it's part reusable and too simple to require separation of concerns. Having VM here will just increase amount of boilerplate and make it less easy to understand. We also have plenty of other dialogs without VMs and I did think it's because of reasons above. Did we change approach and now require to follow same pattern no matter of what?

bmarty: I agree it's a bit more boilerplate code, but the idea is to have the logic in the ViewModel also to unit test it. Views (Fragment, etc.) should only be responsible of displaying things.
I guess for this particular part it can stay like that. Maybe better to spend effort to use DataStore.

vector/src/main/java/im/vector/app/features/home/room/list/home/layout/HomeLayoutSettingBottomDialogFragment.kt
fedrunov: Changes are applied when user press DONE, so if user closes dialog in any other way this changes will just be lost. I don't have any strong opinion on this topic. I just felt that seeing changes in ui, while you switch options, feels more interactive, and since here it came with zero costs I decided to implement that way.

TLDR; Both options are good for me and there are no technical issues with both of them, so I let Amsha decide :)

amshakal: While we don't have done button in settings, we can be a bit more explicit with a bottom sheet. I couldn't find anything for or against it in the material guidelines.

I am open to having interactive options but since there is an overlay at the back anyway, the user won't really be able to see the changes properly. 

Let's start with this simple approach and we can iterate on it incase it feels like an odd pattern.

I had a chat with Nikita around this. Benoit, if this feels like a massive blocker/anti pattern. I am more than happy to remove the DONE button.

bmarty: I guess users are used to close bottom sheets either by swiping to the bottom or by clicking on the `back` button.

As a general guidelines, I think that Buttons on BottomSheet are not a good UX pattern. I do not see lots of Button in the example of BottomSheet [here](https://material.io/develop/android/components/bottom-sheet-dialog-fragment) and also in all the apps I use on my phone.

I would definitely remove the DONE button.

amshakal: Got it. In that case Nikita, let's go with benoit's suggestion of removing the CTA. Sorry about the back and forth

ericdecanini (approved):

vector/src/main/res/layout/bottom_sheet_home_layout_settings.xml
```diff
@@ -0,0 +1,67 @@
...
```
ElementBot, line 6: - :warning: Possible overdraw: Root element paints background `?colorSurface` with a theme that also paints a background (inferred theme is `@style/Theme.Vector.Light`)

bmarty (approved):
Thanks for the update!

PR #6857, opened 2022-08-17
Try to fix Danger job

bmarty:
Copied from the doc in this PR:

> PRs from forks do not always have access to the secret `secrets.DANGER_GITHUB_API_TOKEN`, so `secrets.GITHUB_TOKEN` is also provided to the job environment. If `secrets.DANGER_GITHUB_API_TOKEN` is available, it will be used, so user `ElementBot` will comment the PR. Else `secrets.GITHUB_TOKEN` will be used, and bot `github-actions` will comment the PR.

Doing some test, it seems fine for internal and external PR.

Let see with this solution how it will behave.

No changelog required, but happy to see that @ElementBot is commenting correctly this PR!

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -16,3 +16,5 @@ jobs:
           args: "--dangerfile tools/danger/dangerfile.js"
         env:
           DANGER_GITHUB_API_TOKEN: ${{ secrets.DANGER_GITHUB_API_TOKEN }}
+          # Fallback for forks
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


bmarty:
> No changelog required, but happy to see that @ElementBot is commenting correctly this PR!

Actually the comment have been deleted, I do not know why... :/

jmartinesp:
> > No changelog required, but happy to see that @ElementBot is commenting correctly this PR!
> 
> Actually the comment has been deleted, I do not know why... :/

The comment was not posted because of this error:

> Could not add a commit status, the GitHub token for Danger does not have access rights.

Re-running the job fixed it, but I'm not sure if it was just a temporary issue, some cached value, or there's an issue that sometimes makes it fail.

bmarty:
> Could not add a commit status, the GitHub token for Danger does not have access rights.

Is another error. Comment can be posted, even when this error is displayed. You can check it in the latest job execution, and the comment is back:

image

GitHub Actions are quite hard to configure, and as you said, consecutive runs are not always outputting the same result :/

jmartinesp (approved):
Since the issue doesn't look like a configuration issue, but rather either a GH or Danger bug we can probably merge this.

PR #6878, opened 2022-08-18
Add polling option to gplay variant

chagai95:
## Type of change
Add the polling option to gplay variant. 

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content
Added all the code specific to FDroid to the gplay variant and edited the strings to be more generic.

## Motivation and context
This is useful for people who might have problems with push notifications on their phone or in general to speed up the process of calls and messages if a user so wish.

## Tests

- I changed the push server URL to be invalid and then allowed polling and I got notifications not via push.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

Signed-off-by: Chagai Friedlander me@chagai.website

### DIFF
--- a/changelog.d/6878.feature
+++ b/changelog.d/6878.feature
@@ -1,0 +1,1 @@
+Add polling option to gplay variant.
--- a/vector/src/fdroid/java/im/vector/app/push/fcm/FdroidFcmHelper.kt
+++ b/vector/src/fdroid/java/im/vector/app/push/fcm/FdroidFcmHelper.kt
@@ -31,7 +31,6 @@ import javax.inject.Inject
  */
 class FdroidFcmHelper @Inject constructor(
         private val context: Context,
-        private val backgroundSyncStarter: BackgroundSyncStarter,
 ) : FcmHelper {
 
     override fun isFirebaseAvailable(): Boolean = false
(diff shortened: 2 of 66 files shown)


chagai95:
I am not sure how to change the strings that are translates so I just refactored them, if that's no good I can do something else.
For some reason after closing the app the default mode is set but that's not a big deal, would just be nice to fix that.

chagai95:
Not sure why FDroid is not building, I tried for a bit to find out what's wrong and I can try again soon hopefully.

chagai95:
Hey @bmarty , just wondering if this is something you're planning on considering...

bmarty:
Hello, sorry for the late reply.

Please do not update translation files in a PR. You can find how to manage string resources at this location: https://github.com/vector-im/element-android/blob/main/CONTRIBUTING.md#internationalisation

Regarding this PR content:
- during a call, I think we already keep the `SyncThread` alive, using `ActiveCallHandler`. So call signalling is fast and not relying on Push to work.

I do not think we want to have a long polling fallback on GPlay variant. We rely on the Push to work properly.

People who want to have long poll can install F-Droid version of the app (which now also support UnifiedPush, along with background sync)

chagai95:
Hey, thx for taking a look at this! 

The issue is not during a call but the fact that a call only arrives way too late, an instant message arriving too late is one thing but a call arriving too late is a whole other story....

I don't understand why you wouldn't want to have this as an opt in feature, you can hide it behind labs or whatever but calls are extremely difficult to use on many devices and messages are also arriving late. 

Most users don't know and won't install an F-Droid variant so I think that's not really a helpful solution...

Sorry if I seem strict about this but I've been using Element for 2 years now on many devices and I've seen many people use it and this just seems like something that would help a lot and it's a pity to see this dismissed like this...

PR #6884, opened 2022-08-19
Ensure sync thread is started

### LINKED ISSUE #6782: Messages are not arriving
### Steps to reproduce

1. Be actively chatting with someone in a DM
2. Sometimes see typing notifications for a long time, sometimes not see anything to indicate that the other person is sending a message
3. See the timeline not updating
4. Swipe the app up to close it and then open it again (see green loading screen)
5. Open DM and see messages that were sent minutes ago

#### What did you expect?
I can have a live conversation

#### What happened instead?
I had to keep closing/opening the app to see my messages

### Your phone model

_No response_

### Operating system version

Android 11

1.4.31 playstore

### Homeserver

matrix.org

### Will you send logs?

Yes

No

bmarty:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Fix #6782. This is this commit: 0b1b228. - **I am still investigating, because sometimes, the SyncThread stays in the  PAUSE state, which is not correct.**
- Improve SDK Debug API
- Log Thread before rageshake are sent.
- Rename thread names

## Motivation and context

- Fix #6782

Regression from #6710, release in Element and SDK v1.4.31 (2022-08-01)

## Screenshots / GIFs

NA

## Tests

- See the step to repro [here](https://github.com/vector-im/element-android/issues/6782#issuecomment-1219602578).

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/core/di/ActiveSessionHolder.kt
+++ b/vector/src/main/java/im/vector/app/core/di/ActiveSessionHolder.kt
@@ -20,6 +20,7 @@ import android.content.Context
 import arrow.core.Option
 import im.vector.app.ActiveSessionDataSource
 import im.vector.app.core.extensions.configureAndStart
+import im.vector.app.core.extensions.startSyncing
 import im.vector.app.core.pushers.UnifiedPushHelper
 import im.vector.app.core.services.GuardServiceStarter
 import im.vector.app.features.call.webrtc.WebRtcCallManager
@@ -100,10 +101,16 @@ class ActiveSessionHolder @Inject constructor(
     }
 
     suspend fun getOrInitializeSession(startSync: Boolean): Session? {
-        return activeSessionReference.get() ?: sessionInitializer.tryInitialize(readCurrentSession = { activeSessionReference.get() }) { session ->
-            setActiveSession(session)
-            session.configureAndStart(applicationContext, startSyncing = startSync)
-        }
+        return activeSessionReference.get()
+                ?.also {
+                    if (startSync && !it.syncService().isSyncThreadAlive()) {
+                        it.startSyncing(applicationContext)
+                    }
+                }
+                ?: sessionInitializer.tryInitialize(readCurrentSession = { activeSessionReference.get() }) { session ->
+                    setActiveSession(session)
+                    session.configureAndStart(applicationContext, startSyncing = startSync)
+                }
     }
 
     fun isWaitingForSessionInitialization() = activeSessionReference.get() == null && authenticationService.hasAuthenticatedSessions()
--- a/changelog.d/6884.bugfix
+++ b/changelog.d/6884.bugfix
@@ -1,0 +1,1 @@
+Ensure SyncThread is started when the app is launched after a Push has been received.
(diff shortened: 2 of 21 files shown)


vector/src/main/java/im/vector/app/core/di/ActiveSessionHolder.kt
```diff
@@ -100,10 +101,16 @@ class ActiveSessionHolder @Inject constructor(
     }
 
     suspend fun getOrInitializeSession(startSync: Boolean): Session? {
-        return activeSessionReference.get() ?: sessionInitializer.tryInitialize(readCurrentSession = { activeSessionReference.get() }) { session ->
-            setActiveSession(session)
-            session.configureAndStart(applicationContext, startSyncing = startSync)
-        }
+        return activeSessionReference.get()
```
ouchadam, line 104: this might be clashing with https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/VectorApplication.kt#L176

I've always wondered if we should stop syncing when entering the foreground :thinking:

opusforlife2:
> I am still investigating, because sometimes, the SyncThread stays in the PAUSE state, which is not correct.

I don't know about you guys, but I have occasionally encountered the Paused state on my device since forever. I don't remember if it started in a particular version. The solution has always been to remove the app from Recent Apps and then open it again, which causes the sync to resume.

bmarty:
> > I am still investigating, because sometimes, the SyncThread stays in the PAUSE state, which is not correct.
> 
> I don't know about you guys, but I have occasionally encountered the Paused state on my device since forever. I don't remember if it started in a particular version. The solution has always been to remove the app from Recent Apps and then open it again, which causes the sync to resume.

Thanks for the feedback.

On my side I did not repro in the past days.
We will investigate this issue separately.

Moving this PR to ready to review.

vector/src/main/java/im/vector/app/core/di/ActiveSessionHolder.kt
```diff
@@ -100,10 +101,16 @@ class ActiveSessionHolder @Inject constructor(
...
+                ?.also {
+                    if (startSync && !it.syncService().isSyncThreadAlive()) {
```
ouchadam, line 106: for my understanding of the error scenario

- app is not started
- receive a push
- session instance is created with `startSyncing = false` and cached
- next time the is launched the sync is not started because the session instance already exists and we only `configureAndStart` brand new instances


the fix is to always start the sync (if `startStart = true`) on our cached instances if they're not already syncing?

bmarty: Yes, that's it, sorry, the code is not crystal clear.

ouchadam (approved):

bmarty:
`checkCantVerifyPopup[test(AVD) - 9] FAILED` for some other reason.

opusforlife2:
> I don't know about you guys, but I have occasionally encountered the Paused state on my device since forever.

It's still a bit early to be sure, but I don't think I've encountered this after switching to ntfy for push notifications, from F-Droid's 60 second polling. It might have been specific to that particular config.

opusforlife2:
Yep. Too early. Got a Paused even with ntfy. I have no STR, sadly, otherwise I would have created a new issue.

PR #6888, opened 2022-08-19
FTUE - Allow changing server on onboarding start error

### LINKED ISSUE #6718: Users who have no access to matrix.org will NOT be able to  connect to other accessible HS.
### Steps to reproduce

1. Since the login and register flows have been change in v1.4.30, users now who don't have access to matrix.org (in case matrix.org is blocked due to a firewall but some others are not) will lose their access to other accessible HS because they will always get a message says that there was an error connecting to HS when trying to sign up (the 1st screenshot) and log in. (the 2nd screenshot)
register
login

2. But before v1.4.30, v1.4.28 for example, users will always be able to choose their preferred HS when signing up and logging in. (the 3rd screenshot, took in v1.4.28)
OK

3. So I think it will work perfectly that adding a button to both the register and login pages just like before (4th screenshot, button at bottom of the page) but keep the new flows, when they click that button, just redirect them to the new `Select your server` page (the last screenshot) without enforcing users to access matrix.org  and so that fix this problem.
image
image

#### What did you expect?
Being able to directly enter my HS address and receive no network IO exceptions when because of trying to access matrix.org.

#### What happened instead?
I have no way to directly enter my HS address and I *have to* connect to matrix.org, then as a result, running into network problems.

### Your phone model

Huawei Y9

### Operating system version

Android 9

Element v1.4.30

### Homeserver

matrix.org (problem HS), https://network.lan (my own HS in lan)

### Will you send logs?

Yes

Yes

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Checks for network errors on the initial homeserver selection (starting the login or account creation step) and forwards the user to Edit server screen instead to allow them to choose a different homeserver
- Removes the dedicated deeplink failure dialog in favour of reusing the edit server flow 

## Motivation and context

Fixes #6718 - Allowing the homeserver server to be changed on failure when starting the onboarding (matches iOS) 

## Screenshots / GIFs

|Before|After|
|-|-|
|before-edit-server-on-error|edit-server-on-error|
 

## Tests

- Block matrix.org or within the code change the default matrix.org address to an invalid url
- Attempt to login or create and account
- Notice the error dialog

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
@@ -701,9 +701,8 @@ class OnboardingViewModel @AssistedInject constructor(
 
     private fun onAuthenticationStartError(error: Throwable, trigger: OnboardingAction.HomeServerChange) {
         when {
-            error.isHomeserverUnavailable() && applicationContext.inferNoConnectivity(sdkIntProvider) -> _viewEvents.post(
-                    OnboardingViewEvents.Failure(error)
-            )
+            error.isHomeserverUnavailable() && applicationContext.inferNoConnectivity(sdkIntProvider) -> _viewEvents.post(OnboardingViewEvents.Failure(error))
+            isUnableToSelectServer(error, trigger) -> handle(OnboardingAction.PostViewEvent(OnboardingViewEvents.EditServerSelection))
             deeplinkUrlIsUnavailable(error, trigger) -> _viewEvents.post(
                     OnboardingViewEvents.DeeplinkAuthenticationFailure(
                             retryAction = (trigger as OnboardingAction.HomeServerChange.SelectHomeServer).resetToDefaultUrl()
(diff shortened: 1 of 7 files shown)


vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
```diff
@@ -715,26 +716,28 @@ class OnboardingViewModel @AssistedInject constructor(
...
-            deeplinkUrlIsUnavailable(error, trigger) -> _viewEvents.post(
-                    OnboardingViewEvents.DeeplinkAuthenticationFailure(
-                            retryAction = (trigger as OnboardingAction.HomeServerChange.SelectHomeServer).resetToDefaultUrl()
-                    )
-            )
+            error.isHomeserverUnavailable() && applicationContext.inferNoConnectivity(sdkIntProvider) -> _viewEvents.post(OnboardingViewEvents.Failure(error))
+            isUnableToSelectServer(error, trigger) -> {
+                withState { state ->
+                    when {
+                        canEditServerSelectionError(state) -> handle(OnboardingAction.PostViewEvent(OnboardingViewEvents.EditServerSelection))
```
ouchadam, line 723: when selecting a homeserver (starting the sign in or sign up flow) fails due to a network error we offer the user the ability to edit the initial homeserver url

this also means the deeplink error handling can be removed as we can reuse the edit url flow

bmarty (approved):
LGTM (I did not test the change)

iz4tow:
Same issue on Desktop application.

onurays (approved):
LGTM!

PR #6986, opened 2022-09-01
empty state for new invites screen

### LINKED ISSUE #6876: AppLayout: Invites Empty States
New invites screen also has empty state

it should be shown if invites was resolved while screen was loading (e.g. cancelled, or accepted from another session) or when user rejects last invite in the list. 

NOTE: when user accepts last invite in the list, invites activity should be finished before user is navigate to a chat screen. In this case we don't show empty state since user will be navigated back to home screen

[figma](https://www.figma.com/file/Hw4gjP3pknMZUir8jQqmWH/%5BMobile%5D-IA-early-proposals?node-id=3966%3A171689)

fedrunov:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

new invites screen should also present empty state

## Motivation and context

closes #6876
part of #6502

## Screenshots / GIFs

|Ligh|Dark|Black|
|-|-|-|
|light|dark|black|

output-optimised


vector/src/main/java/im/vector/app/features/home/room/list/home/invites/InvitesViewModel.kt
```diff
@@ -52,6 +65,11 @@ class InvitesViewModel @AssistedInject constructor(
 
     companion object : MavericksViewModelFactory<InvitesViewModel, InvitesViewState> by hiltMavericksViewModelFactory()
 
+    private val _invites = MutableSharedFlow<InvitesContentState>(replay = 1)
+    val invites = _invites.asSharedFlow()
+
+    var invitesCount = -1
```
ouchadam, line 71: I'm wondering if this count should be part of the state, what do you think?

vector/src/main/java/im/vector/app/features/home/room/list/home/invites/InvitesViewModel.kt
```diff
@@ -129,8 +142,26 @@ class InvitesViewModel @AssistedInject constructor(
                 sortOrder = RoomSortOrder.ACTIVITY
         )
 
-        setState {
-            copy(pagedList = pagedList)
-        }
+        pagedList.asFlow()
+                .map {
+                    if (it.isEmpty()) {
```
ouchadam, line 147: out of interest, is an empty result only returned after the `pagedList` has finished loading?

or is it possible for it to emit an empty state instantly?

vector/src/main/java/im/vector/app/features/home/room/list/home/invites/InvitesFragment.kt
```diff
@@ -62,13 +68,32 @@ class InvitesFragment : VectorBaseFragment<FragmentInvitesBinding>(), RoomListLi
...
-    }
 
-    private fun handleClose() {
-        requireActivity().finish()
+        viewModel.invites.onEach {
+            when (it) {
+                is InvitesContentState.Content -> {
+                    views.invitesStateView.state = StateView.State.Content
+                    Suppress("UNCHECKED_CAST")
+                    controller.submitList(it.content as? PagedList<RoomSummary>)
```
ouchadam, line 79: is this cast (and suppress) needed?

vector/src/main/java/im/vector/app/features/home/room/list/home/invites/InvitesViewModel.kt
fedrunov: this is not used by view, only inside ViewModel, so i decided to not put it to view state

bmarty: it should be declared `private` then.

vector/src/main/java/im/vector/app/features/home/room/list/home/invites/InvitesViewModel.kt
```diff
@@ -129,8 +142,26 @@ class InvitesViewModel @AssistedInject constructor(
...
+                        InvitesContentState.Empty(
+                                title = stringProvider.getString(R.string.invites_empty_title),
+                                image = drawableProvider.getDrawable(R.drawable.ic_invites_empty),
+                                message = stringProvider.getString(R.string.invites_empty_message)
+                        )
+                    } else {
+                        invitesCount = it.loadedCount
```
bmarty, line 154: This will work since we will test only if this is equal to 1, but I think it's more correct to use `it.size()`

vector/src/main/java/im/vector/app/features/home/room/list/home/invites/InvitesViewState.kt
```diff
@@ -16,13 +16,24 @@
...
 import com.airbnb.mvrx.MavericksState
 import org.matrix.android.sdk.api.session.room.members.ChangeMembershipState
 import org.matrix.android.sdk.api.session.room.model.RoomSummary
 
 data class InvitesViewState(
-        val pagedList: LiveData<PagedList<RoomSummary>>? = null,
         val roomMembershipChanges: Map<String, ChangeMembershipState> = emptyMap(),
 ) : MavericksState
+
+sealed class InvitesContentState {
```
bmarty, line 29: nit: it could be a `sealed interface`

vector/src/main/res/layout/fragment_invites.xml
```diff
@@ -20,17 +20,24 @@
...
-    <androidx.recyclerview.widget.RecyclerView
-        android:id="@+id/invites_recycler"
+    <im.vector.app.core.platform.StateView
+        android:id="@+id/invites_state_view"
         android:layout_width="0dp"
         android:layout_height="0dp"
-        android:fastScrollEnabled="true"
-        android:overScrollMode="always"
-        android:scrollbars="vertical"
-        app:layout_behavior="@string/appbar_scrolling_view_behavior"
```
bmarty, line 30: To double check, is it OK to remove this layout_behavior?

bmarty (commented):
LGTM, some minor remarks though.

vector/src/main/res/layout/fragment_invites.xml
fedrunov: yes, we don't need it for state view

vector/src/main/java/im/vector/app/features/home/room/list/home/invites/InvitesViewModel.kt
fedrunov: `it.size()` will also return placeholders (if there are any), while here we are interested in exact count without them

ericdecanini (approved):

vector/src/main/java/im/vector/app/features/home/room/list/home/invites/InvitesViewModel.kt
bmarty: from my understanding, the exact count is returned by `size()`. `loadedCount` does not include placeholders (= not loaded data, but which exist). If we have for instance 120 invites, there is maybe a risk that `loadedCount` return 0, or 1 (and so we have a bug), while `size()` returns 120, during the time the data are not yet loaded.
Anyway the risk is very low I think, and probably temporary, since more data will be loaded to be displayed.

bmarty:
Test `provideVoiceRecorderOnAndroidQReturnsQRecorder` will be fixed separately.

bmarty (approved):
LGTM

PR #6988, opened 2022-09-01
space switcher empty spaces

### LINKED ISSUE #6754: Space Switching: Space Bottom Sheet Empty States
When there are no spaces at the root or children of the current space, we want to show an empty state to notify the user. Designs of this are TBD.

As per https://github.com/vector-im/element-android/issues/6771, we also want to make the toolbar non-collapsible in an empty state.

Figma design: https://www.figma.com/file/Hw4gjP3pknMZUir8jQqmWH/%5BMobile%5D-IA-early-proposals?node-id=2523%3A135211

Parent epic: https://github.com/vector-im/element-android/issues/6605

fedrunov:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

empty state for new space switching dialog

## Motivation and context
 closes #6754

## Screenshots / GIFs

|Light|Dark|Black|
|-|-|-|
|light|dark|black|


vector/src/main/java/im/vector/app/features/spaces/SpaceListFragment.kt
```diff
@@ -71,6 +73,7 @@ class SpaceListFragment :
         homeActivitySharedActionViewModel = activityViewModelProvider[HomeSharedActionViewModel::class.java]
         roomListSharedActionViewModel = activityViewModelProvider[RoomListSharedActionViewModel::class.java]
         views.stateView.contentView = views.groupListView
+        views.spacesEmptyButton.onClick { onAddSpaceSelected() }
```
bmarty, line 76: This is maybe not new, but when the user clicks on this button, do we want to close the bottomsheet?

vector/src/main/java/im/vector/app/features/spaces/SpaceListFragment.kt
```diff
@@ -147,13 +150,22 @@ class SpaceListFragment :
...
+            is Success -> {
+                views.stateView.state = StateView.State.Content
+                if (spaces.invoke().isEmpty()) {
+                    views.spacesEmptyGroup.isVisible = true
+                    views.groupListView.isVisible = false
+                } else {
+                    views.spacesEmptyGroup.isVisible = false
+                    views.groupListView.isVisible = true
+                }
+            }
```
bmarty, line 168: About StateView, the correct code would be:

```kotlin
            is Success -> {
                if (spaces.invoke().isEmpty()) {
                    views.stateView.state = StateView.State.Empty(...)
                } else {
                    views.stateView.state = StateView.State.Content
                }
            }
```


But maybe there are specific needs here? If this is the `spaces_empty_button`, maybe we could improve the StateView to integrate it. (can be done later, to not block this.)

bmarty (commented):
Thanks, no blocking remarks here.

vector/src/main/java/im/vector/app/features/spaces/SpaceListFragment.kt
fedrunov: @ericdecanini could you please verify?

ericdecanini: Ideally we don't want to close it immediately in case the user navigates back but we want to close it following a successful space creation. This may already work as it is but it would be good to verify this

vector/src/main/java/im/vector/app/features/spaces/SpaceListFragment.kt
fedrunov: `StateView` has exact implementation for `Empty` state and I hadn't time to extend it to accept custom views (it rises some problems like callbacks for button). So in this case I had to place custom empty state near content view and switch between them. `StateView` her is mostly for loading state

bmarty: OK; thanks for explaining.

bmarty (approved):
Can be merged!

PR #7003, opened 2022-09-05
applayout release experience

### LINKED ISSUE #6909: App Layout: Release Experience
### Your use case

https://github.com/vector-im/design/issues/25

We need for current users who update the app to be shown an in-app release experience that tells them what's different in this version. It should be shown once and not to new users. 

_No response_

### Additional context

_No response_

No

fedrunov:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

we want to onboard users to new app layout

## Motivation and context

closes #6909

## Screenshots / GIFs

there are different margins for tall and short screens

|Tall|Short|
|-|-|
|tall|short|

|Light|Dark|Black|
|-|-|-|
|short|dark|black|


vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesFragment.kt
```diff
@@ -0,0 +1,126 @@
...
+        } else {
+            views.releaseNotesButtonNext.setText(R.string.action_next)
+        }
+    }
+
+    override fun onDestroyView() {
+        tabLayoutMediator?.detach()
+        tabLayoutMediator = null
+
+        views.releaseNotesCarousel.adapter = null
```
bmarty, line 123: You should also invoke `views.releaseNotesCarousel.unregisterOnPageChangeCallback`. Not sure they will be a memory leak, but let's not take the risk.

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesPreferencesStore.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    val appLayoutOnboardingShown: Flow<Boolean> = context.dataStore.data
+            .map { preferences -> preferences[isAppLayoutOnboardingShown].orFalse() }
+            .distinctUntilChanged()
+
+    suspend fun setAppLayoutOnboardingShown(isShown: Boolean) {
+        context.dataStore.edit { settings ->
+            settings[isAppLayoutOnboardingShown] = isShown
+        }
+    }
```
bmarty, line 47: Are you sure the Activity will not be show for user launching Element for the first time? This is a requirement of the issue. I do not see the logic here.

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesViewModel.kt
```diff
@@ -0,0 +1,66 @@
...
+    interface Factory : MavericksAssistedViewModelFactory<ReleaseNotesViewModel, VectorDummyViewState> {
+        override fun create(initialState: VectorDummyViewState): ReleaseNotesViewModel
+    }
+
+    companion object : MavericksViewModelFactory<ReleaseNotesViewModel, VectorDummyViewState> by hiltMavericksViewModelFactory()
+
+    private var selectedPageIndex = 0
+
+    init {
+        _viewEvents.post(ReleaseNotesViewEvents.SelectPage(selectedPageIndex))
```
bmarty, line 43: For clarity, I would write

```suggestion
        _viewEvents.post(ReleaseNotesViewEvents.SelectPage(0))
```

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesViewModel.kt
```diff
@@ -0,0 +1,66 @@
...
+) : VectorViewModel<VectorDummyViewState, ReleaseNotesAction, ReleaseNotesViewEvents>(initialState) {
+
+    @AssistedFactory
+    interface Factory : MavericksAssistedViewModelFactory<ReleaseNotesViewModel, VectorDummyViewState> {
+        override fun create(initialState: VectorDummyViewState): ReleaseNotesViewModel
+    }
+
+    companion object : MavericksViewModelFactory<ReleaseNotesViewModel, VectorDummyViewState> by hiltMavericksViewModelFactory()
+
+    private var selectedPageIndex = 0
```
bmarty, line 40: This should be moved to the ViewState (to create)

vector/src/main/java/im/vector/app/features/popup/PopupAlertManager.kt
```diff
@@ -307,6 +308,7 @@ class PopupAlertManager @Inject constructor(
                 activity !is PinActivity &&
                 activity !is SignedOutActivity &&
                 activity !is AnalyticsOptInActivity &&
+                activity !is ReleaseNotesActivity &&
```
bmarty, line 311: 👍

vector/src/main/res/layout/bottom_sheet_release_notes.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:layout_gravity="bottom"
+    android:background="?colorSurface"
+    android:orientation="vertical">
```
bmarty, line 9: Please remove `android:orientation="vertical"`

bmarty (commented):
A few first remarks, I will test it when the PR will be updated.

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesViewModel.kt
fedrunov: I actually put it out from there, because there was a confusion

- sometimes page change is induced by button and in that case it comes from ViewModel
- sometimes page change is induced by user and it comes from Fragment

so, my initial attempt was to keep this index in State and update page indicator and button via view events, (since we can't react on each change, because some of them are induced by user (fragment)), but the problem is that view events and view state are both async and delivered by different streams, so there was a race problem

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesPreferencesStore.kt
bmarty: You can (maybe) test the nullability of `HomeActivityViewState.authenticationDescription` https://github.com/vector-im/element-android/blob/main/vector/src/main/java/im/vector/app/features/home/HomeActivityViewState.kt#L25 to have the info.

vector/src/main/res/layout/item_release_carousel.xml
```diff
@@ -0,0 +1,62 @@
...
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginBottom="16dp"
+        android:gravity="center"
+        android:textColor="?vctr_content_primary"
+        android:maxLines="2"
+        app:layout_constraintBottom_toTopOf="@id/carousel_item_body"
+        app:layout_constraintEnd_toEndOf="@id/splashCarouselGutterEnd"
+        app:layout_constraintStart_toStartOf="@id/splashCarouselGutterStart"
+        tools:text="Login version" />
```
bmarty, line 47: Better to  use `@string/onboarding_new_app_layout_welcome_title`

vector/src/main/res/layout/item_release_carousel.xml
```diff
@@ -0,0 +1,62 @@
...
+        style="@style/Widget.Vector.TextView.Subtitle"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:gravity="center"
+        android:textColor="?vctr_content_secondary"
+        android:maxLines="3"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="@id/splashCarouselGutterEnd"
+        app:layout_constraintStart_toStartOf="@id/splashCarouselGutterStart"
+        tools:text="Login version" />
```
bmarty, line 60: Maybe use `@string/onboarding_new_app_layout_welcome_message`

vector/src/main/res/layout/item_release_carousel.xml
```diff
@@ -0,0 +1,62 @@
...
+        android:id="@+id/carousel_item_image"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:scaleType="centerInside"
+        android:layout_marginBottom="@dimen/release_notes_vertical_margin_large"
+        android:contentDescription="@null"
+        app:layout_constraintStart_toStartOf="@id/splashCarouselGutterStart"
+        app:layout_constraintEnd_toEndOf="@id/splashCarouselGutterStart"
+        app:layout_constraintBottom_toTopOf="@id/carousel_item_title" />
```
bmarty, line 33: Could be nice to add `tools:src="@drawable/app_layout_onboarding_welcome"`

bmarty (commented):
A few more remarks on the layout

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -110,9 +112,19 @@ class HomeActivityViewModel @AssistedInject constructor(
         checkSessionPushIsOn()
         observeCrossSigningReset()
         observeAnalytics()
+        observeReleaseNotes()
         initThreadsMigration()
     }
 
+    private fun observeReleaseNotes() {
+        releaseNotesPreferencesStore.appLayoutOnboardingShown.onEach { isAppLayoutOnboardingShown ->
```
bmarty, line 120: Must not be done if `vectorFeatures.isNewAppLayoutEnabled()` is `false`.

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -110,9 +112,19 @@ class HomeActivityViewModel @AssistedInject constructor(
...
+            if (!isAppLayoutOnboardingShown) {
+                _viewEvents.post(HomeActivityViewEvents.ShowReleaseNotes)
+                releaseNotesPreferencesStore.setAppLayoutOnboardingShown(true)
```
bmarty, line 123: I would move this line at above the previous one, just to be 100% sure that the viewModel scope is not cancelled before we set this boolean. Action `HomeActivityViewEvents.ShowReleaseNotes` will start another Activity.

ericdecanini (approved):
Benoit has said most already. The only thing I can add is maybe you can change the status bar color in this new activity to match the background to look a bit nicer. Consider it optional

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesPreferencesStore.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.matrix.android.sdk.api.extensions.orFalse
+import javax.inject.Inject
+
+private val Context.dataStore: DataStore<Preferences> by preferencesDataStore(name = "release_notes")
+
+class ReleaseNotesPreferencesStore @Inject constructor(
+        private val context: Context
+) {
+
+    private val isAppLayoutOnboardingShown = booleanPreferencesKey("SETTINGS_APP_LAYOUT_ONBOARDING_SHOWN")
```
bmarty, line 37: Maybe better to use an int preference, so that we can reuse for future release, to show off what's new.

You could defined constants like

```kotlin
const val RELEASE_1_NEW_APP_LAYOUT = 1
```

or something like that. Maybe no need to set up the whole framework now, but at least store an Int (will be painful to change that later)

WDYT?

fedrunov: I think we'll need more sophisticated system, maybe also with priorities or aggregation of notes. I don't think it will be a problem to switch to ints later, when we'll have final requirement for this system. Unless until we have another flag

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
```diff
@@ -110,9 +112,27 @@ class HomeActivityViewModel @AssistedInject constructor(
...
+        // we don't want to show release notes fore new users or after relogin
+        if (state.authenticationDescription == null) {
+            releaseNotesPreferencesStore.appLayoutOnboardingShown.onEach { isAppLayoutOnboardingShown ->
+                if (!isAppLayoutOnboardingShown) {
+                    _viewEvents.post(HomeActivityViewEvents.ShowReleaseNotes)
+                    releaseNotesPreferencesStore.setAppLayoutOnboardingShown(true)
+                }
+            }.launchIn(viewModelScope)
+        } else {
+            //we assume that users which came from auth flow either have seen updates already (relogin) or don't need them (new user)
```
ElementBot, line 129: - :no_entry_sign: Missing space after //

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesPreferencesStore.kt
fedrunov: fixed - https://github.com/vector-im/element-android/pull/7003/files#diff-7153f5b6cf4c510293d53c73474e854a3a84e6003c88341472c0808377bf3462R121

also fixed problem when this value was null after login - https://github.com/vector-im/element-android/pull/7003/files#diff-31d1cb8d12c34d852a52405908bc8f1a427a1f08485a662527a6ff5c11e7053eR630

vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt
ElementBot, line 129: - :no_entry_sign: Missing space after //

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseCarouselItem.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    @EpoxyAttribute
+    lateinit var item: ReleaseCarouselData.Item
+
+    override fun bind(holder: Holder) {
+        super.bind(holder)
+
+        holder.image.setImageResource(item.image)
+        holder.title.setText(item.title)
+        holder.body.setText(item.body)
```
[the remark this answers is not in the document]
bmarty: Not possible for `setText(@StringRes)`.

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesPreferencesStore.kt
bmarty: > also fixed problem when this value was null after login - https://github.com/vector-im/element-android/pull/7003/files#diff-31d1cb8d12c34d852a52405908bc8f1a427a1f08485a662527a6ff5c11e7053eR630

I think this fixes the issue, but I am wondering if this is the best place to set the `state.selectedAuthenticationState`. Not blocking this PR, but maybe @ouchadam can have a quick look later on this change.

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesFragment.kt
```diff
@@ -51,11 +53,21 @@ class ReleaseNotesFragment : VectorBaseFragment<BottomSheetReleaseNotesBinding>(
...
 
-        views.releaseNotesCarousel.registerOnPageChangeCallback(object : ViewPager2.OnPageChangeCallback() {
+        val pageCallback = object : ViewPager2.OnPageChangeCallback() {
             override fun onPageSelected(position: Int) {
                 viewModel.handle(ReleaseNotesAction.PageSelected(position))
                 updateButtonText(position)
             }
+        }
+
+        viewLifecycleOwner.lifecycle.addObserver(object : DefaultLifecycleObserver {
```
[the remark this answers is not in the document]
bmarty: OK, thanks

bmarty (approved):
LGTM, thanks!

vector/src/main/java/im/vector/app/features/home/room/list/home/release/ReleaseNotesPreferencesStore.kt
ouchadam: for context, the reason why this was `null` previously is because the value was only being used to track _signup_ events in the [HomeActivityViewModel](https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/home/HomeActivityViewModel.kt#L156) 

for the change, I think it's fine, we could eventually remove the `selectedAuthenticationState` parameter and instead rely on updating the state before calling `onSessionCreated`

PR #7081, opened 2022-09-08
restart HomeActivity when app layout flag is changed

fedrunov:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

we should restart app when app layout flag is changed


bmarty (approved):
Thanks!

Issue #7087, opened 2022-09-09
Quickly double clicking the space button always crashes the app


bmarty:
> Thread: main, Exception: java.lang.IllegalStateException: Fragment already added: SpaceListBottomSheet{c83880a} (1a201cc5-17ed-481c-8d34-4f5a958465c5 tag=SpacesBottomSheet)

I can repro.

https://github.com/matrix-org/element-android-rageshakes/issues/42645

daniellekirkwood:
Is this specific to the app layout version?

ericdecanini:
Given the error message, it would be yeah. I managed to reproduce this myself and I'm coming with a fix

> Is this specific to the app layout version?

bmarty:
Fixed in 1.4.38

PR #7102, opened 2022-09-12
Fixes Crash On Double Click Of Space FABs

ericdecanini:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Fixes Crash On Double Click Of Space FABs

## Motivation and context

Fixes https://github.com/vector-im/element-android/issues/7087

## Screenshots / GIFs

N/A

## Tests

Double quick really quickly on the space list or new chat FABs in the new layout. These would have previously crashed the app.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 12

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/NewHomeDetailFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/NewHomeDetailFragment.kt
@@ -201,13 +201,12 @@ class NewHomeDetailFragment :
     private fun setupFabs() {
         showFABs()
 
-        views.newLayoutCreateChatButton.setOnClickListener {
-            newChatBottomSheet.show(requireActivity().supportFragmentManager, NewChatBottomSheet.TAG)
+        views.newLayoutCreateChatButton.debouncedClicks {
+            newChatBottomSheet.takeIf { !it.isAdded }?.show(requireActivity().supportFragmentManager, NewChatBottomSheet.TAG)
         }
 
-        views.newLayoutOpenSpacesButton.setOnClickListener {
-            // Click action for open spaces modal goes here
-            spaceListBottomSheet.show(requireActivity().supportFragmentManager, SpaceListBottomSheet.TAG)
+        views.newLayoutOpenSpacesButton.debouncedClicks {
+            spaceListBottomSheet.takeIf { !it.isAdded }?.show(requireActivity().supportFragmentManager, SpaceListBottomSheet.TAG)
         }
     }
 
--- a/changelog.d/7102.bugfix
+++ b/changelog.d/7102.bugfix
@@ -1,0 +1,1 @@
+Fixes crash when quickly double clicking FABs in the new app layout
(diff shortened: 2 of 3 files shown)


fedrunov (approved):
LGTM

vector/src/main/java/im/vector/app/features/home/NewHomeDetailFragment.kt
```diff
@@ -201,13 +201,12 @@ class NewHomeDetailFragment :
...
-            newChatBottomSheet.show(requireActivity().supportFragmentManager, NewChatBottomSheet.TAG)
+        views.newLayoutCreateChatButton.debouncedClicks {
+            newChatBottomSheet.takeIf { !it.isAdded }?.show(requireActivity().supportFragmentManager, NewChatBottomSheet.TAG)
         }
 
-        views.newLayoutOpenSpacesButton.setOnClickListener {
-            // Click action for open spaces modal goes here
-            spaceListBottomSheet.show(requireActivity().supportFragmentManager, SpaceListBottomSheet.TAG)
+        views.newLayoutOpenSpacesButton.debouncedClicks {
+            spaceListBottomSheet.takeIf { !it.isAdded }?.show(requireActivity().supportFragmentManager, SpaceListBottomSheet.TAG)
```
bmarty, line 209: I think the root cause of the issue is to instantiate the BottomSheets in advance [at this point](https://github.com/vector-im/element-android/blob/main/vector/src/main/java/im/vector/app/features/home/NewHomeDetailFragment.kt#L93). I would have fixed this issue differently. Using `debouncedClicks` is OK, but using `takeIf { !it.isAdded }` looks a bit like a workaround.

WDYT?

bmarty: Also there is no need to inject the constructor [here](https://github.com/vector-im/element-android/blob/main/vector/src/main/java/im/vector/app/features/home/room/list/home/NewChatBottomSheet.kt#L30).

ericdecanini: I believe the approach you're leading towards is to instantiate the bottom sheet as we show it? (`SpaceListBottomSheet.show()`)

I considered this approach but doing so would just reduce the bug from a crash to having two bottom sheets show together. `debouncedClicks` may help with this, but I wanted to add the `!it.isAdded` check as a guaranteed method.

Research also shows me that making this check (or using `bottomSheet.isShowing` in other cases) is a common practice.

bmarty (approved):

vector/src/main/java/im/vector/app/features/home/NewHomeDetailFragment.kt
bmarty: One more point:
When `spaceListBottomSheet` is displayed, if the user rotates it's device, the Fragment `SpaceListBottomSheet` will be restored by the System (the BottomSheet will still be visible), but `NewHomeDetailFragment` will not have a correct reference to this Fragment, since it will just instantiate a new instance in the constructor.

So this code:
https://github.com/vector-im/element-android/blob/develop/vector/src/main/java/im/vector/app/features/home/NewHomeDetailFragment.kt#L197

will not work (I did not check though).

You should try to get the reference (by tag for instance) to the restored bottom sheet Fragment instead of creating a new one.

PR #7108, opened 2022-09-13
Use `buildjet-4vcpu-ubuntu-2204` runner instead of `macos-latest` to …

bmarty:
…build and run the integration tests.

Draft, to see how the CI will digest this change.

post build result (using `4vcpu` runner):

|Before|After|
|-|-|
|(from #7015) 74 minutes image|23 minutesimage|

Just for this specific job. I will also remove the second attempt as discussed today, since it looks not very useful.

### DIFF
--- a/.github/workflows/post-pr.yml
+++ b/.github/workflows/post-pr.yml
@@ -31,7 +31,7 @@ jobs:
   ui-tests:
     name: UI Tests (Synapse)
     needs: should-i-run
-    runs-on: macos-latest
+    runs-on: buildjet-4vcpu-ubuntu-2204
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/post-pr.yml
+++ b/.github/workflows/post-pr.yml
@@ -31,7 +31,7 @@ jobs:
   ui-tests:
     name: UI Tests (Synapse)
     needs: should-i-run
-    runs-on: buildjet-2vcpu-ubuntu-2204
+    runs-on: buildjet-4vcpu-ubuntu-2204
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/post-pr.yml
```diff
@@ -31,7 +31,7 @@ jobs:
   ui-tests:
     name: UI Tests (Synapse)
     needs: should-i-run
-    runs-on: macos-latest
+    runs-on: buildjet-4vcpu-ubuntu-2204
```
bmarty, line 34: We start with this runner and will compare later the perf of other available (but more not with the same price) runners.

manuroe: Can we try first with  `buildjet-2vcpu-ubuntu-2204`?
And use beefier machines only if we can prove there are better?

bmarty: Yes. 0685fb1e1a16cb2dce12ed2e2785db31e8efab46

bmarty:
It seems that with `buildjet-2vcpu-ubuntu-2204` we do not have enough RAM. See https://github.com/vector-im/element-android/actions/runs/3080897698/jobs/4980174874 failure.

Can we revert to `4vcpu` which got 16 GB of RAM (2x the RAM available on 4vcpu) @manuroe ?

changelog.d/7108.misc
```diff
@@ -0,0 +1 @@
+Move some GitHub action to buildjet runner, and remove the second attempt to run integration tests.
```
manuroe, line 1:
```suggestion
Move some GitHub actions to buildjet runner, and remove the second attempt to run integration tests.
```

manuroe (approved):
We need to change the PR title again. Then, it is all good to me 🚀🚀🚀

fedrunov:
It shows "all checks haves passed", while `all-tests` has actually failed. Probably because of `continue-on-error: true` : https://github.com/community/community/discussions/15452

PR #7129, opened 2022-09-14
Example screenshot test

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Motivation and context

An example screenshot test to show the CI pipelines working and to use as a reference for future tests

## Screenshots / GIFs
N/A

## Tests
N/A

## Tested devices

- [x] CI

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: Archive Screenshot Results on Error
         if: failure()
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v3
         with:
           name: screenshot-results
           path: |
--- a/build.gradle
+++ b/build.gradle
@@ -312,6 +312,7 @@ ext.initScreenshotTests = { project ->
     if (hasScreenshots) {
         project.apply plugin: 'app.cash.paparazzi'
     }
+    project.dependencies { testCompileOnly "app.cash.paparazzi:paparazzi:1.0.0" }
     project.android.testOptions.unitTests.all {
         def screenshotTestCapture = "**/*ScreenshotTest*"
         if (hasScreenshots) {
--- a/vector/src/test/java/im/vector/app/screenshot/PaparazziExampleScreenshotTest.kt
+++ b/vector/src/test/java/im/vector/app/screenshot/PaparazziExampleScreenshotTest.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2022 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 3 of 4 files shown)


build.gradle
```diff
@@ -312,6 +312,7 @@ ext.initScreenshotTests = { project ->
     if (hasScreenshots) {
         project.apply plugin: 'app.cash.paparazzi'
     }
+    project.dependencies { testCompileOnly "app.cash.paparazzi:paparazzi:1.0.0" }
```
ouchadam, line 315: the paparazzi jar is automatically applied by plugin however we only apply the plugin when running the screenshot tasks (because the jar causes runtime class clashes when running the unit tests)

but... we need to always have the jar on the compilation class path so that we can compile the unit tests source set, hence the `testCompileOnly` declaration

vector/src/test/java/im/vector/app/screenshot/PaparazziExampleScreenshotTest.kt
```diff
@@ -31,6 +31,7 @@ class PaparazziExampleScreenshotTest {
     val paparazzi = Paparazzi(
             deviceConfig = PIXEL_3,
             theme = "Theme.Vector.Light",
+            maxPercentDifference = 0.0,
```
bmarty, line 34: Ah, I thought it will be a global setting :/

bmarty:
(there is a Danger false positive warning that I will handle later)

vector/src/test/java/im/vector/app/screenshot/PaparazziExampleScreenshotTest.kt
ouchadam: we could create our own function to create the Paparazzi instance (to act as a global)

bmarty: Yes, I have done it on a stash :), using a parent class:

image

Also maybe update [the doc](https://github.com/vector-im/element-android/blob/7740404b1430d32282ed7289db2c1bba8e107af7/docs/screenshot_testing.md#example) with the new param.

.github/workflows/tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
 
       - name: Archive Screenshot Results on Error
         if: failure()
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v3
```
bmarty, line 40: Strange that Dependabot did not propose to upgrade this.

bmarty (approved):
Perfect, thanks!

.github/workflows/tests.yml
ouchadam: not sure where I copied it from :facepalm:  it should fix the missing artifacts in https://github.com/vector-im/element-android/pull/7137 (which now correctly fails) :crossed_fingers:

PR #7137, opened 2022-09-15
[Ignore] Forcing the screenshot test to fail

ouchadam:
Showcasing how screenshot test failures present themselves  - As artifacts of the `tests.yml` workflow


bmarty:
The CI seems happy?

ouchadam:
@bmarty yeah it's strange.. I'm looking into it, changing the text to `fail1` fails the test but not `fail` or empty

ouchadam:
> @bmarty yeah it's strange.. I'm looking into it, changing the text to `fail1` fails the test but not `fail` or empty

turns out this was due to the `maxPercentDifference` option defaulting to 0.1%, dropping to 0% catches the changes but has the risk failing due to navigation and status bar pixel rendering quirks (will need to monitor)

ouchadam:
Tests are failing and the artifacts are correctly uploaded https://github.com/vector-im/element-android/actions/runs/3134733720 (can close!)

PR #7140, opened 2022-09-15
Excluding legacy support library annotation usages

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Excludes usages of the support annotation library, these are compile time hints and safe to remove.
- Removes unused `progressPieIndicator` dependency
- Excludes the appcompat library from jitsi's `android-scalablevideoview`, the library is pre-jetified but includes a redundant legacy support library transitive dependency  

## Motivation and context

The time has come to avoid using the Jetifier (the process of converting libraries from the legacy support library to androidx). The `Paparazzi` integration is blocked due to the plugin requiring the jetifier to be disabled for the base android libraries which causes the unit tests with android imports to fail. 

There is one more library to update `im.dlg:android-dialer:1.2.5` which will need to be manually included as the upstream project no longer exists, although it's mainly a wrapper around legacy AOSP classes.

Will slightly speed up the build and may improve the _cachability_ of the module dependencies (as they'll be statically cached from their maven repositories)

## Screenshots / GIFs
No UI changes

## Tests

Smoke test around the app 

- View markdown text
- Make video calls

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

### DIFF
--- a/changelog.d/7140.misc
+++ b/changelog.d/7140.misc
@@ -1,0 +1,1 @@
+Exclude legacy android support annotation library
--- a/dependencies_groups.gradle
+++ b/dependencies_groups.gradle
@@ -69,8 +69,6 @@ ext.groups = [
                         'com.gabrielittner.threetenbp',
                         'com.getkeepsafe.relinker',
                         'com.github.bumptech.glide',
-                        'com.github.filippudak',
-                        'com.github.filippudak.progresspieview',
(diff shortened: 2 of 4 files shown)


bmarty (approved):
LGTM, thanks!

PR #7145, opened 2022-09-15
room summary now has constant height

### LINKED ISSUE #7079: New App Layout | PagedList is buggy (long placeholders, jumpt to top)
When using the new layout (nighlty) I have several problems with the room list:
- Placeholders appearing fast and stay quite long (can be forever). It's not happening in production app
- When placehodlers are replaced the list sometimes jump to top

fedrunov:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Currently room summary item height is not constant - it can vary depending if there is enough text in the last message/event to go on a second line. This cause some flickering and jumping scrolls when placeholders are replaced with actual items. To mitigate this I've forced last event text view to have constant amount of lines (2 normally, reduced to 1 if font scale is bigger than "Large"). Also Placeholder's items had height defined in dp, which caused them to have wrong height if font size is scaled. 

## Motivation and context

closes #7079

## Screenshots / GIFs

before changes:

before

After changes:

normal

with font scaling (Huge):

huge

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryItem.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryItem.kt
@@ -103,6 +103,9 @@ abstract class RoomSummaryItem : VectorEpoxyModel<RoomSummaryItem.Holder>(R.layo
     @EpoxyAttribute
     var showSelected: Boolean = false
 
+    @EpoxyAttribute
+    var useSingleLineForLastEvent: Boolean = false
+
     override fun bind(holder: Holder) {
         super.bind(holder)
 
@@ -122,6 +125,10 @@ abstract class RoomSummaryItem : VectorEpoxyModel<RoomSummaryItem.Holder>(R.layo
         holder.roomAvatarFailSendingImageView.isVisible = hasFailedSending
         renderSelection(holder, showSelected)
         holder.roomAvatarPresenceImageView.render(showPresence, userPresence)
+
+        if (useSingleLineForLastEvent) {
+            holder.subtitleView.setLines(1)
+        }
     }
 
     private fun renderDisplayMode(holder: Holder) = when (displayMode) {
--- a/vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryListController.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryListController.kt
@@ -17,18 +17,26 @@
 package im.vector.app.features.home.room.list
 
 import im.vector.app.features.home.RoomListDisplayMode
+import im.vector.app.features.settings.FontScalePreferences
 import org.matrix.android.sdk.api.session.room.model.RoomSummary
 
 class RoomSummaryListController(
         private val roomSummaryItemFactory: RoomSummaryItemFactory,
-        private val displayMode: RoomListDisplayMode
+        private val displayMode: RoomListDisplayMode,
+        fontScalePreferences: FontScalePreferences
 ) : CollapsableTypedEpoxyController<List<RoomSummary>>() {
 
     var listener: RoomListListener? = null
+    private val shouldUseSingleLine: Boolean
+
+    init {
+        val fontScale = fontScalePreferences.getResolvedFontScaleValue()
+        shouldUseSingleLine = fontScale.scale > FontScalePreferences.SCALE_LARGE
+    }
 
     override fun buildModels(data: List<RoomSummary>?) {
         data?.forEach {
-            add(roomSummaryItemFactory.create(it, emptyMap(), emptySet(), displayMode, listener))
+            add(roomSummaryItemFactory.create(it, emptyMap(), emptySet(), displayMode, listener, shouldUseSingleLine))
         }
     }
 }
--- a/vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryPagedController.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryPagedController.kt
@@ -20,18 +20,26 @@ import com.airbnb.epoxy.EpoxyModel
 import com.airbnb.epoxy.paging.PagedListEpoxyController
 import im.vector.app.core.utils.createUIHandler
 import im.vector.app.features.home.RoomListDisplayMode
+import im.vector.app.features.settings.FontScalePreferences
 import org.matrix.android.sdk.api.session.room.members.ChangeMembershipState
 import org.matrix.android.sdk.api.session.room.model.RoomSummary
 
 class RoomSummaryPagedController(
         private val roomSummaryItemFactory: RoomSummaryItemFactory,
-        private val displayMode: RoomListDisplayMode
+        private val displayMode: RoomListDisplayMode,
+        fontScalePreferences: FontScalePreferences
 ) : PagedListEpoxyController<RoomSummary>(
         // Important it must match the PageList builder notify Looper
         modelBuildingHandler = createUIHandler()
 ), CollapsableControllerExtension {
 
     var listener: RoomListListener? = null
+    private val shouldUseSingleLine: Boolean
+
+    init {
+        val fontScale = fontScalePreferences.getResolvedFontScaleValue()
+        shouldUseSingleLine = fontScale.scale > FontScalePreferences.SCALE_LARGE
+    }
 
     var roomChangeMembershipStates: Map<String, ChangeMembershipState>? = null
         set(value) {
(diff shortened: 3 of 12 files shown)


changelog.d/7079.bugfix
```diff
@@ -0,0 +1 @@
+Fix text margin in QR code view when no display name is set
```
bmarty, line 1: Wrong comment?

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryItem.kt
```diff
@@ -122,6 +125,10 @@ abstract class RoomSummaryItem : VectorEpoxyModel<RoomSummaryItem.Holder>(R.layo
         holder.roomAvatarFailSendingImageView.isVisible = hasFailedSending
         renderSelection(holder, showSelected)
         holder.roomAvatarPresenceImageView.render(showPresence, userPresence)
+
+        if (useSingleLineForLastEvent) {
+            holder.subtitleView.setLines(1)
+        }
```
bmarty, line 131: Since items can be recycled, we need to add a `else` block here. But since the value is constant in the controller, I think this is fine.

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryPagedController.kt
```diff
@@ -57,8 +65,14 @@ class RoomSummaryPagedController(
     }
 
     override fun buildItemModel(currentPosition: Int, item: RoomSummary?): EpoxyModel<*> {
-        // for place holder if enabled
-        item ?: return RoomSummaryItemPlaceHolder_().apply { id(currentPosition) }
-        return roomSummaryItemFactory.create(item, roomChangeMembershipStates.orEmpty(), emptySet(), displayMode, listener)
+        return if (item == null) {
+            val host = this
+            RoomSummaryItemPlaceHolder_().apply {
```
bmarty, line 70: Nit: `roomSummaryItemPlaceHolder{ }` could be used
```suggestion
            roomSummaryItemPlaceHolder {
```

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeFilteredRoomsController.kt
```diff
@@ -67,7 +76,14 @@ class HomeFilteredRoomsController @Inject constructor(
     }
 
     override fun buildItemModel(currentPosition: Int, item: RoomSummary?): EpoxyModel<*> {
-        item ?: return RoomSummaryItemPlaceHolder_().apply { id(currentPosition) }
-        return roomSummaryItemFactory.create(item, roomChangeMembershipStates.orEmpty(), emptySet(), RoomListDisplayMode.ROOMS, listener)
+        return if (item == null) {
+            val host = this
+            RoomSummaryItemPlaceHolder_().apply {
```
bmarty, line 81: Nit: `roomSummaryItemPlaceHolder{ }` could be used
```suggestion
            roomSummaryItemPlaceHolder {
```

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryListController.kt
```diff
@@ -17,18 +17,26 @@
...
+        private val displayMode: RoomListDisplayMode,
+        fontScalePreferences: FontScalePreferences
 ) : CollapsableTypedEpoxyController<List<RoomSummary>>() {
 
     var listener: RoomListListener? = null
+    private val shouldUseSingleLine: Boolean
+
+    init {
+        val fontScale = fontScalePreferences.getResolvedFontScaleValue()
+        shouldUseSingleLine = fontScale.scale > FontScalePreferences.SCALE_LARGE
```
bmarty, line 34: A tiny detail, but I would have used `>=`
```suggestion
        shouldUseSingleLine = fontScale.scale >= FontScalePreferences.SCALE_LARGER
```

bmarty (approved):
I am thinking of fixing this issue since a long time, thanks for doing it! (I think I have a local branch with some attempt somewhere).
I was wondering if it would be possible to define a `dimen` resource for the cell height, which could be with unit `sp`, so that we ensure that the room item and the placeholder have the same height, and so we avoid regression. WDYT?

ericdecanini (approved):

changelog.d/7079.bugfix
fedrunov: now I wonder where I put the correct one 😅

vector/src/main/java/im/vector/app/core/utils/FirstItemUpdatedObserver.kt
```diff
@@ -0,0 +1,47 @@
...
+import androidx.recyclerview.widget.RecyclerView
+
+/**
+ * This observer detects when item was added or moved to the first position of the adapter, while recyclerView is scrolled to the top. This is necessary
+ * to force recycler to scroll to the top to make such item visible, because by default it will keep items on screen, while adding new item to the top,
+ * outside of the viewport
+ * @param layoutManager - [LinearLayoutManager] of the recycler view, which displays items
+ * @param onItemUpdated - callback to be called, when observer detects event
+ */
+class FirstItemUpdatedObserver(
```
ElementBot, line 29: - :warning: Documentation of FirstItemUpdatedObserver is outdated

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryListController.kt
fedrunov: I don't feel ok checking equality of two floating-point numbers. We don't do any calculations here, so they should be the same, but I still concerned. 🤔

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryPagedController.kt
fedrunov: `roomSummaryItemPlaceHolder` adds model to collector, but here we need to return item instead, so we can't really use this inside `buildItemModel` method :(

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeFilteredRoomsController.kt
fedrunov: `roomSummaryItemPlaceHolder` adds model to collector, but here we need to return item instead, so we can't really use this inside `buildItemModel` method :(

vector/src/main/java/im/vector/app/core/utils/FirstItemUpdatedObserver.kt
ElementBot, line 29: - :warning: Documentation of FirstItemUpdatedObserver is outdated

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryPagedController.kt
bmarty: Oh, true, my bad.

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeFilteredRoomsController.kt
bmarty: fair enough :)

vector/src/main/java/im/vector/app/features/home/room/list/RoomSummaryListController.kt
bmarty: I see your point, OK

vector/src/main/java/im/vector/app/core/utils/FirstItemUpdatedObserver.kt
bmarty: (you have to use `@property` instead of `@param` for class members documentation, in this case `onItemUpdated` )

bmarty (approved):

vector/src/main/java/im/vector/app/core/utils/FirstItemUpdatedObserver.kt
```diff
@@ -0,0 +1,47 @@
...
+import androidx.recyclerview.widget.RecyclerView
+
+/**
+ * This observer detects when item was added or moved to the first position of the adapter, while recyclerView is scrolled to the top. This is necessary
+ * to force recycler to scroll to the top to make such item visible, because by default it will keep items on screen, while adding new item to the top,
+ * outside of the viewport
+ * @property layoutManager - [LinearLayoutManager] of the recycler view, which displays items
+ * @property onItemUpdated - callback to be called, when observer detects event
+ */
+class FirstItemUpdatedObserver(
```
ElementBot, line 29: - :warning: Documentation of FirstItemUpdatedObserver is outdated

PR #7154, opened 2022-09-16
Deferred DMs - Handle the local rooms within the new AppLayout

### LINKED ISSUE #7153: Deferred DMs - Locals are not deleted anymore within the new AppLayout
### Steps to reproduce

First bug:
1. Start creating a DM with the start DM feature flag and the new App Layout enabled
2. Abort the DM creation and go back to the home list
3. Activate the recent rooms display from the AppLayout configuration
4. Then, the local room appears in the recent rooms

Second bug:
1. Start creating a DM with the start DM feature flag and the new App Layout enabled
2. Write some text without sending it
3. Abort the DM creation and go back to the home list
4. Start creating again a DM with the same person
4. Then, the local room is opened again with the drafted message

#### What did you expect?
The local room should not be persisted anymore when the user is back to the room list

### Your phone model

_No response_

### Operating system version

_No response_

_No response_

### Homeserver

_No response_

### Will you send logs?

No

Yes

Florian14:
## Type of change

- [ ] Feature
- [X] Bugfix
- [ ] Technical
- [ ] Other :

## Content

This PR adds the deletion of the local rooms from the database when the room list is shown within the new AppLayout feature enabled.

## Motivation and context

Close #7153

## Tests

- Start creating a DM with the start DM feature flag and the new App Layout enabled
- Activate the recent rooms display from the AppLayout configuration
- Write some text without sending it
- Abort the DM creation and go back to the home list
- Verify that the local rooms does not appear in the recent rooms
- Start creating again a DM with the same person
- Verify that the local room has not been retrieved (the draft is not persisted) 

We can also observe logs in the console about the deletion of the local rooms when the home room list is shown again.

## Tested devices

- [ ] Physical
- [X] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/accountdata/UpdateBreadcrumbsTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/accountdata/UpdateBreadcrumbsTask.kt
@@ -17,6 +17,7 @@
 package org.matrix.android.sdk.internal.session.user.accountdata
 
 import com.zhuinden.monarchy.Monarchy
+import org.matrix.android.sdk.api.session.room.model.localecho.RoomLocalEcho
 import org.matrix.android.sdk.internal.database.model.BreadcrumbsEntity
 import org.matrix.android.sdk.internal.database.query.get
 import org.matrix.android.sdk.internal.di.SessionDatabase
@@ -41,6 +42,8 @@ internal class DefaultUpdateBreadcrumbsTask @Inject constructor(
 ) : UpdateBreadcrumbsTask {
 
     override suspend fun execute(params: UpdateBreadcrumbsTask.Params) {
+        // Do not add local rooms to the recent rooms list as they should not be known by the server
+        if (RoomLocalEcho.isLocalEchoId(params.newTopRoomId)) return
         val newBreadcrumbs =
                 // Get the breadcrumbs entity, if any
                 monarchy.fetchCopied { BreadcrumbsEntity.get(it) }
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/accountdata/UpdateBreadcrumbsTask.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/accountdata/UpdateBreadcrumbsTask.kt
@@ -17,7 +17,6 @@
 package org.matrix.android.sdk.internal.session.user.accountdata
 
 import com.zhuinden.monarchy.Monarchy
-import org.matrix.android.sdk.api.session.room.model.localecho.RoomLocalEcho
 import org.matrix.android.sdk.internal.database.model.BreadcrumbsEntity
 import org.matrix.android.sdk.internal.database.query.get
 import org.matrix.android.sdk.internal.di.SessionDatabase
@@ -42,8 +41,6 @@ internal class DefaultUpdateBreadcrumbsTask @Inject constructor(
 ) : UpdateBreadcrumbsTask {
 
     override suspend fun execute(params: UpdateBreadcrumbsTask.Params) {
-        // Do not add local rooms to the recent rooms list as they should not be known by the server
-        if (RoomLocalEcho.isLocalEchoId(params.newTopRoomId)) return
         val newBreadcrumbs =
                 // Get the breadcrumbs entity, if any
                 monarchy.fetchCopied { BreadcrumbsEntity.get(it) }
--- a/vector/src/main/java/im/vector/app/features/home/room/list/RoomListViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/list/RoomListViewModel.kt
@@ -97,7 +97,6 @@ class RoomListViewModel @AssistedInject constructor(
 
     init {
         observeMembershipChanges()
-        observeLocalRooms()
 
         spaceStateHandler.getSelectedSpaceFlow()
                 .distinctUntilChanged()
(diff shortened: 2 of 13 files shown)


vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
```diff
@@ -40,6 +40,7 @@ import im.vector.app.databinding.FragmentRoomListBinding
 import im.vector.app.features.analytics.plan.ViewRoom
 import im.vector.app.features.home.room.list.RoomListAnimator
 import im.vector.app.features.home.room.list.RoomListListener
+import im.vector.app.features.home.room.list.RoomListViewState
```
ElementBot, line 43: - :no_entry_sign: Unused import

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
```diff
@@ -97,6 +97,11 @@ class HomeRoomListFragment :
                 is HomeRoomListViewEvents.Done -> Unit
             }
         }
+
+        roomListViewModel.onEach(HomeRoomListViewState::localRoomIds) {
+            // Local rooms should not exist anymore when the room list is shown
+            roomListViewModel.deleteLocalRooms(it)
```
onurays, line 103: I think `HomeRoomListViewModel` should decide this business logic.

Florian14: My problem is that the VM is still alive while the local room is opened, so moving this logic to the VM means that the local room will be deleted too early. I trigger it from the fragment to delete the rooms only when the room list is visible to the user (ie. the local timeline is closed).
It would be nice to have a foreground/background state to decide when a ViewModel should be "alive".
Do you have any suggestions about it?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/accountdata/UpdateBreadcrumbsTask.kt
```diff
@@ -41,6 +42,8 @@ internal class DefaultUpdateBreadcrumbsTask @Inject constructor(
 ) : UpdateBreadcrumbsTask {
 
     override suspend fun execute(params: UpdateBreadcrumbsTask.Params) {
+        // Do not add local rooms to the recent rooms list as they should not be known by the server
+        if (RoomLocalEcho.isLocalEchoId(params.newTopRoomId)) return
```
bmarty, line 46: Good point, but maybe this could be filtered before, at https://github.com/vector-im/element-android/blob/c73c36eb26149390167d23944b9fa3264cbc10f4/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/DefaultRoomService.kt#L175, WDYT?

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
bmarty: You could add an Action to the ViewModel, triggered in the `onResume` of the Fragment, which delete all the local rooms. So `HomeRoomListViewState.localRoomIds` could even be removed I think.

Also we do not add public API to the ViewModels, please call `handle()` and add an action.

So we will have in the Fragment:

```kollin
    override fun onResume() {
        super.onResume()
        roomListViewModel.handle(HomeRoomListAction.DeleteAllLocalRoom)
    }
```

bmarty (commented):
Thanks, some feedback. If you take my suggestion, please also update the "legacy" code.

vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListFragment.kt
Florian14: Done https://github.com/vector-im/element-android/pull/7154/commits/8999b40c1af5f6a9e11868f2b1169c487007697c I moved to `onStart` instead of `onResume`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/user/accountdata/UpdateBreadcrumbsTask.kt
Florian14: Done https://github.com/vector-im/element-android/pull/7154/commits/648498e2de4e8af16b8a576c576c3f6737771812
I was hesitating between both places, I choose the task to be sure it would never be sent to the server but the task is only called by the service until now

Florian14:
I also added this [commit](https://github.com/vector-im/element-android/pull/7154/commits/5e504942cad4e8182ce3cd23e277e43692c9a36b) to remove the read receipts from the DB (they're not sent to the server as they are only attached to local events but they were persisted in the database)

vector/src/main/java/im/vector/app/features/home/room/list/RoomListViewModel.kt
```diff
@@ -338,4 +320,16 @@ class RoomListViewModel @AssistedInject constructor(
...
     }
+
+    private fun handleDeleteLocalRooms() {
+        val localRoomIds = session.roomService()
+                .getRoomSummaries(roomSummaryQueryParams { roomId = QueryStringValue.Contains(RoomLocalEcho.PREFIX) })
+                .map { it.roomId }
+
+        viewModelScope.launch {
+            localRoomIds.forEach {
+                session.roomService().deleteLocalRoom(it)
```
bmarty, line 331: Not for this PR, but having a method `RoomService.deleteAllLocalRooms()` would be handy.

bmarty (approved):
LGTM, thanks!

PR #6788, opened 2022-08-09
`:vector-app` level variants 

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

- Lifts the debug, release, gplay and fdroid source-sets to the application module. Allows the majority of the application to be cached between variant builds
- Enables the gradle build cache by default

## Motivation and context

To help improve the compilation time

## Screenshots / GIFs
No UI changes

## Tests

- Smoke test with all the different variants/build types

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): 28

PR #7169, opened 2022-09-19
Fix regression on our dependency, due to merge of #6788.

bmarty:
We do not use `android-embedded_fcm_distributor` anymore (since #7068). The code was compiling because `android-embedded_fcm_distributor` has a dependency on `firebase-messaging`.

No need for a changelog, since the code is not released.


ganfra (approved):

ouchadam:
ahh great catch! this is the danger of introducing new files, there's no conflicts :cry:

PR #7197, opened 2022-09-21
Fix tests

bmarty:
PR created just to trigger the CI.


bmarty:
We will handle that later...

PR #7198, opened 2022-09-21
Allow configurable sync timeout

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Allows the sync timeout to be configured via the `MatrixConfiguration`

## Motivation and context

Our instrumentation tests can use a shorter timeout (5 seconds instead of 30 seconds) 

Before: **23m 43s**
After: **16m 39s**

**30% reduction** in runtime, saving around 7 minutes~

## Screenshots / GIFs

|Before|After|
|-|-|
|2022-09-21T15:28:49,352822323+01:00|2022-09-21T15:03:35,811981115+01:00|
 
## Tests

Run the `./gradlew matrix-sdk-android:connectedAndroidTest`, or all the `matrix-sdk-android androidTest` tests from within Android Studio 

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 31

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixConfiguration.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixConfiguration.kt
@@ -73,5 +73,5 @@ data class MatrixConfiguration(
         /**
          * Sync configuration.
          */
-        val syncConfig: SyncConfig = SyncConfig()
+        val syncConfig: SyncConfig = SyncConfig(),
 )
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixConfiguration.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixConfiguration.kt
@@ -70,4 +70,8 @@ data class MatrixConfiguration(
          * List of network interceptors, they will be added when building an OkHttp client.
          */
         val networkInterceptors: List<Interceptor> = emptyList(),
+        /**
+         * Sync configuration.
+         */
+        val syncConfig: SyncConfig = SyncConfig()
 )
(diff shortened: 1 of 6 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/MatrixConfiguration.kt
```diff
@@ -70,4 +70,8 @@ data class MatrixConfiguration(
          * List of network interceptors, they will be added when building an OkHttp client.
          */
         val networkInterceptors: List<Interceptor> = emptyList(),
+        /**
```
ElementBot, line 73: - :warning: The first sentence of this KDoc does not end with the correct punctuation.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/SyncConfig.kt
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.api
+
+data class SyncConfig(
+        /**
```
ElementBot, line 20: - :warning: The first sentence of this KDoc does not end with the correct punctuation.

bmarty (approved):
Thanks. Faster tests!
OOI do you know why the tests are faster? Do we have to wait for a sync request to terminate to finish each test doing sync request? Because if there thing on the server for the client, the client should have a fast answer.

Small remark: you could use more trailing comma (not blocking though).

PR #7210, opened 2022-09-22
Missing buildjet Emulator network

ouchadam:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

Fixes the `CantVerifyTest` failing when running on the buildjet github action runner, the cause seems to be related to the emulator not having access to the wider network, which meant the initial matrix.org requests would timeout.

For reasons unknown to me, switching the emulator image to use the play store version ensures that the outside network is accessible.

## Motivation and context

To re-enable an ignored test

## Screenshots / GIFs

No production code changes

## Tests

Run on the CI!

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ on:
 # Enrich gradle.properties for CI/CD
 env:
   GRADLE_OPTS: -Dorg.gradle.jvmargs="-Xmx3072m -Dfile.encoding=UTF-8 -XX:+HeapDumpOnOutOfMemoryError" -Dkotlin.daemon.jvm.options="-Xmx2560m" -Dkotlin.incremental=false
-  CI_GRADLE_ARG_PROPERTIES: --stacktrace -PpreDexEnable=false --max-workers 2 --no-daemon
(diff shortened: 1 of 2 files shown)


mnaturel (approved):
Thanks for looking into this!

.github/workflows/tests.yml
```diff
@@ -54,34 +54,14 @@ jobs:
...
-      - name: AVD cache
-        uses: actions/cache@v3
-        id: avd-cache
-        with:
-          path: |
-            ~/.android/avd/*
-            ~/.android/adb*
-          key: avd-${{ matrix.api-level }}
-
-      - name: create AVD and generate snapshot for caching
```
ouchadam, line 66: caching the AVD is causing strange problems for the `playstore` emulator image where the expected AVD path changes between runs 

```
PANIC: Broken AVD system path. Check your ANDROID_SDK_ROOT value [/usr/local/lib/android/sdk]!
```

Creating a new AVD image each run fixes the problem but does mean the initial start time becomes slower

PR #7212, opened 2022-09-22
RoomList: fix space switching when changing order

ganfra:
Hotfix for the 1.5.0 release


bmarty:
I think we can close it now.

PR #7222, opened 2022-09-23
Ignore flaky test "CantVerifyTest" and "fix" VoiceRecorderLTests

bmarty:
To give us time to work on it and unblock other PRs so they can be merged.


ouchadam (approved):

mnaturel (approved):

jmartinesp:
It seems like changing the test order in `VoiceRecorderL` didn't really fix the issue? I could try creating a PR that fixes this by not using the actual `OggOpusEncoder` but a fake one, but for that I'd need to also make some new changes to its implementation [here](https://github.com/vector-im/libopusencoder-android/blob/main/opusencoder/src/main/java/io/element/android/opusencoder/OggOpusEncoder.kt), since `mockk` can't actually mock `external fun`s.

bmarty:
Yes, strange, on this PR #7197 it seems to be fixed. Let's just ignore the test here. I have force pushed the branch.

PR #7238, opened 2022-09-26
tests workflow will now update gradle cache when run against develop

fedrunov:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [x] Technical
- [ ] Other :

## Content

we want to clean up gradle cache from `build-gradel-action` which we use currently only in `tests` workflow, so now

- cache will be cleaned up when `tests` workflow is run against `develop` branch. This will not wipe entire cache, but rather remove what is not used (up to the doc)
- cache will be updated when `tests` workflow is run against `develop` branch (previously it was `main` as by default)


ouchadam (approved):
:crossed_fingers:

fedrunov:
Should I also add `env: GRADLE_BUILD_ACTION_CACHE_DEBUG_ENABLED`?

jmartinesp:
> Should I also add `env: GRADLE_BUILD_ACTION_CACHE_DEBUG_ENABLED`?

Maybe test it first only on this branch, then if the difference in build time is not really big do it for other branches? The README mentions that it could prevent some caching operations from running on parallel.

ouchadam:
> > Should I also add `env: GRADLE_BUILD_ACTION_CACHE_DEBUG_ENABLED`?
> 
> Maybe test it first only on this branch, then if the difference in build time is not really big do it for other branches? The README mentions that it could prevent some caching operations from running on parallel.

we would need this workflow to run on develop in order to insert the cache (for other PRs to be able to read from the cache)

we need to create another branch after this is merged (and the workflow finishes) to test it's working :sweat_smile:

bmarty (approved):
Thanks, but it should be good to try to track the effects of this change. Maybe a screenshot showing the current cache size is enough.

bmarty:
(and Danger seems broken somehow)

Running locally, I get the expected warning about the missing changelog:

```sh
$> bundle exec danger pr https://github.com/vector-im/element-android/pull/7238 --dangerfile=./tools/danger/dangerfile.js
Starting Danger PR on vector-im/element-android#7238
You don't have a DANGER_GITHUB_API_TOKEN set up, this is optional, but TBH, you want to do this
Check out: http://danger.systems/js/guides/the_dangerfile.html#working-on-your-dangerfile

Danger: ✓ found only warnings, not failing the build
## Warnings
Please add a changelog. See instructions [here](https://github.com/vector-im/element-android/blob/develop/CONTRIBUTING.md#changelog)
```

jmartinesp (approved):
LGTM!

fedrunov:
> Thanks, but it should be good to try to track the effects of this change. Maybe a screenshot showing the current cache size is enough.

This is good idea, but for some reason on [this](https://github.com/vector-im/element-android/actions/runs/3129173343/jobs/5077931796) run action decided to update cache, even though clean up was disabled.

It also looks like writing cache was enabled on develop even without defying `cache-read-only` at all, which is not reflected in the doc of the action.

PR #7246, opened 2022-09-28
[Crash] Autocomplete crashes on open 

ouchadam:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- re-enabling the jetifier, it's required by the autocomplete library
- the upstream library source doesn't require the jetifier however it hasn't been released  https://github.com/natario1/Autocomplete/issues/51

## Motivation and context

Fixes a crash when opening the autocomplete popup (typing a #) 

- Originally the jetifier was disable due to an incompatibility between the screenshot and unit tests, however this was resolved by switching the gradle tasks to be lazy instead of eager.

## Screenshots / GIFs

|Before|After|
|-|-|
|before-autocomplete|after-autocomplete|

## Tests

- In a room
- Type a `#`
- Notice a crash

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,7 +15,8 @@ org.gradle.vfs.watch=true
 org.gradle.caching=true
 
 # Android Settings
-android.enableJetifier=false
+android.enableJetifier=true
+android.jetifier.ignorelist=android-base-common,common
 android.useAndroidX=true
 
 #Project Settings


yostyle (approved):
LGTM

BillCarsonFr (approved):

Issue #7152, opened 2022-09-15
Leaving a room doesn't remove it from "All Chat" list with New App Layout Enabled


torrybr:
### Steps to reproduce

1. Enable new app layout underneath the labs flag
2. Go into a chatroom **yueworuewo**
3. Leave the chatroom, click okay
4. You should be automatically directed back to the homepage
5. Notice the room ( **yueworuewo** ) is not removed from the "All Chats" list.

https://user-images.githubusercontent.com/16907963/190526220-59e24b41-2ca3-4ca7-8bc0-a272c8e13373.mov

#### What did you expect?
Leaving a room should take to back to the "All Chats" page with while showing the correct rooms I am apart of.
#### What happened instead?
The room I just left is still in the list of chats.

### Your phone model

Pixel 6

### Operating system version

Android 13

### Application version and app store

1.4.36

### Will you send logs?

Yes

### Are you willing to provide a PR?

No

daniellekirkwood:
@fedrunov this doesn't feel great - do you think it's a blocker to release? I'm unsure... 

* Would the user be able to see previews of new messages since leaving? (That's probably a blocker)
* If the user was kicked/banned, does this still happen? (That's probably a blocker)

daniellekirkwood:
Discussed with @gsouquet 

- you will not receive new messages
- kick/ban and leave work the same
- this is a regression and therefore a blocker

PR #7278, opened 2022-10-03
fixing bug when room list is not updated being on background

fedrunov:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

we have a problem back when room list is not updated after navigating back to home screen after leaving room

## Motivation and context

#7152


vector/src/main/java/im/vector/app/features/home/room/list/home/HomeRoomListViewModel.kt
```diff
@@ -240,9 +240,6 @@ class HomeRoomListViewModel @AssistedInject constructor(
     }
 
     private fun observeRooms(currentFilter: HomeRoomFilter, isAZOrdering: Boolean) {
-        filteredPagedRoomSummariesLive?.livePagedList?.let { livePagedList ->
```
ganfra, line 243: We are missing the removeObserver here then

ganfra (approved):
LGTM

PR #7343, opened 2022-10-12
Gradle managed device - Draft

bmarty:
Let's see how the CI react.


bmarty:
This does not work, let's investigate later.

PR #7370, opened 2022-10-14
Device Manager Notification and Pusher Fixes

ericdecanini:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

- Pushers refresh when opening Session Overview screen
- Push toggle shows for non-pusher related sessiosn (e.g. web)
- Fixes bug where launching the app overrides any previously existing pushers for that device's session

## Motivation and context

Closes https://element-io.atlassian.net/browse/PSG-841
Closes https://element-io.atlassian.net/browse/PSG-843
Closes https://element-io.atlassian.net/browse/PSG-844

## Screenshots / GIFs

N/A

## Tests

Go through the bugs listed in the 3 tickets and see that they no longer exist

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s): Android 13

## Checklist


ericdecanini:
No changelog as fixes an unreleased bug

ericdecanini:
No changelog as this fixes unreleased issues

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
```diff
@@ -218,7 +218,7 @@ class HomeActivity :
                 fcmHelper.ensureFcmTokenIsRetrieved(
```
mnaturel, line 218: I think here we should only have the `activity` as parameter for that method. `PusherManager` can be injected in `GoogleFcmHelper` and the check to add a pusher or not could be moved to the `GoogleFcmHelper` as well. What do you think?

vector/src/main/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModel.kt
```diff
@@ -74,11 +79,16 @@ class SessionOverviewViewModel @AssistedInject constructor(
...
 
     init {
+        refreshPushers()
         observeSessionInfo(initialState.deviceId)
         observeCurrentSessionInfo()
         observePushers(initialState.deviceId)
     }
 
+    private fun refreshPushers() {
+        session.pushersService().refreshPushers()
```
mnaturel, line 89: Could you remove the `session` parameter from the `ViewModel` and replace it by using the existing `activeSessionHolder`?. As a general rule, it is safer to inject and use this holder.

vector/src/main/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModel.kt
```diff
@@ -213,10 +232,16 @@ class SessionOverviewViewModel @AssistedInject constructor(
 
     private fun handleTogglePusherAction(action: SessionOverviewAction.TogglePushNotifications) {
         viewModelScope.launch {
-            val devicePushers = awaitState().pushers.invoke()?.filter { it.deviceId == action.deviceId }
```
mnaturel, line 216: I think we should encapsulate this code inside a dedicated use case such as `TogglePushNotificationUseCase`, what do you think?

vector/src/main/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModel.kt
```diff
@@ -213,10 +232,16 @@ class SessionOverviewViewModel @AssistedInject constructor(
...
-            devicePushers?.forEach { pusher ->
+            val devicePusher = session.pushersService().getPushers().firstOrNull { it.deviceId == initialState.deviceId }
+            devicePusher?.let { pusher ->
                 session.pushersService().togglePusher(pusher, action.enabled)
             }
+
+            val newNotificationSettingsContent = mapOf("is_silenced" to !action.enabled)
+            session.accountDataService().updateUserAccountData(
```
mnaturel, line 241: I think we may check there is an existing `AccountData` event for the given device before trying to update the value. The event will be created by platforms not supporting `Pusher`: Web sessions F-Droid sessions.

vector/src/main/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModel.kt
```diff
@@ -213,10 +232,16 @@ class SessionOverviewViewModel @AssistedInject constructor(
...
     private fun handleTogglePusherAction(action: SessionOverviewAction.TogglePushNotifications) {
         viewModelScope.launch {
-            val devicePushers = awaitState().pushers.invoke()?.filter { it.deviceId == action.deviceId }
-            devicePushers?.forEach { pusher ->
+            val devicePusher = session.pushersService().getPushers().firstOrNull { it.deviceId == initialState.deviceId }
+            devicePusher?.let { pusher ->
                 session.pushersService().togglePusher(pusher, action.enabled)
             }
+
+            val newNotificationSettingsContent = mapOf("is_silenced" to !action.enabled)
```
mnaturel, line 240: Is it possible to use the `LocalNotificationSettingsContent` instead of the map with the `toContent` extension method?

vector/src/test/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModelTest.kt
```diff
@@ -477,10 +484,28 @@ class SessionOverviewViewModelTest {
...
+        val pushers = listOf(
+                aPusher(deviceId = A_SESSION_ID_1, enabled = false),
+                aPusher(deviceId = "another id", enabled = false)
+        )
+        fakeSession.pushersService().givenPushersLive(pushers)
+
+        val viewModel = createViewModel()
+        viewModel.handle(SessionOverviewAction.TogglePushNotifications(A_SESSION_ID_1, true))
+
+        fakeSession.accountDataService().verifyUpdateUserAccountDataEventSucceeds(
```
mnaturel, line 506: We should also check that the ViewState is correctly updated.

vector/src/test/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModelTest.kt
```diff
@@ -125,7 +132,7 @@ class SessionOverviewViewModelTest {
                 deviceId = A_SESSION_ID_1,
                 deviceInfo = Success(deviceFullInfo),
                 isCurrentSessionTrusted = true,
-                pushers = Loading(),
+                notificationsEnabled = true,
```
mnaturel, line 135: I am wondering how it can be `true` in the `ViewState` as it is `false` by default and I don't see any mock of the pushers or account data event for the given SessionId in several tests. It should normally fail.

vector/src/test/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModelTest.kt
```diff
@@ -477,10 +484,28 @@ class SessionOverviewViewModelTest {
...
+                aPusher(deviceId = "another id", enabled = false)
+        )
+        fakeSession.pushersService().givenPushersLive(pushers)
+
+        val viewModel = createViewModel()
+        viewModel.handle(SessionOverviewAction.TogglePushNotifications(A_SESSION_ID_1, true))
+
+        fakeSession.accountDataService().verifyUpdateUserAccountDataEventSucceeds(
+                UserAccountDataTypes.TYPE_LOCAL_NOTIFICATION_SETTINGS + A_SESSION_ID_1,
+                mapOf("is_silenced" to false),
```
mnaturel, line 508: Could you also use the `LocalNotificationSettingsContent` class instead of a map?

mnaturel (changes requested):
I have tested it and it is working well. I have added some comments mainly to improve code architecture.

vector/src/test/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModelTest.kt
ericdecanini: I'm not entirely sure but it looks like the account data event is mapping a default local notification event content even when it doesn't receive such an event. I tried fixing this but because the mapping to content is non-null it doesn't seem so straightforward. It doesn't break anything though, and I would like to revisit this separately as we are on a tight deadline

vector/src/main/java/im/vector/app/features/settings/devices/v2/overview/TogglePushNotificationUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+
+package im.vector.app.features.settings.devices.v2.overview
+
+import im.vector.app.core.di.ActiveSessionHolder
+import org.matrix.android.sdk.api.account.LocalNotificationSettingsContent
+import org.matrix.android.sdk.api.session.accountdata.UserAccountDataTypes
+import org.matrix.android.sdk.api.session.events.model.toContent
+import javax.inject.Inject
+
+class TogglePushNotificationUseCase @Inject constructor(
```
mnaturel, line 25: Should we name the use case `EnablePushNotificationUseCase` instead? Since we are passing the `enabled` state as an argument, it is not really a toggle we are doing, what do you think?

vector/src/main/java/im/vector/app/features/settings/devices/v2/overview/TogglePushNotificationUseCase.kt
```diff
@@ -0,0 +1,45 @@
...
+        private val activeSessionHolder: ActiveSessionHolder,
+) {
+
+    suspend fun execute(deviceId: String, enabled: Boolean) {
+        val session = activeSessionHolder.getActiveSession()
```
mnaturel, line 30: Using `getActiveSession` may throw an exception in case of no active session. We can either encapsulate the implementation of `execute()` method in a `runCatching` or use the `getSafeActiveSession()` method which may return `null`.

vector/src/main/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModel.kt
```diff
@@ -74,11 +78,16 @@ class SessionOverviewViewModel @AssistedInject constructor(
...
 
     init {
+        refreshPushers()
         observeSessionInfo(initialState.deviceId)
         observeCurrentSessionInfo()
         observePushers(initialState.deviceId)
     }
 
+    private fun refreshPushers() {
+        activeSessionHolder.getActiveSession().pushersService().refreshPushers()
```
mnaturel, line 88: Maybe we should use `activeSessionHolder.getSafeActiveSession()` instead of all the `activeSessionHolder.getActiveSession()` to avoid any crash?

vector/src/test/java/im/vector/app/features/settings/devices/v2/overview/SessionOverviewViewModelTest.kt
mnaturel: Okay let's refactor if needed in another PR then.

vector/src/main/java/im/vector/app/features/home/HomeActivity.kt
ericdecanini: Adding the check to `GoogleFcmHelper` was my first idea but this check involves the session and pushers service. I think we don't need to add these dependencies to `GoogleFcmHelper`. Can we also revisit this later as although I agree that we should try to reduce the number of parameters we pass, I don't want to change too much at this late stage

mnaturel: Okay, no problem let's update it later.

vector/src/main/java/im/vector/app/features/settings/devices/v2/overview/TogglePushNotificationUseCase.kt
ericdecanini: I think toggle is still more accurate here, since toggle can be enable or disable. This usecase's execution isn't always an enable as it can be disable too

vector/src/test/java/im/vector/app/features/settings/devices/v2/overview/TogglePushNotificationUseCaseTest.kt
```diff
@@ -0,0 +1,68 @@
...
+        )
+
+        togglePushNotificationUseCase.execute(sessionId, true)
+
+        activeSessionHolder.fakeSession.accountDataService().verifyUpdateUserAccountDataEventSucceeds(
+                UserAccountDataTypes.TYPE_LOCAL_NOTIFICATION_SETTINGS + sessionId,
+                LocalNotificationSettingsContent(isSilenced = false).toContent(),
+        )
+    }
+
```
ElementBot, line 67: - :no_entry_sign: Unexpected blank line(s) before "}"

mnaturel (approved):
Thanks for the updates and the fixes!

PR #7392, opened 2022-10-18
Fix message sent even if the step `ui-tests` is successful.

bmarty:
Jobs `codecov-units` and `integration-tests` do not exist anymore.

This should fix wrong message that we receive like this one:

image

-> failed even when successful

No need for a changelog here.


mnaturel (approved):
LGTM, thanks!

PR #7436, opened 2022-10-24
[Rich text editor] Implement full screen editor mode (simple approach)

jmartinesp:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

A new 'maximise' button has been added to the rich text editor. This button will toggle the full screen mode of the editor, overlaying the timeline to get more room to write and focus. To achieve this we had to use a new set of constraints in `layout/fragment_timeline_fullscreen.xml`, a new `MessageComposerViewAction` and an extra callback.

## Motivation and context

The full screen mode is a request from one of our clients.

## Screenshots / GIFs

|Before|After|
|-|-|
|Screenshot_1666715277|Screenshot_1666715217|

## Tests

- Go to settings, enable 'rich text editor' in Labs.
- Open any room, tap on the 'full screen' button.
- Test both modes with reply, edit, quote, etc.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 13, 10.

## Checklist


vector/src/main/res/layout/fragment_timeline_fullscreen.xml
```diff
@@ -0,0 +1,252 @@
...
+            android:text="@string/timeline_error_room_not_found"
+            android:textColor="?vctr_content_primary"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@id/roomNotFoundIcon" />
+
+    </androidx.constraintlayout.widget.ConstraintLayout>
+
+</androidx.constraintlayout.widget.ConstraintLayout>
```
bmarty, line 252: It's annoying to duplicate this whole file, there is no other option? Maybe add a comment on top of the two files to warn developers about the fact that they must update the 2 files.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/MessageComposerAction.kt
```diff
@@ -34,6 +34,8 @@ sealed class MessageComposerAction : VectorViewModelAction {
     data class SlashCommandConfirmed(val parsedCommand: ParsedCommand) : MessageComposerAction()
     data class InsertUserDisplayName(val userId: String) : MessageComposerAction()
 
+    object ToggleFullScreen : MessageComposerAction()
```
bmarty, line 37: `ToggleFullScreen` maybe be replaced to a `data class SetFullScreen(val fullscreen: Boolean)`, because I think in the usage there is a View to enter fullscreen and a view to exit fullscreen. Also when the message is sent, it's maybe better to use `SetFullScreen(false)` rather than `if(fullscreen) ToggleFullScreen`.

But I have not tested to implement the change, this is maybe not better, I let you choose.

vector/src/main/java/im/vector/app/features/home/room/detail/composer/RichTextComposerLayout.kt
```diff
@@ -56,12 +50,13 @@ class RichTextComposerLayout @JvmOverloads constructor(
...
     private val views: ComposerRichTextLayoutBinding
 
-    override var callback: PlainTextComposerLayout.Callback? = null
+    override var callback: Callback? = null
 
     private var currentConstraintSetId: Int = -1
-
     private val animationDuration = 100L
 
+    private var isFullScreen = false
```
bmarty, line 58: In case of configuration change, this value (if true) may be lost.
We have 2 source of truth, this is not ideal. Can we rely only on the ViewState?

bmarty (commented):
Nice work, first remarks. I will test the feature.

bmarty:
- [ ] When in fullscreen, there is an issue with the scroll to bottom button:

image

It should maybe be hidden.

vector/src/main/res/layout/fragment_timeline_fullscreen.xml
jmartinesp: We'd have to change those constraints programmatically, I used the same approach that we use for `composer_text_layout.xml`. I agree it's not great, but I think the alternative would be worse. I'll add the comment.

jmartinesp:
> There is an issue with the background color at the top. Maybe using a padding instead of a margin would fix. But this is maybe expected?

If it's about covering the appbar, that's in the designs, but it's a bit weird to be honest. I could probably remove the scrim view elevation.

> The menu item matrix app blinks as well when toggle fullscreen

I think it's because we change its alpha value (the menu is handled again) while we animate the constraintset transition and gets animated too instead of instantly applying the values. I'll try to find a way to fix it.

jmartinesp:
Sorry, there's no way to fix the blinking app icon issue without re-building this from scratch as far as I could see. We're already working on an alternative using a custom `CoordinatorLayout.Behavior` to replace this, but it's a lot more complex so it might take time.

As for the shadow covering the app bar but not the statusbar, I'd have to change the way the activity works so it draws the status bar background and play with the insets... I can try to do it, but it could take some time and this is a bit urgent.

I could also remove the shadow view altogether and cover the screen up to the appbar, but that wouldn't really match the designs.

bmarty:
> As for the shadow covering the app bar but not the statusbar, I'd have to change the way the activity works so it draws the status bar background and play with the insets... I can try to do it, but it could take some time and this is a bit urgent.

I am talking about this banner (between my ugly red brackets):

Screenshot 2022-10-26 at 14 59 27

bmarty (approved):
Thanks for the update!

PR #7521, opened 2022-11-04
Doc: Update the synapse installation section with poetry usage

manuroe:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other : Doc update

## Content

Update the synapse installation doc to set up a local synapse locally for integration tests or dev.


amitkma:
I think we should also mention that `demo/start.sh` should be modified from `echo "public_baseurl: http://localhost:$port/` to `echo "public_baseurl: http://10.0.2.2:$port/`.

Otherwise emulator doesn't connect as public_baseurl returned from api tells `localhost` instead of `10.0.2.2`

bmarty (changes requested):

PR #7611, opened 2022-11-18
Be less restritive when pruning redacted events

Florian14:
## Type of change

- [ ] Feature
- [ ] Bugfix
- [ ] Technical
- [x] Other :

## Content

- Be less restrive when handling redacted events in the sdk
- Some cleanup using `StableUnstableId` object for event types with unstable ids
- Display debug notice for voice broadcast hidden events (when hidden events are shown in the timeline) 

## Motivation and context

Before, the sdk was only pruning events from a limited list, now the sdk will prune the events which are **not** in that list.

The custom events which were unknown from the sdk are now correctly pruned.   

Could be even less restrictive like Element-iOS but I decided to keep some state events / crypto events to limitate potential regressions

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/prune/RedactionEventProcessor.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/prune/RedactionEventProcessor.kt
@@ -61,45 +61,33 @@ internal class RedactionEventProcessor @Inject constructor() : EventInsertLivePr
         val isLocalEcho = LocalEcho.isLocalEchoId(redactionEvent.eventId ?: "")
         Timber.v("Redact event for ${redactionEvent.redacts} localEcho=$isLocalEcho")
 
-        val eventToPrune = EventEntity.where(realm, eventId = redactionEvent.redacts).findFirst()
-                ?: return
+        val eventToPrune = EventEntity.where(realm, eventId = redactionEvent.redacts).findFirst() ?: return
 
         val typeToPrune = eventToPrune.type
         val stateKey = eventToPrune.stateKey
         val allowedKeys = computeAllowedKeys(typeToPrune)
-        if (allowedKeys.isNotEmpty()) {
-            val prunedContent = ContentMapper.map(eventToPrune.content)?.filterKeys { key -> allowedKeys.contains(key) }
-            eventToPrune.content = ContentMapper.map(prunedContent)
-        } else {
-            when (typeToPrune) {
-                EventType.ENCRYPTED,
-                EventType.MESSAGE,
-                in EventType.STATE_ROOM_BEACON_INFO.values,
-                in EventType.BEACON_LOCATION_DATA.values,
-                in EventType.POLL_START.values -> {
-                    Timber.d("REDACTION for message ${eventToPrune.eventId}")
-                    val unsignedData = EventMapper.map(eventToPrune).unsignedData
-                            ?: UnsignedData(null, null)
+        when {
+            allowedKeys.isNotEmpty() -> {
+                val prunedContent = ContentMapper.map(eventToPrune.content)?.filterKeys { key -> allowedKeys.contains(key) }
+                eventToPrune.content = ContentMapper.map(prunedContent)
+            }
+            canPruneEventType(typeToPrune) -> {
+                Timber.d("REDACTION for message ${eventToPrune.eventId}")
+                val unsignedData = EventMapper.map(eventToPrune).unsignedData ?: UnsignedData(null, null)
 
-                    // was this event a m.replace
+                // was this event a m.replace
 //                    val contentModel = ContentMapper.map(eventToPrune.content)?.toModel<MessageContent>()
 //                    if (RelationType.REPLACE == contentModel?.relatesTo?.type && contentModel.relatesTo?.eventId != null) {
 //                        eventRelationsAggregationUpdater.handleRedactionOfReplace(eventToPrune, contentModel.relatesTo!!.eventId!!, realm)
 //                    }
 
-                    val modified = unsignedData.copy(redactedEvent = redactionEvent)
-                    // Deleting the content of a thread message will result to delete the thread relation, however threads are now dynamic
-                    // so there is not much of a problem
-                    eventToPrune.content = ContentMapper.map(emptyMap())
-                    eventToPrune.unsignedData = MoshiProvider.providesMoshi().adapter(UnsignedData::class.java).toJson(modified)
-                    eventToPrune.decryptionResultJson = null
-                    eventToPrune.decryptionErrorCode = null
+                val modified = unsignedData.copy(redactedEvent = redactionEvent)
+                eventToPrune.content = ContentMapper.map(emptyMap())
+                eventToPrune.unsignedData = MoshiProvider.providesMoshi().adapter(UnsignedData::class.java).toJson(modified)
+                eventToPrune.decryptionResultJson = null
+                eventToPrune.decryptionErrorCode = null
 
-                    handleTimelineThreadSummaryIfNeeded(realm, eventToPrune, isLocalEcho)
-                }
-//                EventType.REACTION -> {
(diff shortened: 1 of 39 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -324,9 +325,11 @@ class MessageItemFactory @Inject constructor(
             informationData: MessageInformationData,
             highlight: Boolean,
             attributes: AbsMessageItem.Attributes
-    ): MessageVoiceItem? {
+    ): BaseEventItem<*>? {
         // Do not display voice broadcast messages
-        if (params.event.root.asMessageAudioEvent().isVoiceBroadcast()) return null
+        if (params.event.root.asMessageAudioEvent().isVoiceBroadcast()) {
+            return noticeItemFactory.create(params)
+        }
```
giomfo, line 332: @Florian14 What is the purpose of this change here?

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/VoiceBroadcastItemFactory.kt
```diff
@@ -47,16 +48,19 @@ class VoiceBroadcastItemFactory @Inject constructor(
...
             messageContent: MessageVoiceBroadcastInfoContent,
             highlight: Boolean,
             attributes: AbsMessageItem.Attributes,
-    ): AbsMessageVoiceBroadcastItem<*>? {
+    ): BaseEventItem<*>? {
         // Only display item of the initial event with updated data
-        if (messageContent.voiceBroadcastState != VoiceBroadcastState.STARTED) return null
+        if (messageContent.voiceBroadcastState != VoiceBroadcastState.STARTED) {
+            return noticeItemFactory.create(params)
+        }
```
giomfo, line 63: Same remark

giomfo (commented):
commented - SGTM

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
Florian14: Display a notice if the event is hidden and the flag "show hidden events in the timeline" enabled in the labs

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/VoiceBroadcastItemFactory.kt
Florian14: same answer

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/prune/RedactionEventProcessor.kt
```diff
@@ -61,45 +61,33 @@ internal class RedactionEventProcessor @Inject constructor() : EventInsertLivePr
...
-                EventType.MESSAGE,
-                in EventType.STATE_ROOM_BEACON_INFO,
-                in EventType.BEACON_LOCATION_DATA,
-                in EventType.POLL_START -> {
-                    Timber.d("REDACTION for message ${eventToPrune.eventId}")
-                    val unsignedData = EventMapper.map(eventToPrune).unsignedData
-                            ?: UnsignedData(null, null)
-
-                    // was this event a m.replace
+        when {
```
bmarty, line 69: Maybe add a `else` case to log a warning, which can help to detect possible issue  in the future.
Something like:

```kotlin
            else -> {
                Timber.w("Not pruning event (type $typeToPrune)")
            }
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/prune/RedactionEventProcessor.kt
```diff
@@ -167,4 +155,34 @@ internal class RedactionEventProcessor @Inject constructor() : EventInsertLivePr
...
+                    eventType == EventType.STATE_ROOM_RELATED_GROUPS ||
+                    eventType == EventType.STATE_ROOM_PINNED_EVENT ||
+                    eventType == EventType.STATE_ROOM_ENCRYPTION ||
+                    eventType == EventType.STATE_ROOM_SERVER_ACL ||
+                    eventType == EventType.ROOM_KEY_REQUEST ||
+                    eventType == EventType.FORWARDED_ROOM_KEY ||
+                    eventType in EventType.ROOM_KEY_WITHHELD.values ||
+                    eventType == EventType.REQUEST_SECRET ||
+                    eventType == EventType.SEND_SECRET ||
+                    eventType == EventType.REACTION -> false
```
bmarty, line 184: As discussed with @BillCarsonFr some event types in this list are only toDevice events, so should never be used here. The fact that all Event types are defined in `EventType` does not help.

Not blocking though.

bmarty (approved):
LGTM, thanks!

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/prune/RedactionEventProcessor.kt
Florian14: Done https://github.com/vector-im/element-android/pull/7611/commits/f913b527501f39dfd2b0e53897d7b66585bb1d63

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/prune/RedactionEventProcessor.kt
Florian14: I removed some to-device event types from the list, tell me if I am wrong https://github.com/vector-im/element-android/pull/7611/commits/57ef4fec12a4d07e6359e2fe3d8d52f09a44ace9

Florian14: Also I think the tombstone event can be redacted, in this case we should probably consider it? It can be done later as this is not a regression and common usecase.

Florian14:
@bmarty Since the sticker event is now also pruned, I observe in the log some exceptions (not crashing) when moshi tries to provide the model of the event content. The expected body should be not null but it became null after the redaction. This is also not a regression because it was already visible after deleting a sticker and doing a clear cache but it's more visible since the event is directly pruned. Wdyt about it? Should we add a default value to the body?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/prune/RedactionEventProcessor.kt
bmarty: > I removed some to-device event types from the list, tell me if I am wrong [f59be73](https://github.com/vector-im/element-android/commit/f59be7334cd66f056dd791b2ee44a046257b5f9d)

This is LGTM, thanks!

bmarty:
> @bmarty Since the sticker event is now also pruned, I observe in the log some exceptions (not crashing) when moshi tries to provide the model of the event content. The expected body should be not null but it became null after the redaction. This is also not a regression because it was already visible after deleting a sticker and doing a clear cache but it's more visible since the event is directly pruned. Wdyt about it? Should we add a default value to the body?

Yes, maybe just add a default value to avoid such logs at https://github.com/vector-im/element-android/blob/4c95aafe1037819c3ef2afd8e5e9534ebb089284/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/room/model/message/MessageStickerContent.kt#L37:

```kotlin
@Json(name = "body") override val body: String = "",
```

Thanks!

bmarty (approved):
LGTM, thanks!

PR #7627, opened 2022-11-22
saving sync filter changed

### LINKED ISSUE #7626: Make sync filter consider homeserver capabilities
Currently sync filter doesn't take in account home server capability, so we can't add params which depends on homeserver version like `unread_thread_notifications`

fedrunov:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Currently sync filter doesn't take in account home server capability, so we can't add params which depends on homeserver version like `unread_thread_notifications`, also we need to update filter when those capabilities changing and so we need to store filter parameters to be able to build filter with new capabilities and compare with stored one

## Motivation and context

closes #7626


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/sync/filter/SyncFilterParams.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.api.session.sync.filter
+
+internal data class SyncFilterParams(
```
bmarty, line 19: if it's `internal` it should be moved to the `internal` package.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/mapper/FilterParamsMapper.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    fun map(params: SyncFilterParams): SyncFilterParamsEntity {
+        return SyncFilterParamsEntity(
+                useThreadNotifications = params.useThreadNotifications,
+                lazyLoadMembersForMessageEvents = params.lazyLoadMembersForMessageEvents,
+                lazyLoadMembersForStateEvents = params.lazyLoadMembersForStateEvents,
+                listOfSupportedEventTypes = params.listOfSupportedEventTypes.toRealmList(),
+                listOfSupportedEventTypesHasSet = params.listOfSupportedEventTypes != null,
+                listOfSupportedStateEventTypes = params.listOfSupportedStateEventTypes.toRealmList(),
+                listOfSupportedStateEventTypesHasSet = params.listOfSupportedEventTypes != null,
```
bmarty, line 54: Copy paste bug here:

```suggestion
                listOfSupportedStateEventTypesHasSet = params.listOfSupportedStateEventTypes != null,
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/SyncFilterParams.kt
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * This entity stores Sync Filter configuration data, provided by the client
+ */
+internal open class SyncFilterParamsEntity(
+        var lazyLoadMembersForStateEvents: Boolean? = null,
+        var lazyLoadMembersForMessageEvents: Boolean? = null,
+        var useThreadNotifications: Boolean? = null,
+        var listOfSupportedEventTypes: RealmList<String>? = null,
+        var listOfSupportedEventTypesHasSet: Boolean = false,
```
bmarty, line 30: Maybe rename to

```suggestion
        var listOfSupportedEventTypesHasBeenSet: Boolean = false,
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/SyncFilterParams.kt
```diff
@@ -0,0 +1,36 @@
...
+        var listOfSupportedStateEventTypes: RealmList<String>? = null,
+        var listOfSupportedStateEventTypesHasSet: Boolean = false,
```
bmarty, line 32: Maybe rename to

```suggestion
        var listOfSupportedStateEventTypesHasBeenSet: Boolean = false,
```

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/DefaultFilterRepository.kt
```diff
@@ -17,74 +17,71 @@
...
         return monarchy.awaitTransaction {
             FilterEntity.getOrCreate(it).roomEventFilterJson
         }
     }
+
+    override suspend fun getStoredFilterParams(): SyncFilterParams? {
+        return monarchy.awaitTransaction { realm ->
+            realm.where<SyncFilterParamsEntity>().findFirst()?.let {
+                val params = FilterParamsMapper().map(it)
+                 params
```
bmarty, line 76: maybe simplify those two lines to

`FilterParamsMapper().map(it)`

(will also remove the indentation issue)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/DefaultFilterRepository.kt
```diff
@@ -17,74 +17,71 @@
...
+    override suspend fun getRoomFilterBody(): String {
         return monarchy.awaitTransaction {
             FilterEntity.getOrCreate(it).roomEventFilterJson
         }
     }
+
+    override suspend fun getStoredFilterParams(): SyncFilterParams? {
+        return monarchy.awaitTransaction { realm ->
+            realm.where<SyncFilterParamsEntity>().findFirst()?.let {
+                val params = FilterParamsMapper().map(it)
```
bmarty, line 75: `FilterParamsMapper` could be injected in the constructor.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/DefaultFilterRepository.kt
```diff
@@ -17,74 +17,71 @@
...
+                 params
+            }
+        }
+    }
+
+    override suspend fun storeFilterParams(params: SyncFilterParams) {
+        return monarchy.awaitTransaction { realm ->
+            val entity = FilterParamsMapper().map(params)
```
bmarty, line 83: Use injected `FilterParamsMapper`

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/DefaultFilterService.kt
```diff
@@ -16,20 +16,36 @@
...
+        private val filterRepository: FilterRepository,
+        private val homeServerCapabilitiesDataSource: HomeServerCapabilitiesDataSource,
 ) : FilterService {
 
     // TODO Pass a list of support events instead
-    override fun setFilter(filterPreset: FilterService.FilterPreset) {
-        saveFilterTask
-                .configureWith(SaveFilterTask.Params(filterPreset))
-                .executeBy(taskExecutor)
+    override fun setSyncFilter(filterBuilder: SyncFilterBuilder) {
```
bmarty, line 35: Maybe take the opportunity to make this fun suspendable?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/SaveFilterTask.kt
```diff
@@ -26,44 +25,32 @@ import javax.inject.Inject
...
+        val filter = params.filter
+        val filterResponse = executeRequest(globalErrorReceiver) {
+            // TODO auto retry
+            filterAPI.uploadFilter(userId, filter)
         }
+
+        filterRepository.storeSyncFilter(
+                filter = filter,
+                filterId = filterResponse.filterId,
+                roomEventFilter = FilterFactory.createDefaultRoomFilter()
```
bmarty, line 52: It's weird to call `createDefaultRoomFilter`. Can you explain please?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/SyncFilterBuilder.kt
```diff
@@ -0,0 +1,126 @@
...
+        return SyncFilterParams(
+                useThreadNotifications = useThreadNotifications,
+                lazyLoadMembersForMessageEvents = lazyLoadMembersForMessageEvents,
+                lazyLoadMembersForStateEvents = lazyLoadMembersForStateEvents,
+                listOfSupportedEventTypes = listOfSupportedEventTypes,
+                listOfSupportedStateEventTypes = listOfSupportedStateEventTypes,
+        )
+    }
+
+    internal fun build(homeServerCapabilities: HomeServerCapabilities): Filter {
```
bmarty, line 64: I understand the logic, but from an app developer, it's weird to provide a builder without calling build().
So maybe refactor a bit and create a `SyncFilterParam` with a `SyncFilterParamBuilder` in the API.

then we can have an internal `SyncFilterBuilder` which takes a `SyncFilterParam` and build with `HomeServerCapabilities`, retrieved using an injected `HomeServerCapabilitiesDataSource` in the constructor.

WDYT?

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/sync/DefaultGetCurrentFilterTask.kt
```diff
@@ -0,0 +1,100 @@
...
+private const val A_FILTER_ID = "filter-id"
+private val A_HOMESERVER_CAPABILITIES = HomeServerCapabilities()
+private val A_SYNC_FILTER_PARAMS = SyncFilterParams(
+        lazyLoadMembersForMessageEvents = true,
+        lazyLoadMembersForStateEvents = true,
+        useThreadNotifications = true
+)
+
+@ExperimentalCoroutinesApi
+class DefaultGetCurrentFilterTaskTest {
```
bmarty, line 40: Please rename the file to match this class name.

bmarty (changes requested):
Nice work, thanks! A few remarks, also there are some errors reported by detekt. You can run locally by calling `./gradlew detekt`.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/SyncFilterBuilder.kt
fedrunov: I also had doubts making it internal, but then I decided that introducing extra entities could make it more difficult to understand the flow and maintain it in the future. I don't have any strong opinion on that though, more like personal preference, so if you think it will be better to separate to two builders - I'll do the change

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/SaveFilterTask.kt
fedrunov: There is currently no difference between `ElementRoomFilter` and `DefaultRoomFilter` so basically there is no need to introduce any options here. It's also looks like we could utilise Filter.room.timeline filter for the purpose of `roomEventFilter`, which will mean that we don't need this separate filter at all. But currently no one can say for sure if we could/should do so or no, so for now I decided to do as little change to this area as possible, since the task was mostly about SyncFilter. I just wanted to remove referencing Element from the sdk, since there was no real difference between Default and Element room filters

matrix-sdk-android/src/test/java/org/matrix/android/sdk/internal/sync/DefaultGetCurrentFilterTaskTest.kt
```diff
@@ -0,0 +1,100 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.matrix.android.sdk.internal.sync
+
+import kotlinx.coroutines.ExperimentalCoroutinesApi
```
ElementBot, line 19: - :no_entry_sign: Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/filter/DefaultFilterService.kt
```diff
@@ -17,19 +17,32 @@
...
+
+        // don't upload/store filter until homeserver capabilities are fetched
+        homeServerCapabilitiesDataSource.getHomeServerCapabilities()?.let { homeServerCapabilities ->
+            saveFilterTask
+                    .configureWith(
+                            SaveFilterTask.Params(
+                                    filter = filterBuilder.build(homeServerCapabilities)
+                            )
+                    )
+                    .executeBy(taskExecutor)
```
bmarty, line 45: Since you are in a suspend fun, I think you do not need to use the taskExecutor (it can be removed from the constructor).
Just call
```kotlin
saveFilterTask.execute(
    SaveFilterTask.Params(
        filter = filterBuilder.build(homeServerCapabilities)
    )
)
```

Should be fine

bmarty (approved):
Thanks for the update. One last remark.

bmarty (approved):

Issue #7127, opened 2022-09-14
Add the Voice Broadcast option in the room functionalities


giomfo:
#### What/Why would you like to do?
As a user I want to be able to send a voice broadcast to all users of a room, so that I can easily provide information by just talking. Compared to already existing voice messages it should be possible to listen to voice broadcast at the time they are being recorded. Think of this being a live-podcast over Matrix.

Some use-case scenarios for this feature can be to verbally provide updates, trainings, or talk to a large number of people.

#### How would you like to achieve it?
With the Voice Broadcast function, it should be possible to share longer audio messages in rooms with a short time delay.

- [x] Add a feature flag to implement this feature. This flag may be used later to enable it via the labs section.
- [ ] Add in the room permission the voice broadcast option
- [x] Add in the build settings or config file:
   - a max voice broadcast length  (240 minutes)
   - a voice broadcast chunk length (TBD)
- [x] Add “Voice broadcast“ to the room composer options
- [x] Make the “Voice broadcast“  option available only if the end user is allowed to use it
- [x] There can be a maximum of one simultaneous audio stream per room.
- [x] The sending client generates individual voice messages of x minutes (configurable via Build Setting), which are sent almost immediately.
- [x] ~~In case of bad network connections, the sending client saves messages locally and automatically sends them when the connection is restored.~~ In case of an interruption of the network during a voice broadcasting, the sending client pauses the broadcast until the connection is restored. The end user is then able to resume the VB or stop it.
- [ ] When activating the recording, the user would have a PIP recorder widget (or a bubble widget) to control the record (stop/pause, audio input device selection…)
- [x] The sender gets a countdown of the remaining recording time is displayed.
- [x] Recording can be paused and resumed at any time. During the paused recording, the app can be exited or other functions in the app can be used. The recording can be continued afterwards.
- [x] During the recording, the other clients are using the same User profile still functional
- [x] On the receiver side you only see a widget in the timeline, which dynamically reloads incoming audio chunks and plays them without any user action without interruption (for the listener).
- [x] There are buttons for play, pause, +/- 30 seconds
- [x] There is a progress slider that can be used to "scroll".
- [x] The volume of the system is to be applied so that the user can also control it with the physical buttons
- [x] The paused position during playback must be preserved in the client, to continue playback at a later point in time, even after reopening the client.
- [x] The live stream must always be visible to the user as long as it is active.
- [x] There can be chat messages in the room of others parallel to this stream be sent.
[...]

PR #7629, opened 2022-11-23
Voice Broadcast - Handle event deletion when listening or recording

Florian14:
## Type of change

- [x] WIP Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Handle the deletion of a voice broadcast state event on the recorder and the listener sides.
- Improve the voice broadcast state observation by handling the event redaction.
- Stop listening to this VB as soon as the `started` event is redacted
- Stop recording this VB  as soon as the `started` event is redacted
- The recorder now reacts to the state event changes (previously, we sent the state event and did the action on the recorder at the same time)

## Motivation and context

Continue #7127 

## Tests

- Show hidden events in the timeline (advanced settings -> developer mode)
- Record a VB with several state events (started/paused/resumed/stopped) and verify that the tile in the timeline is correctly updated on listener and recorder sides
- Redact some state events and verify that the tile is always displayed according to the most recent not redacted state event
- When redacting the started state event (behind the tile), the tile should be removed

Test again during recording and listening to the VB, delete the started event should stop the action.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s):

## Checklist


vector/src/main/java/im/vector/app/features/voicebroadcast/recording/VoiceBroadcastRecorderQ.kt
```diff
@@ -30,10 +39,17 @@ import java.util.concurrent.TimeUnit
 @RequiresApi(Build.VERSION_CODES.Q)
 class VoiceBroadcastRecorderQ(
         context: Context,
+        private val sessionHolder: ActiveSessionHolder,
+        private val getVoiceBroadcastEventUseCase: GetMostRecentVoiceBroadcastStateEventUseCase
 ) : AbstractVoiceRecorderQ(context), VoiceBroadcastRecorder {
 
+    private val session get() = sessionHolder.getActiveSession()
```
mnaturel, line 46: Have you considered it may throw an exception when getting the session variable?

vector/src/main/java/im/vector/app/features/voicebroadcast/recording/usecase/ResumeVoiceBroadcastUseCase.kt
```diff
@@ -31,8 +30,7 @@ import timber.log.Timber
 import javax.inject.Inject
 
 class ResumeVoiceBroadcastUseCase @Inject constructor(
-        private val session: Session,
-        private val voiceBroadcastRecorder: VoiceBroadcastRecorder?,
+        private val session: Session
```
mnaturel, line 33: Should we keep the trailing comma?

vector/src/main/java/im/vector/app/features/voicebroadcast/recording/VoiceBroadcastRecorder.kt
```diff
@@ -31,7 +32,7 @@ interface VoiceBroadcastRecorder : VoiceRecorder {
     /** Current remaining time of recording, in seconds, if any. */
     val currentRemainingTime: Long?
 
-    fun startRecord(roomId: String, chunkLength: Int, maxLength: Int)
+    fun startRecordVoiceBroadcast(voiceBroadcast: VoiceBroadcast, chunkLength: Int, maxLength: Int)
```
mnaturel, line 35: Do you think we need to add `VoiceBroadcast` in the name of the method? I personnally think this is enough with `startRecord` since first parameter is a `VoiceBroadcast`.

Florian14: My problem is that I want to distinguish this method from the startRecord which is inherited from VoiceRecorder to avoid confusion. Ideally, I would like to only expose this one to the outside and change the visibility of the other method to "protected", but afaik I can do it programmatically but this is a bit ugly imo. Wdyt @mnaturel?

vector/src/main/java/im/vector/app/features/voicebroadcast/recording/usecase/StartVoiceBroadcastUseCase.kt
```diff
@@ -79,13 +84,15 @@ class StartVoiceBroadcastUseCase @Inject constructor(
                 ).toContent()
         )
 
-        startRecording(room, eventId, chunkLength, maxLength)
+        val voiceBroadcast = VoiceBroadcast(roomId = room.roomId, voiceBroadcastId = eventId)
+        room.flow().liveTimelineEvent(eventId).unwrap().first() // wait for the event come back from the sync
```
mnaturel, line 88: I think we need to update the related unit tests to take this change into account.

vector/src/main/java/im/vector/app/features/voicebroadcast/usecase/GetMostRecentVoiceBroadcastStateEventUseCase.kt
```diff
@@ -0,0 +1,163 @@
...
+import org.matrix.android.sdk.api.session.getRoom
+import org.matrix.android.sdk.api.session.room.Room
+import org.matrix.android.sdk.api.util.Optional
+import org.matrix.android.sdk.api.util.toOptional
+import org.matrix.android.sdk.flow.flow
+import org.matrix.android.sdk.flow.mapOptional
+import timber.log.Timber
+import javax.inject.Inject
+
+class GetMostRecentVoiceBroadcastStateEventUseCase @Inject constructor(
```
mnaturel, line 47: I think this use case needs some unit tests.

vector/src/main/java/im/vector/app/features/voicebroadcast/usecase/GetMostRecentVoiceBroadcastStateEventUseCase.kt
```diff
@@ -0,0 +1,163 @@
...
+        // observe started event changes
+        return startedEventFlow
+                .mapOptional { it.root.asVoiceBroadcastEvent() }
+                .flatMapLatest { startedEvent ->
+                    if (startedEvent.hasValue().not() || startedEvent.get().root.isRedacted()) {
+                        // if started event is null or redacted, send null
+                        flowOf(Optional.empty())
+                    } else {
+                        // otherwise, observe most recent event changes
+                        getMostRecentRelatedEventFlow(room, voiceBroadcast)
```
mnaturel, line 77: Have you think about this way of writing this logic?
```
getMostRecentRelatedEventFlow(room, voiceBroadcast)
                                .transformWhile { mostRecentEvent ->
                                    val hasValue = mostRecentEvent.hasValue()
                                    if(hasValue) {
                                        // keep the most recent event
                                        emit(mostRecentEvent)
                                    } else {
                                        // no most recent event, fallback to started event
                                        emit(startedEvent)
                                    }
                                    hasValue
                                }
```

vector/src/main/java/im/vector/app/features/voicebroadcast/usecase/GetMostRecentVoiceBroadcastStateEventUseCase.kt
```diff
@@ -0,0 +1,163 @@
...
+                    .liveStateEvent(VoiceBroadcastConstants.STATE_ROOM_VOICE_BROADCAST_INFO, QueryStringValue.Equals(stateKey))
+                    .mapOptional { it.asVoiceBroadcastEvent() }
+                    // drop first event sent by the matrix-sdk, we compute manually this first event
+                    .drop(1)
+                    // start with the computed most recent event
+                    .onStart { emit(mostRecentEvent) }
+                    // handle event if null or related to the given voice broadcast
+                    .filter { it.hasValue().not() || it.get().voiceBroadcastId == voiceBroadcast.voiceBroadcastId }
+                    // observe changes while event is not null
+                    .transformWhile { event ->
```
mnaturel, line 114: Have you tried this shorter notation?
`.takeWhile { it.hasValue() }`

vector/src/main/java/im/vector/app/features/voicebroadcast/usecase/GetMostRecentVoiceBroadcastStateEventUseCase.kt
```diff
@@ -0,0 +1,163 @@
...
+                        if (newMostRecentEvent.hasValue()) {
+                            // observe most recent event changes
+                            newMostRecentEvent.get().flow()
+                                    .transformWhile { event ->
+                                        // observe changes until event is null or redacted
+                                        emit(event)
+                                        event.hasValue() && event.get().root.isRedacted().not()
+                                    }
+                                    .flatMapLatest { event ->
+                                        if (event.getOrNull()?.root?.isRedacted().orFalse()) {
```
mnaturel, line 128: What was your intention when event is `null`? The comment says it should be in this `if` branch but using `orFalse` means it  will go through the `else` branch.

vector/src/main/java/im/vector/app/features/voicebroadcast/recording/VoiceBroadcastRecorder.kt
mnaturel: I see. I didn't see the method in the `VoiceRecorder` Interface. Let's keep it like this then.

mnaturel (commented):
I have added some small comments. 
Just a general thought about redaction of voice broadcast events. For the default user, when pressing the "Remove" button on a voice broadcast, it seems it only redacts the started event. It is intended? I expected it redacts every associated events of the given voice broadcast (message events + state events).

vector/src/main/java/im/vector/app/features/voicebroadcast/usecase/GetMostRecentVoiceBroadcastStateEventUseCase.kt
Florian14: I definitely prefer this format, updated https://github.com/vector-im/element-android/pull/7629/commits/aa53105f1713b237b06c059329175cc400784d5e

mnaturel (approved):
LGTM. Thanks for the updates!
I have just a general question: I have seen when removing from the tile of a voice broadcast, it actually doesn't remove all the related events like messages or related state events to the original event. As a user, I would expect it redacts every related events to the selected voice broadcast. Can you confirm it will be implemented this way in the future?

Florian14:
> LGTM. Thanks for the updates! I have just a general question: I have seen when removing from the tile of a voice broadcast, it actually doesn't remove all the related events like messages or related state events to the original event. As a user, I would expect it redacts every related events to the selected voice broadcast. Can you confirm it will be implemented this way in the future?

Sorry, I forgot to answer this question, indeed it is not done for the moment but we'll implement the proper redaction of all the related events in a dedicated PR by using this [MSC](https://github.com/matrix-org/matrix-spec-proposals/pull/3912)

PR #7474, opened 2022-10-27
added read receipts for threads

fedrunov:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

with [MSC3771](https://github.com/matrix-org/matrix-spec-proposals/pull/3771) we have a possibility to display read receipts for messages in a thread

## Motivation and context

https://github.com/vector-im/element-android/issues/6996

## Screenshots / GIFs

output

PR #7639, opened 2022-11-25
Revert "added read receipts for threads"

fedrunov:
Reverts vector-im/element-android#7474 due to missing migration


fedrunov:
Since fix is relatively fast it was decided to just do hotfix PR

PR #7641, opened 2022-11-25
ignore push for a thread if it's currently visible to user

### LINKED ISSUE #7634: Do not show push notification for thread, when thread timeline is observed by user
Currently we don't show push notifications for room which is observed by user, but we don't take in account threads. We should instead only hide notifications for a timeline which is observed (main or thread)

fedrunov:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

we don't show push notification for a room which is currently observed by the user, but we should also take in account which thread timeline user observes

## Motivation and context

closes #7634

### DIFF
--- a/vector/src/main/java/im/vector/app/core/pushers/VectorPushHandler.kt
+++ b/vector/src/main/java/im/vector/app/core/pushers/VectorPushHandler.kt
@@ -28,6 +28,7 @@ import im.vector.app.core.network.WifiDetector
 import im.vector.app.core.pushers.model.PushData
 import im.vector.app.core.resources.BuildMeta
 import im.vector.app.features.notifications.NotifiableEventResolver
+import im.vector.app.features.notifications.NotifiableMessageEvent
 import im.vector.app.features.notifications.NotificationActionIds
 import im.vector.app.features.notifications.NotificationDrawerManager
 import im.vector.app.features.settings.VectorDataStore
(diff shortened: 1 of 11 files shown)


vector/src/main/java/im/vector/app/core/pushers/VectorPushHandler.kt
```diff
@@ -142,11 +143,6 @@ class VectorPushHandler @Inject constructor(
         pushData.roomId ?: return
         pushData.eventId ?: return
 
-        // If the room is currently displayed, we will not show a notification, so no need to get the Event faster
-        if (notificationDrawerManager.shouldIgnoreMessageEventInRoom(pushData.roomId)) {
```
bmarty, line 146: I guess we do not have the threadId in the `pushData`, true?

vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
```diff
@@ -123,6 +124,22 @@ class NotificationDrawerManager @Inject constructor(
         }
     }
 
+    /**
+     * Should be called when the application is currently opened and showing timeline for the given threadId.
+     * Used to ignore events related to that thread (no need to display notification) and clean any existing notification on this room.
+     */
+    fun setCurrentThread(threadId: String?) {
```
bmarty, line 131: Maybe merge with the method `setCurrentRoom`, just adding the parameter `threadId: String?` to it?

bmarty (approved):
LGTM, thanks!
Just a few remarks, but no blockers.

vector/src/main/java/im/vector/app/core/pushers/VectorPushHandler.kt
fedrunov: We don't :(

vector/src/main/java/im/vector/app/features/notifications/NotificationDrawerManager.kt
bmarty: OK

PR #7646, opened 2022-11-25
Voice Broadcast - Fix playback stuck in buffering

Florian14:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Improvement related to the voice broadcast playback, fix issues with buffering state
Also fixed an issue when the live broadcast is ended, the playback was not automatically reset after listening to the last chunk.

## Motivation and context

Sometimes, the (live) playback was stuck in a buffering state after a pause/resume from the recorder or when receiving new chunks after reaching the current duration

## Tests

Verify that the playback is not stuck in buffering when starting listening to a live voice broadcast (from the beggining & after reaching the end)
The playback should be automatically reset when the listening is over and the live VB has ended.

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
+++ b/vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
@@ -376,6 +376,11 @@ class VoiceBroadcastPlayerImpl @Inject constructor(
             // Notify live mode change to all the listeners attached to the current voice broadcast id
             listeners[voiceBroadcastId]?.forEach { listener -> listener.onLiveModeChanged(isLiveListening) }
         }
+
+        // Live has ended and last chunk has been reached, we can stop the playback
+        if (!isLiveListening && playingState == State.BUFFERING && playlist.currentSequence == mostRecentVoiceBroadcastEvent?.content?.lastChunkSequence) {
+            stop()
+        }
     }
 
     private fun onNextMediaPlayerStarted(mp: MediaPlayer) {
--- a/vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
+++ b/vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
@@ -72,7 +72,7 @@ class VoiceBroadcastPlayerImpl @Inject constructor(
         @MainThread
         set(value) {
             if (field != value) {
-                Timber.w("## Voice Broadcast | isLiveListening: $field -> $value")
+                Timber.d("## Voice Broadcast | isLiveListening: $field -> $value")
                 field = value
                 onLiveListeningChanged(value)
             }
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
```diff
@@ -73,7 +72,7 @@ class VoiceBroadcastPlayerImpl @Inject constructor(
         @MainThread
         set(value) {
             if (field != value) {
-                Timber.w("isLiveListening: $field -> $value")
+                Timber.w("## Voice Broadcast | isLiveListening: $field -> $value")
```
onurays, line 75: This will produce many logs, right? Can we avoid if not necessary?

vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
```diff
@@ -373,6 +376,19 @@ class VoiceBroadcastPlayerImpl @Inject constructor(
             // Notify live mode change to all the listeners attached to the current voice broadcast id
             listeners[voiceBroadcastId]?.forEach { listener -> listener.onLiveModeChanged(isLiveListening) }
         }
+
+        // Live has ended and last chunk has been reached, we can stop the playback
+        if (!isLiveListening && playingState == State.BUFFERING && playlist.currentSequence == mostRecentVoiceBroadcastEvent?.content?.lastChunkSequence) {
```
onurays, line 381: To avoid possible bugs, we should encapsulate these logic in state data classes and create use case classes with unit test.

onurays (approved):
These changes are LGTM but this class has heavy logics. We need to separate them into use case class and add some unit tests.

vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
Florian14: Not so many, it will only log if the user is listening to a voice broadcast.
Then, if the VB is ended, it will log once (or not if the default value does not change). If the VB is live, it will log if the user scrolls in the playback or if the recorder pause/resume the recording, but in a quite acceptable frequency.
Btw, it doesn't need to warn, I changed it to debug

Same for the log about `playingState`, It can be a little more frequent but also only when listening to a VB. The frequency is still acceptable imo, and the log is useful for debugging.
Also changed to debug level

vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
Florian14: I agree with you but I am not sure to follow exactly what you have in mind about the use cases. There are a lot of combinations with several parameters to take into account, For the moment, I don't know how many combinations there are and if I would be able to enumerate all of them.

I already refactored this class but I confirm it is still not perfect, it will require new refactoring and unit tests to prevent regression as every change can be dangerous. I created an [issue](https://github.com/vector-im/element-android/issues/7654) to track the missing tests on this part. I'll try to improve this mechanism by introducing a state machine at this moment.

PR #7691, opened 2022-12-02
Rich text editor: improve performance when changing composer mode

jmartinesp:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Store the current `MessageComposerMode` so we can diff and avoid unnecessary updates. 

Also, only change some layout params for the editor if we don't want to animate the full screen change (keeps compatibility with both Element Android and Element X).

## Motivation and context

Element X stopped properly responding to input as the `renderComposerMode` method was being triggered for each change (on Element Android this is not so noticeable). It also had some issues with resizing the editor properly.

As both apps share most of this code, it's better to also handle it properly here.

## Tests

It's quite difficult to test, as it wasn't a noticeable issue in Element Android.

- Enable the Rich Text Editor in Labs.
- Open a room.
- Select a message and edit or quote it, or reply to it.
- Exit the editing/replying/quoting mode.

There should be no issues when writing or 'flashes'.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): 13

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/composer/RichTextComposerLayout.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/composer/RichTextComposerLayout.kt
@@ -66,6 +66,7 @@ internal class RichTextComposerLayout @JvmOverloads constructor(
     // There is no need to persist these values since they're always updated by the parent fragment
     private var isFullScreen = false
     private var hasRelatedMessage = false
+    private var composerMode: MessageComposerMode? = null
 
     var isTextFormattingEnabled = true
         set(value) {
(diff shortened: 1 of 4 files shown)


bmarty (approved):
LGTM!

vector/src/main/java/im/vector/app/features/home/room/detail/composer/RichTextComposerLayout.kt
```diff
@@ -363,6 +370,9 @@ class RichTextComposerLayout @JvmOverloads constructor(
     }
 
     override fun renderComposerMode(mode: MessageComposerMode) {
+        if (this.composerMode == mode) return
```
bmarty, line 373: It's worth noting that `MessageComposerMode.Quote` and `MessageComposerMode.Reply` are not data class, so maybe `==` will not work as expected here.

When replying to someone, then (without sending the message), reply to someone else, does the original sender name updates properly?

jmartinesp: > It's worth noting that MessageComposerMode.Quote and MessageComposerMode.Reply are not data class, so maybe == will not work as expected here.

I just changed it so they're data classes too (I don't really remember why they weren't in the first place).

> When replying to someone, then (without sending the message), reply to someone else, does the original sender name updates properly?

Yes, both before and after making `Reply` a data class.

bmarty (approved):
LGTM!

PR #7768, opened 2022-12-12
Don't dismiss bottom sheet if verification is required

### LINKED ISSUE #7735: Check verification required behaviour in ElementR
Check the behaviour of verification in cases when it is required by server.

amitkma:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content
When verification is required by the server, self verification bottom sheet should not be dismissable.

## Motivation and context

[Closes #7735](https://github.com/vector-im/element-android/issues/7735)

## Tests

- Set `isVerificationRequired` to true in `SelfVerificationViewModel`.
- Try to do self verification, bottom-sheet should not be dismissable.

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): API 11 and 12

## Checklist


vector/src/main/java/im/vector/app/features/crypto/verification/self/SelfVerificationViewModel.kt
```diff
@@ -204,8 +204,18 @@ class SelfVerificationViewModel @AssistedInject constructor(
                 }
             }
             is VerificationAction.GotItConclusion -> {
-                // just dismiss
-                _viewEvents.post(VerificationBottomSheetViewEvents.Dismiss)
+                withState { state ->
```
BillCarsonFr, line 207: I think there is more to do for that use case.
You would have to intercept back (swipe) navigation events. For a bottom sheet probably by overriding onCreateDialog() and add a KeyListener (should be in the old VerificationBottomSheet)

Then when intercept you delegate to view model (queryCancel()) that depending on the state would do different things.
E.g if there is an outgoing request you might want to prompt for cancel, eventually cancel and go back to main screen where you can use 4S/Reset etc..

BillCarsonFr (changes requested):

PR #7806, opened 2022-12-16
[Voice Broadcast] Update live indicator color

yostyle:


giomfo (approved):
Works like a charm
Test ok

PR #7827, opened 2022-12-20
Hotfix/1.5.14

bmarty:
WIP - just to trigger the CI.


Tycho-S:
This hotfix seems to have caused a weird edge case where the activity thread crashes and never restarts until you force kill the app.

Is there any way this could be looked at? It's really annoying because I can't rely on the app anymore. I've reverted to 1.5.13 which still works great.

The discussion of the issue is here: https://github.com/vector-im/element-android/issues/7876

Please let me know if you need any info, I can test etc if it's being looked at. For now I've reverted but i can install the latest again if any logs or tests are required :)

PR #7843, opened 2022-12-23
sync filter params moved to matrix config

fedrunov:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

currently sync filter params are stored in session realm which can be cleared. This leads to incorrect filter being used (e.g. without lazyLoadEvents param).

We need to move params to somewhere else, so they won't be cleared 

## Tests

- Step 1
- Step 2
- Step ...

## Tested devices

- [ ] Physical
- [ ] Emulator
- OS version(s):

## Checklist


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/SessionRealmModule.kt
```diff
@@ -72,7 +72,6 @@ import org.matrix.android.sdk.internal.database.model.threads.ThreadSummaryEntit
             SpaceParentSummaryEntity::class,
             UserPresenceEntity::class,
             ThreadSummaryEntity::class,
-            SyncFilterParamsEntity::class,
```
bmarty, line 75: Do we need a database migration because of this change?

bmarty (approved):
LGTM, thanks, just a question about the database migration.
Also did you test to clear cache and see what was the sync filter applied to the sync request?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/SessionRealmModule.kt
fedrunov: I've tested upgrading develop build with this branch and there were no problems

fedrunov:
> Also did you test to clear cache and see what was the sync filter applied to the sync request?

Yes, I've tested enabling and disabling threads which clears session realm and also tested login with different account. All works fine

BillCarsonFr:
After a clear cache homeServerCapabilities will be null. That means that the initial sync will have a filter without useThreadNotifications. Is this expected?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/SessionRealmModule.kt
BillCarsonFr: I added the migration to cleanup the database from the entity added in 45

PR #7851, opened 2022-12-27
[Poll] Warning message on decryption failure of some events (PSG-1025)

### LINKED ISSUE #7824: [Poll] Warning message on decryption failure of some events
When at least one related event (`m.reference` relation type) to a given poll cannot be decrypted, the associated message contain a warning message saying the result may be inaccurate.

A poll message is an aggregation of multiple events. It means when some events fail to be decrypted, we may end up with inaccurate results.

The message content should be “Due to decryption errors, some votes may not be counted”

_No response_

### Additional context

_No response_

Yes

mnaturel:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Making the encrypted events as processable. Processing the events with `relatesTo` field. When an event relates to an existing poll, it is saved into the related summary entity so that a warning can be displayed to the user.

## Motivation and context

Closes #7824 

## Tests

The decryption failure is rather hard to test. The UI has been tested by mocking decryption failure.
We should ensure poll is working as before.

- Create open or closed poll in a room
- Vote on the poll
- Check the UI for the message is correct
- End the poll
- Check the UI for the message is correct

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s): Android 11

## Checklist


matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/aggregation/poll/DefaultPollAggregationProcessor.kt
```diff
@@ -226,4 +230,10 @@ internal class DefaultPollAggregationProcessor @Inject constructor(
             fetchPollResponseEventsTask.execute(params)
         }
     }
+
+    private fun removeEncryptedRelatedEventIdIfNeeded(aggregatedPollSummaryEntity: PollResponseAggregatedSummaryEntity, eventId: String) {
+        if(aggregatedPollSummaryEntity.encryptedRelatedEventIds.contains(eventId)) {
```
ElementBot, line 235: - :no_entry_sign: Missing spacing after "if"

BillCarsonFr (changes requested):
I think i'd prefer to modify the exisiting EventInsertEntity and EventInsertLiveObserver to handle the UTD. And pass them once to EventInsertLiveProcessor with a new method processUnableToDecryptEvent()

mnaturel:
> I think i'd prefer to modify the exisiting EventInsertEntity and EventInsertLiveObserver to handle the UTD. And pass them once to EventInsertLiveProcessor with a new method processUnableToDecryptEvent()

In fact, I retried to make it work with existing `EventInsertEntity` but I see problems depending on the solution:
- setting `canBeProcessed` field to `true` for encrypted events may lead to multiple processing of the same events. When the event is finally decrypted we need to [create or update the entity](https://github.com/vector-im/element-android/blob/5c442296efa532bd48a42b3ae80b7dbb6d517b9f/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/EventEntity.kt#L98) but we cannot know whether the entity is currently beeing processed due to access of the database at the same time...
- when adding another observer using a filter on `canBeProcessed` to `false`, we may end up processing multiple times the same encrypted event as in that case we will not delete the entity after process since it is still needed for the other existing observer on `canBeProcessed` to `true`

These are the reasons I introduced a new `Entity` with a separate observer. Do you have an idea of a way to use `EventInsertEntity`?

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/EventEntityQueries.kt
```diff
@@ -32,10 +31,9 @@ internal fun EventEntity.copyToRealmOrIgnore(realm: Realm, insertType: EventInse
             .equalTo(EventEntityFields.ROOM_ID, roomId)
             .findFirst()
     return if (eventEntity == null) {
-        val canBeProcessed = type != EventType.ENCRYPTED || decryptionResultJson != null
```
mnaturel, line 35: Is it possible to have a case where the type is not  `EventType.ENCRYPTED` and `decryptionResultJson` is set to `null`? If this case exists, I wonder if we should make it processable?

mnaturel:
> > I think i'd prefer to modify the exisiting EventInsertEntity and EventInsertLiveObserver to handle the UTD. And pass them once to EventInsertLiveProcessor with a new method processUnableToDecryptEvent()
> 
> In fact, I retried to make it work with existing `EventInsertEntity` but I see problems depending on the solution:
> 
>     * setting `canBeProcessed` field to `true` for encrypted events may lead to multiple processing of the same events. When the event is finally decrypted we need to [create or update the entity](https://github.com/vector-im/element-android/blob/5c442296efa532bd48a42b3ae80b7dbb6d517b9f/matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/model/EventEntity.kt#L98) but we cannot know whether the entity is currently beeing processed due to access of the database at the same time...
> 
>     * when adding another observer using a filter on `canBeProcessed` to `false`, we may end up processing multiple times the same encrypted event as in that case we will not delete the entity after process since it is still needed for the other existing observer on `canBeProcessed` to `true`
> 
> 
> These are the reasons I introduced a new `Entity` with a separate observer. Do you have an idea of a way to use `EventInsertEntity`?

I have reverted to use `EventInsertEntity` in [18ace4b](https://github.com/vector-im/element-android/pull/7851/commits/18ace4beb29d6a4e6d29807e22988722ab5392d1). Due to usage of a Realm transaction in the observer, it seems we won't have any race conditions issues.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
```diff
@@ -73,6 +74,7 @@ internal class EventRelationsAggregationProcessor @Inject constructor(
         private val sessionManager: SessionManager,
         private val liveLocationAggregationProcessor: LiveLocationAggregationProcessor,
         private val pollAggregationProcessor: PollAggregationProcessor,
+        private val encryptedReferenceAggregationProcessor: EncryptedReferenceAggregationProcessor,
         private val editValidator: EventEditValidator,
         private val clock: Clock,
 ) : EventInsertLiveProcessor {
```
BillCarsonFr, line 80: I wonder if we should not update the `shouldProcess()` methods and for UTD events just look if there is a `m.relates_to` field? 
Actually it could even be done at the `EventInsertLiveObserver` level? not sure why we would like to observe UTD events that are not relations?

BillCarsonFr (changes requested):
Looks good, I just have 2 comments.

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/session/room/EventRelationsAggregationProcessor.kt
mnaturel: That's a good point, I agree, I will update the `EventInsertLiveObserver`.

mnaturel: Done in [ee3cbd9](https://github.com/vector-im/element-android/pull/7851/commits/ee3cbd988481d0d071726831c720389fd60c176a)

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/item/MessageInformationData.kt
```diff
@@ -90,7 +90,8 @@ data class PollResponseData(
         val votes: Map<String, PollVoteSummaryData>?,
         val totalVotes: Int = 0,
         val winnerVoteCount: Int = 0,
-        val isClosed: Boolean = false
+        val isClosed: Boolean = false,
+        val hasDecryptionError: Boolean = false,
```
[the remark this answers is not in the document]
mnaturel: Yes exactly, I have renamed it to `hasEncryptedRelatedEvents` instead. See [d1ce15b](https://github.com/vector-im/element-android/pull/7851/commits/d1ce15bf18ea4119ece633dc4b42cf08cc5a3316)

matrix-sdk-android/src/main/java/org/matrix/android/sdk/internal/database/query/EventEntityQueries.kt
mnaturel: Thanks for the explanation. It made me realize I needed to change some logic in `EventInsertLiveObserver`. I have updated it in [60e838a](https://github.com/vector-im/element-android/pull/7851/commits/60e838a82c4d82b26e1141f79cb7028bab50a7c6).

BillCarsonFr (approved):
LGTM

PR #7885, opened 2023-01-04
Fix lint false positive

bmarty:
Sometimes lint reports false positives. I do not want to ignore the corresponding rules, instead I have added a baseline with all the false positive so that they are ignore now.

This is a temporary solution, but I do not know why lint is changing behavior. AGP version hasn't been changed recently (7.3.1).

This is a recurring issue though...

### DIFF
--- a/tools/lint/lint.xml
+++ b/tools/lint/lint.xml
@@ -77,6 +77,7 @@
     <issue id="UseValueOf" severity="error" />
     <issue id="ObsoleteSdkInt" severity="error" />
     <issue id="Recycle" severity="error" />
+    <issue id="KotlinNullnessAnnotation" severity="error" />
     <issue id="KotlinPropertyAccess" severity="error" />
     <issue id="DefaultLocale" severity="error" />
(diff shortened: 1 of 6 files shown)


tools/lint/lint.xml
```diff
@@ -114,4 +115,11 @@
 
     <!-- DI -->
     <issue id="JvmStaticProvidesInObjectDetector" severity="error" />
+
+    <!-- Temporary ignoring errors... TODO Unignore at some point -->
+    <issue id="LintError" severity="error">
+        <ignore path="**/CreateRoomViewModel.kt" />
+        <ignore path="**/MessageComposerViewModel.kt" />
+    </issue>
```
bmarty, line 123: Adding the issues in the baseline file does not work, so I added the exclusion here...

mnaturel:
I have raised #7891 as an alternative to fix Lint issues.

bmarty:
OK, thanks, I will drop the 2nd commit then.

mnaturel (approved):

PR #7945, opened 2023-01-12
Voice Broadcast - only send a notification on the first chunk

### LINKED ISSUE #7845: Ignore the voice broadcast chunks at the notifications level
During the notifications handling, we should notify only the first voice chunk of a voice broadcast, and ignore the other events related to the voice broadcast chunks

These events are not displayed in the room timeline, they should not be taken into account in the notifications count.
They are voice messages marked as voice broadcast chunks (they refer to a voice message state event)

There is no case where the user may want to be notified on these chunks, they are supposed to be invisible for them. That is why we can hardcode this mechanism client side without adding config in push rules

We have to ignore them locally, because these events are encrypted, we could not ignore them server side

_No response_

### Additional context

_No response_

Yes

Florian14:
## Type of change

- [x] WIP Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

During the handling of the notifications, notify only the first voice chunk of a voice broadcast, and ignore the other events related to the voice broadcast chunks

## Motivation and context

Close #7845 

## Screenshots / GIFs

https://user-images.githubusercontent.com/11990514/212134974-33339bae-49ca-40d7-b801-c58ee13231a2.mp4

## Tests

- Enable all notifications on Android device
- Start a voice broadcast from another device
- Pause or wait 2 min to send a chunk
- Verify that a voice broadcast notification appears on Android device
- Resume & Pause again the VB or wait for a second chunk
- Verify that there is no new notification
- Verify that the other notifications (text or simple voice message) are still working

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/vector/src/main/java/im/vector/app/features/notifications/FilteredEventDetector.kt
+++ b/vector/src/main/java/im/vector/app/features/notifications/FilteredEventDetector.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2023 New Vector Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package im.vector.app.features.notifications
+
+import im.vector.app.ActiveSessionDataSource
+import im.vector.app.features.voicebroadcast.isVoiceBroadcast
+import im.vector.app.features.voicebroadcast.sequence
+import org.matrix.android.sdk.api.session.events.model.isVoiceMessage
+import org.matrix.android.sdk.api.session.getRoom
+import org.matrix.android.sdk.api.session.room.getTimelineEvent
+import org.matrix.android.sdk.api.session.room.model.message.asMessageAudioEvent
+import org.matrix.android.sdk.api.session.room.timeline.TimelineEvent
+import javax.inject.Inject
+
+class FilteredEventDetector @Inject constructor(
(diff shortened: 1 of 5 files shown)


vector/src/main/java/im/vector/app/features/notifications/FilteredEventDetector.kt
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Returns true if the given event should be ignored.
+     * Used to skip notifications if a non expected message is received.
+     */
+    fun shouldBeIgnored(notifiableEvent: NotifiableEvent): Boolean {
+        val session = activeSessionDataSource.currentValue?.orNull() ?: return false
+
+        if (notifiableEvent is NotifiableMessageEvent) {
+            val room = session.getRoom(notifiableEvent.roomId) ?: return false
+            val timelineEvent = room.getTimelineEvent(notifiableEvent.eventId) ?: return false
```
jonnyandrew, line 41: If each of these might be null and fall back to returning false, do we risk dropping notifications unnecessarily?

Would it be safer to fall back to returning true?

jonnyandrew (commented):
Looks good! I have one concern about the general impact on notifications.

jonnyandrew (approved):

vector/src/main/java/im/vector/app/features/notifications/FilteredEventDetector.kt
Florian14: I think this is fine, returning false means **not** ignore the notification, so it should not drop the notification in this case, wdyt?

Florian14: I did the same as https://github.com/vector-im/element-android/blob/f4b7161db216991a94f1a93a2391d77511fbbf7f/vector/src/main/java/im/vector/app/features/notifications/OutdatedEventDetector.kt#L31, in this detector, messages are not considered as outdated if the room or session is not retrieved

jonnyandrew: That makes sense - I misread the method name. I've edited my comment and approved

Issue #7830, opened 2022-12-20
Pause voice broadcast listening on new VB recording


giomfo:
### Your use case

Your use case
What would you like to do?
When I am listening to a Live VB, and I start recording a live voice broadcast. We should pause voice broadcast listening

Why would you like to do it?
To prevent us from recording and broadcasting the audio playback

### Are you willing to provide a PR?

Yes

PR #7973, opened 2023-01-19
Pause voice broadcast listening on new VB recording

yostyle:
#7830

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
@@ -626,13 +626,17 @@ class TimelineViewModel @AssistedInject constructor(
         viewModelScope.launch {
             when (action) {
                 VoiceBroadcastAction.Recording.Start -> {
+                    voiceBroadcastHelper.pausePlayback()
                     voiceBroadcastHelper.startVoiceBroadcast(room.roomId).fold(
                             { _viewEvents.post(RoomDetailViewEvents.ActionSuccess(action)) },
                             { _viewEvents.post(RoomDetailViewEvents.ActionFailure(action, it)) },
                     )
                 }
                 VoiceBroadcastAction.Recording.Pause -> voiceBroadcastHelper.pauseVoiceBroadcast(room.roomId)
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
```diff
@@ -626,13 +626,17 @@ class TimelineViewModel @AssistedInject constructor(
         viewModelScope.launch {
             when (action) {
                 VoiceBroadcastAction.Recording.Start -> {
+                    voiceBroadcastHelper.pausePlayback()
```
Florian14, line 629: I'm not sure if it would have been better to do this in `StartVoiceBroadcastUseCase.startVoiceBroadcast` (after checking if the user is allowed to start a VB). wdyt?

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
```diff
@@ -626,13 +626,17 @@ class TimelineViewModel @AssistedInject constructor(
...
                     voiceBroadcastHelper.startVoiceBroadcast(room.roomId).fold(
                             { _viewEvents.post(RoomDetailViewEvents.ActionSuccess(action)) },
                             { _viewEvents.post(RoomDetailViewEvents.ActionFailure(action, it)) },
                     )
                 }
                 VoiceBroadcastAction.Recording.Pause -> voiceBroadcastHelper.pauseVoiceBroadcast(room.roomId)
-                VoiceBroadcastAction.Recording.Resume -> voiceBroadcastHelper.resumeVoiceBroadcast(room.roomId)
+                VoiceBroadcastAction.Recording.Resume -> {
+                    voiceBroadcastHelper.pausePlayback()
```
Florian14, line 637: Same here

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
yostyle: to do that I need to inject the `VoiceBroadcastHelper` in `StartVoiceBroadcastUseCase` but `StartVoiceBroadcastUseCase` is already inject in `VoiceBroadcastHelper`

Florian14 (approved):

Issue #7895, opened 2023-01-04
Send voice message should not be allowed during a voice broadcast recording


giomfo:
### Steps to reproduce

1 start a voice broadcast
2 send some voice messages

#### What did you expect?

<img width="720" alt="image" src="https://user-images.githubusercontent.com/8969772/210623968-d985a71e-ca5e-4f8d-95af-62483b86ebc5.png">

#### What did you observe?
the voice message are sent in parallel to the VB NOK

### Will you send logs?

No

### Are you willing to provide a PR?

Yes

PR #7974, opened 2023-01-19
Send voice message should not be allowed during a voice broadcast recording

yostyle:
#7895

### DIFF
--- a/changelog.d/7895.bugfix
+++ b/changelog.d/7895.bugfix
@@ -1,0 +1,1 @@
+Send voice message should not be allowed during a voice broadcast recording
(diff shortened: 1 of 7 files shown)


jmartinesp:
Should we also consider the case when the voice broadcast is paused, you start a voice message and lock it so it keeps recording, then resume the broadcast? I think both recordings will be running at the same time.

https://user-images.githubusercontent.com/480955/213427686-303e9566-64b6-4663-a382-01c9aed68bc2.mov

giomfo:
> Should we also consider the case when the voice broadcast is paused, you start a voice message and lock it so it keeps recording, then resume the broadcast? I think both recordings will be running at the same time.

Yes we should block voice message even when voice broadcast is paused

jmartinesp (changes requested):
Based on @giomfo 's comment, I think we need to add some extra checks to prevent voice broadcasts and voice messages from running at the same time.

jmartinesp (approved):
LGTM, thanks!

Issue #7807, opened 2022-12-16
Support reactions on Voice Broadcast


giomfo:
### Your use case

We should be allowed to react on a VB - currently we're treating this as any other message in the timeline so it'd make sense to have it for consistency

### Are you willing to provide a PR?

Yes

giomfo:
The demand here is to react only on started VB state events

PR #7976, opened 2023-01-19
Support reactions on Voice Broadcast

yostyle:
#7807

### DIFF
--- a/vector/src/main/java/im/vector/app/core/extensions/TimelineEvent.kt
+++ b/vector/src/main/java/im/vector/app/core/extensions/TimelineEvent.kt
@@ -18,6 +18,8 @@ package im.vector.app.core.extensions
 
 import im.vector.app.features.voicebroadcast.VoiceBroadcastConstants
 import im.vector.app.features.voicebroadcast.model.MessageVoiceBroadcastInfoContent
+import im.vector.app.features.voicebroadcast.model.VoiceBroadcastState
+import im.vector.app.features.voicebroadcast.model.asVoiceBroadcastEvent
 import org.matrix.android.sdk.api.session.events.model.EventType
 import org.matrix.android.sdk.api.session.events.model.toModel
 import org.matrix.android.sdk.api.session.room.model.message.MessageContent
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/core/extensions/TimelineEvent.kt
```diff
@@ -27,7 +27,8 @@ import org.matrix.android.sdk.api.session.room.timeline.getLastMessageContent
 
 fun TimelineEvent.canReact(): Boolean {
     // Only event of type EventType.MESSAGE, EventType.STICKER and EventType.POLL_START are supported for the moment
-    return root.getClearType() in listOf(EventType.MESSAGE, EventType.STICKER) + EventType.POLL_START.values + EventType.POLL_END.values &&
+    return root.getClearType() in listOf(EventType.MESSAGE, EventType.STICKER, VoiceBroadcastConstants.STATE_ROOM_VOICE_BROADCAST_INFO) +
```
Florian14, line 30: we should add a check on the event content to only allow reactions on the started event (the stopped is also displayed in the timeline)

yostyle: I added this condition

vector/src/main/java/im/vector/app/core/extensions/TimelineEvent.kt
```diff
@@ -26,9 +28,10 @@ import org.matrix.android.sdk.api.session.room.timeline.TimelineEvent
 import org.matrix.android.sdk.api.session.room.timeline.getLastMessageContent
 
 fun TimelineEvent.canReact(): Boolean {
-    // Only event of type EventType.MESSAGE, EventType.STICKER and EventType.POLL_START are supported for the moment
-    return root.getClearType() in listOf(EventType.MESSAGE, EventType.STICKER, VoiceBroadcastConstants.STATE_ROOM_VOICE_BROADCAST_INFO) +
-            EventType.POLL_START.values + EventType.POLL_END.values &&
+    // Only event of type EventType.MESSAGE, EventType.STICKER and EventType.POLL_START, and started voice broadcast are supported for the moment
+    return (root.getClearType() in listOf(EventType.MESSAGE, EventType.STICKER) + EventType.POLL_START.values + EventType.POLL_END.values ||
+            root.getClearType() == VoiceBroadcastConstants.STATE_ROOM_VOICE_BROADCAST_INFO &&
```
Florian14, line 33: You don't need anymore, if the event is not a VB, `root.asVoiceBroadcastEvent()` will return `null`.
Also, do you think we should add parenthesis?

Florian14 (approved):
LGTM, thanks

Issue #7929, opened 2023-01-10
Tapping slightly left or right of the 30s buttons highlights the whole cell instead of registering as button presses


giomfo:
### Your use case

Tapping slightly left or right of the 30s buttons highlights the whole cell instead of registering as button presses. Maybe we could increase the tap target? I ran into this quite a few times on my phone

### Are you willing to provide a PR?

Yes

PR #7978, opened 2023-01-19
Tapping slightly left or right of the 30s buttons highlights the whole cell instead of registering as button presses

yostyle:
#7929


Florian14 (approved):
SGTM

PR #8012, opened 2023-01-26
[Voice Broadcast] Use internal playback timer to compute the playback position

Florian14:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Rely on an internal playback timer to display the current playback position instead of relying on the position given by the android media player.

~Note:
In my opinion, this is not an ideal solution, I reused and improved the existing [CountUpTimer](https://github.com/vector-im/element-android/blob/3acaa16c56d9f843688f1342bb58d1d13f34d373/library/core-utils/src/main/java/im/vector/lib/core/utils/timer/CountUpTimer.kt) which is not based on real elapsed time. Under the hood, it's a `tickerFlow` which notifies at a fixed interval (50 ms in our case) except if we pause it. It can shift a bit for about +/- the `intervalMs` on each pause/resume depending on the timing.
Btw, I noticed that a too-low interval (eg. 10 ms) will break the timer. I also noticed perfs issues after enabling the debugger on android studio, the Flow takes too much time to notify. I don't know if a busy device can have the same kind of issue.~

## Motivation and context

Voice broadcasts sent by Element-Web were causing trouble in the playback position computation.
The chunks sent from the web are encoded in ogg format with aggregated metadata which is not the case for the other VB sent from iOS or Android.

Also, when the recorder is paused on the web, the aggregation is reset, so we cannot rely on the ogg file metadata to calculate the current playback position. The android media player also gives the aggregated position and duration of these files causing glitches in the playback rendering.

I did not find a way to get the relative position in the android MediaPlayer API. Moreover, there are some glitches when we switch to the next file, `MediaPlayer.currentPosition` starts returning the relative position (eg. 0ms/250ms/375ms) and then jumps to the aggregatedPosition (eg. 15_000ms...) but always returns the aggregated duration (eg. 45_000ms for a chunk of 15_000ms, supposing we are listening to the third chunk).

## Tests

- Have several VB with several chunks (you can set the chunk duration to 15s on web in the `/devtools` options) recording from the web and iOS or Android
- Verify that the playback position is correcty updated on each VB
- Verify that the playback position is still correct after moving the cursor
- Stress a bit the playback by switching to another one or playing with the controls (pause/resume)

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

- [x] Pull request is based on the ~develop~ main branch


library/core-utils/src/main/java/im/vector/lib/core/utils/timer/Clock.kt
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package im.vector.lib.core.utils.timer
+
+interface Clock {
```
Florian14, line 19: I did not find a better way instead of duplicating this file. I have issues with DI after moving the existing clock interface from the vector module into this one. Also, I cannot move the CountUpTimer to the vector module because another lib is using it.

library/core-utils/src/main/java/im/vector/lib/core/utils/timer/CountUpTimer.kt
```diff
@@ -55,17 +56,25 @@ class CountUpTimer(private val intervalInMs: Long = 1_000) {
...
 
     fun stop() {
+        tickListener?.onTick(addAndGetElapsedTime())
         coroutineScope.cancel()
     }
 
+    private fun addAndGetElapsedTime(): Long {
+        val now = clock.epochMillis()
+        return elapsedTime.addAndGet(now - lastTime.getAndSet(now))
+    }
```
jonnyandrew, line 76: Would it be more accurate to keep track of the start time and calculating each time:

```
elapsed time = now - start
```

vector/src/main/java/im/vector/app/features/voicebroadcast/listening/VoiceBroadcastPlayerImpl.kt
```diff
@@ -270,6 +282,8 @@ class VoiceBroadcastPlayerImpl @Inject constructor(
                 playbackTracker.updatePausedAtPlaybackTime(voiceBroadcast.voiceBroadcastId, positionMillis, positionMillis.toFloat() / duration)
             }
             playingState == State.Playing || playingState == State.Buffering -> {
+                stopPlayer()
```
jonnyandrew, line 285: Is it right that when the playing state is `State.Playing` that we call `stopPlayer()`?

Florian14: Yes, as the user requested to listen from another position (potentially in another chunk file), we stop the current player and will start a new one just two lines later. If we don't stop it, the playback position which is saved in the next line could be erased by the current player

Florian14: ... but you're right, this is not necessary, it is already done in `startPlayback` method. I'll remove it from this place

library/core-utils/src/main/java/im/vector/lib/core/utils/timer/CountUpTimer.kt
```diff
@@ -28,24 +28,25 @@ import java.util.concurrent.atomic.AtomicBoolean
...
+    private var lastTime: AtomicLong = AtomicLong()
+    private val elapsedTime: AtomicLong = AtomicLong(initialTime)
+
     init {
         startCounter()
     }
 
     private fun startCounter() {
-        tickerFlow(coroutineScope, intervalInMs / 10)
+        tickerFlow(coroutineScope, intervalInMs)
```
bmarty, line 45: `/ 10` was used for this reason: https://github.com/vector-im/element-android/commit/c69bc126377729de1483d084a88d1aa5fbd36327

Are you sure you want to revert this?

bmarty (approved):
Just one question, not blocking.

library/core-utils/src/main/java/im/vector/lib/core/utils/timer/CountUpTimer.kt
Florian14: @bmarty The elapsed time and the current time are updated when the timer is paused & resumed (see https://github.com/vector-im/element-android/blob/941d0e5cf4c509fbc1ef6d3f29fa4242266dc74e/library/core-utils/src/main/java/im/vector/lib/core/utils/timer/CountUpTimer.kt#L58-L66), it should be more precise now and the previous trick should not be necessary anymore. Moreover, the timer does not work anymore if the interval is too low (eg. under 10 ms), the ticks are too close and it does not notify in the expected delay (it can take more than 2 seconds to notify 1 second...), so it is better to keep the real interval value and do not divide the value by ten to prevent this bad behaviour.

Florian14:
Some minor improvements since the last review: https://github.com/vector-im/element-android/pull/8012/files/3d87b796b46441e1c08fd8bd772565ecce587715..ecc0cb5f0bb7ac32124be744b7cbd9d0a550d1cb

yostyle (approved):
SGTM. I did some tests.

library/core-utils/src/main/java/im/vector/lib/core/utils/timer/CountUpTimer.kt
bmarty: Reading the code again, I think user can observe an extra delay if we keep a tick per second.
The pb is that the `tickerFlow` ticks every second, and not when the UI has to be udpated.

bmarty: As discussed together IRL, @Florian14 will make a dedicated PR to improve `CountUpTimer` implementation, and make it respect its expected contract. Ideally with some tests around it.

Issue #4199, opened 2021-10-10
Animated webp is not working.


DUCKCHI:
#### What would you like to do?
animated webp plays well on element web, pc clients. However, on Android clients, animation is not played and there is a problem that looks like a static image. Not only the server I'm using, but also other servers, including matrix.org, have the same problem.

I hope that animated webp will continue to play on Android clients like web clients.

#### Why would you like to do it?
I also want to use the animated sticker of the webp on the mobile client.

williamkray:
duplicate of #2695

Issue #2695, opened 2021-01-18
Support animated webp files.


kevincox:
**Is your feature request related to a problem? Please describe.**
Right now webp animations will not play. This can be very confusing as it fails to convey what the sender intended.

**Describe the solution you'd like**
webp animations should be treated identically to animated gifs.

kojid0:
Duplicate of #2668?

kevincox:
I don't think so. At the least this is more specific. But that bug seems
more about selecting a media file where as this is about making it play.

HarHarLinks:
This here is about allowing playback of the webp mime type at all.
2668 is about **stickers** that use animated picture mime types that are already supported (like gif) and potentially going to be supported (like webp), which are currently bugged in that they show a play button on top of the sticker similar to the gif, however it is impossible to press the play button which usually works to open some kind of video player that plays back a non-sticker gif.

PR #8120, opened 2023-02-12
Fix animated webp playback

alexmaras:
## Type of change

- [x] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Allowed Animated WebP to play correctly.

This involves making Glide transformations optional in order to prevent mimetypes that Glide doesn't support from throwing exceptions. This has a very minor impact - making rounded corners not work for animated WebP files in the timeline. However, I see this as an improvement over it not working at all.

## Motivation and context

Animated WebP files are delivered by default from the Facebook Bridge (https://github.com/mautrix/facebook) so this has been a pain point for me.

Element Android issues:
https://github.com/vector-im/element-android/issues/2695
https://github.com/vector-im/element-android/issues/4199

Animated WebP is supported on android versions natively. The only issue is that the Glide library doesn't support it outside of rendering:
https://github.com/bumptech/glide/issues/571

This could alternatively potentially be fixed by using https://github.com/zjupure/GlideWebpDecoder
However, this feels like a change that could potentially introduce more issues, as that library only supports a couple of recent versions of Glide, which would necessitate keeping those in-line. I'm not sure how well maintained the library is, and this could be worse than just accepting the minor differences in missing rounded corners for animated WebP files.

## Screenshots / GIFs

|Before|After|
|-|-|
|animated-webp-before|animated-webp-after|

There is a minor annoyance introduced by this - non-animated webp files will have a "play" button on them. This is unavoidable without some major changes to the image viewer. We'd need to pull the image and read the VP8X header for the ANIM bit to see if it's an animated webp. The same issue is already present with non-animated Gifs. 

## Tests

- Run in Android Studio
- Open a room with animated WebP files
- Tap into the animated WebP, ensure it still plays
- Tap into an animated gif, ensure it still plays
- Tap into a still WebP, ensure there are no changes (due to now being marked as `playable`)

## Tested devices

- [x] Physical
- [x] Emulator
- OS version(s): Pixel 6 API 21 on Emulator w/ Lollipop, Pixel 4A API 31 on Emulator, Nothing Phone One (physical) w/ Android 12

The changes don't do anything particularly useful on older versions of Android - but it's not more broken than it was. Android requires API v28 to play Animated WebP by the looks of it. At the moment, it would just be still images, whereas with this change they can at least see that they're supposed to be animated. 

## Checklist


vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -509,7 +509,7 @@ class MessageItemFactory @Inject constructor(
                 allowNonMxcUrls = informationData.sendState.isSending()
         )
 
-        val playable = messageContent.mimeType == MimeTypes.Gif
+        val playable = messageContent.mimeType == MimeTypes.Gif || messageContent.mimeType == MimeTypes.Webp
```
alexmaras, line 512: This is a relatively poor check, but it's the best we can do without being able to read the file headers. 

In theory, we could also use 
```suggestion
        val playable = messageContent.mimeType == MimeTypes.Gif || (messageContent.mimeType == MimeTypes.Webp && Build.VERSION.SDK_INT >= Build.VERSION_CODES.P)
```
to not show the "play" button on older versions of Android, but it doesn't really make much of an appreciable difference - they're broken either way, and were before.

jmartinesp (commented):
I think most of the code makes sense except for the `playable` attribute being true for all webp images.

Also, to have the same behaviour in the attachment screen as you do, I had to add the `|| it.mimeType == MimeTypes.Webp` to both [DataAttachmentRoomProvider.getAttachmentInfoAt](https://github.com/vector-im/element-android/blob/928183ff64f71d8f29f745a7cde58cafa0b7833f/vector/src/main/java/im/vector/app/features/media/DataAttachmentRoomProvider.kt#L52) and [RoomEventsAttachmentProvider.getAttachmentInfoAt](https://github.com/vector-im/element-android/blob/9f9c6fbc5b76a54f18f7ec2b84d308209f2fffc5/vector/src/main/java/im/vector/app/features/media/RoomEventsAttachmentProvider.kt#L74), otherwise the webp images didn't animate in the attachment viewer screen.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
jmartinesp: I don't have any statistics, but I don't think animated webps are commonly used compared to static webp images, so making all webp images 'playable' might not be the best approach. 

The `playable` attribute is only used to display the 'play' button above the image, but if you have the autoplay animated images option enabled the image will already be playing in the timeline, so given these 2 possible cases:

1. Users not seeing the play button in an animated webp image and not having autoplay images enabled, thus not knowing that it's an animated image.
2. Users seeing play buttons over every single webp, most of them being static.

I think it would be better to go with the first option, that is: remove the `|| messageContent.mimeType == MimeTypes.Webp`.

alexmaras:
Thanks @jmartinesp - apologies, I had those previously. I must have checked them out during my pre-commit clean-up. I'll add them back in and revert WebP being marked as playable and re-test.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
alexmaras: Sadly, autoplay in thumbnail doesn't work unless `playable` is enabled.

https://github.com/vector-im/element-android/blob/de50577ac38ebd9f2b69715142de1520bf9299eb/vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt#L552

It does work when tapping into it, which is still a better result than it not working at all. 
If we want to actually have it autoplay in preview mode, we'd need to know beforehand that it's actually an animated webp.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
```diff
@@ -528,6 +528,8 @@ class MessageItemFactory @Inject constructor(
         )
 
         val playable = messageContent.mimeType == MimeTypes.Gif
+        // don't show play button because detecting animated webp isn't possible via mimetype
+        val playableIfAutoplay = playable || messageContent.mimeType == MimeTypes.Webp
```
alexmaras, line 532: This works to autoplay animated WebP in the timeline if autoplay is enabled. If it _isn't_ enabled, no play button shows (because `playable` is `false`) which will prevent confusion with static WebP. Static WebP functions fine in autoplay mode still.

vector/src/main/java/im/vector/app/features/home/room/detail/timeline/factory/MessageItemFactory.kt
jmartinesp: I'm actually seeing the opposite issue. `playable` is disabled, autoplay is disabled too, but thumbnails for animated webp images are autoplayed anyway. I don't think it's directly related to your changes, but I also have no idea why in your case it seems like the opposite is happening.

<details><summary>See recording</summary>
Top and bottom items are WebP, middle is a gif.

[Play recoding](https://user-images.githubusercontent.com/480955/220705383-2e8a13a1-7478-4ba2-8f31-0e01beb6d57e.webm)
</details>

alexmaras: I've been able to replicate both behaviors now. I'm going to do some testing to figure out what causes it to autoplay _sometimes_ when autoplay is turned off. In your recording, you can see that the nyancat gif has squared corners, and I'm currently looking at my emulator on API 33 and it has rounded corners and doesn't play until I click into it, even if I tap in and then go back out, and if I turn autoplay on, go back, then go turn it off again.

However, my laptop is now showing the same behavior as yours is.

I'll let you know when I have an explanation for this. Thanks for spotting!

jmartinesp (approved):
Oh, sorry, I just realised you made some changes to address the issues, I should have reviewed this months ago...

LGTM. Thanks for the PR and sorry for the long delay.

PR #8130, opened 2023-02-14
[Poll] Synchronize polls and message push rules after creation (PSG-1137)

Florian14:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Add an observer on the push rule updates in the account data and realign the push rules with their parent if needed.

## Motivation and context

#8114 introduces the new push rules in the notification settings. The new rules should be aligned with their parent rules as described in the following table:

image

The purpose of this PR is to verify and fix if necessary the consistency of the new rules with their parent when changes are detected in the account data.

## Tests

- Manually change and break the poll rules in the account data using the devtools on the web interface
- Verify that the Android client automatically realign the rules with the parent one (by refreshing the account data after an incremental sync)

## Tested devices

- [x] Physical
- [ ] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushrules/RuleIds.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushrules/RuleIds.kt
@@ -61,22 +61,36 @@ object RuleIds {
     const val RULE_ID_FALLBACK = ".m.rule.fallback"
 
     const val RULE_ID_REACTION = ".m.rule.reaction"
+}
 
-    fun getSyncedRules(ruleId: String): List<String> {
-        return when (ruleId) {
-            RULE_ID_ONE_TO_ONE_ROOM -> listOf(
-                    RULE_ID_POLL_START_ONE_TO_ONE,
-                    RULE_ID_POLL_START_ONE_TO_ONE_UNSTABLE,
-                    RULE_ID_POLL_END_ONE_TO_ONE,
-                    RULE_ID_POLL_END_ONE_TO_ONE_UNSTABLE,
-            )
-            RULE_ID_ALL_OTHER_MESSAGES_ROOMS -> listOf(
-                    RULE_ID_POLL_START,
-                    RULE_ID_POLL_START_UNSTABLE,
-                    RULE_ID_POLL_END,
-                    RULE_ID_POLL_END_UNSTABLE,
-            )
-            else -> emptyList()
-        }
+fun RuleIds.getSyncedRules(ruleId: String): List<String> {
+    return when (ruleId) {
+        RULE_ID_ONE_TO_ONE_ROOM -> listOf(
+                RULE_ID_POLL_START_ONE_TO_ONE,
+                RULE_ID_POLL_START_ONE_TO_ONE_UNSTABLE,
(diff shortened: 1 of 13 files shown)


matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushrules/RuleIds.kt
```diff
@@ -61,22 +61,36 @@ object RuleIds {
...
-            )
-            RULE_ID_ALL_OTHER_MESSAGES_ROOMS -> listOf(
-                    RULE_ID_POLL_START,
-                    RULE_ID_POLL_START_UNSTABLE,
-                    RULE_ID_POLL_END,
-                    RULE_ID_POLL_END_UNSTABLE,
-            )
-            else -> emptyList()
-        }
+fun RuleIds.getParentRule(ruleId: String): String? {
```
jonnyandrew, line 84: Could be worth adding a comment here to explain what a parent rule is?

jonnyandrew (approved):

matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/pushrules/RuleIds.kt
Florian14: Yeah, I'm also wondering if this should be moved outside of the sdk, this relation between the push rules is mentioned in the MSC but their connection is done at the application side (the clients are not forced to apply this relation, but on the other hand it could also be interesting to share this information at the sdk level... wdyt?)

jonnyandrew: I couldn't find any mention of parent rules in the [MSC3930](https://github.com/matrix-org/matrix-spec-proposals/pull/3930). Is that the right place to look?

Florian14: The term "parent" is not mentioned, I was talking about these references to `m.room.message` rule:
>In order to have polls behave similar to message events, the following underride push rules are defined. Note that the push rules are mirrored from those available to m.room.message events.

>Typically these rules would be kept in sync with the m.room.message rules they are based upon, however there is no requirement to do so. A client's approach might be to only keep them in sync while setting the values, rather than monitoring for changes to push rules.

jonnyandrew: It sounds like more of a recommendation from the MSC about how the push rules might be related. In a different app, for example one that is based around polls rather than messages) I guess the parent relation might be reversed?

So I'm leaning towards moving this out of the SDK.

Florian14: You're totally right, I moved the extensions to the vector module https://github.com/vector-im/element-android/pull/8130/commits/dcd43d6e7fc8ab1949fb5163bf6ddedf26b44559

jonnyandrew (approved):
Thanks for the changes!

PR #8152, opened 2023-02-21
Fix always jumps to the bottom of the timeline #8090

### LINKED ISSUE #8090: Element Android always jumps to the bottom of the timeline
It seems like it is a design decision for leaving a room and opening it back up (although maybe that should be reconsidered), but opening an image, a profile, threads, the room settings and so on should not reset the position in the timeline.

Reproduce:
- scroll up
- open image
- close image
- you are at the bottom of the timeline again

ganfra:
Fix #8090 
The regression came from this https://github.com/vector-im/element-android/pull/7734 
I've checked it was still fixing the related notification issue ^

### DIFF
--- a/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
+++ b/vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
@@ -53,6 +53,7 @@ import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import androidx.transition.TransitionManager
 import com.airbnb.epoxy.EpoxyModel
+import com.airbnb.epoxy.EpoxyVisibilityTracker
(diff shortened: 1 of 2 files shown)


jmartinesp (approved):
LGTM, we just need to fix the lint issues.

vector/src/main/java/im/vector/app/features/home/room/detail/TimelineFragment.kt
```diff
@@ -1038,7 +1039,6 @@ class TimelineFragment :
         timelineEventController.callback = this
         timelineEventController.timeline = timelineViewModel.timeline
 
-        views.timelineRecyclerView.trackItemsVisibilityChange()
```
bmarty, line 1041: I am wondering if this extension is correct since we never detach the tracker...

bmarty (commented):
Thanks a lot!

PR #8202, opened 2023-03-03
Fix CI

bmarty:
Let's see what's happening here.


bmarty:
@mnaturel do you know a better way to fix the issue on test rather than https://github.com/vector-im/element-android/pull/8202/commits/ceda85f6249a20972b811fb37a0522396c47767e (i.e. avoiding making the tag public)?

yostyle (approved):
LGTM

mnaturel:
> @mnaturel do you know a better way to fix the issue on test rather than [ceda85f](https://github.com/vector-im/element-android/commit/ceda85f6249a20972b811fb37a0522396c47767e) (i.e. avoiding making the tag public)?

I think there are different possible alternatives:
- annotate the field with `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)`
to warn if we try to use the field externally
- use the `spyk` annotation to use a mix between real and mocked object
- use a relaxed mock to have default behavior

Issue #8286, opened 2023-03-28
Replace "Terms & Conditions" by "Acceptable Use Policy" in settings


manuroe:
It will open a new link: https://element.io/acceptable-use-policy-terms.

PR #8295, opened 2023-03-31
Replace Terms and Conditions with Acceptable Use Policy.

ganfra:
Handles #8286

### DIFF
--- a/library/ui-strings/src/main/res/values/strings.xml
+++ b/library/ui-strings/src/main/res/values/strings.xml
@@ -1002,6 +1002,7 @@
     <string name="settings_version">Version</string>
     <string name="settings_olm_version">olm version</string>
     <string name="settings_app_term_conditions">Terms &amp; conditions</string>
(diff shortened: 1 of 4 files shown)


library/ui-strings/src/main/res/values/strings.xml
```diff
@@ -1002,6 +1002,7 @@
     <string name="settings_version">Version</string>
     <string name="settings_olm_version">olm version</string>
     <string name="settings_app_term_conditions">Terms &amp; conditions</string>
+    <string name="settings_acceptable_use_policy">Acceptable Use Policy</string>
```
jmartinesp, line 1005: Aren't we using localazy for this?

jmartinesp: My bad, I automatically thought this was for EAX 🤦

jmartinesp (approved):
LGTM!

PR #8310, opened 2023-04-05
Fix multiple read receipts for the same user in timeline #7882

### LINKED ISSUE #7882: Two "Seen by" dots shown
### Steps to reproduce

In the messages view, in a conversation with one person, I see two "Seen by" dots - one after my most recent message the other user has seen, and one after the last message they sent. Something like this (x represents the "Seen by" dot):
```
Me: hello
Them: How's it going?
                                     x
Me: Pretty good, you?
                                     x
Me: You there?
```

#### What did you expect?
I expect to see only one "Seen by" dot, after the most recent of my messages the other user has seen.

#### What happened instead?
I see two.

### Your phone model

Google Pixel 7

### Operating system version

Android 13

Element version 1.5.11

### Homeserver

non-public, synapse

### Will you send logs?

Yes

No

ganfra:
Fixes #7882 
As we can now have multiple read receipts for the same user because of threads, we need to ensure we only show the most recent in a timeline.
We also need to handle clients not handling threads.
It could be handled in the sdk but it's easier to manage in the presentation layer.

### DIFF
--- a/changelog.d/7882.bugfix
+++ b/changelog.d/7882.bugfix
@@ -1,0 +1,1 @@
+Fix multiple read receipts for the same user in timeline.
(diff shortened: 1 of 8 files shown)


bmarty (approved):
LGTM at first sight. Thanks!

yostyle (approved):
LGTM
tested on my device

RayBB:
Thank you so much!!!!!! A+ can't wait for the release!

PR #8327, opened 2023-04-12
Add variation selector 16 to all applicable emojis

### LINKED ISSUE #2776: Duplicate thumbs | Mobile reactions for 👍 and 👎 are not the same as web

tulir:
Actually fixes #2776 for all emojis. #2777 only fixed quick reactions as mentioned in a comment on that PR.


bmarty (approved):
LGTM, thanks!

PR #8401, opened 2023-05-05
Kotlin 1.8.21

bmarty:
**WARNING** There is a db migration, so we must wait for migration to Rust Crypto PR - which also contain DB migration - to be merged first and work again on this PR, to update the migration.

Replaces:

- #8374
- #8369
- #8367


bmarty:
We will only upgrade a limited set of dependencies now. See https://github.com/vector-im/element-android/issues/8518 for more details.

PR #8519, opened 2023-06-09
Only upgrade our libraries to limit the number of open PRs by Dependabot

### LINKED ISSUE #8518: Deactivate Dependabot upgrades
## Context

The project is in maintenance mode, we are focusing on [Element X](https://github.com/vector-im/element-x-android) now.

## Problematic

Dependabot is creating lots of PR that we cannot handle since sometimes there are some API break, or it's too risky to upgrade the dependency without taking the risk to break something.

## Solution

Disable Dependabot upgrade except for our own libraries:
- RichTextEditor (`io.element.android:wysiwyg`)
- Crypto Rust SDK (`org.matrix.rustcomponents:crypto-android`)

All [open PRs](https://github.com/vector-im/element-android/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies) by Dependabot can be closed once this is issue has been handled.

We will do upgrade of dependencies only if some security issue is reported.

bmarty:
Fixes #8518

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,7 @@
 # To get started with Dependabot version updates, you'll need to specify which
 # package ecosystems to update and where the package manifests are located.
 # Please see the documentation for all configuration options:
-# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
 
 version: 2
 updates:
@@ -13,7 +13,7 @@ updates:
     reviewers:
       - "vector-im/element-android-reviewers"
     ignore:
-      - dependency-name: "*github-script*"
+      - dependency-name: "*"
   # Updates for Gradle dependencies used in the app
   - package-ecosystem: gradle
     directory: "/"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -22,5 +22,6 @@ updates:
     open-pull-requests-limit: 200
     reviewers:
       - "vector-im/element-android-reviewers"
-    ignore:
-      - dependency-name: com.google.zxing:core
+    allow:
+      - dependency-name: "io.element.android:wysiwyg"
+      - dependency-name: "org.matrix.rustcomponents:crypto-android"
```
jonnyandrew, line 27: I'm not sure this will work as according to [the docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow)

> a dependency that is matched by both an allow and an ignore will be ignored.

bmarty: Sin I have removed the `ignore` section, I think this will be fine 🤞

bmarty: (because we are in the `package-ecosystem: gradle`)

jonnyandrew: Ah yes, I missed that. Looks good to me then!

jonnyandrew (approved):

PR #8598, opened 2023-08-02
fix wrong reading key or passphrase from well-known configuration

bagvand:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content
Bug: secure_backup_setup_methods from well-know config not working
Fix: After apply these changes secure_backup_setup_methods start working


bmarty:
@BillCarsonFr what do you think about this change?

PR #8683, opened 2023-11-15
Unified Push: Ignore the potential SSL error when the custom gateway is testing

giomfo:
When the Unified Push is enabled, the application checks the potential custom gateway before applying it. If an SSL error happens, the application may ignore this error and keep using this custom gateway. The actual SSL check will be done server side where this gateway is actually used.

Signed-off-by: giomfo gforet@matrix.org


bmarty (approved):
OK, thanks!

PR #8698, opened 2023-11-30
Implement Mobile Device Manager feature with 3 keys.

bmarty:
## Type of change

- [x] Feature
- [ ] Bugfix
- [ ] Technical
- [ ] Other :

## Content

Add support for MDM with the following keys:
- default homeserver URL - `im.vector.app.serverConfigDefaultHomeserverUrlString`
- push gateway URL - `im.vector.app.serverConfigSygnalAPIUrlString`
- permalink base URL - `im.vector.app.clientPermalinkBaseUrl`

iOS PR: https://github.com/vector-im/element-ios/pull/7667

## Motivation and context

Support MDM.

Internal issue: https://github.com/vector-im/element-internal/issues/527

## Tests

Cannot test using a real MDM configuration, but can be tested by hacking the method getData to return a custom string.
I have tested by returning a value for `MdmData.DefaultHomeserverUrl` different from `matrix.org` and the application was correctly defaulting to the provided homeserver.

Also currently when the MDM configuration change, this is a no-op on the application. We may consider logging out the user.

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

### DIFF
--- a/vector-app/src/main/java/im/vector/app/core/di/SingletonModule.kt
+++ b/vector-app/src/main/java/im/vector/app/core/di/SingletonModule.kt
@@ -112,6 +112,9 @@ import javax.inject.Singleton
     abstract fun bindEmojiSpanify(emojiCompatWrapper: EmojiCompatWrapper): EmojiSpanify
 
     @Binds
+    abstract fun bindMdmService(service: DefaultMdmService): MdmService
+
+    @Binds
     abstract fun bindFontScale(fontScale: FontScalePreferencesImpl): FontScalePreferences
 
     @Binds
@@ -171,11 +174,6 @@ import javax.inject.Singleton
     @Singleton
     fun providesMatrix(context: Context, configuration: MatrixConfiguration): Matrix {
         return Matrix(context, configuration)
-    }
-
-    @Provides
-    fun providesMdmService(context: Context): MdmService {
-        return DefaultMdmService(context)
     }
 
     @Provides
--- a/vector/src/main/java/im/vector/app/features/mdm/DefaultMdmService.kt
+++ b/vector/src/main/java/im/vector/app/features/mdm/DefaultMdmService.kt
@@ -25,7 +25,9 @@ import androidx.core.content.getSystemService
 import dagger.hilt.android.qualifiers.ApplicationContext
 import timber.log.Timber
 import javax.inject.Inject
+import javax.inject.Singleton
 
+@Singleton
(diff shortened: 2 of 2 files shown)


vector-app/src/main/java/im/vector/app/core/di/SingletonModule.kt
```diff
@@ -168,6 +173,11 @@ import javax.inject.Singleton
         return Matrix(context, configuration)
     }
 
+    @Provides
```
ganfra, line 176: You should use `@Singleton ` otherwise listeners won't be kept.
Also use @Binds instead of @Provides

ganfra (approved):
One change on the DI otherwise LGTM

vector-app/src/main/java/im/vector/app/core/di/SingletonModule.kt
bmarty: Done in 6452b5c2b450da8de1f6ee4717b9940b49616b1e, hence I think it was fine since the listener was added and removed by the same entity (the Activity).
OK for you?

vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
```diff
@@ -143,7 +146,7 @@ class OnboardingViewModel @AssistedInject constructor(
     }
 
     private val matrixOrgUrl = stringProvider.getString(R.string.matrix_org_server_url).ensureTrailingSlash()
-    private val defaultHomeserverUrl = matrixOrgUrl
+    private val defaultHomeserverUrl = mdmService.getData(MdmData.DefaultHomeserverUrl, matrixOrgUrl)
```
giomfo, line 149: @bmarty don't you think we should move this mdmService call where the `defaultHomeserverUrl` is used?

giomfo (commented):
just one question

vector/src/main/java/im/vector/app/features/onboarding/OnboardingViewModel.kt
bmarty: It should not change during the lifetime of the `OnBoardingViewModel`, but I prefer to avoid this scenario to occur. So better to read it once, and then it cannot change during the lifetime of the `OnBoardingViewModel`.

giomfo (approved):
SGTM

PR #8700, opened 2023-12-01
Support Functional members #3736

### LINKED ISSUE #3736: Support Functional members
#### Is your feature request related to a problem? Please describe.
At Element we want to add bots to nameless DM rooms. These bots should not affect the dynamic room name.

A DM room with Alice, Reminder Bot and me should be called "Alice" and not "Alice and Reminder bot", given the state described below is present.

#### Describe the solution you'd like.
When calculating a room name:
1. Read the room state of type `io.element.functional_members`. (no state key)
2. In the content, look for an array of strings called.
3. Exclude those user ids from the name generation.

```json
{
    "service_members": [
        "@helperbot:localhost",
        "@reminderbot:alice.tdl"
    ]
}
```

Whenever the state event content gets overwritten, the room name needs to be recalculated.

#### Describe alternatives you've considered.
[MSC2199: Canonical DMs](https://github.com/matrix-org/matrix-doc/blob/travis/msc/immutable-dms/proposals/2199-canonical-dms.md) defines `unimportant` users, which are likely going to accomplish the same.
However, we have an upcoming EMS product which needs this feature faster and it's still uncertain where MSC2199 is going.

#### Additional context
MSC-like description:
https://github.com/vector-im/element-meta/blob/develop/spec/functional_members.md

Implementation for Element Web:
https://github.com/matrix-org/matrix-js-sdk/pull/1771

Issue for Element iOS:
https://github.com/vector-im/element-ios/issues/4609

ganfra:
Closes #3736

### DIFF
--- a/vector-config/src/main/java/im/vector/app/config/Config.kt
+++ b/vector-config/src/main/java/im/vector/app/config/Config.kt
@@ -42,6 +42,10 @@ object Config {
     const val ENABLE_LOCATION_SHARING = true
     const val LOCATION_MAP_TILER_KEY = "fU3vlMsMn4Jb6dnEIFsx"
 
+    /// Whether to read the `io.element.functional_members` state event
+    // and exclude any service members when computing a room's name and avatar.
+    const val SUPPORT_FUNCTIONAL_MEMBERS = true
+
     /**
      * The maximum length of voice messages in milliseconds.
      */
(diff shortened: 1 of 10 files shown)


vector-config/src/main/java/im/vector/app/config/Config.kt
```diff
@@ -42,6 +42,10 @@ object Config {
     const val ENABLE_LOCATION_SHARING = true
     const val LOCATION_MAP_TILER_KEY = "fU3vlMsMn4Jb6dnEIFsx"
 
+    /// Whether to read the `io.element.functional_members` state event
```
ElementBot, line 45: - :no_entry_sign: Missing space after //

PR #8749, opened 2024-02-05
Location avatar

bmarty:
## Type of change

- [ ] Feature
- [x] Bugfix
- [ ] Technical
- [ ] Other :

## Content

This change ensure that the user Avatar used in location Event rendering is the same avatar than the avatar of the user in the room. Previously the application was using the Avatar from the user profile, but it can happen that the profile is not loaded into the DB, and so the fallback avatar was used (with the userId first later). As a consequence, if the user has set a avatar just on the current room, it will also be used.

This PR also improves how the cache works in `LocationPinProvider`. Rely on Glide cache, limit size, and invalidate cache in case of error, then success retrieving avatar data. Also use `UserItem` for the cache key, to invalidate cache if user avatar / display name change. So this PR also fix a refresh issue if user change their avatar.

This PR also fix a rendering issue of the initial avatar when long clicking on a location Event.

## Motivation and context

User user avatar in Location share Event.

Parity with Element Web.

## Tests

- Have a room with 2 members Alice and Bob.
- Invite Charlie to the room, Charlie has a profile avatar
- Charlie share their location to the room.
- Without this change, Alice and Bob do not see Charlie's avatar in the location Event, because Charlie's profile is not loaded. Charlie avatar is correctly rendered at the sender details above the message.
- With this change, Charlies's avatar is correctly rendered on the map, since the sender avatar is known and is used to render the location Event

## Tested devices

- [ ] Physical
- [x] Emulator
- OS version(s):

## Checklist

- [ ] Accessibility has been taken into account. See https://github.com/element-hq/element-android/blob/develop/CONTRIBUTING.md#accessibility
- [ ] Pull request includes a new file under ./changelog.d. See https://github.com/element-hq/element-android/blob/develop/CONTRIBUTING.md#changelog
- [ ] If you have modified the screen flow, or added new screens to the application, you have updated the test [UiAllScreensSanityTest.allScreensTest()](https://github.com/element-hq/element-android/blob/main/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt#L73)

### DIFF
--- a/changelog.d/8749.bugfix
+++ b/changelog.d/8749.bugfix
@@ -1,0 +1,1 @@
+Fix issues about location Event avatar rendering.
--- a/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/SessionExtensions.kt
+++ b/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/SessionExtensions.kt
@@ -19,6 +19,7 @@ package org.matrix.android.sdk.api.session
 import org.matrix.android.sdk.api.session.room.Room
 import org.matrix.android.sdk.api.session.room.model.RoomSummary
 import org.matrix.android.sdk.api.session.user.model.User
+import timber.log.Timber
 
 /**
  * Get a room using the RoomService of a Session.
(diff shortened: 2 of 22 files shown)


ganfra (approved):
LGTM, thanks!

PR #8770, opened 2024-03-04
Fix send button blinking with RTE

ganfra:
Add a small debounce operation in the `TextWatcher` so we don't process the temporary empty text triggered by the RTE.


jmartinesp (approved):
Looks good, thanks.

PR #8790, opened 2024-03-27
Add dependency on `com.github.matrix-org:matrix-analytics-events`

bmarty:
Add dependency on `com.github.matrix-org:matrix-analytics-events` and remove local plan and management of its import.


BillCarsonFr (approved):
LGTM

PR #8798, opened 2024-04-02
When reporting a user, use the membership state eventId for the eventId.

bmarty:
Complement to #8797.


vector/src/main/java/im/vector/app/features/home/room/detail/TimelineViewModel.kt
```diff
@@ -1147,7 +1147,22 @@ class TimelineViewModel @AssistedInject constructor(
         if (room == null) return
         viewModelScope.launch {
             val event = try {
-                room.reportingService().reportContent(action.eventId, -100, action.reason)
+                if (action.user && action.senderId != null) {
```
ganfra, line 1150: Can you just create a reportUser method to avoid repeating?

ganfra (approved):

PR #8841, opened 2024-05-31
Fix redacted events not grouped correctly

### LINKED ISSUE #8840: Redacted messages are not grouped correctly
When there are hidden events between redacted messages, they are not grouped correctly.

ganfra:
Closes #8840

The algorithm was only checking for `m.room.redaction` but this could happen for any invisible events.


bmarty (approved):
LGTM, thanks!

PR #8915, opened 2024-09-24
 Change license to AGPL

bmarty:
Change license to AGPL.
Please read https://element.io/blog/sustainable-licensing-at-element-with-agpl/ for more details.

Should not impact production code nor script


bro2020:
The article states that this mostly applies to forks.  And what about forks that were created only to contribute to the main repository using a pull request?

Issue #8964, opened 2024-12-19
Version 1.6.24 - App Crashes when Phone Locked and an incoming call comes in


ToddCrimson:
### Steps to reproduce

We updated today to latest version 1.6.24. Note we tried this on Four of our phones (Two Pixel 8 Pro, Pixel 7A, Pixel 9). All Failed. We tried with Background Sync and with NTFY push server. We built the app from your code to test it. Same issue. If phones are locked, and a call comes in, you hear the bell for a millisecond, and it goes quiet. 

1. Lock Pixel 8 Pro - Call from another device (Pixel 7A or Pixel 9)
2. Pixel 8 Pro rings for a millisecond
3. Unlock the Pixel 8 Pro, open app, it shows app crashed.

NOTE: We use the F-Droid version, as your GPLAY version does not have Background Sync option.

#### What did you expect?
Call to continue to ring, and being able to answer

#### What happened instead?
Call rings for a millisecond or less. Unlocking the phone, and starting the app reveals that app crashed.

### Your phone model

Pixel 8 Pro, Pixel 7 A, Pixel 9

### Operating system version

Android 14 and 15

### Application version and app store

Element version 1.6.24 from F-Droid

### Homeserver

Latest Synapse private server

### Will you send logs?

Yes

### Are you willing to provide a PR?

Yes

AdamMajer:
Same issue on Nokia (Android One phone, Android 14) and Elements installed from Play store.

atuctuc:
i experience the same. since i updated element from 1.6.20 to 1.6.24, every incomming call crashes element. the appears on two android phones, on both the same. i use f-droid version with unified push

ToddCrimson:
> i experience the same. since i updated element from 1.6.20 to 1.6.24, every incomming call crashes element. the appears on two android phones, on both the same. i use f-droid version with unified push

I would suggest switching back to 1.6.18 from F-Droid store.
This is the direct link to the Android Phone version - https://f-droid.org/repo/im.vector.app_40106182.apk
Version 1.6.18 (40106182) - Added on Jul 06, 2024  --  arm64-v8a 

PS: 1.6.20 has the issue with Microphone access revoking if you are on a call, and you switch to another app (putting element in background) or lock your screen.

foegra:
Same here, on Xiaomi 12 with 1.6.26

ToddCrimson:
[...]

PR #8989, opened 2025-01-13
#8964 Fix app crash on incoming call when running Android 14+

christianrowlands:
## Type of change

- [ ] Feature
- [X] Bugfix
- [ ] Technical
- [ ] Other :

## Content

I removed the call to start the MicrophoneAccessService on Android 14+.

## Motivation and context

This change is in response to this GitHub Issue: https://github.com/element-hq/element-android/issues/8964

When running on Android 14+, if you have the microphone permission granted to the app, incoming calls while the app is in the background causes the app to crash, and the incoming call to not come through. The Android documentation for that is here: https://developer.android.com/develop/background-work/services/fgs/service-types#microphone

and here: https://developer.android.com/develop/background-work/services/fgs/restrictions-bg-start#wiu-restrictions

## Tests

I tested by making a call to a device running Android 14 and my changes while the device is locked, as well as unlocked with the app open.

## Tested devices

- [X] Physical
- [ ] Emulator
- OS version(s): Android 14

## Checklist

- [ - ] Changes has been tested on an Android device or Android emulator with API 21
- [ - ] UI change has been tested on both light and dark themes
- [ - ] Accessibility has been taken into account. See https://github.com/element-hq/element-android/blob/develop/CONTRIBUTING.md#accessibility
- [X] Pull request includes a new file under ./changelog.d. See https://github.com/element-hq/element-android/blob/develop/CONTRIBUTING.md#changelog
- [ - ] Pull request includes screenshots or videos if containing UI changes
- [ - ] If you have modified the screen flow, or added new screens to the application, you have updated the test [UiAllScreensSanityTest.allScreensTest()](https://github.com/element-hq/element-android/blob/main/vector/src/androidTest/java/im/vector/app/ui/UiAllScreensSanityTest.kt#L73)

Signed-off-by: Christian Rowlands <craxiomdev [at] gmail.com>

### DIFF
--- a/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
+++ b/vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
@@ -246,15 +246,20 @@ class VectorCallActivity :
                 == PackageManager.PERMISSION_GRANTED) {
             // Only start the service if the app is in the foreground
             if (isAppInForeground()) {
-                // Starting in Android 14, you can't create a microphone foreground service while your app is in
-                // the background. If we call startForegroundService the app will crash.
-                // https://github.com/element-hq/element-android/issues/8964
-                if (Build.VERSION.SDK_INT < Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
-                    Timber.tag(loggerTag.value).v("Starting microphone foreground service")
-                    val intent = Intent(this, MicrophoneAccessService::class.java)
-                    ContextCompat.startForegroundService(this, intent)
-                } else {
-                    Timber.tag(loggerTag.value).v("App is in running Android 14+; cannot start microphone service")
+                withState(callViewModel) {
+                    // Starting in Android 14, you can't create a microphone foreground service while your app is in
+                    // the background. If we call startForegroundService while the call state is ringing (i.e. the
+                    // user has not interacted with the device at all) the app will crash. Make sure the call has
+                    // already been answered before starting the MicrophoneAccessService
+                    // https://github.com/element-hq/element-android/issues/8964
+                    val callState = it.callState.invoke()
+                    if (callState !is CallState.LocalRinging && callState !is CallState.Ended) {
+                        Timber.tag(loggerTag.value).v("Starting microphone foreground service")
+                        val intent = Intent(this, MicrophoneAccessService::class.java)
+                        ContextCompat.startForegroundService(this, intent)
+                    } else {
+                        Timber.tag(loggerTag.value).v("Call is in ringing or ended state; cannot start microphone service. callState: $callState")
+                    }
                 }
             } else {
                 Timber.tag(loggerTag.value).v("App is not in foreground; cannot start microphone service")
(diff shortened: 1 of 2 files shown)


vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
```diff
@@ -246,9 +246,16 @@ class VectorCallActivity :
                 == PackageManager.PERMISSION_GRANTED) {
             // Only start the service if the app is in the foreground
             if (isAppInForeground()) {
-                Timber.tag(loggerTag.value).v("Starting microphone foreground service")
-                val intent = Intent(this, MicrophoneAccessService::class.java)
-                ContextCompat.startForegroundService(this, intent)
+                // Starting in Android 14, you can't create a microphone foreground service while your app is in
+                // the background. If we call startForegroundService the app will crash.
```
bmarty, line 250: Yes, but in this block the app is in the foreground (since `isAppInForeground()` returned `true`), or am I misreading something?

bmarty (commented):
Thanks for the fix.
This code has been added in https://github.com/element-hq/element-android/pull/8914 in order to fix a problem when the app is "backgrounded".

Maybe the service still need to be started, but later?

vector/src/main/java/im/vector/app/features/call/VectorCallActivity.kt
christianrowlands: I assumed that too, but when I added logging it indicated the app is in the foreground. I am guessing this is the case because by the time this code is reached the app has been brought to the foreground with an incoming call UI.

As for my comment about creating the microphone foreground service while your app is in the background, it depends on how the activity was triggered. A user interaction with the foreground such as clicking on a UI button or on a notification will allow the microphone to be accessed. However, if the chain is triggered entirely from the background then the exception will occur.

I believe this approach works because now the app does not use the microphone permission until the "answer" button is clicked, which is a user interaction with the UI.

See the documentation here: https://developer.android.com/develop/background-work/services/fgs/restrictions-bg-start#background-start-restriction-exemptions

And the relevant section: 

> The user performs an action on a UI element related to your app. For example, they might interact with a [bubble](https://developer.android.com/guide/topics/ui/bubbles), [notification](https://developer.android.com/develop/ui/views/notifications), [widget](https://developer.android.com/guide/topics/appwidgets/overview), or activity.

christianrowlands:
I think you are right. The service likely still needs to be started, but later. I will see if there is a later point that we can start it (has to be after the user clicks the answer button).

christianrowlands:
@bmarty , I took another stab at this fix. Because the `VectorCallActivity` is paused and resumed before the incoming call has been answered, my approach is to only start the MicrophoneAccessService if the call has already been answered.

christianrowlands:
@bmarty , can you take another look at this PR? I think it is good now, or at least really close.

bmarty (approved):
Thanks.
I confirm that the patch fixes the crash.
One important thing to know, to reproduce the crash, the USB cable had to be unplugged. This is maybe something I did not test enough when checking the previous fix.

KeithCavill:
@bmarty and @christianrowlands  - do you guys' think that this fix can be ported over to ios side as well? That app is still showing the exact same behavior. Crashes if phone is locked.

christianrowlands:
@KeithCavill , I might take a look at the iOS version next week if I get some free time, but iOS development is not something I have done before (other than Flutter) so it will take me a bit to get situated in the code.

Issue #9086, opened 2025-09-24
[SDK] 16 KB memory page size support for Matrix SDK


jaydeep-p-simform:
### Description

https://developer.android.com/guide/practices/page-sizes

I’ve updated the Matrix SDK to the latest version 1.6.36, but I still see the following files are not aligned with the 16 KB memory page size requirement:

```
lib/arm64-v8a/libmatrix_sdk_crypto_ffi.so
lib/arm64-v8a/librealm-jni.so
```

I noticed that the 16 KB memory page size issue was addressed in [#9041](https://github.com/element-hq/element-android/issues/9041) about a week ago, but it doesn’t appear to be included in the latest SDK release. 

Could you please confirm whether the fix also covers the files mentioned above? If so, could you please let me know when you are planning to publish a new release that includes it?

shining-grimace:
Side note: the latest SDK release on Maven Central is 1.6.40. The version in GitHub releases when looking at the repo can be out of date sometimes, though the maven-central badge in the readme does show the correct latest version.

Regardless, 1.6.40 doesn't support 16 KB page size, so a new release is still required. Hoping this comes soon!

ynsok:
Starting November 1st, 2025, all app updates submitted to Google Play that target Android 15 and above must comply with the 16 KB memory page size requirement. This is an urgent matter. Do you know when the SDK will start to meet this requirement?

@shining-grimace @bmarty

MarcinKochanowski:
io.realm:realm-gradle-plugin:10.18.0 would have to be bumped to 10.19.0, release notes state that only change between these 2 is the 16KB support
bigger problem is org.matrix.rustcomponenets:crypto-android:0.6.0 which is some old version as current one I see is 0.11.1, unfortunetely this one also is not 16KB aligned :( so new binaries would have to be built
@shining-grimace @bmarty Do you know if there is any plan to meet the 16KB requirement? Thx!

bmarty:
The current Kotlin SDK will be deprecated in favor of the [Rust SDK](https://github.com/matrix-org/matrix-rust-sdk) one. We do not have the bandwidth to work on this right now. That said, we may review PR that will fix this issue. One would need to upgrade the crypto ffi sdk and the Realm dependencies.

Just to let you know, the PlayStore allows developers to require an additional time in order to comply to the 16 KB requirement. The limit can be extended to May 31, 2026.

JohannesTeichert:
is this deprecation official?

[...]

PR #9115, opened 2026-02-02
Update sentry to version 8.31.0

bmarty:
Update Sentry to the latest version.

Part of #9086

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -26,7 +26,7 @@ def glide = "4.16.0"
 def bigImageViewer = "1.8.1"
 def jjwt = "0.11.5"
 def vanniktechEmoji = "0.16.0"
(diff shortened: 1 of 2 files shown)


jmartinesp (approved):
LGTM, but it seems like Sentry silently upgraded the lifecycle library and there are breaking API changes.

bmarty:
> LGTM, but it seems like Sentry silently upgraded the lifecycle library and there are breaking API changes.

The code is compiling, which breaking API changes are you talking about?

jmartinesp:
I was talking about the issues in https://github.com/element-hq/element-android/actions/runs/21594590156/job/62223427159?pr=9115, but I just remembered we're just ignoring those 🤦 .

PR #9117, opened 2026-02-02
Remove flipper from the codebase.

bmarty:
The library is not maintained anymore, and we are not using it anymore.

Part of #9086, this is removing these libraries from the (debug) build:
 - libcrypto.so
 - libevent-2.1.so
 - libevent_core-2.1.so
 - libevent_extra-2.1.so
 - libflipper.so
 - libssl.so


jmartinesp (approved):
Thanks!

